PR #6, opened 2019-02-11
Keep the catalogue docs next to the catalogue

kenoir:


alexwlchan:
Definitely think we should have these, but I’ve cancelled the build until we get the Travis logic in.

PR #14, opened 2019-02-27
Support for selectable queries

mklander:
Add support for additional ElasticSearch queries, selectable through an additional query parameter `queryType`

https://github.com/wellcometrust/platform/issues/3242

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -143,12 +143,12 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
   private def queryTypeToWorkQuery(
     queryString: String,
     maybeQueryType: Option[String]): WorkQuery = {
-    maybeQueryType.map(_.toUpperCase) match {
-      case Some("JB") => JustBoostQuery(queryString)
-      case Some("BB") => BroaderBoostQuery(queryString)
-      case Some("SP") => SlopQuery(queryString)
-      case Some("MM") => MinimumMatchQuery(queryString)
-      case _          => SimpleQuery(queryString)
+    maybeQueryType.map(_.toLowerCase) match {
+      case Some("justboost")    => JustBoostQuery(queryString)
+      case Some("broaderboost") => BroaderBoostQuery(queryString)
+      case Some("slop")         => SlopQuery(queryString)
+      case Some("minimummatch") => MinimumMatchQuery(queryString)
+      case _                    => SimpleQuery(queryString)
     }
   }
 
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -12,6 +12,7 @@ import uk.ac.wellcome.display.models._
 import uk.ac.wellcome.models.work.internal._
 import uk.ac.wellcome.platform.api.ContextHelper.buildContextUri
 import uk.ac.wellcome.platform.api.elasticsearch.ElasticErrorHandler
+import uk.ac.wellcome.platform.api.models.WorkQuery._
 import uk.ac.wellcome.platform.api.models._
 import uk.ac.wellcome.platform.api.requests._
 import uk.ac.wellcome.platform.api.responses.{
@@ -129,11 +130,25 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
     def searchFunction: (Index, WorksSearchOptions) => Future[
       Either[ElasticError, ResultList]] =
       request.query match {
-        case Some(queryString) => worksService.searchWorks(queryString)
-        case None              => worksService.listWorks
+        case Some(queryString) =>
+          worksService.searchWorks(
+            queryTypeToWorkQuery(queryString, request._queryType))
+        case None =>
+          worksService.listWorks
       }
 
     searchFunction(index, worksSearchOptions)
+  }
+
+  private def queryTypeToWorkQuery(queryString: String,
+                                   maybeQueryType: Option[String]): WorkQuery = {
+    maybeQueryType.map(_.toUpperCase) match {
+      case Some("JB") => JustBoostQuery(queryString)
+      case Some("BB") => BroaderBoostQuery(queryString)
+      case Some("SP") => SlopQuery(queryString)
+      case Some("MM") => MinimumMatchQuery(queryString)
+      case _          => SimpleQuery(queryString)
+    }
   }
 
   private def handleWorksServiceResult[T](maybeResult: Either[ElasticError, T],
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
@@ -1,0 +1,76 @@
+package uk.ac.wellcome.platform.api.models
+
+import com.sksamuel.elastic4s.http.ElasticDsl._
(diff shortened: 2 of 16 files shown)


alexwlchan:
reviewing

api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -129,13 +130,28 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
...
 
+  private def queryTypeToWorkQuery(
+    queryString: String,
+    maybeQueryType: Option[String]): WorkQuery = {
+    maybeQueryType.map(_.toUpperCase) match {
+      case Some("JB") => JustBoostQuery(queryString)
+      case Some("BB") => BroaderBoostQuery(queryString)
+      case Some("SP") => SlopQuery(queryString)
+      case Some("MM") => MinimumMatchQuery(queryString)
+      case _          => SimpleQuery(queryString)
```
alexwlchan: What’s the thinking behind these query parameters?

Personally I’d lean towards more expressive terms (justboost, broaderboost, slop, minmatch), but I don't feel strongly about it.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
```diff
@@ -0,0 +1,72 @@
...
+  MultiMatchQuery,
+  MultiMatchQueryBuilderType
+}
+
+sealed trait WorkQuery {
+  val queryString: String
+  def query(): Query
+}
+
+object WorkQuery {
```
alexwlchan: Are these queries written down anywhere? Is there a ticket or doc or something I can look at?

api/api/src/main/scala/uk/ac/wellcome/platform/api/requests/Requests.scala
```diff
@@ -28,6 +29,7 @@ case class V1MultipleResultsRequest(
   @QueryParam includes: Option[V1WorksIncludes],
   @QueryParam query: Option[String],
   @QueryParam _index: Option[String],
+  @QueryParam _queryType: Option[String],
```
alexwlchan: I’d double-check with Jonathan, but I’d be inclined to skip the V1 API. You can replicate the results (roughly) with the filters, and we’re not really doing much work on it these days. At some point we’ll be freezing its code entirely.

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -229,6 +240,147 @@ class ElasticsearchServiceTest
...
 
+  describe("searches using JustBoostQuery") {
+    it("finds results for a JustBoostQuery search") {
+      withLocalWorksIndex { index =>
+        // Longer text used to ensure signal in TF/IDF
+        val text = "Text that contains Aegean"
+
+        val work = createIdentifiedWorkWith()
+        val matchingTitle =
+          createIdentifiedWorkFrom(work)(title = f"$text title")
```
alexwlchan: Another way to write this that might be a little clearer (and save you defining a method with a stack of arguments) would be:

```scala
val matchingTitle = work.copy(title = s"$text title")
```

Also I’m reasonably sure f-strings are Python, not Scala, no?

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -229,6 +240,147 @@ class ElasticsearchServiceTest
...
+          genres = List(createGenreWith(f"$text genre")))
+        val matchingDescription = createIdentifiedWorkFrom(work)(
+          description = Some(f"$text description"))
+        val matchingLettering =
+          createIdentifiedWorkFrom(work)(lettering = Some(f"$text lettering"))
+        val matchingContributor =
+          createIdentifiedWorkFrom(work)(
+            contributors =
+              List(createPersonContributorWith(f"$text contributor")))
+        val noMatch = createIdentifiedWorkFrom(work)(title = "Before a Bengal")
```
alexwlchan: As these are repeated, I’d be inclined to pull them out of the test bodies and define them as constants on the test class/inside a `describe` block. Reduce the boilerplate.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -41,15 +35,15 @@ class ElasticsearchService @Inject()(elasticClient: ElasticClient)(
   def listResults: (Index, ElasticsearchQueryOptions) => Future[
     Either[ElasticError, SearchResponse]] =
     executeSearch(
-      maybeQueryString = None,
+      maybeWorkQuery = None,
       sortDefinitions = List(fieldSort("canonicalId").order(SortOrder.ASC))
     )
 
   def simpleStringQueryResults(
```
alexwlchan: At this point `simpleStringQueryResults()` isn’t a very accurate name for this method, please change it.

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksTest.scala
```diff
@@ -369,4 +369,36 @@ class ApiV2WorksTest extends ApiV2WorksTestBase {
         }
     }
   }
+
+  it("supports the query _type parameter") {
```
alexwlchan:
```suggestion
  it("supports the _queryType parameter") {
```

api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
kenoir: Yeah, i'm with Alex - just makes it a little clearer to anyone not directly involved.

api/api/src/main/scala/uk/ac/wellcome/platform/api/requests/Requests.scala
kenoir: Yes do.

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
kenoir: 🐍

mklander: I don't think copy is the same because `createIdentifiedWorkFrom` is a shorthand for
`val matchingTitle = work.copy(canonicalId: String = createCanonicalId, title = s"$text title")`
not
`val matchingTitle = work.copy(title = s"$text title")`

However, it is okay with random works if the query text is long enough so I have changed it.

`f` [formatted strings](https://docs.scala-lang.org/overviews/core/string-interpolation.html) are valid Scala though the expressiveness isn't needed here, so have removed that too.

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksTest.scala
mklander: Thanks, old name :)

api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
mklander: 👍

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
mklander: I don't think so, they are a work in progress.

PR #17, opened 2019-02-28
Relative Location Header in redirect

kenoir:
Currently redirects are absolute and the host points at `localhost` which is useless to clients.

This uses a relative scheme, correcting that issue.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -16,7 +16,10 @@ import uk.ac.wellcome.platform.api.elasticsearch.ElasticErrorHandler
 import uk.ac.wellcome.platform.api.models.WorkQuery._
 import uk.ac.wellcome.platform.api.models._
 import uk.ac.wellcome.platform.api.requests._
-import uk.ac.wellcome.platform.api.responses.{ResultListResponse, ResultResponse}
+import uk.ac.wellcome.platform.api.responses.{
+  ResultListResponse,
+  ResultResponse
+}
 import uk.ac.wellcome.platform.api.services.{WorksSearchOptions, WorksService}
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -232,17 +230,38 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
...
+      s"/${work.canonicalId}",
+      s"/${work.redirect.canonicalId}"
+    )
+
+    val params = originalRequest.params.toMap
+      .filterNot {
+        case (k,_) => k == "id"
+      }
+      .map {
+        case (k,v) => s"$k=$v"
```
alexwlchan: Have you checked this handles URL encoding/decoding of parameters in queries correctly?

kenoir: No. It may well not - I'd rather not add more custom code to do URL param conversion as that should definitely be handled by whatever library is parsing these requests. Sadly Finatra does not expose that class (it exists,but is private).

api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -232,17 +233,38 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
...
+                                  contextUri: String) = {
+
+    val path = originalRequest.path.replaceAll(
+      s"/${work.canonicalId}",
+      s"/${work.redirect.canonicalId}"
+    )
+
+    val params = originalRequest.params.toMap
+      .filterNot {
+        case (k, _) => k == "id"
```
jamesgorrie: From reading this is the URL `/works/{id}` converted to `/works?id=10`?

kenoir: If you don't filter this param you can go from a search -> redirected item -> actual item with ?id=previous_id, which seems undesirable.

jamesgorrie: I didn't know we had an id param.

PR #19, opened 2019-03-04
Get the API testing in Travis again

alexwlchan:
I'm hoping that warm caches alleviate the memory pressure.


alexwlchan:
This ran to completion (even without the fix!), so merging.

PR #21, opened 2019-03-05
Add query type query

jamesgorrie:
For some reason, as a standard, we've gone with a simple string query to Elastic Search, where a [query string query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html) would be really powerful, especially on the front end to test different querying interfaces on the web-app. 

I might need a little help on the tests, as even creating new contributors (to test the query string stuff) on a work is proving tricky.

__TODO__
- [x] write tests to check query string queries

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -149,6 +149,7 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
       case Some("broaderboost") => BroaderBoostQuery(queryString)
       case Some("slop")         => SlopQuery(queryString)
       case Some("minimummatch") => MinimumMatchQuery(queryString)
+      case Some("querystring")  => QueryString(queryString)
       case _                    => SimpleQuery(queryString)
     }
   }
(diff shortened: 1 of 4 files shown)


alexwlchan:
One of the reasons we went with a SimpleStringQuery over a QueryStringQuery in the initial implementation was that an SSQ will always return a result from Elasticsearch, whereas QSQ can return an error if you pass a malformed query – which we'd need to translate into a 400 error.

jamesgorrie:
> think about how you'd handle that on the front end if the user puts in an invalid query.

Let's see if it's useful, then more than happy to make it robust and lovely.

mklander:
If I understand this change correctly I think it is the wrong approach.  Queries can be prototyped and tested directly on the ES cluster (preferably a non-live cluster).  The API has a role in protecting the backend storage, which is subverted by allowing unconstrained queries .

alexwlchan:
I’m with @mklander on this one. And it exposes weirdness of our internal model to clients – whereas if we wrap these specific queries in filters or our own search parameters, we present a consistent view of the world.

jamesgorrie:
> And it exposes weirdness of our internal model to clients

Which could be hidden behind interfaces if built properly. The point in this would be to allow us to quickly build out something and validate that people want it before creating an end-to-end feature - we could then upgrade it to a more fully-fledged tested and filter / aggregation etc.

> The API has a role in protecting the backend storage

This is true, maybe a more robust solution would be to put a query parser in front of the `QueryStringQuery`, which would allow us to whitelist queries, and fail elegantly on invalid ones, and potentially hide the internal model raised by @alexwlchan above.

Maybe something like [parboiled](https://github.com/sirthias/parboiled/wiki)?

PR #27, opened 2019-03-28
Make the transformer a little more tolerant of incorrect 264 fields

alexwlchan:
Pretty much all the recent Sierra transformer errors have been because the MARC record contains 260 and 264 fields, except the 264 field doesn't tell us anything useful:

```
260    2019 
264  1 :|b,|c 
```

We can detect these and not throw an error pretty easily, which should make the pipeline a bit less chatty.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraProduction.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraProduction.scala
@@ -165,6 +165,14 @@ trait SierraProduction {
       case _ => false
     }
 
+  private def marc264IsOnlyPunctuation(
+    marc264fields: List[VarField]): Boolean =
+    marc264fields
+      .map { vf: VarField =>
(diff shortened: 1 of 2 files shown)


alexwlchan:
> 264 is preferred now over 260 (but this discards it if it's unhelpfully populated)?

Yep.

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraProduction.scala
```diff
@@ -165,6 +165,14 @@ trait SierraProduction {
       case _ => false
     }
 
+  private def marc264IsOnlyPunctuation(
```
mklander: I wonder a bit about the name, the test nicely shows this matches `:` `,` or `blank` but the name suggests it won't match blank and will match other punctuation.  Maybe something more generic?

jamesgorrie: Could split it into `isEmpty`, and `isPunctuation` to be more explicit.

PR #28, opened 2019-03-28
Fix the name of one of the Miro contributors

alexwlchan:
For wellcometrust/platform#3523.


jamesgorrie:
@alexwlchan only if it's efficient, fancy running me through the reindex process?

pollecuttn:
@jamesgorrie Did this reindex happen? If not will this change be picked up in the reindex for the changes to the 089?

alexwlchan:
@pollecuttn No, yes.

PR #32, opened 2019-04-09
Reduce the logging in the API tests

alexwlchan:


api/api/src/test/resources/logback-test.xml
```diff
@@ -8,4 +8,11 @@
   <root level="DEBUG">
     <appender-ref ref="STDOUT" />
   </root>
+
+  <logger name="com.sksamuel.elastic4s" level="INFO" />
```
jamesgorrie: Why is this one `INFO`?

PR #40, opened 2019-04-23
Extend label to be subfields a,b,c,d,t and remove prefix / numeration from person 

jamesgorrie:
Does the `Person` /  `Organisation` transformation bit for https://github.com/wellcometrust/platform/issues/3591

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAgents.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAgents.scala
@@ -89,7 +89,7 @@ trait SierraAgents {
 
   def getLabel(subfields: List[MarcSubfield]): Option[String] =
     subfields.filter { s =>
-      List("a", "b", "c", "t").contains(s.tag)
+      List("a", "b", "c", "d", "t").contains(s.tag)
     } map (_.content) match {
       case Nil          => None
       case nonEmptyList => Some(nonEmptyList mkString " ")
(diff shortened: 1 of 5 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAgents.scala
```diff
@@ -35,14 +21,14 @@ trait SierraAgents {
       if (normalisePerson) {
         Person.normalised(
           label = label,
-          prefix = prefixString,
-          numeration = numeration
+          prefix = None,
+          numeration = None
```
alexwlchan: I thought the implication from Jonathan was that we were scrapping this field entirely, not just setting it to `None`?

(Note that doing so would change the index structure, so we'd need to do wellcometrust/platform#3547)

jamesgorrie: Yea, this would be the first step in the transformers. I'd like to get the pipeline changed first, then the data (reindex) and then change the index shape.

Sound like a sensible plan?

jamesgorrie:
@alexwlchan made another change post review, fancy another set of 👁s on it?
https://github.com/wellcometrust/catalogue/pull/40/commits/366291a4d23e99e1a24175e0124e1bb8115a825a

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraContributorsTest.scala
```diff
@@ -461,18 +338,31 @@ class SierraContributorsTest
...
       val varFields = List(
         createVarFieldWith(
-          marcTag = "710",
-          subfields = List(MarcSubfield(tag = "a", content = name))
+          marcTag = "110",
+          subfields = List(
+            MarcSubfield(tag = "a", content = name),
+            MarcSubfield(tag = "b", content = subordinateUnit),
+            MarcSubfield(tag = "d", content = date),
+            MarcSubfield(tag = "c", content = place)
```
alexwlchan: Worth adding another subfield with a different tag and asserting that it gets ignored?

jamesgorrie: Done. I'd like to run through these tests too as it feels like there might be some redundancy about as we're simplifying things, but also don't want to move to many parts.

PR #45, opened 2019-05-09
Allow the reporting AWS account to access the bibs merger topic

jamesgorrie:
As the reporting account will have to access the platform (catalogue) account, we need to compose a new policy into the default policy (You can't add multiple policies - [this seems to be the suggested way for now](https://www.terraform.io/docs/providers/aws/r/sns_topic_subscription.html)).

I have also added the ability to grant cross account read access to the VHS S3 buckets.

### DIFF
--- a/infrastructure/critical/vhs/s3.tf
+++ b/infrastructure/critical/vhs/s3.tf
@@ -5,6 +5,8 @@ resource "aws_s3_bucket" "bucket" {
   lifecycle {
     prevent_destroy = true
   }
+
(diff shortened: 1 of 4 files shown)


kenoir:
This is probably the simplest way right now (and maybe eventually), but it occurs to me we should think about avoiding interdependencies between stacks that end up forming a complex graph. I wonder if there is a way to reimagine this as a "shared resource held by central account" that delegates access to reporting and catalogue, so specifically a tree.

jamesgorrie:
@kenoir - , so something like `catalogue_sierra_vhs_events` => `shared_events` <= `reporting_sierra_vhs_subscribe`?

PR #54, opened 2019-05-22
Use the new storage/messaging libraries

alexwlchan:
Closes wellcometrust/platform#3628

Now we can run lots of stuff in-memory, so it's faster and easier to debug. Containers we no longer need to run:

* goobi_adapter: no DynamoDB
* reindexer: no SNS
* ingestor: no S3

Rough edges are being documented in wellcometrust/platform#3630


alexwlchan:
Closing in favour of #59.

PR #57, opened 2019-05-28
parse dates on production period

jamesgorrie:
We need to be able to take a label (free string) - parse that into a `from` `Instant` and then parse that into an `InstantRange` with an associate parser (as we add things like `inferred` and might have different `to` values dependant on the pattern). 

This does that.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
@@ -1,6 +1,6 @@
 package uk.ac.wellcome.models.work.internal
 
-import java.time.{LocalDateTime, Year}
+import java.time.{Instant, LocalDateTime, Year, ZoneOffset}
 import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder}
 import java.time.temporal.ChronoField
 
@@ -22,11 +22,22 @@ object Concept {
 }
 
 case class DateRange(label: String,
-                     from: LocalDateTime,
-                     to: LocalDateTime,
+                     from: Instant,
+                     to: Instant,
                      inferred: Boolean)
     extends AbstractConcept
 object DateRange {
+  // We use this apply as it's easier to work with date math on LocalDateTime than it is on Instant
+  def apply(label: String,
+            from: LocalDateTime,
+            to: LocalDateTime,
+            inferred: Boolean): DateRange =
+    DateRange(
+      label = label,
+      from = from.toInstant(ZoneOffset.UTC),
+      to = to.toInstant(ZoneOffset.UTC),
+      inferred = inferred)
+
   type GetDateRange = (String, LocalDateTime) => DateRange
   type DateRangeParser = (DateTimeFormatter, GetDateRange)
 
@@ -45,13 +56,14 @@ object DateRange {
       (label: String, from: LocalDateTime) =>
         DateRange(
           label,
-          from,
+          from = from,
           to = from.plusYears(1).minusNanos(1),
           inferred = true)
     )
   )
 
-  private def formatterWithDefaults(pattern: String): DateTimeFormatter =
+  private def formatterWithDefaultingFallbacks(
+    pattern: String): DateTimeFormatter =
     new DateTimeFormatterBuilder()
       .appendPattern(pattern)
       .parseDefaulting(ChronoField.NANO_OF_SECOND, 0)
@@ -65,9 +77,11 @@ object DateRange {
 
   def parse(label: String): Option[DateRange] = {
     parsers.toStream.map {
-      case (pattern, getDateRange) =>
-        Try(LocalDateTime.parse(label, formatterWithDefaults(pattern)))
+      case (pattern, getDateRange) => {
+        Try(
+          LocalDateTime.parse(label, formatterWithDefaultingFallbacks(pattern)))
           .map(getDateRange(label, _))
+      }
     } find (_.isSuccess) map (_.get)
   }
 }
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
@@ -7,7 +7,7 @@ import java.time.temporal.ChronoField
 import uk.ac.wellcome.models.work.text.TextNormalisation._
 
 import scala.annotation.tailrec
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 sealed trait AbstractRootConcept {
   val label: String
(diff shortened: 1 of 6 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -15,6 +21,57 @@ object Concept {
...
+      .parseDefaulting(ChronoField.MONTH_OF_YEAR, 1)
+      .parseDefaulting(ChronoField.YEAR_OF_ERA, Year.now().getValue())
+      .toFormatter()
+
+  def parse(label: String): Option[DateRange] = {
+    parsers map {
+      case (pattern, getDateRange) =>
+        Try(LocalDateTime.parse(label, formatterWithDefaults(pattern)))
+          .map(getDateRange(label, _))
+    } find (_.isSuccess) map (_.get)
```
alexwlchan: This is all a bit mushed together – it looks like it goes through the list, then calls `formatterWithDefaults` (what does that do?), and then does some `.isSuccess` and `.get`. Could we break it up to make it easier to read?

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/internal/DateRangeTest.scala
```diff
@@ -0,0 +1,38 @@
...
+        label = "[1918]",
+        start = LocalDateTime.parse("1918-01-01T00:00"),
+        end = LocalDateTime.parse("1918-12-31T23:59:59.999999999"),
+        inferred = true
+      )
+
+      DateRange.parse(label) shouldBe Some(expected)
+
+    }
+  }
```
alexwlchan: Missing tests for the unhappy path.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -15,6 +21,57 @@ object Concept {
   }
 }
 
+case class DateRange(label: String,
+                     start: LocalDateTime,
+                     end: LocalDateTime,
+                     inferred: Boolean)
+    extends AbstractConcept
+object DateRange {
```
alexwlchan: I would name this something separate (e.g. `DateRangeParsers`), pull it into a separate file and emphasise that it's a distinct thing that we can test really rigorously.

jamesgorrie: Yea, I was wondering if the standard we followed was keeping things like parsing / sanitising / validating etc with the actual case class themselves (I quite like that, but it can be easy to overload).

Should we be calling the companion objects for the other type `ThingyTypeNormaliser` then?

Or maybe instead of parse, I could use `fromString`, but that is parsing... ah naming.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
jamesgorrie: Yea, I might take a look at turning into a `tailref`, I don't really like this.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
jamesgorrie: Also - might think about DateRange, as it is `DateTime` first of all.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -15,6 +22,83 @@ object Concept {
...
+    parsers: List[(DatePattern, GetInstantRange)]): Option[InstantRange] = {
+
+    parsers match {
+      case (pattern: DatePattern, getInstantRange: GetInstantRange) :: tail =>
+        val tryLocalDateTime = Try(
+          LocalDateTime.parse(label, formatterWithDefaults(pattern)))
+
+        if (tryLocalDateTime.isSuccess)
+          tryLocalDateTime.toOption.map(ldt => getInstantRange(label, ldt))
+        else findParser(label, tail)
```
alexwlchan: This sort of `if (tryValue.isSuccess) … else ` isn't very idiomatic Scala, or at least compared to the rest of our codebase! We do a lot of pattern matching, e.g.:

```scala
tryLocalDateTime match {
  case Success(localDateTime) => Some(getInstanceRange(label, localDateTime))
  case Failure(_) => findParser(label, tail)
}
```

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/InstantRangeGenerators.scala
```diff
@@ -0,0 +1,19 @@
+package uk.ac.wellcome.models.work.generators
+
+import java.time.{LocalDateTime, ZoneOffset}
+
+import uk.ac.wellcome.models.work.internal._
+
+trait InstantRangeGenerators {
+  // This is used to make the tests more readable.
```
alexwlchan: While true, this comment isn't adding much – we have dozens of `generators/create` methods lying around, none of which have this comment.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/internal/InstantRangeTest.scala
```diff
@@ -0,0 +1,45 @@
...
+      InstantRange.parse(label) shouldBe Some(parsedInstantRange)
+    }
+  }
+
+  it("does not parse invalid date labels") {
+    forAll(
+      Table(
+        "label",
+        "nineteen sixty what, nineteen sixty who",
+        "[21233]",
```
alexwlchan: other possible test cases:

* four characters, not all digits
* four characters in square brackets, not all digits
* less than four characters, with and without square brackets

project/Common.scala
```diff
@@ -5,9 +5,8 @@ object Common {
   val settings: Seq[Def.Setting[_]] = Seq(
     scalaVersion := "2.12.6",
     organization := "uk.ac.wellcome",
-
-    resolvers ++= Seq("S3 releases" at "s3://releases.mvn-repo.wellcomecollection.org/"),
-
+    resolvers ++= Seq(
+      "S3 releases" at "s3://releases.mvn-repo.wellcomecollection.org/"),
```
alexwlchan: nit: I know this is probably just your IDE having opinions, but I like to keep this sort of unrelated change out of PRs to keep the diffs tight and easy to review.

jamesgorrie: Same, but it's should have been picked up by scalafmt anyway, so I am going to leave it.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
jamesgorrie: I knew I was missing something.

jamesgorrie:
@alexwlchan I've taken a look at your comments and made respective changes, thanks for the help.

I'm a little loathed to change the parsing code too much, so far it looks like we have a pipeline of
`text => startInstant => dependatnOnPatternInstantRange` but would like to get some more examples to see if the current implementation scales, which it might not - so don't want to waste time on it.

Clarity is important though, let me know if it needs a bit more polishing ✨ .

PR #68, opened 2019-07-02
Filter works by date range in API

warrd:
Issue: https://github.com/wellcometrust/platform/issues/3705

The API accepts two new query parameters, `_dateFrom` and `_dateTo`, both accepting ISO 8601 date format. This does not include the time stamp, date only e.g. `1905-05-19`

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
@@ -1,5 +1,7 @@
 package uk.ac.wellcome.platform.api.models
 
+import java.time.LocalDate
+
 sealed trait WorkFilter
 
 case class ItemLocationTypeFilter(locationTypeIds: Seq[String])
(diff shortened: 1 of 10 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
```diff
@@ -16,3 +18,6 @@ case object WorkTypeFilter {
   def apply(workTypeId: String): WorkTypeFilter =
     WorkTypeFilter(workTypeIds = Seq(workTypeId))
 }
+
+case class DateRangeFilter(fromDate: Option[DateTime], toDate: Option[DateTime])
```
jamesgorrie: Should these be non-optional?

warrd: I was thinking for cases where the user wants e.g any work created before the 20th century, where they might enter something like this:

```
/api/works/?date_to=1900
```

It makes sense to me to make either `fromDate` or `toDate` optional because of this (note this implementation is still unfinished). Thoughts?

jamesgorrie: Ah, I see, makes perfect sense and adds value.

PR #69, opened 2019-07-02
upgrade elastic4s to v7

jamesgorrie:
https://github.com/wellcometrust/platform/issues/3692

In prep of upgrading Elastic up.

I think maybe I have done this the wrong way round where there is a step of upgrading ES => 7 first, but thought this might be worth doing in the mean time.


common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/WorksIndexTest.scala
```diff
@@ -116,7 +115,7 @@ class WorksIndexTest
     implicit encoder: Encoder[T]): Future[Response[IndexResponse]] =
     elasticClient
       .execute {
-        indexInto(index.name / index.name).doc(toJson(t).get)
+        indexInto(index.name).doc(toJson(t).get)
```
warrd: Is this changing the name of the index, and if so, does that matter?

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndex.scala
```diff
@@ -112,6 +111,9 @@ object WorksIndex {
     objectField("agent").fields(location(), keywordField("ontologyType"))
   )
 
+  def englishTextField(name: String) =
+    textField(name).fields(textField("english").analyzer("english"))
```
warrd: :+1: nice

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticClientBuilder.scala
```diff
@@ -31,9 +32,9 @@ object ElasticClientBuilder {
       .setHttpClientConfigCallback(new ElasticCredentials(username, password))
       // Needed for the snapshot_generator.
       // TODO Make this a flag
-      .setMaxRetryTimeoutMillis(2000)
+//      .setMaxRetryTimeoutMillis(2000)
```
warrd: Is this not needed anymore? Maybe just delete this rather than commenting out if so?

common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/WorksIndexTest.scala
jamesgorrie: Na, that's the syntax for `{index_name}/{doc_type}`.
We were previously giving the `doc_type` the same value as `index_name`.

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticClientBuilder.scala
jamesgorrie: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/breaking-changes-7.0.html#maxretrytimeout-removed

Thanks for spotting.

PR #70, opened 2019-07-16
Parse dates in Marc 260 / 264 fields

warrd:
Currently we only accept marc dates in year formats (e.g. `1972` or `[1972]`).

This PR uses the [FastParse](https://www.lihaoyi.com/fastparse/) library to parse a wider range of textual dates, and can easily be extended to iniclude more.

Some examples of dates that can be parsed correctly:

```
1972
31 July 1980
July 31 1980
July 31st 1980
[July 31st 1980]
10/02/1913
10/02/1913]
Apr 1920
1870-1873
2 Aug 1914 - 16 Apr 1915
2 Aug 1914-16 Apr 1915
August 1914 - April 1915
2-10 April 1968

```


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -22,101 +18,49 @@ object Concept {
...
-// actually want it to be part of the Display
-// model as yet before we've started testing.
-// extends AbstractConcept
+                        label: String = "",
+                        inferred: Boolean = false) {
+
+  def withInferred(inferred: Boolean): InstantRange =
+    InstantRange(from, to, label, inferred)
+
+  // TODO: is label necessary? appears to always be same as Period.label
```
jamesgorrie: The reason for the `label` is it's required by the [`AbstractConcept`](https://github.com/wellcometrust/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala#L12), which this will probably extend from.

It's what the rendering app requires so as not to have to work with the structured data for rendering (although I imagine they will on occasion).

ping @gestchild and @davidpmccormick to check in if that was their understanding?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
+package uk.ac.wellcome.models.parse
+
+import java.time.LocalDate
+
+import uk.ac.wellcome.models.work.internal.InstantRange
+
+/**
+  *  Models freeform text representing some date or period
```
kenoir: We generally do not document our code like this, but it's a pretty good idea in any case 👍

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
...
+  *
+  *  @param date The date or date range
+  *  @param inferred Whether the date is known exactly or inferred
```
kenoir: Unless `inferred` has a specific use case, I would be careful about keeping it around if "it might be useful" is the reason for having it as I find it makes understanding what the code _is for_ harder.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
...
+  */
+case class FreeformDate(date: Either[FuzzyDate, DateRange],
```
kenoir: Feels like `FuzzyDate` & `DateRange` should share a common trait enabling you to disambiguate them without an `Either`.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
...
+  def withInferred(inferred: Boolean): FreeformDate =
+    FreeformDate(date, inferred)
+}
+
+/**
+  *  A exact or ambigous date
+  */
+sealed trait FuzzyDate extends DateHelpers {
+
+  def instantRange: Option[InstantRange] =
```
kenoir: I would extract this to a companion object to make the code clearer, also having the effect that the data types have less logic.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
kenoir: ... though I would only do it if it's going to be consistent.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
...
+}
+
+case class ExactDate(date: LocalDate) extends FuzzyDate
+case class Year(year: Int) extends FuzzyDate
+case class MonthAndYear(month: Int, year: Int) extends FuzzyDate
+case class Month(day: Int) extends FuzzyDate
+case class Day(day: Int) extends FuzzyDate
+
+/**
+  *  A continuos period over some days / months / years
```
kenoir: nit: continuous

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
```diff
@@ -0,0 +1,180 @@
...
+  def dateRange[_: P] =
+    DateRangeParser.parser map { Right(_) }
+}
+
+/**
+  *  Attempts to parse freeform text into a date range
+  */
+object DateRangeParser extends Parser[DateRange] with DateParsingUtils {
+
+  def parser[_: P] =
```
kenoir: Why `[_: P]`?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
```diff
@@ -0,0 +1,180 @@
...
+    "nov" -> 11,
+    "dec" -> 12,
+  )
+
+  def dayDigits[_: P] =
+    digit
+      .rep(min = 1, max = 2)
+      .!
+      .map(_.toInt)
+      .filter(value => value >= 1 && value <= 31)
```
kenoir: Can you replace `fliter` & `map` with `collect`?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
```diff
@@ -0,0 +1,180 @@
...
+    digit
+      .rep(min = 1, max = 2)
+      .!
+      .map(_.toInt)
+      .filter(value => value >= 1 && value <= 12)
+
+  def yearDigits[_: P] =
+    digit
+      .rep(exactly = 4)
+      .!
```
kenoir: Wat? `!` Wat?

common/internal_model/src/test/scala/uk/ac/wellcome/models/parse/DateParserTest.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.scalatest.{FunSpec, Matchers}
+
+import java.time.LocalDate
+
+import uk.ac.wellcome.models.work.internal.InstantRange
+import uk.ac.wellcome.models.parse.parsers.DateParser
+
+class DateParserTest extends FunSpec with Matchers {
+
+  it("parses year") {
```
kenoir: Would http://www.scalatest.org/user_guide/table_driven_property_checks help to make these tests more succinct?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
warrd: This is not a standard scala collection, but a `P` type from the [FastParse(https://www.lihaoyi.com/fastparse/) lib

project/Dependencies.scala
```diff
@@ -187,6 +188,10 @@ object ExternalDependencies {
   val swaggerDependencies = Seq(
     "com.jakehschwartz" %% "finatra-swagger" % versions.finatra
   )
+
+  val parseDependencies = Seq(
+    "com.lihaoyi" %% "fastparse" % versions.fastparse
```
kenoir: Are we confident that this parser is the right one to use long term?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
warrd: Same as above. The `!` is used to capture a value from a parser (so here any further operations such as `map` will receive a string of 4 digits)

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
...
+  *
+  *  @param from The start date
+  *  @param to The end date
+  */
+case class DateRange(from: FuzzyDate, to: FuzzyDate) extends DateHelpers {
+
+  def instantRange: Option[InstantRange] =
```
kenoir: As above, i'd prefer a companion object.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
+package uk.ac.wellcome.models.parse
```
kenoir: I think you could probably split this file up a bit.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
warrd: This is a context bound. Am following the FastParse documentation for this, without it we get the error:

```
could not find implicit value for parameter ctx: fastparse.P[_]
```

project/Dependencies.scala
warrd: I am not particularly attached to it, would happy if people would rather use https://github.com/sirthias/parboiled2 or something else?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
kenoir: Eek. Ok.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
warrd: Sounds good

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
kenoir: Is there a plaintext alternative method that makes it clearer that this is doing?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
warrd: Yep I agree, cheers

project/Dependencies.scala
kenoir: I know parboiled better, but up to you & @jamesgorrie to work out what's best.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
kenoir: Yup - i'm familiar with type bounds - guessed it was probably "so the compiler doesn't complain" but it's just a bit confusing.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
warrd: Not as far as I am aware :stuck_out_tongue_closed_eyes:

common/internal_model/src/test/scala/uk/ac/wellcome/models/parse/DateParserTest.scala
warrd: I did use that at first, but I ended up preferring this method for readability. Don't mind too much either way.

jamesgorrie: Feels like consistency might be better here as a method of testing, unless it's really unreadable?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
```diff
@@ -0,0 +1,103 @@
...
+case class Month(day: Int) extends FuzzyDate
+case class Day(day: Int) extends FuzzyDate
+
+/**
+  *  A continuos period over some days / months / years
+  *
+  *  @param from The start date
+  *  @param to The end date
+  */
+case class DateRange(from: FuzzyDate, to: FuzzyDate) extends DateHelpers {
```
jamesgorrie: I wonder if calling this a `FuzzyDateRange` might have helped with drawing the relationships between these types?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
jamesgorrie: Could we leave a note with this so we won't have to look it up when confused later?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
warrd: Yes played around a lot with the name of these thing but I think that could be clearer, cheers :+1

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
warrd: As this syntax is used absolutely everywhere in the parsers, where would the best place to leavev it be? In the parser trait [here](https://github.com/wellcometrust/catalogue/pull/70/files#diff-f833ec2c3fb7d5264ebc6086d702749bR15)?

warrd:
@kenoir I have tried out your suggestion of using the type system to dictate what can be converted to `InstantRange`, rather than using an `Option[InstantRange]`.

In the process the types have been cleaned up a bit (so no need for `FreeformDate` now or the `Either`, everything including `DateRange` extends from `FuzzyDate`).

The changes are in commit 0d2d28b, I'm sure there are ways it could be improved further so let me know what you think (and @jamesgorrie too :slightly_smiling_face:).

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -22,101 +18,49 @@ object Concept {
...
+}
+
 object InstantRange {
-  // We use this apply as it's easier to work with date math on LocalDateTime than it is on Instant
-  def apply(label: String,
-            from: LocalDateTime,
-            to: LocalDateTime,
-            inferred: Boolean): InstantRange =
+
+  def apply(date: LocalDate): InstantRange = InstantRange(date, date)
```
jamesgorrie: It looks like we only ever [use this apply once](https://github.com/wellcometrust/catalogue/pull/70/files#diff-9c256946414f264a2892d54d6e336e37R31).

Would it make more sense to do the conversion there (using the DateHelpers) to keep the case class simpler, and conversion logic in the parsing?

warrd: :+1: makes sense to me, it probably was used more frequently before the recent refactor

common/internal_model/src/test/scala/uk/ac/wellcome/models/parse/DateParserTest.scala
warrd: I have had a think and actually think this is better than using the table based tests for these test cases.

If we use table based test the error messages will be quite hard to parse, whereas here the textual descriptions make it clear (e.g. seeing "parses written date ranges spanning multiple years" has failed rather than "`Some(InstantRange(LocalDateTime(...))` does not equal `Some(InstantRange(LocalDateTime(...))`").

jamesgorrie: You could have one of the table columns being the description of the text you are trying to parse... not sure if that's wedging it in though.

jamesgorrie:
```scala
Table(
  ("date ranges spanning multiple years", InstantRange(LocalDateTime(...), LocalDateTime(...))),
  ....
) { (description, instant) => ... }
```

Assuming these tests are going to fail now too (they're crap tests)
https://github.com/wellcometrust/catalogue/blob/master/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/internal/InstantRangeTest.scala

warrd: Have tried out the `Table` syntax and discussed with @jamesgorrie. We have decided to keep the syntax as is, due to slightly better error reporting and readability.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
warrd: Turns out this is used all over the tests too, but will remove as will probably make the  tests more explicit anyway

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDate.scala
warrd: Have changed to `FuzzyDateRange` now :)

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
warrd: :point_up: done

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  def month[_: P] = writtenMonth map (Month(_))
+
+  def day[_: P] = dayDigits map (Day(_))
+}
+
+/**
+  *  Date parsing utilities
+  */
+trait DateParserUtils extends ParserUtils {
```
warrd: Completely unnecessary comment here :see_no_evil:

project/Dependencies.scala
warrd: Have implemented this with parboiled and written some comparisons: https://github.com/wellcometrust/catalogue/pull/73

PR #73, opened 2019-07-19
[TO DISCUSS] Parse dates with parboiled rather than fastparse

warrd:
This contains a rewrite of the date parsing code using [parboiled2](https://github.com/sirthias/parboiled2) rather than [FastParse](https://www.lihaoyi.com/fastparse/).

Issue: https://github.com/wellcometrust/platform/issues/3755

# Comparison

## Rule definition

In FastParse we have to add `[_ : P]` everywhere to please the compiler, which makes the definitions a bit confusing.

FastParse:

https://github.com/wellcometrust/catalogue/blob/5cbcc1b06a35aba85a061c7f72d2c93c5a173cba/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala#L13-L14

ParBoiled:

https://github.com/wellcometrust/catalogue/blob/b9df66bcd500c0179b218527bfaeba8472ea6fdc/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala#L10

## Predicates

The way to specify a predicate in ParBoiled is slightly more involved, having to subsequently [push](https://github.com/sirthias/parboiled2#id33) the value onto the stack rather than just filtering:

FastParse:

https://github.com/wellcometrust/catalogue/blob/5cbcc1b06a35aba85a061c7f72d2c93c5a173cba/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala#L160

Parboiled:

https://github.com/wellcometrust/catalogue/blob/b9df66bcd500c0179b218527bfaeba8472ea6fdc/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala#L100

## Meta-rules

In Parboiled, when defining functions which take some rule and return another rule, we have to modify any rule we want to pass to be a `() => rule` and pass that rather than the rule itself (as documented [here](https://github.com/sirthias/parboiled2#meta-rules)):

https://github.com/wellcometrust/catalogue/blob/b9df66bcd500c0179b218527bfaeba8472ea6fdc/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala#L48

This ends up with us having to write things like this :crying_cat_face::

https://github.com/wellcometrust/catalogue/blob/b9df66bcd500c0179b218527bfaeba8472ea6fdc/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala#L18-L21

Rather than this:

https://github.com/wellcometrust/catalogue/blob/5cbcc1b06a35aba85a061c7f72d2c93c5a173cba/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala#L20-L23

## Efficiency

According to metrics both are similar efficiency. See [here](https://www.lihaoyi.com/fastparse/#Performance).

## Project usage

Both projects appear to be approximately as popular as each other, and are relatively well maintained.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
@@ -1,163 +1,129 @@
 package uk.ac.wellcome.models.parse
 
 import uk.ac.wellcome.models.work.internal.InstantRange
-import fastparse._, NoWhitespace._
-import ToInstantRange._
-import DateParserImplicits._
+import org.parboiled2._
+import org.parboiled2.{Parser => ParboiledParser}
 
-/**
-  *  Attempts to parse freeform text into a date or date range
-  */
-object FreeformDateParser extends Parser[InstantRange] with DateParserUtils {
+class FreeformDateParser(val input: ParserInput)
+  extends ParboiledParser with DateHelpers {
 
-  def parser[_: P] =
-    Start ~ (inferredTimePeriod | timePeriod) ~ End
+  def parser = rule(inferredTimePeriod | timePeriod ~ EOI)
 
-  def inferredTimePeriod[_: P] =
-    ("[".? ~ timePeriod ~ "]") map (_ withInferred true)
+  def inferredTimePeriod =
+    rule(
+      ("[".? ~ timePeriod ~ "]") ~> (_ withInferred true))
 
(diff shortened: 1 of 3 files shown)


warrd:
Have discussed with @jamesgorrie and  @kenoir and have decided to stick with FastParse for now

PR #79, opened 2019-07-24
add period to rootConcept es mapping

jamesgorrie:
To better reflect that a concept can be a period.
More and better work to be done here, but this will do for now.


warrd:
:+1: 

Also looks to me like `place` should be a valid `concept` value too, but this should fix the error messages for now

jamesgorrie:
See https://github.com/wellcometrust/platform/issues/3754 for addressing this in the long run.

PR #80, opened 2019-07-24
Use Marc 008 data for production events when 260/264 unavailable

warrd:
Issue: https://github.com/wellcometrust/platform/issues/3752

When there is no `260` / `264` data we fall back to parsing the marc `008` field. When using `008` we generate a new label for each date (e.g.`1920`, `1760-1779`, `2004/04/02`). We also pull out place information from the marc `008` string.

We also use `008` when there is `260` / `264` data, but without date information. Here we fill in the dates, but leave the rest intact.


common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/Marc008DateParser.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  // TODO : should these dates be marked as inferred?
+  def questionableDate[_: P] =
+    ("q" ~ year ~ year) map { case (from, to) => FuzzyDateRange(from, to) }
+
+  def differingReleaseAndProduction[_: P] =
+    ("p" ~ year ~ year) map { case (release, production) => release }
+
+  // TODO : should century / century and decade be marked inferred
+  def partialYear[_: P] =
```
jamesgorrie: Is there a reason we use `.toInstantRange` on these parsers, and not do this when setting `val parser`?

Just made a bit tricky to read and understand why [some parsers were being converted, and other had already been converted](https://github.com/wellcometrust/catalogue/pull/80/files#diff-340b42773e0489ec2e33b6d0ea6499eeR17-R20) - but not that confusing.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -37,15 +37,22 @@ case class InstantRange(from: Instant,
 object InstantRange {
 
   def apply(from: LocalDate, to: LocalDate): InstantRange =
+    InstantRange(from, to, "")
```
jamesgorrie: I can't really see when we create InstantRange's without a label - where is this?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/Marc008DateParser.scala
warrd: If we don't convert these parsers here, i.e we have

```scala
def partialYear[_: P] = year | century | centuryAndDecade
```

then the return type of this method is `P[FuzzyDate]` rather than `P[InstantRange]`.

There is no implicit `ToInstantRange[FuzzyDate]` (as some `FuzzyDate` types are not specific unless to convert to `InstantRange`) so we wouldn't then be able to call the `toInstantRange` method on the result of this.

It is slightly annoying, but is required to do it this way if we want the strong typing of what can be converted to an `InstantRange` (otherwise we would have to return `Option[InstantRange]` like we were doing previously).

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
warrd: All these tests [here](https://github.com/wellcometrust/catalogue/blob/master/common/internal_model/src/test/scala/uk/ac/wellcome/models/parse/DateParserTest.scala) create an `InstantRange` without a label. The reason for these tests expecting an empty label is that the `DateParser` adds the label as an empty string, and it isn't until [here](https://github.com/wellcometrust/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala#L53) that the label is added.

Maybe it would be better if the date parser filled in the label itself (with the label being the whole string being parsed), but I couldn't find a way to do capture the string without the parser consuming the input.

Now you have brought this all up I'm not 100% happy with the way it is, so might take another look..

warrd: @jamesgorrie have pushed a commit updating this

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/Marc008DateParser.scala
jamesgorrie: Yea - it was just a readability thing, I think keeping the strong type system makes more sense though - my 2¢.

PR #90, opened 2019-07-31
Add remove work script

jamesgorrie:
I suppose I could have a docker container that runs the make script and then runs the script, meaning you wouldn't even need python etc, but time.


scripts/remove_work/run.py
```diff
@@ -0,0 +1,300 @@
...
+                "canonicalId": existing_work["canonicalId"],
+                "sourceIdentifier": existing_work["sourceIdentifier"],
+                "type": "IdentifiedInvisibleWork",
+                # We bump the version so any in-flight works won't overwrite
+                # this one.
+                "version": existing_work["version"] + 1,
+            }
+            print("··· Replacing work with an IdentifiedInvisibleWork")
+            resp = requests.put(
+                "%s%s/%s/%s" % (es_host, index_name, index_name, catalogue_id),
```
warrd: Not sure what version of python people are using (I've on 3.7), but using the [new string interpolation](https://www.python.org/dev/peps/pep-0498/) introduced in 3.6 makes these sorts of expressions much more readable IMO:

```python
resp = requests.put(
    f"{es_host}{index_name}/{index_name}/{catalogue_id}"
    ...
)
```

scripts/remove_work/run.py
```diff
@@ -0,0 +1,300 @@
...
+        es_host = "%s://%s:%s/" % (
+            app_secrets["es_protocol"],
+            app_secrets["es_host"],
+            app_secrets["es_port"],
+        )
+
+        for index_name in (app_env_vars["es_index_v1"], app_env_vars["es_index_v2"]):
+            print("··· Looking up %s in index %s" % (catalogue_id, index_name))
+            resp = requests.get(
+                f"{es_host}{index_name}/_doc/{catalogue_id}",
```
warrd: Ah you have used it here :tada: (see my comment https://github.com/wellcometrust/catalogue/pull/90/files#r309175621)

jamesgorrie: Thanks! This was largely a copy . paste job (but nice to clean it up).

PR #95, opened 2019-08-01
Include part number and part name from 245

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3587

### Description

Here we add `partNumber` and `partName` fields to the work model and extract them from Marc field 245 if they are present.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraPartNameTest.scala
+++ b/pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraPartNameTest.scala
@@ -36,8 +36,7 @@ class SierraPartNameTest
   val transformer = new SierraPartName {}
 
   private def getPartName(varFields: List[VarField]) =
-    transformer getPartName (createSierraBibDataWith(
-      varFields = varFields))
+    transformer getPartName (createSierraBibDataWith(varFields = varFields))
 
   private def createVarField(
     content: String,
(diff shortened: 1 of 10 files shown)


pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraPartNameTest.scala
```diff
@@ -0,0 +1,50 @@
...
+  SierraDataGenerators
+}
+
+class SierraPartNameTest
+    extends FunSpec
+    with Matchers
+    with MarcGenerators
+    with SierraDataGenerators {
+
+  val partName = "Part Name"
```
jamesgorrie: Any chance we could use real data just to give us an indication as to what this might be?

jamesgorrie: Some examples I have found:

1. `Vol. 1, no. 4,`
1. `Section A.`
1. `I, Folk medicine in Northeast Anatolia`
1. `IV,`

@jtweed @jtweed how are these going to be useful on the frontend? They aren't really renderable out of context.

Should we not be delivering the the title statement and then starting to turn it into queryable structured data?

We have gone down this process before of exploding fields and then having to stitch them together and it would be nice to avoid that if that's whate we can foresee happening?

e.g. for `4` above, the label is
`Traditional medicine in Turkey. I, Folk medicine in Northeast Anatolia E. Sezik ... [et al.]`

jamesgorrie: These are actually for partNumber ☝️ 

Part name gets:
1. Precision medicine /
1. Some historical notes on tetanus and commentaries thereon /
1. Bioenergy
1. ...

jennpb: Those definitely feel like part of the title to me. If it's currently rendered to the user as a title in Sierra, as in [this one](http://search.wellcomelibrary.org/iii/encore/record/C__Rb1318082?lang=eng), I'd go for keeping them together, rather than trying to make the 'part' a thing.

PR #96, opened 2019-08-02
Add associated dates to contributors

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3592

### Description

Rather than add the associated dates (subfield d) to contributor labels, this adds them as a separate date field on `Contributor`. The field is of type `Period`, where the existing date parser is used to to attempt to parse the field into an instant range.


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAgents.scala
```diff
@@ -89,7 +89,7 @@ trait SierraAgents {
 
   def getLabel(subfields: List[MarcSubfield]): Option[String] =
     subfields.filter { s =>
-      List("a", "b", "c", "d", "t").contains(s.tag)
+      List("a", "b", "c", "t").contains(s.tag)
```
jamesgorrie: I don't think we should remove this from the label as it will land up with us having screwy punctuation on the frontend again.

Ping @jtweed @jennpb

jamesgorrie:
```suggestion
      List("a", "b", "c", "d", "t").contains(s.tag)
```

jamesgorrie:
@warrd do we know what we're doing with this?

warrd:
I can't remember what the plan for this was, maybe @jtweed knows?

jtweed:
I think we decided not to do it.

PR #100, opened 2019-08-06
[WIP] Log results of transformers

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3784

### Description

This PR adds logging so it is possible to see the results of the transformer. For relevant fields it adds logging output, as displayed below.

*This is still a work in progress, and may need some discussion.*

### Example log output

The following has been split across multiple lines to make it easier to read:

```
14:53:48.463 [pool-77-thread-10-ScalaTest-running-SierraProductionTest] INFO  u.a.w.p.t.s.t.SierraProduction$ -
DATA TRANSFORMED:
BibID=6988525
Transformer=SierraProduction
Info=Production events extracted from field 260 (dates not extractable, used 008 for dates), 
Output=List(ProductionEvent(London,List(Place(London)),List(),List(Period(1757,Some(InstantRange(1757-01-01T00:00:00Z,1757-12-31T23:59:59.999999999Z,1757,false)))),None,ProductionEvent))
```

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
@@ -15,16 +15,20 @@ trait SierraTransformer extends Logging {
 
   type TransformInfo <: Info
 
-  lazy val transformerName = this.getClass.getSimpleName.dropRight(1)
-
-  def apply(bibId: SierraBibNumber, bibData: SierraBibData): Output = {
-    val (output, transformInfo) = transform(bibId, bibData)
-    info(
+  case class Log(bibId: SierraBibNumber, transformerName: String, transformInfo: TransformInfo, output: Output) {
+    override def toString: String =
       "DATA TRANSFORMED: " +
         s"BibID=${bibId}, " +
         s"Transformer=${transformerName}, " +
         s"Info=${transformInfo.info}, " +
-        s"Output=${output}")
+        s"Output=${output}"
+  }
+
+  lazy val transformerName = this.getClass.getSimpleName.dropRight(1)
+
+  def apply(bibId: SierraBibNumber, bibData: SierraBibData): Output = {
+    val (output, transformInfo) = transform(bibId, bibData)
+    info(Log(bibId, transformerName, transformInfo, output))
     output
   }
 
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
@@ -5,17 +5,28 @@ import grizzled.slf4j.Logging
 import uk.ac.wellcome.models.transformable.sierra.SierraBibNumber
 import uk.ac.wellcome.platform.transformer.sierra.source.SierraBibData
 
+trait Info {
+  def info: String
+}
+
 trait SierraTransformer extends Logging {
 
   type Output
 
+  type TransformInfo <: Info
+
   lazy val transformerName = this.getClass.getSimpleName.dropRight(1)
 
   def apply(bibId: SierraBibNumber, bibData: SierraBibData): Output = {
-    val output = transform(bibId, bibData)
-    info(s"DATA TRANSFORMED: BibID=${bibId}, Transformer=${transformerName}, Output=${output}")
+    val (output, transformInfo) = transform(bibId, bibData)
+    info(
+      "DATA TRANSFORMED: " +
+        s"BibID=${bibId}, " +
+        s"Transformer=${transformerName}, " +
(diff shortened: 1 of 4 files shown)


jamesgorrie:
@warrd as I'm looking at it - could you give an example of what you might want this to look like as a JSON structured log?

warrd:
@jamesgorrie well as an example, the above output might look like this if it was represented as JSON:

```json
{
  "logLevel": "INFO",
  "timestamp": "14:53:48.463",
  "logType":"transformer-output",
  "data": {
    "bibId": "6988525",
    "transformer": "SierraProduction",
    "infoString": "Production events extracted from field 260 (dates not extractable, used 008 for dates)",
    "info": {
      "primaryMarcField": "260",
      "dateMarcField": "008"
    },
    "output": "List(ProductionEvent(London,List(Place(London)),List(),List(Period(1757,Some(InstantRange(1757-01-01T00:00:00Z,1757-12-31T23:59:59.999999999Z,1757,false)))),None,ProductionEvent)"
  }
}
```

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
```diff
@@ -0,0 +1,33 @@
...
+
+trait Info {
+  def info: String
+}
+
+trait SierraTransformer extends Logging {
+
+  type Output
+
+  type TransformInfo <: Info
```
jamesgorrie: Is there a reason to not just have `def info` that returns a `String`?

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
```diff
@@ -0,0 +1,33 @@
+package uk.ac.wellcome.platform.transformer.sierra
+
+import grizzled.slf4j.Logging
+
+import uk.ac.wellcome.models.transformable.sierra.SierraBibNumber
+import uk.ac.wellcome.platform.transformer.sierra.source.SierraBibData
+
+trait Info {
+  def info: String
```
jamesgorrie: I wonder if info could return a type of Log, that then has a toString, which we could JSONify later?

e.g.
```scala
sealed trait Log { def toString: String }
trait Info {
  def info: Log
}

case class SierraTransformerLog(bibId: String, tranformerName: string, info: string, output: String) extends Log {
  def toString = "DATA TRANSFORMED: " +
        s"BibID=${bibId}, " +
        s"Transformer=${transformerName}, " +
        s"Info=${transformInfo.info}, " +
        s"Output=${output}"
}
```

I've missed pieces out, but hopefully that gives the gist.... just a thought as we've been talking about GELP etc.

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
warrd: One problem is that we are generating the info inside the `transform` method, so if we had a separate method just for an `info` string, it would essentially have to duplicate the logic inside `transform`.

We could build a string directly inside `transform`, and return `(Output, String)` (instead of `(Output, TransformInfo)`), but having an associated type I think is cleaner and is easier to e.g. export as json if we wanted to represent it differently.

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
warrd: @jamesgorrie have done something similar to what you suggested.

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
```diff
@@ -0,0 +1,37 @@
...
+}
+
+trait SierraTransformer extends Logging {
+
+  type Output
+
+  type TransformInfo <: Info
+
+  case class Log(bibId: SierraBibNumber, transformerName: String, transformInfo: TransformInfo, output: Output) {
```
jamesgorrie: Should this be a `SierraTransformerLog` as we will probably be logging from elsewhere?
Although maybe best to look at the structure of that when we're doing it.

warrd: Well the `Log` is defined within the `SierraTransformer` class so to access it from outside you would have to do `SierraTransformer.Log`, which is maybe nicer than `SierraTransformer.SierraTransformerLog`?

Might make sense to define a more general `Log` trait like you suggested, tho until we do `toJson` or something it will just have `toString` (which is kind of redundant as every type already defines that)

warrd:
Closing this for now

PR #115, opened 2019-08-12
Add some notes about URL patterns

alexwlchan:
Some old notes I found lying around; figured they're marginally more useful here than anywhere else.

Is there a good place for this sort of implementation note?

### DIFF
--- a/docs/marc_tag_962_entries.md
+++ b/docs/marc_tag_962_entries.md
@@ -1,0 +1,73 @@
+# MARC tag 962 subfield $u matching entries.md
+
+This is an old note of some of the URL patterns I came across in MARC tag 962 subfield $u, and the regexes I used to match them.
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Any chance of adding context as to where these are from, and where they are used?

jamesgorrie:
Happy to have you talk me through it and I can add it.

alexwlchan:
Just discovered this is already in the code comments on https://github.com/wellcometrust/catalogue/blob/0d5832c0466802e6ef1fc2c21aa31eb31b5cacee/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/parsers/WellcomeImagesURLParser.scala

Closing as not necessary.

PR #121, opened 2019-08-16
Update sierra transformer messaging and storage libs

warrd:
## Issues

https://github.com/wellcometrust/platform/issues/3628
https://github.com/wellcometrust/platform/issues/3819

## Description

Upgrades `sierra_transformer` to use:
* `scala-messaging`  v5.3.1
* `scala-storage` v7.19.0


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/Main.scala
```diff
@@ -1,37 +1,78 @@
...
+
+  type IndexEntry = HybridIndexedStoreEntry[ObjectLocation, EmptyMetadata]
+
+  type HashEntry = DynamoHashEntry[String, Int, IndexEntry]
+
   runWithConfig { config: Config =>
+    // TODO: from where do we get the correct values for this?
+    val objectLocationPrefix = ObjectLocationPrefix("namespace", "path")
+    val dynamoConfig =
+      DynamoBuilder.buildDynamoConfig(config, namespace = "namespace")
```
warrd: Any ideas where I get these from? The `Config`? :thinking:

alexwlchan: Where you get what from? The prefix?

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/fixtures/HybridRecordReceiverFixture.scala
```diff
@@ -1,34 +1,119 @@
...
+    sierraTransformable: SierraTransformable,
+    version: Int = 1,
+    id: String = Random.alphanumeric take 10 mkString): HybridRecord = {
+
+    hybridStore.put(Version(id, version))(
+      HybridStoreEntry(sierraTransformable, EmptyMetadata()))
+    HybridRecord(
+      id = id,
+      version = version,
+      location = ObjectLocation("namespace", "path")
```
alexwlchan: Don’t you want to use the location you get from the hybridStore.put()?

(That does return the location, right?)

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/fixtures/HybridRecordReceiverFixture.scala
```diff
@@ -1,34 +1,119 @@
...
+    val typeId = s"${id}/${version}"
+    val stream = new ByteArrayInputStream("{\"x\": 1}".getBytes)
+    memoryIndexedStore.put(Version(id, version))(
+      HybridIndexedStoreEntry(typeId, EmptyMetadata()))
+    memoryStreamStore.put(typeId)(
+      new InputStreamWithLengthAndMetadata(stream, 8L, Map.empty))
+    HybridRecord(
+      id = id,
+      version = version,
+      location = ObjectLocation("namespace", "path")
```
alexwlchan: Ditto above.

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/Main.scala
warrd: The `namespace` and `path` strings

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/fixtures/HybridRecordReceiverFixture.scala
warrd: `hybridStore.put()` will return a `Version[String, Int]` (which is what can be used to access the object again from the store).

`location` is used to encode the location in S3 (as received from the adaptor), but these tests aren't using a dynamo/S3 hybrid store store (uses in memory equivilent), so it is kind of irrelevant what is in that field.

Does that make sense?

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/Main.scala
warrd: These are presumably needed for accessing the correct S3 location? Needed by `new DynamoHybridStore` below.

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/fixtures/HybridRecordReceiverFixture.scala
warrd: It's a bit confusing, so will add some more clarification:

* `HybridRecord` is the message received from the adaptor (bad naming probably?)

* The  old code used the `location` field directly to access S3

* In the discussion the other day we decided we should just access the hybrid store in the standard way (using `Version[String, Int]` rather  than going directly to S3

* At some stage we can change the message put on the queue by the adaptor to just be `Version[String, Int]`, but for now we have to put something in the `location` field even though it is ignored / irrelevant

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/Main.scala
warrd: In relation to the discussion below, is the only way we know the `ObjectLocation` from the incoming message? If so then the strategy of using the hybrid store rather than S3 directly is not going to work :see_no_evil:

kenoir: I think the value of `ObjectLocationPrefix` in this case is stable and can be provided via configuration.

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/fixtures/HybridRecordReceiverFixture.scala
jamesgorrie: Could just set this location to a val with a comment explaining what / why it is.
val `irrelevantLocation = ObjectLocation("namespace", "path") // blah blah`

warrd:
Closing this now, see #145 instead

PR #122, opened 2019-08-16
Only use PHRASE simple syntax in query

jamesgorrie:
The syntax uses a whole bunch of stuff researches might be looking for.
We also don't make this clear anywhere.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax

We could do something like Kibana if we wanted to add special syntax.
screenshot-8cf19dece3804980bc0ad36afe28c7f6 eu-west-1 aws found io_9243-2019 08 16-14_58_26

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
@@ -34,6 +34,9 @@ object WorkQuery {
         fields = defaultBoostedFields,
         lenient = Some(true),
         minimumShouldMatch = Some(defaultMSM),
+        // PHRASE is the only syntax that researchers know and understand, so we use this exclusively
+        // so as not to have unexpected results returned when using simple query string syntax.
+        // See: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax
         flags = Seq(SimpleQueryStringFlag.PHRASE)
       )
     }
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
@@ -2,6 +2,7 @@ package uk.ac.wellcome.platform.api.models
 
 import com.sksamuel.elastic4s.requests.searches.queries.{
   Query,
+  SimpleQueryStringFlag,
   SimpleStringQuery
 }
 
@@ -32,7 +33,8 @@ object WorkQuery {
         queryString,
         fields = defaultBoostedFields,
         lenient = Some(true),
-        minimumShouldMatch = Some(defaultMSM)
+        minimumShouldMatch = Some(defaultMSM),
+        flags = Seq(SimpleQueryStringFlag.PHRASE)
       )
     }
   }
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
@@ -376,12 +376,12 @@ class WorksServiceTest
         )
 
         val workLooseTitle = createIdentifiedWorkWith(
-          title = "An loose match of a title"
+          title = "A loose match of a title"
         )
 
         // Should return both
         assertSearchResultIsCorrect(
-          query = "A exact match of a title"
+          query = "An exact match of a title"
         )(
           allWorks = List(workExactTitle, workLooseTitle),
           expectedWorks = List(workExactTitle, workLooseTitle),
@@ -390,7 +390,7 @@ class WorksServiceTest
 
         // Should return only the exact match
         assertSearchResultIsCorrect(
-          query = "\"A exact match of a title\""
+          query = "\"An exact match of a title\""
         )(
           allWorks = List(workExactTitle, workLooseTitle),
           expectedWorks = List(workExactTitle),
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
@@ -269,6 +269,49 @@ class WorksServiceTest
       )
     }
 
+    it("Should simple_query_syntax into account and not fail when it's invalid") {
+      val workEmu = createIdentifiedWorkWith(
+        title = "a b c"
+      )
+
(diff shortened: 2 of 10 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
```diff
@@ -32,7 +33,8 @@ object WorkQuery {
         queryString,
         fields = defaultBoostedFields,
         lenient = Some(true),
-        minimumShouldMatch = Some(defaultMSM)
+        minimumShouldMatch = Some(defaultMSM),
+        flags = Seq(SimpleQueryStringFlag.PHRASE)
```
alexwlchan: Maybe a comment here explaining why we set this flag, and the sort of queries that break if we don’t?

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
```diff
@@ -269,6 +269,49 @@ class WorksServiceTest
       )
     }
 
+    it("Should simple_query_syntax into account and not fail when it's invalid") {
+      val workEmu = createIdentifiedWorkWith(
+        title = "a b c"
+      )
+
+      // This would land up with a too_many_clauses error due to the precedence syntax
```
alexwlchan: This sentence feels incomplete. “This would land up” – but it doesn’t, so we must be doing something to stop it!

Also, “the precedence syntax” isn’t very clear – unless you know how ES syntax works, it’s not obvious that `()` are precedence operators. Would be useful to have something that explains what originally happened, and what we’ve changed to ensure it doesn’t crash here.

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
```diff
@@ -269,6 +269,49 @@ class WorksServiceTest
       )
     }
 
+    it("Should simple_query_syntax into account and not fail when it's invalid") {
```
alexwlchan: nit: we don’t tend to prefix test descriptions with “should”, it’s implicit

Also, “not fail when it’s invalid” describes a lot of possible test! Could we be more specific? Say:

```scala
it("doesn't throw a too_many_clauses error from precedence syntax")
```

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
```diff
@@ -269,6 +269,49 @@ class WorksServiceTest
...
+      assertSearchResultIsCorrect(
+        query = "(a b c d e) h"
+      )(
+        allWorks = List(workEmu),
+        expectedWorks = List(workEmu),
+        expectedTotalResults = 1
+      )
+    }
+
+    it("Should work with simple_query_syntax PHRASE syntax (\"term\")") {
```
alexwlchan: Including an example of the syntax is much more useful!

```suggestion
    it("matches results with the PHRASE syntax (\"term\")") {
```

jamesgorrie:
@alexwlchan any clearer?

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
```diff
@@ -338,6 +338,66 @@ class WorksServiceTest
         result.left.get shouldBe a[ElasticError]
       }
     }
+
+    // See: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax
+    describe("simple query string syntax") {
+      it("uses none but PHRASE simple query syntax") {
```
alexwlchan: “uses none but”?

“none what”?

no other operators?

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
```diff
@@ -338,6 +338,66 @@ class WorksServiceTest
...
+        val work = createIdentifiedWorkWith(
+          title =
+            "+a -title | with (all the simple) query~4 syntax operators in it*"
+        )
+
+        // prefix
```
alexwlchan: "* at the end of a term signifies a prefix query"

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
```diff
@@ -338,6 +338,66 @@ class WorksServiceTest
...
+        )
+      }
+
+      it("doesn't throw a too_many_clauses exception when passing an invalid simple query syntax query") {
+        val workEmu = createIdentifiedWorkWith(
+          title = "a b c"
+        )
+
+        assertSearchResultIsCorrect(
+          query = "(a b c d e) h"
```
alexwlchan: Why is this query invalid?

jamesgorrie: Says in the description, this would normally throw a `too_many_clauses` error.

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
jamesgorrie: Left over from somewhere else.

api/terraform/locals.tf
```diff
@@ -10,7 +10,7 @@ locals {
 
   production_api     = "romulus"
   pinned_nginx       = "760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/nginx_api-gw:bad0dbfa548874938d16496e313b05adb71268b7"
-  pinned_remus_api   = "760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/api:31d558140f900af7c996f2b7062471b823768f33"
+  pinned_remus_api   = "760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/api:837a9a819942204aeb0a30d07f0a506b0a7ea633"
```
jamesgorrie: This shouldn't be here...

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
jamesgorrie: updated description

jamesgorrie:
Merging, happy to make further changes requests on this PR as and when.

PR #125, opened 2019-08-20
allow genre to be root concept

jamesgorrie:
Not immediately obvious how the hierarchy works here (root seems to imply that that's all the shapes a concept can be in).

Related to some talk here:
https://github.com/wellcometrust/platform/issues/3754

Will relook at this once it's in to close of the things it's blocking.

Probably need to find a way to test data against this, although I am sure I have seem 1 somewhere. The problem arises when not producing data that is parsed into structural data. https://github.com/wellcometrust/catalogue/issues/126

### DIFF
--- a/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndex.scala
+++ b/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndex.scala
@@ -50,13 +50,13 @@ object WorksIndex {
 
   val period = Seq(
     textField("label"),
+    keywordField("ontologyType"),
     objectField("range").fields(
       textField("label"),
       dateField("from"),
       dateField("to"),
       booleanField("inferred")
-    ),
-    keywordField("ontologyType")
+    )
   )
 
   val concept = Seq(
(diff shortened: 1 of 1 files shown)


common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndex.scala
```diff
@@ -95,7 +95,7 @@ object WorksIndex {
   def genre(fieldName: String) = objectField(fieldName).fields(
     textField("label"),
     keywordField("ontologyType"),
-    identified("concepts", concept)
+    identified("concepts", rootConcept)
```
warrd: So `genre.concept` can now be a `concept`, `agent` or `period`?

Not quite sure from the description what issue this is addressing, could you just clarify for me?

jamesgorrie: It is a concept, of which agent and period inherit their properties.
So the mapping needs to allow for fields from all of those types.

So this is basically saying allow genre to have any of the fields from concept, agent or period.

Not very clear, I will refactor once this is in and we have the data working.

PR #128, opened 2019-08-21
[MEGA-WIP] pipeline steps

jamesgorrie:
Thought being a `PipelineStep` knows only about the `DataType`s of what will come in from somewhere, and then what will go out.

Then you might have a `PipelineLogger` that is more aware of the `PipelineStates` varying states.
Then you might have a `PipelineMetrics` that is more aware of the `PipelineStates` varying states.

The PipelineStates I would like to be able to keep the `DataIn` available to them, as well as some contextual information relating to the step itself.

Still loads to think about - but putting it here for 👀 
The implementation of the thinking is a little behind.

### DIFF
--- a/pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/models/PipelineStep.scala
+++ b/pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/models/PipelineStep.scala
@@ -1,0 +1,62 @@
+package uk.ac.wellcome.platform.ingestor.models
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
+
+sealed trait PipelineResult
+case class PipelineSucceeded() extends PipelineResult
+case class PipelineFailedDetermanistically() extends PipelineResult
+case class PipelineFailedUndetermanistically() extends PipelineResult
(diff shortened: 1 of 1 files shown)


pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/models/PipelineStep.scala
```diff
@@ -0,0 +1,62 @@
...
+import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
+
+sealed trait PipelineResult
+case class PipelineSucceeded() extends PipelineResult
+case class PipelineFailedDetermanistically() extends PipelineResult
+case class PipelineFailedUndetermanistically() extends PipelineResult
+case class PipelineGetError() extends PipelineResult
+
+sealed trait PipelineStepState[+DataIn]
```
kenoir: image

💃

kenoir: 🎰 FINITE STATE MACHINES 🎰

kenoir: image

PR #131, opened 2019-08-21
👋  to V1

jamesgorrie:
This was the lightest touch for the removal of this.
Once we don't get any complaints - we can look at how we might remove all the V1 stuff.

<sup>Maybe including Finatra</sup>


api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
```diff
@@ -89,6 +89,16 @@ trait ApiWorksTestBase
       "description": "$description"
     }"""
 
+  def goneRequest(apiPrefix: String, description: String) =
```
alexwlchan: If you’re creating this, might be worth modifying `gone()` (L130–8) to use this.

jamesgorrie: Updated `gone` to be `deleted` to be more specific.

PR #134, opened 2019-08-22
put both APIs on the same index and container version

jamesgorrie:
Causing issues with being able to bump versions because the data is now out of date and not backwards incompatible.


jamesgorrie:
@warrd Merging so we don't land up with deployment issues. We need to look at tooling as there is still a lot of room for human error.

PR #135, opened 2019-08-27
Improve compilation times of pipeline apps

warrd:
I have done some profiling and discovered slow compilation times are due to derivation of implicits taking a huge amount of time.

In this PR expensive implicits are cached so that they are not derived unnecessarily. There are huge improvements in compilation times for all pipeline apps:

| Project                 | Before       | After     | Speed-up  |
| --- | --- | --- | --- |
| `internal_model`          | 1.5 secs     | 15 secs   | 0.1 x     |
| `transformer_sierra`      | 49 secs      | 1.5 secs  | 33 x      |
| `transformer_sierra-test` | 121 secs     | 12 secs   | 10 x      |
| `recorder`                | 89 secs      | 0.8 secs  | 111 x     |
| `recorder-test`           | 299 secs     | 0.6 secs  | 498 x     |
| `matcher`                 | 55 secs      | 0.6 secs  | 91 x      |
| `matcher-test`            | 367 secs     | 3 secs    | 122 x     |
| `merger`                  | 111 secs     | 0.6 secs  | 185 x     |
| `merger-test`             | 418 secs     | 1.5 secs  | 279 x     |
| `ingestor`                | 73 secs      | 0.3 secs  | 243 x     |
| `ingestor-test`           | 143 secs     | 0.6 secs  | 238 x     |
| `id_minter`               | 0.6 secs     | 0.3 secs  | 2 x       |
| `id_minter-test`          | 181 secs     | 0.7 secs  | 259 x     |

Note that `internal_model` compilation is 10 times slower in the above table, as it is here that the implicits are being derived. If this is an annoyance we can always move the implicit calculations to it's own seperate app.


common/internal_model/src/main/scala/uk/ac/wellcome/models/Implicits.scala
```diff
@@ -0,0 +1,161 @@
...
+  MergeCandidate,
+  WorkType,
+  Displayable,
+  Identified,
+  Subject,
+}
+import uk.ac.wellcome.models.matcher.MatcherResult
+import uk.ac.wellcome.json.JsonUtil._
+
+object Implicits {
```
warrd: This file is pretty ugly :see_no_evil: 

One option to make it more readable and maintainable would be to write a macro to generate the implicits below. Not sure if worth the effort right now though :thinking:

kenoir:
👏

kenoir:
Very nice.

PR #137, opened 2019-08-28
add first part of workType aggs

jamesgorrie:
Generally getting `aggs` in.

We will have it as a `_` test property first, half thinking we could create a new index to query so as not to risk crashing it, but this aggregation seems to perform pretty well.

We will also have them as a separate request from the general query, thus setting the size to `0` so on the client it will be an async call. This will allow us to send over the normal data that we are already, and for the test, only send over what we need.

Ping @wellcometrust/weco-experience-devs to see if ☝️ makes sense for what you're doing.

## Example response
```JS
{
  // This is empty for the first pass till we have monitoring etc on
  "results": [],
  "aggregations": {
    "workType": {
      "buckets": [
        { "key": "a", "count": 10009 },
        { "key": "b", "count": 65638 }]
    }
  }
}
```

## Next
I'd like to generify the types so you can just have something like
```scala
trait AggregationRequest extends GettableByString
case class WorkTypeAggregationRequest() extends AggregationRequest
```

And then have a parser for `List[GettableByString]` - but that can happen after this (and why the tests are a little lacking for the parser).


api/api/src/test/scala/uk/ac/wellcome/platform/api/models/AggregationResultsTest.scala
```diff
@@ -0,0 +1,21 @@
...
+    // val searchResponse: com.sksamuel.elastic4s.requests.searches.SearchResponse
+
+    // This mimics the behaviour of searchResponse.aggregationsAsMap
+    val singleAggregationsMap = Map("workType" -> Map.empty)
+
+    //  And this mimics the searchResponse.aggregationsAsString
+    val responseString =
+      """
+        |{"workType":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[{"key":"b","doc_count":2},{"key":"a","doc_count":1},{"key":"m","doc_count":1}]}}
+        |""".stripMargin
```
warrd: Maybe parse this to a circe `Json` object for the test, so that it is not dependent on formatting, and can therefore use indentation when defining it? Should hopefully result in nicer error message upon failure too.

See here if not sure what I mean: https://github.com/wellcometrust/catalogue/pull/146/files#diff-1cd03f8075a9a86e5d6cb75387924cc6R129-R138

common/display/src/main/scala/uk/ac/wellcome/display/serialize/AggregationRequestSerializer.scala
```diff
@@ -0,0 +1,59 @@
...
+  override def deserialize(
+    p: JsonParser,
+    ctx: DeserializationContext): List[AggregationRequest] = {
+
+    // The value that comes through here is similar to `["foo,bar"]`
+    // So we jump into the array and then grab that text
+    p.nextValue()
+    val commaSeparatedString = p.getText()
+
+    // We create a (key: String, agg: Option[Aggregation]) tuple so we can error nicely telling people which values are invalid
```
warrd: The auto-formatter hasnt appeared to break this line up. Maybe do it manually?

common/display/src/main/scala/uk/ac/wellcome/display/serialize/AggregationRequestSerializer.scala
```diff
@@ -0,0 +1,59 @@
...
+    val validations: Map[Boolean, List[(String, Option[AggregationRequest])]] =
+      commaSeparatedString
+        .split(",")
+        .toList
+        .map(_.trim)
+        .map(key => (key, AggregationRequest(key))) groupBy {
+        case (_, maybeAggregationRequest) => maybeAggregationRequest.isEmpty
+      }
```
warrd: Not a biggy, but potentially things might be a bit simpler here if we have a `Either[AggregationRequest, String]` to represent valid / invalid values: when it is valid we are not interested in the `key`, whereas when it is invalid the second value is always `None`, so modelling as a tuple does not make as much sense to me.

api/api/src/test/scala/uk/ac/wellcome/platform/api/models/AggregationResultsTest.scala
jamesgorrie: I've reformatted the string, but the point of this test it to check that we can parse an Elasticsearch search JSON string into our case classes, as Elastic4S doesn't model those out, so I've left it as a string.

There was an error in the test that probably didn't help explain this which has been fixed.

common/display/src/main/scala/uk/ac/wellcome/display/serialize/AggregationRequestSerializer.scala
jamesgorrie: Yea - `Either`s are definitely what we should use for validation here.
Changed the `AggregationRequest.apply` to  return `Either[InvalidAggregationRequest, AggregationRequest]` which tidies things up quite nicely - thanks!

common/display/src/main/scala/uk/ac/wellcome/display/serialize/AggregationRequestSerializer.scala
```diff
@@ -20,27 +20,22 @@ class AggregationRequestDeserializer
...
-    val validations: Map[Boolean, List[(String, Option[AggregationRequest])]] =
-      commaSeparatedString
-        .split(",")
-        .toList
-        .map(_.trim)
-        .map(key => (key, AggregationRequest(key))) groupBy {
-        case (_, maybeAggregationRequest) => maybeAggregationRequest.isEmpty
-      }
+    // We create a (key: String, agg: Option[Aggregation]) tuple so we can error nicely
+    // telling people which values are invalid
```
warrd: This message not relevant any more?

jamesgorrie: Nope

PR #129, opened 2019-08-21
Update recorder messaging and storage libs

warrd:
## Issues

https://github.com/wellcometrust/platform/issues/3628
https://github.com/wellcometrust/platform/issues/3819

## Description

Upgrades `recorder` to use:
* `scala-messaging`  v5.3.1
* `scala-storage` v7.19.0

PR #145, opened 2019-08-30
Update sierra transformer messaging and storage libs

warrd:
It made sense to do this again and bring in the all the nicer changes from #129.

This closes #121.


warrd:
Forced a merge of this, the build is failing due to `api` stuff which is unrelated, but `master` is also.

PR #150, opened 2019-09-03
Release aggs

jamesgorrie:
Seems there's there's a whole array of places you need to add the aggregations.
I'd like to simplify this if we could at some stage...


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayResultList.scala
```diff
@@ -23,13 +30,18 @@ case object DisplayResultList {
     resultList: ResultList,
     toDisplayWork: (IdentifiedWork, W) => T,
     pageSize: Int,
-    includes: W): DisplayResultList[T] =
+    includes: W): DisplayResultList[T] = {
+    println("-==================================")
+    println(toJson(resultList.aggregations))
+    println("-==================================")
```
warrd: Is this for debugging?

PR #152, opened 2019-09-03
Use WorkType data type for worktype aggregations

jamesgorrie:
Adds structure to returned aggregation results. Mostly there to follow spec.

__points__ (another PR as it does't affect the client)
- Change `ontologyType` to `type` - the current analogy of `DisplayType` and `Type` is OK, but I'd like to look at it.
- We could create parsers from ES JSON to our `AggregationSet` type, which I had done, but it felt a lot less clear with `.downStream("buckets").blah.blah`.
- Not sure how we are more succinct with going from a `AggregationRequest` to knowing what fields we need to request as part of the composite aggregation, or if there will be subaggregation problems further down the line. When looking at genres, as they are way more complex, it might be easier to answer those questions.


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/AggregationSet.scala
```diff
@@ -0,0 +1,87 @@
...
+import io.circe.{Decoder, Json}
+import uk.ac.wellcome.models.work.internal.WorkType
+import uk.ac.wellcome.json.JsonUtil._
+
+import scala.util.{Failure, Success}
+
+// TODO: Return a `ParseError`s where applicable
+case class Genre(label: String)
+case class AggregationSet(workType: Option[Aggregation[WorkType]],
+                          genre: Option[Aggregation[Genre]])
```
warrd: What do you think about removing the `Option[_]` here and just have `Aggregation[_]`? It seems to just contain a list of buckets so could have an empty list when none exist, or does the front-end need to receive `null` in that case?

jamesgorrie: Ah, I just re-read this - we need the option as we can get a request of:
`?aggreagtions=workType` | `?aggreagtions=genre` | `?aggreagtions=workType,genre` and we only attach the aggregations you've requested.

Having an empty list might imply that you've requested it, there just aren't any buckets e.g. none of your results have any genres, or there are no results.

So:

No key on object = client hasn't requested it.
Key with no results = client has requested it, it is just empty.

PR #156, opened 2019-09-06
Fix backwards compatibility with adaptor

warrd:
The adaptor stores `ObjectLocation` with an older format to what is used by the
recent storage libs: namely the `path` is stored as `key`.

This PR introduces an index store for VHS that uses the old format internally whilst exposing the modern API.


warrd:
@alexwlchan that would be ideal, but unfortunately would require us to update all the adaptor code to use the new storage libs. The plan was to not to that (only the pipeline), but if others think we should do that work now then am happy to.

warrd:
(another alternative would be to have compatibility code in the adaptor, but not much of an improvement on this I don't reckon)

PR #168, opened 2019-09-13
add sort parameter to the API

jamesgorrie:

Adds the sort parameter to the API.

There's probably some tightening up around the types we can do, especially around the comma separated list, but I'd like to get this in, and and ✨ after.

`sortBy` will need to be added next.

--- 

__notes__
I tried to figure out a more generic way to parse comma separated lists, but jackson is not keen on generics, and starts to do weird things with the order of the parser it choses dependant on however it is deriving the parser from the type parameter.


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -101,11 +105,23 @@ class ElasticsearchService @Inject()(elasticClient: ElasticClient)(
...
 
+    val sort = queryOptions.sort flatMap {
+      case _: ProductionDateFromSortRequest =>
+        Some(FieldSort("production.dates.range.from"))
+      case _: ProductionDateToSortRequest =>
+        Some(FieldSort("production.dates.range.to"))
+      case _ => None
+    }
+
+    println(sort)
```
warrd: Debug statement?

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -101,11 +105,23 @@ class ElasticsearchService @Inject()(elasticClient: ElasticClient)(
       case _ => None
     }
 
+    val sort = queryOptions.sort flatMap {
+      case _: ProductionDateFromSortRequest =>
+        Some(FieldSort("production.dates.range.from"))
+      case _: ProductionDateToSortRequest =>
+        Some(FieldSort("production.dates.range.to"))
+      case _ => None
```
warrd: Does this allow the user to specify ascending / descending sort?

jamesgorrie: Not yet - see the description:

> sortBy will need to be added next.

warrd: Ah sorry missed that

jamesgorrie:
@warrd Sort by will be completely controllable by the person doing the research, I don't think we can assume what order they would want it in. I imagine we would chose a default, and let the person chose, and not swap no which field is selected.

You land up with pleasing half the people and annoying the other half in that regard. Keep it simple, with as little assumption and magic as possible, and allow the person to control it.

This PR has exposed the ability to order by `from` and `to`.

jamesgorrie:
Not making assumptions now and allowing the client to chose explicitly also allows us to test these assumptions, and if something does come up as a winner, we'll codify it in the API.

jamesgorrie:
> I think we should discuss adding another sort parameter with simply production.dates (using the strategy outlined above or something else sensible).

💯 but I'd like some user research to back up what that does.

PR #171, opened 2019-09-13
Date sort ordering

warrd:

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/SortRequest.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/SortRequest.scala
@@ -17,3 +17,7 @@ object SortRequest {
       case _                       => Left(InvalidSortRequest(str))
     }
 }
+
+sealed trait SortingOrder
+case object SortAscending extends SortingOrder
+case object SortDescending extends SortingOrder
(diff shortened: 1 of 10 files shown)


common/display/src/main/scala/uk/ac/wellcome/display/models/SortRequest.scala
```diff
@@ -6,14 +6,21 @@ case class InvalidSortRequest(key: String) extends InvalidStringKeyException
...
-      case "production.dates.to"   => Right(ProductionDateToSortRequest())
+      case "production.dates.from" => Right(ProductionDateFromSortRequest)
+      case "production.dates.to"   => Right(ProductionDateToSortRequest)
       case _                       => Left(InvalidSortRequest(str))
     }
 }
+
+sealed trait SortingOrder
+
+object SortingOrder {
```
jamesgorrie: Do we use this syntax anywhere else? What do we get from wrapping it in an object?

warrd: It just means we write `SortingOrder.Ascending` instead of `SortingOrderAscending`. I prefer it in this particular case as is very explicit that the type is a subtype of `SortingOrder`.

I think it is quite a common pattern for sealed traits in scala (It is used for exactly the same purpose, for example, [in elasticsearch here](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/sort/FieldSort.scala#L30-L37).).

PR #173, opened 2019-09-16
Proposed method for sorting in API by "production.dates"

warrd:
## Issue

## Description

In #168 there was some discussion regarding the date sorting in the API.

The current implementation allows the user to explicitly sort by `production.dates.from` or `production.dates.to`. This is good for users who want flexibility, but this exposes an implementation detail that maybe should be hidden: namely, dates are stored as an `InstantRange`, which might be confusing as users would likely conceive of the vast majority of works being produced at a particular point, such as within a single year.

Here a further sort parameter `production.dates` is added. The strategy used is to

* sort by the `to` date when in `ascending` order
* sort by the `from` date when in `descending` order

This is done to stop works with broad date ranges always appearing high in the search results ([see test case for an example](https://github.com/wellcometrust/catalogue/pull/173/files#diff-15b3dabe1602921af2819a2a3456db34R555-R557)).

There are likely other strategies worth investigating, as this might not necessarily be what users want (the main issue being the results aren't reversed exactly when changing between `ascending` and `descending`), so some discussion and user research is probably required.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
@@ -18,6 +18,7 @@ import uk.ac.wellcome.display.models.{
   AggregationRequest,
   ProductionDateFromSortRequest,
   ProductionDateToSortRequest,
+  ProductionDateSortRequest,
   SortRequest,
   SortingOrder,
   WorkTypeAggregationRequest
@@ -113,14 +114,16 @@ class ElasticsearchService @Inject()(elasticClient: ElasticClient)(
     }
 
     val sort = queryOptions.sortBy
-      .flatMap {
-        case ProductionDateFromSortRequest =>
-          Some(FieldSort("production.dates.range.from"))
-        case ProductionDateToSortRequest =>
-          Some(FieldSort("production.dates.range.to"))
-        case _ => None
+      .map {
+        case ProductionDateFromSortRequest => "production.dates.range.from"
+        case ProductionDateToSortRequest => "production.dates.range.to"
+        case ProductionDateSortRequest =>
+          queryOptions.sortOrder match {
+            case SortingOrder.Ascending  => "production.dates.range.to"
+            case SortingOrder.Descending => "production.dates.range.from"
+          }
       }
-      .map { _.order(sortOrder) }
+      .map { FieldSort(_).order(sortOrder) }
 
     val limit = if (aggregations.nonEmpty) 0 else queryOptions.limit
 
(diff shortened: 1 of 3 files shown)


jamesgorrie:
This is already something that the client could do on the interface to test.

i.e. have a link that is
`?sortBy=asc&sort=production.dates.to`
`?sortBy=desc&sort=production.dates.from`

Especially where we're not sure what people are expecting, or how this will translate against real data - we're trying to avoid building things in that might not be used.

This is essentially choosing a default on what we think might be useful without any evidence. 

@jtweed might need to make the call here

warrd:
@jamesgorrie yes the client could do this, I suppose it depends on whether backend or frontend is responsible for deciding exactly how the results are ordered (to me it seems like the backend would be responsible for this, being it that queries elasticsearch and having knowledge about the data model, but I'm sure there a arguments for the other way).

> This is essentially choosing a default on what we think might be useful without any evidence.

Yes, but the same thing could be said about ordering by `production.dates.from` or `production.dates.to`. All we can do is make an informed decision about what we think might work well (ideally followed by user research to see how people respond to a few different options)

jamesgorrie:
> All we can do is make an informed decision about what we think might work well

We have the ability to do what we know we need to do e.g. sort by date and query on date ranges.
We have a substantial amount of research saying that is what we need.

I would suggest then we have a question for the experience team as to what ordering makes sense, setup some user testing, and maybe some quantitive testing (not sure how we might do that) and then look at the finding, and decide on a default, if there is one that stands out.

jamesgorrie:
Having run through it with @jennpb, it seems to make sense using the following examples.

Let's say we have (structure of dates taken from real data):
1. `1900 - 1910`
1. `July 1980`
1. `1905`
1. `7th July 1980`

We would expect:
__asc__
1. `1900 - 1910`
1. `1905`
1. `July 1980`
1. `7th July 1980`

__desc__
1. `7th July 1980`
1. `July 1980`
1. `1905`
1. `1900 - 1910`

This doesn't correspond with the code in this PR (not a critique, because it was never stated).

And starts to point in the direction of sorting by `from` in all cases. The point is taken that we can just make a decision here, as echoed by @jennpb.

warrd:
So we have discussed the issue sorting by `from` cause before. (it is actually dependent on how #170 is resolved).

Say we have the following data:

1. `1900-2000`
2. `1934`
3. `1930`
4. `1930-40`

Assuming all works are returned when filtering `1930-40`, if we filter by `from` then the ordering will be:

1. `1925-1941`
2. `1930`
3. `1930-40`
4. `1934`

Which obviously doesn't make sense. Filtering by only `from` is fine if we decide to filter out the first result in the above query, but that results in that results with larger date ranges being hidden from a lot of results, which may or may not be desirable.

warrd:
Have discussed this with @jennpb and @jamesgorrie and the outcome is that we don't want to display works with broader date ranges than the query, so this means we can close both #173 and #170 :tada:

PR #175, opened 2019-09-17
Date aggregations

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/3573

## Description

Allows aggregating on year via `?aggregations=year`

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
@@ -8,7 +8,7 @@ import com.sksamuel.elastic4s.requests.searches.queries.{
 
 sealed trait WorkQuery {
   val queryString: String
-  def query(): Query
+  def query: Query
 }
 
 object WorkQuery {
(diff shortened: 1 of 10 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkQuery.scala
```diff
@@ -8,7 +8,7 @@ import com.sksamuel.elastic4s.requests.searches.queries.{
 
 sealed trait WorkQuery {
   val queryString: String
-  def query(): Query
+  def query: Query
```
warrd: :point_up: I've removed the `()`cos whenever I see a method being called with not arguments I think I assume it is causing a side effect (obvs not a major issue and a bit of a bike shed, just thought I would mention why)

common/display/src/main/scala/uk/ac/wellcome/display/models/AggregationRequest.scala
```diff
@@ -6,14 +6,28 @@ case class InvalidAggregationStringKeyRequest(key: String)
...
+  case class Date(interval: DateInterval = DateInterval.Year)
+      extends AggregationRequest
+
   def apply(str: String)
     : Either[InvalidAggregationStringKeyRequest, AggregationRequest] =
     str match {
-      case "workType" => Right(WorkTypeAggregationRequest())
-      case _          => Left(InvalidAggregationStringKeyRequest(str))
+      case "workType" => Right(AggregationRequest.WorkType)
+      case "year" =>
```
jamesgorrie: Are we sure we want to go with year here and not `production.dates.period` and use the year as a default aggregation bucket size? 

Ping @jtweed

jamesgorrie:
Merged as I want to use some of the work here.

PR #178, opened 2019-09-23
SierraTransformer trait

warrd:
This PR introduces a simple `SierraTransformer` trait that is used by objects transforming incoming bib data to particular fields on a work (similar to what was suggested in #100, which was closed for other reasons):

```scala
trait SierraTransformer {

  type Output

  def apply(bibId: SierraBibNumber, bibData: SierraBibData): Output
}
```
This in my opinion is cleaner and more structured than using mixins to provide transformation methods. The `SierraTransformableTransformer` logic is also tidied up a little so slightly easier to understand.

Will be used on the upcoming notes work (issue https://github.com/wellcometrust/platform/issues/3631)


warrd:
@jamesgorrie yes that was what we were doing originally in the transformer logging PR (#100). Didn't include that here for now, but def agree would be a good addition

PR #183, opened 2019-09-24
Updates to physical description

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3589
https://github.com/wellcometrust/platform/issues/3585

### Description

* Uses `563` in addition to `300` for `physicalDescription`.
* Uses `300` subfields `$a`, `$b` and `$c` for `physicalDescription`
* Removes `extent` and `dimensions` fields (:thinking: anyone know whether this will break the frontend somewhere?)


jamesgorrie:
> Removes extent and dimensions fields (🤔 anyone know whether this will break the frontend somewhere?)

~Na - those were added and never worked, so removed.~

Hold up, they are used, but @alexwlchan's comment stands that the client will be able to deal with it.

https://github.com/wellcometrust/wellcomecollection.org/pull/4744

PR #185, opened 2019-09-25
Meeting and brand name subjects

warrd:
### Issues

https://github.com/wellcometrust/platform/issues/3633
https://github.com/wellcometrust/platform/issues/3634

### Description

Add "meeting" (as `Meeting`) and "brand name" (as `Concept`) subjects from marc fields 611 / 652.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraMeetingSubjects.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraMeetingSubjects.scala
@@ -27,19 +27,24 @@ object SierraMeetingSubjects
 
   val subjectVarFields = List("611")
 
+  val labelSubfields = List("a", "b", "c")
+
   def getSubjectsFromVarFields(bibId: SierraBibNumber,
                                varFields: List[VarField]) =
-    varFields.map { varField =>
-      val label = createLabel(varField, subfieldTags = List("a", "c", "d"))
-
-      val subject = Subject(
-        label = label,
-        concepts = List(Unidentifiable(Meeting(label = label)))
-      )
-
-      varField.indicator2 match {
-        case Some("0") => identify(varField.subfields, subject, "Meeting")
-        case _         => Unidentifiable(subject)
+    varFields.flatMap { varField =>
+      createLabel(varField, subfieldTags = List("a", "c", "d")) match {
+        case "" => None
+        case label =>
+          val subject = Subject(
+            label = label,
+            concepts = List(Unidentifiable(Meeting(label = label)))
+          )
+          Some(
+            varField.indicator2 match {
+              case Some("0") => identify(varField.subfields, subject, "Meeting")
+              case _         => Unidentifiable(subject)
+            }
+          )
       }
     }
 }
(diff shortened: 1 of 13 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraMeetingSubjects.scala
```diff
@@ -0,0 +1,50 @@
...
+// Use MARC field "611".
+//
+// *  Populate the platform "label" with the concatenated values of
+//    subfields a, c, and d.
+//
+// *  Populate "concepts" with a single value:
+//
+//    -   Create "label" from subfields a, c and d
+//    -   Set "type" to "Meeting"
+//    -   Use subfield 0 to populate "identifiers", if present.  Note the
```
jamesgorrie: This comment doesn't look like it's been implemented below?

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraSubjectsTransformer.scala
```diff
@@ -0,0 +1,53 @@
...
+    *
+    */
+  def createLabel(varField: VarField, subfieldTags: List[String]): String =
+    varField.subfields
+      .filter { subfield =>
+        subfieldTags.contains(subfield.tag)
+      }
+      .sortBy { _.tag }
+      .map { _.content }
+      .mkString(" ")
```
jamesgorrie: I thought we weren't meant to be adding any values? Does the marc fields not already contain ` `?

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraMeetingSubjectsTest.scala
```diff
@@ -0,0 +1,120 @@
...
+      ),
+      Unidentifiable(
+        Subject(
+          label = "Second",
+          concepts = List(Unidentifiable(Meeting(label = "Second")))
+        )
+      )
+    )
+  }
+}
```
jamesgorrie: We could wrap most these tests up into 1 table driven test.
If we're worried about the logging, we should probably try to fix the logging with the table test libs we use.

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraSubjectsTransformer.scala
warrd: Not sure I understand. This method already existed [here](https://github.com/wellcometrust/catalogue/blob/master/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraOrganisationSubjects.scala#L76-L77), I have just moved it up to the base trait.

It concatenates the content of various subfields into a single label (which we seem to do all over the place in the transformer)

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraMeetingSubjectsTest.scala
warrd: Agree that these tests maybe a bit verbose, but still not convinced that table driven tests are the way to go. Each of the tests are actually testing different things, so even if the structure of the tests is similar I reckon makes sense to have separate unit tests.

PR #187, opened 2019-09-27
SierraQueryOps helpers in the transformation code

warrd:
This introduces some extra operations (kind of a mini DSL) to help transformation of the sierra data, replacing the existing `MarcUtils` mixin.

IMO it makes the transformation code clearer than before, and is often more succinct.

For example:

```scala
// Before

val physicalDescriptionFields = List(
    "300" -> "a",
    "300" -> "b",
    "300" -> "c",
    "563" -> "a"
  )
physicalDescriptionFields
  .flatMap {
    case (tag, subfieldTag) =>
      getMatchingSubfields(
        bibData = bibData,
        marcTag = tag,
        marcSubfieldTag = subfieldTag
      )
  }
  .map(_.content) match {
  case Nil      => None
  case contents => Some(contents.mkString("\n"))
}

// After

bibData
  .subfieldsWithTags(
    "300" -> "a",
    "300" -> "b",
    "300" -> "c",
    "563" -> "a"
  )
  .contentString("\n")
```


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/source/SierraQueryOps.scala
```diff
@@ -0,0 +1,95 @@
+package uk.ac.wellcome.platform.transformer.sierra.source
+
+
+trait SierraQueryOps {
+
+  implicit class BibDataOps(bibData: SierraBibData) {
+
+    def varfields: List[VarField] = bibData.varFields
```
jamesgorrie: Why `def`?

warrd: :thinking: not sure, will make it `val`

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraCiteAs.scala
```diff
@@ -1,12 +1,17 @@
...
 
-object SierraCiteAs extends SierraTransformer with MarcUtils {
+object SierraCiteAs extends SierraTransformer with SierraQueryOps {
 
   type Output = Option[String]
 
   def apply(bibId: SierraBibNumber, bibData: SierraBibData) =
-    getFirstSubfieldContent(bibData, "524", "a")
+    bibData
+      .subfieldsWithTag("524" -> "a")
```
jamesgorrie: I'm sure I will get used to it, but having an API where you have same named method `subfieldsWithTag` used to get subfields of a certain `MarcField` in the `BibData`, and also from a `VarField`seems odd.

Would it be clearer to have something like `bibData.varfield("524").subfields("a")`?

jamesgorrie: Hold up - I realise this would defeat the point of making this more succinct.
Annoyingly the only thing I can think of it to make the name clearer* - but like I said, I'll get used to it.

* `bibData.varFieldsWithSubFieldsWithTag`?

warrd: The less succinct version is already supported like this:

```scala
bibData
    .varfieldsWithTag("524")
    .subfieldsWithTag("a")
```

However when there are multiple tags you want to capture, such as in

```scala
bibData
  .subfieldsWithTags(
    "300" -> "a",
    "300" -> "b",
    "563" -> "a"
  )
```

That wouldnt work and would have to do a `flatMap` yourself.

:thinking: I agree the naming slightly confusing, but not sure the longer name is ideal either as makes an expression consisting of a few chained operators quite a mouthful to read.

I'm tempted to do have a more generic name in this case, such as using `querySubfields` or `queryVarfields`. What do you think?

jamesgorrie: I think 1 issue is using the word `tags` - it's something that only applies to subfields, but we could drop it all together as we would say "varfield 260, subfield a". I think the word tag appeared as that's what it looks like in the sierra data:

```
{
  "subfields": {
    "tag": "a",
    "content": "halo halo"
  }
}
```

If the `BibData` was the only thing you gave a query DSL - you could then reduce it to two methods (naming will be verbose to get the point across):
```scala
type VarfieldAndSubfieldTuple = (String, String)
def getSubfields(varfieldAndSubfieldTuple: VarfieldAndSubfieldTuple*): List[MarcSubfield]`
def getVarfields(varfield: String*): List[Varfield]
```

Would that cover all use cases?

warrd: > If the BibData was the only thing you gave a query DSL - you could then reduce it to two methods 

I'm not sure how that suggestion would work. The DSL methods should be available at all levels as it's kind of the whole point that there are methods available for `BibData` / `List[VarField]` / `List[MarcSubfield]` and what makes it usable.

> I think 1 issue is using the word tags - it's something that only applies to subfields,

Doesn't `tag` apply to `VarField` in addition to `MarcSubfield`? (they have fields `marcTag` and `tag` respectively).

Options so far:

* `subfieldsWithTag`  / `varfieldsWithTag`
* `getSubfields` / `getVarfields`
* `querySubfields` / `queryVarfields`

I kind of prefer `query` than `get` as less overloaded than `get` for the datatype we are operating on (`List`), and  using `get` can lead to strange phrasings (e.g. say you had a `varfields` list you wanted to filter, would the operation be `varfields.getVarfields("500", "501")` / `varfields.get("500", "501")`? neither makes much sense).

warrd: Another option would be to provide an interface like this:

```scala
bibData.query {
  varfieldTag("500") -> subfieldTag("a")
}

bibData.query {
  varfieldTag("246") && indicator2("6")
}
```

Would be more extendable and expressive than what I have done so far.

jamesgorrie: > I'm not sure how that suggestion would work. The DSL methods should be available at all levels as it's kind of the whole point that there are methods available for BibData / List[VarField] / List[MarcSubfield] and what makes it usable.

My thoughts were that we don't really seem to use these methods on anything but the `BibData` and we either want returned a `List[Varfield]` or a `List[Subfield]` - but maybe I was wrong about that.

> I kind of prefer query than get as less overloaded than get for the datatype we are operating on (List)

I just thought get was more standard, and combined with what you are getting in the method name, it's feels expressive. `query` feels like we're offering more than we are - bikeshedding completely here.  

Another concern is adding implicit complexity over using just simple `List`s for what seems to be a couple of times where we need more than 1 `Varfield` with different `Subfield`s.

I am also not working on this code however, so if it makes it easier and more clear, go for It.

warrd: > My thoughts were that we don't really seem to use these methods on anything but the BibData and we either want returned a List[Varfield] or a List[Subfield] - but maybe I was wrong about that.

Yeh they are not just used on that, agree would make sense to just define the methods on `BibData` directly if that was the case

> Another concern is adding implicit complexity over using just simple Lists for what seems to be a couple of times where we need more than 1 Varfield with different Subfields.

Unfortunately using implicits is the only way of adding extra methods to already defined types in scala (other languages, such as rust, it is much nicer to do this). It does add a bit of complexity in the implementation, but makes the transformers much simpler to read, so agree it is a trade off. It is not just a few places but a repeating pattern in 10+ transformers.

Take your point about `query` being unrepresentative, so just gonna merge as is for now.

PR #190, opened 2019-10-01
[WIP] Structured Note type

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3909

### Description

Instead of just storing notes as strings, we include information about the type of note that is stored. For example

```js
// notes
[
  {
    "content": "note content",
    "type": "GeneralNote"
  },
  ...
]
```

Currently the mapping is:

[`500` "general note"](https://www.loc.gov/marc/bibliographic/bd500.html) -> `GeneralNote`
[`501` "with note"](https://www.loc.gov/marc/bibliographic/bd501.html) -> `GeneralNote`?
[`504` "bibliography etc, note"](https://www.loc.gov/marc/bibliographic/bd504.html) -> `BibliographicalInformation`?
[`518` "date/time and place of event note"](https://www.loc.gov/marc/bibliographic/bd518.html) -> `TimeAndPlaceNote`?
[`536` "funding information note"](https://www.loc.gov/marc/bibliographic/bd536.html) -> `FundingInformation`?
[`545` "biographical or historical data"](https://www.loc.gov/marc/bibliographic/bd545.html) -> `BibliographicalInformation`?
[`547` "former title complexity note"](https://www.loc.gov/marc/bibliographic/bd547.html) -> `GeneralNote`?
[`562` "copy and version identification note"](https://www.loc.gov/marc/bibliographic/bd562.html) -> `GeneralNote`?

There are also upcoming notes fields (used primarily for audiovisual works):

[`505` "formatted contents note"](https://www.loc.gov/marc/bibliographic/bd500.html) -> `ContentsNote`?
[`508` "creation / production credits note"](https://www.loc.gov/marc/bibliographic/bd501.html) -> `CreditsNote`?
[`511` "participant or performer note"](https://www.loc.gov/marc/bibliographic/bd501.html) -> `CreditsNote`?

For reference [here](https://gist.github.com/warrd/8a904f02d3eee5a78601de886091cddb) is the analysis of different notes  fields.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Note.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Note.scala
@@ -5,3 +5,9 @@ sealed trait Note {
 }
 
 case class GeneralNote(val content: String) extends Note
+
+case class BibliographicalInformation(val content: String) extends Note
+
+case class FundingInformation(val content: String) extends Note
+
+case class TimeAndPlaceNote(val content: String) extends Note
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Note.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Note.scala
@@ -1,0 +1,7 @@
+package uk.ac.wellcome.models.work.internal
+
+sealed trait Note {
+  val content: String
+}
+
+case class GeneralNote(val content: String) extends Note
--- a/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndex.scala
+++ b/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndex.scala
@@ -32,6 +32,12 @@ object WorksIndex {
       keywordField("id")
     )
 
+  val notes = objectField("notes")
+    .fields(
+      keywordField("type"),
+      englishTextField("content")
+    )
+
   def location(fieldName: String = "locations") =
     objectField(fieldName).fields(
       keywordField("type"),
@@ -170,7 +176,7 @@ object WorksIndex {
       language,
       location("thumbnail"),
       textField("edition"),
-      englishTextField("notes"),
+      notes,
       englishTextField("dissertation"),
       englishTextField("locationOfOriginal"),
       englishTextField("citeAs"),
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/Implicits.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/Implicits.scala
@@ -25,6 +25,7 @@ import uk.ac.wellcome.models.work.internal.{
   Location,
   MaybeDisplayable,
   MergeCandidate,
+  Note,
   Period,
   Person,
   PhysicalLocation,
@@ -36,7 +37,7 @@ import uk.ac.wellcome.models.work.internal.{
   UnidentifiedInvisibleWork,
   UnidentifiedRedirectedWork,
   UnidentifiedWork,
-  WorkType,
+  WorkType
 }
 import uk.ac.wellcome.models.transformable.SierraTransformable
 import uk.ac.wellcome.models.transformable.SierraTransformable._
@@ -52,6 +53,7 @@ object Implicits {
   // flamegraphs using the scalac-profiling plugin. See this blog post for
   // info: https://www.scala-lang.org/blog/2018/06/04/scalac-profiling.html
 
+  implicit val _dec01: Decoder[Note] = deriveDecoder
   implicit val _dec02: Decoder[SourceIdentifier] = deriveDecoder
   implicit val _dec03: Decoder[Identifiable[AbstractConcept]] = deriveDecoder
   implicit val _dec04: Decoder[Unidentifiable[AbstractConcept]] = deriveDecoder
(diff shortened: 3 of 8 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Note.scala
```diff
@@ -0,0 +1,26 @@
+package uk.ac.wellcome.models.work.internal
+
+import io.circe.{Encoder, Decoder}
+
+sealed trait NoteType
+
+case class Note(content: String, noteType: NoteType)
+
+object NoteType {
```
jamesgorrie: [I think using the `withDescriminator` set here](https://github.com/wellcometrust/scala-json/blob/master/src/main/scala/uk/ac/wellcome/json/JsonUtil.scala#L15-L16), circe will do this automagically.

https://circe.github.io/circe/codecs/adt.html#the-future

warrd: :thinking: to do it like that the code would be more like this:

```scala
sealed trait Note {
  val content: String
}

case class GeneralNote(val content: String) extends Note
```

which would serialise to:

```scala
case class GeneralNote(val content: String) extends Note

{
  "content": ".. note content ..",
  "type": "GeneralNote"
}
```

I agree much better to use the inbuilt serialisation rather than having to implement ourselves. I think I did it this way as was thinking of cases where we encounter a note type we don't recognise (the `Unknown` above), which would happen if we decided to remove a type but still had data in the index. It would error with the default behaviour, but maybe over thinking things and don't need to handle such cases.

jamesgorrie:
Is there a reason we sometime postfix the type with `Note` and sometimes not?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Note.scala
jamesgorrie: Yea, we shouldn't have types we don't recognise in the data - I would really like to avoid that situation, and having a hard error would handle that.

The data and model should always reflect each other, and if we have to remove a type, part of that would be thinking about what we do with the data.

warrd:
> Is there a reason we sometime postfix the type with Note and sometimes not?

I just chose names that I thought were readable and made sense. These are just preliminary names as was waiting for input from you and @jtweed

jtweed:
Do we want to create this proliferation of types?

Or do something a bit more like we do with work types and location types?

Ie. Have a single `Note` type, give it `noteType` property and then give those an `id` like `funding-information`?

I think I might prefer that. Thoughts?

jamesgorrie:
Another example would be language - I think this makes sense and is more inline with what we have.

We could also have `{ "type": "Note", "content": "stuff" }` and not have an ID on there for now - but as we have it maybe it's a matter of putting it there and seeing if it get's used.

warrd:
>  Ie. Have a single Note type, give it noteType property and then give those an id like funding-information?

So what it is currently:

```json
{
  "content": "...",
  "type": "FundingInformation"
}
```

What it would be with this change:

```json
{
  "content": "...",
  "noteType": "funding-information"
}
```

These are pretty similar for users of the API, and I would implement both these in the same way on the backend (just generate the "funding-information" label slightly differently). Is this mainly an issue with the use of the field name `type`, which has special significance elsewhere in the model?

warrd:
Or you mean like this?

```json
{
  "content": "...",
  "noteType": {
    "id": "funding-information",
  }
```

warrd:
After chat with @jtweed there are a few options for this:

```js
// Option 1
"notes": [
  {
    "content": "This short film shows footage of a Brazilian settlement",
    "noteType": { "id": "contents-note", "type": "NoteType" },
    "type": "Note"
  },
  {
    "content": "Time start: 00:00:00:00",
    "noteType": { "id": "contents-note", "type": "NoteType" },
    "type": "Note"
  },
  {
    "content": "Time end: 00:09:23:00",
    "noteType": { "id": "contents-note", "type": "NoteType" },
    "type": "Note"
  },
  {
    "content": "No production credits.",
    "noteType": { "id": "credits-note", "type": "NoteType" }
  }
]

// Option 2
"notes": {
  "contents-note": [
    {
      "content": "This short film shows footage of a Brazilian settlement",
      "type": "Note"
    },
    {
      "content": "Time start: 00:00:00:00",
      "type": "Note"
    },
    {
      "content": "Time end: 00:09:23:00",
      "type": "Note"
    }
  ],
  "credits-note": [
    {
      "content": "No production credits.",
      "type": "Note"
    }
  ]
}

// Option 3
"notes": [
  {
    "noteType": { "id": "contents-note", "type": "NoteType" },
    "type": "Note",
    "contents": [
      "This short film shows footage of a Brazilian settlement",
      "Time start: 00:00:00:00",
      "Time end: 00:09:23:00"
    ]
  },
  {
    "noteType": { "id": "crerdits-note", "type": "NoteType" },
    "type": "Note",
    "contents": [
      "No production credits."
    ]
  }
]
```

Any thoughts @jamesgorrie @jtweed?

jtweed:
I like the last one. It allows us to group things together a bit more easily and should still be easy to render.

@jamesgorrie what do you think, esp about the `contents` name. That's the only bit I'm not so sure about.

warrd:
I agree nicer to group them together, and option 3 doesn't require the user of the API to know the particular note type ids (i.e. the second option is more likely for consumers to hardcode `notes["contents-note"]`, which makes it harder to change names etc).

The name `contents` isn't great (`data` probably too vague too?). Another potential problem with option 3 is that an item of type `Note` is not actually a single note, but a group of notes, which is a bit deceiving. Maybe not a major issue though.

warrd:
This could potentially deal with some of the issues I mentioned above?

```json
"notes": [
  {
    "id": "contents-notes",
    "type": "NoteCategory",
    "items": [
      {
        "content": "This short film shows footage of a Brazilian settlement",
        "type": "Note"
      },
      {
        "content": "Time start: 00:00:00:00",
        "type": "Note"
      },
      {
        "content": "Time end: 00:09:23:00",
        "type": "Note"
      }
    ]
  },
  {
    "id": "credits-notes",
    "type": "NoteCategory",
    "items": [
      {
        "content": "No production credits.",
        "type": "Note"
      }
    ]
  }
]
```

PR #192, opened 2019-10-02
Refactor API to use akka http

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3910

### Checklist

- [x] Setup server with typesafe config
- [x] Routing
- [x] Parsing query string parameters
- [x] Using circe to serialize response
- [x] V1 API gone route
- [x] Health check route
- [x] context.json route
- [x] Multiple works route
- [x] Multiple works pagination
- [x] Single work route
- [x] Server error handling (e.g. on elastic search client error)
- [x] Updates tests
- [x] Error detail on invalid query string parameters
- [x] Swagger docs
- [x] Update snapshot generator
- [x] Update terraform (config vars, docker images etc)


api/api/src/main/scala/uk/ac/wellcome/platform/api/AkkaServer.scala
```diff
@@ -0,0 +1,58 @@
...
+        contextSuffix = config.getOrElse[String]("api.context.suffix")(
+          default = "/context.json"),
+      )
+
+    val router = new Router(elasticClient, elasticConfig, apiConfig)
+
+    new Runnable {
+      def run() =
+        Http()
+          .bindAndHandle(router.routes, apiConfig.host, 8080)
```
warrd: ?8080?

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -0,0 +1,227 @@
...
+      pageNumber = page.getOrElse(1),
+      aggregations = aggregations.getOrElse(Nil),
+      sortBy = sort.getOrElse(Nil),
+      sortOrder = sortOrder.getOrElse(SortingOrder.Ascending),
+    )
+
+  def workQuery: Option[WorkQuery] =
+    query.map { qry =>
+      _queryType.map(_.toLowerCase) match {
+        case _ => WorkQuery.MSMBoostQuery(qry)
```
warrd: add withNotes

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -0,0 +1,227 @@
...
+  private def dateFilter =
+    (`production.dates.from`, `production.dates.to`) match {
+      case (None, None)       => None
+      case (dateFrom, dateTo) => Some(DateRangeFilter(dateFrom, dateTo))
+    }
+}
+
+object MultipleWorksParams extends QueryParamsUtils {
+
+  def parse =
```
warrd: Add comment explaining this is a custom akka-http `Directive`, and what is does

project/Dependencies.scala
```diff
@@ -89,16 +89,16 @@ object WellcomeDependencies {
...
+    val akkaHttpCirce = "1.20.0"
     val akkaStreamAlpakka = "0.20"
     val apacheCommons = "3.7"
     val apacheLogging = "2.8.2"
     val aws = "1.11.504"
     val circe = "0.9.0"
     val elastic4s = "7.1.0"
     val fastparse = "2.1.3"
-    val finatra = "19.8.0"
-    val guice = "4.2.0"
```
kenoir: 👋 Bye Guice

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2ErrorsTest.scala
```diff
@@ -1,42 +1,39 @@
...
-        testWith(server)
-    }
 
   describe("returns a 400 Bad Request for errors in the ?include parameter") {
     it("a single invalid include") {
       assertIsBadRequest(
         "/works?include=foo",
-        description = "include: 'foo' is not a valid include"
+        description =
+          "include: 'foo' is not a valid value. Please choose one of: ['identifiers', 'items', 'subjects', 'genres', 'contributors', 'production', 'notes', 'dissertation', 'alternativeTitles']"
```
kenoir: Can we generate this list somehow?

project/Dependencies.scala
```diff
@@ -141,27 +143,6 @@ object ExternalDependencies {
     "com.sksamuel.elastic4s" %% "elastic4s-testkit" % versions.elastic4s % "test"
   )
 
-  val finatraDependencies = Seq(
```
kenoir: 💥

api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
```diff
@@ -0,0 +1,58 @@
...
+        contextSuffix = config.getOrElse[String]("api.context.suffix")(
+          default = "context.json"),
+      )
+
+    val router = new Router(elasticClient, elasticConfig, apiConfig)
+
+    new Runnable {
+      def run() =
+        Http()
+          .bindAndHandle(router.routes, apiConfig.host, 8080)
```
kenoir: Reckon this means it just presents the same as the current app to the nginx proxy in the task definition. We should look later at making sure we are using the same nginx proxy image across the estate.

api/api/src/main/resources/application.conf
```diff
@@ -0,0 +1,7 @@
+es.host=${?es_host}
```
kenoir: We setting these properly in terraform?

warrd: We appear to be setting these in the terraform here:

https://github.com/wellcometrust/catalogue/blob/ea151ecb666d2fc644bf8fdf72a01ac4ed8c2e12/api/terraform/catalogue_api/services/service/main.tf#L60-L66

I don't really understand how these are actually resolved into the correct envvars  though.

jamesgorrie: They are sent to ECS as references, which are then looked up from secrets store, and pushed in as ENV vars on container startup.

@kenoir as far as I know this is the correct way.

api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
jamesgorrie: Yea, this needs changing.

We use the container / sidecar setup, and nginx is running on port 9000:
https://github.com/wellcometrust/catalogue/blob/master/api/terraform/catalogue_api/variables.tf#L24-L26

Looking at the dockerfile we used to use and the terraform, we are on `8888`.
https://github.com/wellcometrust/catalogue/blob/master/api/terraform/main.tf#L9
https://github.com/wellcometrust/dockerfiles/blob/master/finatra_service_base/Dockerfile#L11

It would be good to use whatever is standard for AKKA HTTP, which looks like `8080`, but we also tend to use the pattern:
nginx => 9000
app => 9001

Not to fussed about what we do, but we do need to change this inline with Docker / Terraform / Code.

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
jamesgorrie: Just a bump on this.

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -0,0 +1,227 @@
...
+      _queryType.map(_.toLowerCase) match {
+        case _ => WorkQuery.MSMBoostQuery(qry)
+      }
+    }
+
+  def validationErrors: List[String] =
+    List(
+      page
+        .filterNot(_ >= 1)
+        .map(_ => "page: must be greater than 1"),
```
jamesgorrie: Annoyingly the error here is between 1 - 10000.
https://api.wellcomecollection.org/catalogue/v2/works?page=10000

warrd: This is actually caught by the error handler [here](https://github.com/wellcometrust/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticErrorHandler.scala#L38-L40), with a 400 being returned. This is how we deal with this currently, can potentially catch it earlier in a subsequent PR but for now lets just leave the current behaviour.

api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
warrd: I have updated this  to `8888` for now

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -0,0 +1,238 @@
...
+  implicit def unmarshaller[T](
+    implicit decoder: Decoder[T]): Unmarshaller[String, T] =
+    Unmarshaller.strict[String, T] { str =>
+      decoder.decodeJson(Json.fromString(str)) match {
+        case Left(err)    => throw new IllegalArgumentException(err.message)
+        case Right(value) => value
+      }
+    }
+
+  implicit val includesDecoder: Decoder[V2WorksIncludes] =
```
jamesgorrie: Looking through this - we seems to have a few different ways of doing things for the same thing.
e.g. Aggregations are just a string enum to case Object enum map, which can be a list. So is Includes.

With includes we seem to have encoded the knowledge of the valid string values into that object. I think I prefer the knowledge to live here with the routing, removing any type-unsafe string checking out of the code. Could we do that with includes?

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -0,0 +1,238 @@
...
+
+  def decodeOneOf[T](values: (String, T)*): Decoder[T] =
+    Decoder.decodeString.emap { str =>
+      values.toMap
+        .get(str)
+        .map(Right(_))
+        .getOrElse(Left(invalidValuesMsg(List(str), values.map(_._1).toList)))
+    }
+
+  def decodeOneOfCommaSeperated[T](values: (String, T)*): Decoder[List[T]] =
```
jamesgorrie: I'm not 100% sure I get the naming of this? Unless I've mixed it up is it more `decodeCommaSeparatedList`?

api/api/src/main/scala/uk/ac/wellcome/platform/api/Responses.scala
```diff
@@ -0,0 +1,111 @@
...
+            .map { fields =>
+              Json.obj(fields ++ obj.filterKeys(_ != "result").toList: _*)
+            }
+        }
+        .getOrElse(json)
+    }
+}
+
+case class MultipleWorksResponse(
+  @JsonKey("@context") context: String,
```
jamesgorrie: Could we somehow make this a static value, as it is always the same? I've given it a go in the old world and couldn't, but I actually think that might be a circe thing.

api/api/src/main/scala/uk/ac/wellcome/platform/api/Responses.scala
```diff
@@ -0,0 +1,111 @@
...
+      results = resultList.results,
+      prevPage = pageLink(currentPage - 1, resultList.totalPages, requestUri),
+      nextPage = pageLink(currentPage + 1, resultList.totalPages, requestUri),
+      aggregations = resultList.aggregations,
+    )
+
+  private def pageLink(page: Int,
+                       totalPages: Int,
+                       requestUri: Uri): Option[String] =
+    if (pageInBounds(page, totalPages))
```
jamesgorrie: What's the need for this method (`pageInBounds`)?

api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -0,0 +1,246 @@
...
+    val index = params._index.map(Index(_)).getOrElse(elasticConfig.indexV2)
+    val includes = params.include.getOrElse(V2WorksIncludes())
+    worksService
+      .findWorkById(id)(index)
+      .map {
+        case Right(Some(work: IdentifiedWork))           => workFound(work, includes)
+        case Right(Some(work: IdentifiedRedirectedWork)) => workRedirect(work)
+        case Right(Some(_))                              => workGone
+        case Right(None)                                 => workNotFound(id)
+        case Left(err)                                   => elasticError(err)
```
jamesgorrie: Out of interest - do we think this is the only error that will pop up?

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
```diff
@@ -0,0 +1,204 @@
+/* IMPORTANT:
```
jamesgorrie: Thought on the test in general.

Do you think with things like paths / query options etc we could get those from the routes, and make sure they're in the `SwaggerDocs.scala`?

What I am trying to guard against is that we add something new to the API and don't document it. The use of tis API is meant to be gearing towards being used by loads of people, so it's a consideration.

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * This file contains code for generating swagger docs from Scala case classes, and is mirrored from:
+ * https://github.com/swagger-akka-http/swagger-scala-module
+ *
+ * I could not get it working by depending on that library, as encountered the following runtime error:
+ * java.lang.NoSuchMethodError: com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
+ * This appears to be due to differences in Jackson version needed for our version of Scala and that
```
alicefuzier: Why do we still have Jackson though? Maybe this has changed while I was away, but I thought that we were using CIrce for everything other than Finatra?

warrd: It's needed for the generation of swagger JSON unfortunately :cry:

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
warrd: @jamesgorrie good point that the query params file should have knowledge about decoding the strings. Does [this commit](https://github.com/wellcometrust/catalogue/pull/192/commits/553da588ae94617c77a920d9e120a1faa7f855fa) work for you

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
warrd: I think it is because the library uses a version of Jackson which uses scala 2.13:

https://github.com/swagger-akka-http/swagger-scala-module/blob/ca1e802b856b86331ff3879cefe13c3f0a2f34de/build.sbt#L28

I am not entirely certain though, so if you want to investigate how to get it working without the copy/paste please feel free!

api/api/src/main/scala/uk/ac/wellcome/platform/api/Responses.scala
warrd: I'm guessing there shouldn't be a `nextPage` link for when there is no more data. Same logic as we had originally [here](https://github.com/wellcometrust/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala)

api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
warrd: No there are other errors that can be raised at different parts of the routing, but here we are matching on an `Either[ElasticError, Option[Work]]`, so it's the only error response we can produce a this point in the code.

api/api/src/main/scala/uk/ac/wellcome/platform/api/Responses.scala
warrd: I have updated it so we just calculate the request URI once when we load up the router, rather than per request. See [here](https://github.com/wellcometrust/catalogue/pull/192/commits/b794660969c4ba52d3e7e8266ed37ace65e734ee).

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
warrd: I think I wanted to get across the fact that each value in the list can be one of a set number of values (there is a corresponding `decodeOneOf` which only decodes a single value rather than a list).

How about `decodeEnum` and `decodeCommaSeperatedEnumList` instead?

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
warrd: :thinking: We would have to inspect the routes and have some rules about what routes need swagger documentation and those which don't (health checks, context.json, swagger.json itself etc).

We don't do anything like this for the pre-refactor Finatra version, so reckon this is beyond the scope of this PR.

api/api/src/main/scala/uk/ac/wellcome/platform/api/Responses.scala
warrd: (We can't make it completely static as is dependent on the `ApiConfig`, which isn't static and is generated from configuration)

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2ErrorsTest.scala
```diff
@@ -1,42 +1,39 @@
...
-      case (_, server: EmbeddedHttpServer) =>
-        testWith(server)
-    }
 
   describe("returns a 400 Bad Request for errors in the ?include parameter") {
     it("a single invalid include") {
       assertIsBadRequest(
         "/works?include=foo",
-        description = "include: 'foo' is not a valid include"
+        description =
```
alicefuzier: Nice! Much better error messages 😄

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
alicefuzier: I see, thanks for explaining. 
I'm not massively keen on the copy paste also because it's not incredibly intelligible to me what it does, but I don't think we should hold off merging because of this. If we can manage to ged rid of this later on then great, otherwise 🤷‍♀

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
alicefuzier: Yep, agree that is beyond the scope of this PR. 
Also, I'm not sure in practice if the hassle is actually worth it: we currently have effectively only one route (works) and although we plan to add items and others later on, we don't really know how that's going to look. It might even be that we split the api into multiple services and this becomes a non-issue. (Not saying we will definitely be doing it, just saying maybe inspecting the routes now is a bit over-engineering)

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
warrd: :+1: yep totally agree that would be better to avoid doing this, copy / pasting code which I don't really understand is not something I want to be in the habit of doing!

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
warrd: Re-reading @jamesgorrie's comment, it seems he meant checking that all query parameters are documented rather than all routes. That is a bit easier, so have added a commit [here](https://github.com/wellcometrust/catalogue/pull/192/commits/780526f3ab89bb078c132283c7a34a84ffd68210).

Turned out to be a good idea, as it picked up a few we hadn't documented yet!

alicefuzier: Nice! I think this PR is massive though and it already adds plenty of value as is... we should get it merged and iterate in subsequent PRs, me thinks.

common/display/src/main/scala/uk/ac/wellcome/display/models/WorksIncludes.scala
```diff
@@ -1,44 +1,62 @@
 package uk.ac.wellcome.display.models
 
+sealed trait WorkInclude
```
jamesgorrie: Do we still need this - this was for when we have V1 / V2 - if it has knock on effects we can change it later.

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
jamesgorrie: https://github.com/wellcometrust/platform/issues/3967

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
jamesgorrie: Maybe let's keep it at this, and we can run through it in person as a separate PR.

PR #194, opened 2019-10-03
add genre aggregations

jamesgorrie:
~I don't really like the way we deserialise to `AggregatedGenre`, and convert to `Genre`.~

~I might move the logic into `Aggregations` so that that case class(`AggregateGenre`) never leaks from there.~ 

I've moved the logic into 1 place, not massively happy with it - but at least it;'s contained and the rest of the app does the same as before.


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
```diff
@@ -3,39 +3,61 @@ package uk.ac.wellcome.platform.api.models
...
+              genres = genres,
               productionDates = getAggregation[Period](date)
             )
           )
+        }
       }
       .getOrElse { None }
 
+  def convertAggregationData[T, V](maybeAgg: Option[Aggregation[T]],
+                                   c: T => V): Option[Aggregation[V]] =
```
warrd: I reckon `convert` or something else less vague might be a better var name for this?

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayAggregations.scala
```diff
@@ -48,7 +48,9 @@ object DisplayAggregations {
       workType = displayAggregation(aggs.workType, DisplayWorkType.apply),
       productionDates =
         displayAggregation(aggs.productionDates, DisplayPeriod.apply),
-      genre = None
+      genres = displayAggregation(
+        aggs.genres,
+        (g: Genre[Displayable[AbstractConcept]]) => DisplayGenre(g, false))
```
warrd: If `includesIdentifiers` default to `false`, then could just put `DisplayGenre.apply` here? Which would make it a bit easier to read imo

jamesgorrie: Yea, I was trying not to let the somewhat strange code of the aggregations meddle with code elsewhere.

I wondered if not having the default in the case class made the code more explicit around the code base.

By setting the default to false, it feels like we're expressing that they dont normally have them, and on special situations they do. 

So I didn't want to make that implication for the sake of readability in 1 place. 

But I've probably completely over thought that.

warrd: In that case, this potentially more readable, but just as bad??

```scala
genres = displayAggregation[Genre[Displayable[AbstractConcept]], DisplayGenre](
  aggs.genres,
  DisplayGenre(_, false)
)
```

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
jamesgorrie: ✨

PR #197, opened 2019-10-07
[WIP] Deploy pipeline with new Sierra data

warrd:


pipeline/terraform/main.tf
```diff
@@ -38,3 +38,44 @@ module "catalogue_pipeline_20190916" {
...
+    "${local.sierra_merged_bibs_topic_name}",
+
+    "${local.sierra_merged_items_topic_name}",
+  ]
+  miro_adapter_topic_names = [
+    "${local.miro_reindexer_topic_name}",
+    "${local.miro_updates_topic_name}",
+  ]
+  # Elasticsearch
+  es_works_index = "v2-20191007"
```
jamesgorrie: We sometimes add a few words saying why we created the new index...  not sure if that's actually useful though.

warrd: I don't find it particularly useful (might be useful immediately after creating a new index, but generally not important why the index was originally created a few weeks down the line). Have already run a full reindex on this, so don't think it is worth changing now, but happy to add a suffix on future ones.

jamesgorrie: 💯 Was just a note which made me realise I was following blindly.

PR #204, opened 2019-10-10
search on work and items ids

jamesgorrie:
I was thinking we could make this standard for all IDs - but would like to approach that piece of work if that is the requirement, as we might land up with a solution of creating the Items API, which is maybe better though about in context of ingesting the METS.

For now, all I need for the requests service is to be able to search for an item ID.


jamesgorrie:
Merging - needed for item requests service

PR #210, opened 2019-10-22
METS transformer reads from bags endpoint

alicefuzier:


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/BagsRetriever.scala
```diff
@@ -0,0 +1,14 @@
+package uk.ac.wellcome.platform.transformer.mets.service
+
+import scalaj.http.Http
```
warrd: Any reason we shouldn't use akka-http for this, as already a dependency? :thinking:

alicefuzier: Not really, I had used this one in the past and the syntax is very simple. But also, I haven't really tried akka-http, I've only just scrolled the docs... I'll give it a go

jamesgorrie: Personally, after working on the item service, I found the `scalaj` more easier to work with if not working within the AKKA HTTP world (I need a client to talk to an API).

That might be my lack of understanding of AKKA HTTP, but it seemed, to do simple HTTP requests, overly verbose.

warrd: :+1: if people prefer this I don't mind, was just checking

alicefuzier: Yeeah that was the feeling I got from scrolling the akka-http docs. But I'll give it a go, maybe it's not so bad

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/model/IngestUpdate.scala
```diff
@@ -1,3 +1,18 @@
 package uk.ac.wellcome.platform.transformer.mets.model
 
-case class IngestUpdate()
+case class IngestUpdate(space: Space, bag: StubBag)
+
+case class Space(id: String)
+trait BaseBag {
+  val info: BagInfo
```
jamesgorrie: We've called it data in the works, not sure if this is taken from the storage vocab though, in which case it feels like following that make sense.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/BagsRetriever.scala
jamesgorrie: I would definitely ask the question of - does any develop who then needs to touch this code need to go learn the AKKA paradigms? Maybe we're happy with that because we use it all over the shop.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/model/IngestUpdate.scala
alicefuzier: Yes, bag info is what it's called in the storage world I think

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/BagsRetriever.scala
warrd: It doesn't look like there are any new paradigms compared to the `scalaj` one though, the main difference is it returns a `Future`?

It seems simple enough to me from the docs, though maybe it is more complicated in practice:

```scala
val responseFuture: Future[HttpResponse] = Http().singleRequest(HttpRequest(uri = "http://akka.io"))
```

I have slight preference for `akka-http`, having a non blocking interface and keeping the dependencies down, but like I said above, have no issues if others prefer `scalaj`.

alicefuzier: I've changed to akka-http [here](https://github.com/wellcometrust/catalogue/pull/210/commits/c897d598e714ea352c1519efd16293a5cb53ec3d). Actually, I prefer akka-http too. The unmarshalling thing to parse the response is a bit awkward but I like that it's non blocking and i think there are circe unmarshallers that we can plug in (haven't done it yet)

alicefuzier: Thanks @warrd for the suggestion

warrd: I have an example of circe support in the API [here](https://github.com/wellcometrust/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala#L15), you would just need an implicit decoder in scope.

PR #212, opened 2019-10-23
Nested WorkData

warrd:
### Waiting on

~~#192 (akka-http)~~


jamesgorrie:
@warrd what are we doing with this?

warrd:
@jamesgorrie this is ready for review, but maybe worth having a chat between me, you, @alicefuzier and @jamieparkinson about whether we think this is the correct way to model this, and the difficult question of how things are named.

kenoir:
But why?

warrd:
@kenoir some context [here](https://gist.github.com/warrd/0b1ee98e6e9bce47e0b236c5358a610f). We are going to have further discussion about this before merging.

jamesgorrie:
@kenoir @warrd I was looking at this again, and I wonder if it could maybe be paused and looked at with a greater context?

We use similar models everywhere, [including on our Concept model](https://github.com/wellcometrust/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentityState.scala), which have some errors in them (note the property `agent`) that we should clean up at the same time.

This is also based on the fact that we are starting to need to work model in other projects, and I would like to take the time to work with a more generic modelling serialiser / deserialiser that can autogenerate code so we are not fixed to using Scala ([see](https://gist.github.com/kenoir/eddab5afc256518d7096c5b7d7f5cd58)).

Saying all of this - it might make the work on the METS stuff easier and more understandable, so I would defer to @alicefuzier and @warrd working on that but with the foresight that we want to look at this modelling later.

warrd:
@jamesgorrie thanks, yep maybe would make sense to take a bit of a step back and look at things. I don't think I am completely understanding everything in the comment (like relation to the `Conept` model, and the stuff about generic serialiser / deserialiser and code generation), so might make sense to talk this through in person.

PR #214, opened 2019-10-28
[WIP] METS XML parsing

warrd:


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/MetsData.scala
```diff
@@ -0,0 +1,47 @@
...
+
+case class MetsData(
+  recordIdentifier: String,
+  accessCondition: Option[String]
+) {
+
+  def toWork(version: Int) =
+    // TODO: change this to UnidentifiedInvisibleWork (UnidentifiedPartialWork?)
+    UnidentifiedWork(
+      title = "??? SHOULD TITLE BE OPTIONAL ???",
```
alicefuzier: Hmm, good point... I think there might be places that rely on this not being optional, probably where we serialize into display models. So making this optional means deciding what should the API do if it finds a work with no title

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/MetsData.scala
```diff
@@ -0,0 +1,47 @@
...
+      mergeCandidates = List(mergeCandidate),
+      otherIdentifiers = Nil,
+    )
+
+  private def item: MaybeDisplayable[Item] =
+    Unidentifiable(
+      agent = Item(
+        locations = List(
+          DigitalLocation(
+            url = "???",
```
alicefuzier: I _think_ [this](https://github.com/wellcometrust/catalogue/blob/f293c62d38d1b6f7b3c3894c9f10fb2ecf1eb58f/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraLocation.scala#L32) is what the URL should look like, but better check with @jtweed

alicefuzier:
Closing as superseded by #259

PR #215, opened 2019-10-29
Brand name subjects should use Marc 652 subfields

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3997


jamesgorrie:
@warrd do you think it would help potentially having some recorded source data to run the tests against, using something similar to what we use in the METS project with `wiremock`?

PR #216, opened 2019-10-29
Merge dissertation, locationOfOriginal, citeAs fields into notes field

jamieparkinson:
Resolves #3996
Resolves #3998


warrd:
Looking at  this file [here](https://github.com/wellcometrust/catalogue/blob/5b8705c9dba377783675c77146e8c00a9847b6e2/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraNotes.scala), it appears like all subfields will be created as seperate notes,

Is that what we want, for all these fields?

jamesgorrie:
@jamieparkinson looks like we have an issue here:
https://travis-ci.org/wellcometrust/catalogue/jobs/604444535#L503

jamesgorrie:
On @warrd's comment - @jtweed 
```JS
[{
  "content": "Wellcome Library;",
  "type": "LocationOfOriginalNote"
},
{
  "content": "GB.",
  "type": "LocationOfOriginalNote"
}]
```

Would then be serialised as:
```JS
[{
  "contents": ["Wellcome Library;", "GB."],
  "noteType": { /*...*/ }
}]
```

This is the way we're treating all notes, we could however add a label which would be `Wellcome Library;GB.` for https://search.wellcomelibrary.org/iii/encore/record/C__Rb1879996?marcData=Y

jtweed:
I think we probably want to join subfields with spaces, but create multiple entries in `contents` for repeated fields?

PR #217, opened 2019-10-29
Add auth capabilities to mets transformer

alicefuzier:
As per discussion in https://github.com/wellcometrust/platform/issues/3921, this is going to be migrated to the mets adapter, but I think it's worth merging as is anyway and move it in another PR


pipeline/transformer/transformer_mets/src/test/resources/mappings/mapping-oauth2-token.json
```diff
@@ -0,0 +1,48 @@
+{
+  "id": "c4c51257-4385-4543-8559-50d46241ee46",
+  "request": {
+    "url": "/oauth2/token",
+    "method": "POST",
+    "headers": {
+      "Authorization": {
+        "equalTo": "Basic Y2xpZW50OnNlY3JldA=="
```
jamesgorrie: We don't care about this being there?

alicefuzier: it's `client:secret`. Not the real credentials

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/TokenService.scala
```diff
@@ -0,0 +1,52 @@
...
+    clientId = clientId,
+    clientSecret = secret,
+    site = URI.create(url),
+    tokenUrl = "/oauth2/token"
+  )
+
+  private val client = Client(config)
+  private val token = new AtomicReference[OAuth2BearerToken]
+
+  actorSystem.scheduler.schedule(initialDelay, interval)(refreshToken())
```
warrd: It appears that this method is deprecated with most recent version of akka:

https://github.com/akka/akka/blob/d4ed47f3091c48f708bfc4a4327c4933e4390e2b/akka-actor/src/main/scala/akka/actor/Scheduler.scala#L393-L400

alicefuzier: oh thanks, I'll change it

alicefuzier: Actually, annoyingly the version of akka in our classpath doesn't have the suggested methods. We should fix it once we update akka but I'm reluctant to do it in this PR

PR #218, opened 2019-10-29
Python script for diffing live APIs

warrd:
### Issue

https://github.com/wellcometrust/platform/issues/3968

### Description

Contains a python script which allows us to test how the responses from `prod` and `stage` APIs differ, which should help us switch more across confidently.

2019-10-29-172729_1164x1398_scrot

Note: It only really makes sense to run this when both are pointing at the same index.

### DIFF
--- a/api/live_testing/main.py
+++ b/api/live_testing/main.py
@@ -18,22 +18,27 @@ routes = [
     ("a22au6yn", {"include": "items,genres,identifiers"}),
 ]
 
+
 class ApiDiffer:
 
-    prod = 'api.wellcomecollection.org'
-    stage = 'api-stage.wellcomecollection.org'
+    prod = "api.wellcomecollection.org"
+    stage = "api-stage.wellcomecollection.org"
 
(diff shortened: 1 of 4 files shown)


api/live_testing/main.py
```diff
@@ -0,0 +1,157 @@
+#!/bin/env python3
+
+import requests
+import click
+import json
+
+routes = [
```
jamieparkinson: Is there any particular reason these routes were chosen to be tested? Can/should they be specified externally to the tool as well?

warrd: I chose them on the basis of trying to provide coverage of as much of our API as possible. I could probably have added a few more, but got bored :see_no_evil: 

I was thinking of providing them in a seperate json / yaml or something. Maybe worth it for a clean separating of code / configuration, but would require overhead of some parsing code.

warrd: @jamieparkinson now specified in `routes.json`

PR #225, opened 2019-10-30
Extract public URI in akka-http routing

warrd:
We need to set the correct `host` and `scheme`, and remove `port`, otherwise we get links which include  `https://localhost:8888/...`

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
@@ -88,13 +88,13 @@ class Router(elasticClient: ElasticClient,
       .map {
         case Left(err) => elasticError(err)
         case Right(resultList) =>
-          extractUri { uri =>
+          extractPublicUri { uri =>
             complete(
               MultipleWorksResponse(
                 resultList,
                 searchOptions,
                 params.include.getOrElse(V2WorksIncludes()),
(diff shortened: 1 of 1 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -221,6 +221,14 @@ class Router(elasticClient: ElasticClient,
       }
     }
 
+  def extractPublicUri =
+    extractUri.map { uri =>
+      uri
+        .withHost(apiConfig.host)
+        .withPort(0) // Port 0 is used here to indicate to akka no explicit port
+        .withScheme(apiConfig.scheme)
+    }
```
alexwlchan: Perhaps worth including the comment from the PR description explaining why we're doing this, and what the consequences of not doing it are? With examples of before/after URIs?

warrd: :+1: done

alexwlchan: What should the URIs look like after you've done this? Also I liked the comment about `withPort(0)` being a way to tell Akka not to set an explicit port.

warrd: Have updated, what about now?

PR #227, opened 2019-10-30
Allow reindexing of specific records

jamieparkinson:
Resolves https://github.com/wellcometrust/platform/issues/4003

### DIFF
--- a/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetter.scala
+++ b/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetter.scala
@@ -6,18 +6,26 @@ import com.amazonaws.services.dynamodbv2.document._
 import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
 
+/** Wraps DynamoDB in order to run a DynamoDB BatchGet for a given list of
+  * item IDs (= hash keys = partition keys) in a given table.
+  *
+  * https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html
+  *
+  * If an ID doesn't exist then no corresponding item is returned in the list.
+  * Returns up to 100 items or 16 MB of data (whichever is met first)
+  */
 class BatchItemGetter(dynamoDBClient: AmazonDynamoDB)(
   implicit ec: ExecutionContext) {
 
-  final val PARTITION_KEY: String = "id"
+  private final val PARTITION_KEY: String = "id"
   val dynamoDB = new DynamoDB(dynamoDBClient)
 
-  def get(recordIds: List[String])(tableName: String): Future[List[String]] =
+  def get(itemIds: List[String])(tableName: String): Future[List[String]] =
     for {
       batchGetResult: BatchGetItemOutcome <- Future {
         dynamoDB.batchGetItem(
           new TableKeysAndAttributes(tableName)
-            .addHashOnlyPrimaryKeys(PARTITION_KEY, recordIds: _*))
+            .addHashOnlyPrimaryKeys(PARTITION_KEY, itemIds: _*))
       }
       tableItems: List[Item] = batchGetResult.getTableItems.asScala
         .get(tableName)
--- a/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetter.scala
+++ b/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetter.scala
@@ -1,0 +1,28 @@
+package uk.ac.wellcome.platform.reindex.reindex_worker.dynamo
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.amazonaws.services.dynamodbv2.document._
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{ExecutionContext, Future}
+
+class BatchItemGetter(dynamoDBClient: AmazonDynamoDB)(
+  implicit ec: ExecutionContext) {
+
+  final val PARTITION_KEY: String = "id"
+  val dynamoDB = new DynamoDB(dynamoDBClient)
+
+  def get(recordIds: List[String])(tableName: String): Future[List[String]] =
+    for {
+      batchGetResult: BatchGetItemOutcome <- Future {
+        dynamoDB.batchGetItem(
+          new TableKeysAndAttributes(tableName)
+            .addHashOnlyPrimaryKeys(PARTITION_KEY, recordIds: _*))
+      }
+      tableItems: List[Item] = batchGetResult.getTableItems.asScala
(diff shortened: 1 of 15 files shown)


reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetter.scala
```diff
@@ -0,0 +1,28 @@
...
+import scala.collection.JavaConverters._
+import scala.concurrent.{ExecutionContext, Future}
+
+class BatchItemGetter(dynamoDBClient: AmazonDynamoDB)(
+  implicit ec: ExecutionContext) {
+
+  final val PARTITION_KEY: String = "id"
+  val dynamoDB = new DynamoDB(dynamoDBClient)
+
+  def get(recordIds: List[String])(tableName: String): Future[List[String]] =
```
alexwlchan: Would be good to get a brief comment on this class explaining what it does, and how it’s used.

Would be good to mention:

* What if I request a really big list of IDs?
* What if I request an ID that doesn't exist?

Naming nits:

* The class is called `BatchItemGetter` but you pass it a list of `recordIds`. Would it be better to pass a list of `itemIds`?
* Everywhere it’s used, it’s called `SpecificItemGetter`. Would that be neater?
* Where does the term `PARTITION_KEY` come from? In DynamoDB we have hash keys; I'm not familiar with that term.

reindexer/reindex_worker/src/test/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetterTest.scala
```diff
@@ -0,0 +1,63 @@
...
+      val specifiedRecord = records.head
+
+      val futureResult =
+        batchItemGetter.get(List(specifiedRecord.id))(table.name)
+
+      whenReady(convertScalaFuture(futureResult)) { result =>
+        convertToAnyShouldWrapper(
+          result.map { fromJson[NamedRecord](_).get }.head shouldBe specifiedRecord)
+      }
+    }
```
alexwlchan: I'd have written this as:

```scala
whenReady(futureResult) { result =>
  result.map { fromJson[NamedRecord](_).get }.head shouldBe specifiedRecord
}
```

Is there something I'm missing with the wrappers?

reindexer/reindex_worker/src/test/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetterTest.scala
```diff
@@ -0,0 +1,63 @@
...
+      createRecords(table, count = 5)
+
+      val futureResult =
+        batchItemGetter.get(List("bananas"))(table.name)
+
+      whenReady(convertScalaFuture(futureResult)) { result =>
+        convertToAnyShouldWrapper(result shouldBe empty)
+      }
+    }
+  }
```
alexwlchan: Some other test cases you could cover:

* What if I ask for a list of IDs, and some of them exist and some don't? (You could argue that returning the ones that exist and ignoring the rest is acceptable, or that errors should never pass silently so you should error out. I don't think it really matters, but a test to check which convention we're using would be helpful.)
* What if the table doesn't exist?

reindexer/start_reindex.py
```diff
@@ -147,7 +151,7 @@ def publish_messages(job_config_id, topic_arn, parameters):
 )
 @click.option(
     "--mode",
-    type=click.Choice(["complete", "partial"]),
+    type=click.Choice(["complete", "partial", "specific"]),
```
alexwlchan: Help text needs updating to match the choices.

reindexer/start_reindex.py
```diff
@@ -166,6 +170,15 @@ def start_reindex(src, dst, mode, reason):
     elif mode == "partial":
         max_records = click.prompt("How many records do you want to send?", default=10)
         parameters = partial_reindex_parameters(max_records)
+    elif mode == "specific":
+        specified_records_str = click.prompt(
+            "Which records do you want to reindex? (separate multiple IDs with spaces)",
+            type=str,
+        )
+        specified_records = specified_records_str.split()
+        if len(specified_records) is 0:
```
alexwlchan: Two notes:

* In Python, empty lists are false-y, so common convention is to check if a list is empty by doing:

    ```python
    if not my_list:
        print("my_list is empty!")
    ```

* `==` compares equality, `is` compares identity. Are these two objects equivalent, or are they at the same memory address?

  In general, prefer `==` unless you're sure you really want to check these are the same objects in-memory. Otherwise you can be lead astray:

    ```pycon
    >>> 1024 is 1024
    True

    >>> [] is []
    False
    >>> [] == []
    True

    >>> 10000 is 10000
    True
    >>> 10000 is int(10000.0)
    False
    >>> 10000 == int(10000.0)
    True
    ```

reindexer/reindex_worker/src/test/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetterTest.scala
jamieparkinson: lol, this is from me button mashing in IntelliJ to show/hide implicits - looks like I pressed something that actually put them in code explicitly!

reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetter.scala
jamieparkinson: Good points, will add documentation comment.

- yep good point re `recordIDs`
- went for `BatchItemGetter` to reflect the name of the DynamoDB operation (as with `ScanSpecScanner`) given that it directly wraps the client. Will probably leave this for now if that's OK with you.
- yeah, I've had this issue before, I have a feeling AWS changed their naming convention at some point: now they almost exclusively refer to partition keys (eg [here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.CoreComponents.html)) although some APIs call them hash keys 🤷‍♂

reindexer/reindex_worker/src/test/scala/uk/ac/wellcome/platform/reindex/reindex_worker/dynamo/BatchItemGetterTest.scala
jamesgorrie: There is definitely some work here around defining what is a Success and Failure of the service, one issue with this is we don't surface it at all - so it's pretty useless building those What ifs in.

jamieparkinson: Have left out the table existence test for both the above reason and because it's not included in existing/equivalent tests. Definitely worth revisiting if we reindex more frequently etc etc

kenoir:
General comment: this is an accomplished PR for a new dev on the code base 👍

PR #228, opened 2019-10-31
Binding information note

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/3995

## Description

Instead of appending Marc 563 to `physicalDescription` field, instead pull it out into a `binding-information` note, as we do with other 5XX fields.

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayNote.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayNote.scala
@@ -64,5 +64,7 @@ object DisplayNoteType {
         DisplayNoteType(
           "location-of-original-note",
           "Location of original note")
+      case BindingInformation(_) =>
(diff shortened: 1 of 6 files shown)


jamesgorrie:
@jtweed this is a 300 field and the frontend uses it - why are we putting it as a note?

warrd:
@jtweed that makes sense. Shall we use the name `BindingInformation` for the new note type, for consistency with the marc field naming?

jtweed:
Sounds good 👍

warrd:
@jtweed suggested in https://github.com/wellcometrust/catalogue/pull/216#issuecomment-547522683 that

> I think we probably want to join subfields with spaces, but create multiple entries in contents for repeated fields?

This doesn't appear to be done yet so gonna add another commit as part of this PR, if everyone thinks that's the correct strategy?

warrd:
Have done the above

PR #232, opened 2019-11-01
Aggregate date buckets dynamically

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/3977

## Description

Instead of returning an unbounded number of aggregation buckets when aggregating by date, here we dynamically group into progressively less granular buckets:

* Year
* Decade
* Half century
* Century

Note we have to do this in scala land rather than elasticsearch, as the broadest date aggregation it supports is year.


warrd:
@jamesgorrie any opinions on this? What bucket ranges do you think make sense, and max number of aggregation buckets returned?

jamesgorrie:
> max number of aggregation buckets returned

I still think returning all of the buckets make sense, or we will have to provide a way to paginate through them and indicate that to people using this otherwise we will be giving people a mis-representation of what we have in the catalogue, which is the opposite of what we're trying to achieve.

 > What bucket ranges do you think make sense

I'd say let's go ahead with decade, and get it into testing.

PR #233, opened 2019-11-04
Aggregation updates

warrd:
## Issues

https://github.com/wellcometrust/platform/issues/3973
https://github.com/wellcometrust/platform/issues/3974
https://github.com/wellcometrust/platform/issues/3975
https://github.com/wellcometrust/platform/issues/3976

## Description

* Don't just return first 10 aggregation buckets
* Order subject and genre aggregations by frequency

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElastsearchSearchRequestBuilder.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElastsearchSearchRequestBuilder.scala
@@ -34,49 +34,56 @@ case class ElastsearchSearchRequestBuilder(
 
   lazy val aggregations = queryOptions.aggregations.map {
     case AggregationRequest.WorkType =>
-      CompositeAggregation("workType").sources(
-        List(
-          TermsValueSource("label", field = Some("workType.label.raw")),
-          TermsValueSource("id", field = Some("workType.id")),
-          TermsValueSource("type", field = Some("workType.ontologyType"))
+      CompositeAggregation("workType")
+        .size(100)
+        .sources(
+          List(
+            TermsValueSource("label", field = Some("workType.label.raw")),
+            TermsValueSource("id", field = Some("workType.id")),
+            TermsValueSource("type", field = Some("workType.ontologyType"))
+          )
         )
-      )
+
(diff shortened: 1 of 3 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElastsearchSearchRequestBuilder.scala
```diff
@@ -34,49 +34,56 @@ case class ElastsearchSearchRequestBuilder(
...
+            TermsValueSource("label", field = Some("workType.label.raw")),
+            TermsValueSource("id", field = Some("workType.id")),
+            TermsValueSource("type", field = Some("workType.ontologyType"))
+          )
         )
-      )
+
     case AggregationRequest.ProductionDate =>
-      // We use `productionDates` here over `production.dates` to match the case classes, which we then serialise to
-      // the JSON path later.
```
warrd: :point_up: removed this comment as this class is just for building a request for elasticsearch, and beyond the scope of this class to know about how serialisation is performed further down the line.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElastsearchSearchRequestBuilder.scala
```diff
@@ -34,49 +34,58 @@ case class ElastsearchSearchRequestBuilder(
...
-      )
+        .subAggregations(sortedByCount)
 
     case AggregationRequest.Language =>
-      CompositeAggregation("language").sources(
-        List(
-          TermsValueSource("id", field = Some("language.id")),
-          TermsValueSource("label", field = Some("language.label.raw"))
+      CompositeAggregation("language")
+        .size(100)
```
jamieparkinson: I think there were something like 170 languages - iirc we want to display all of them (?)

warrd: :+1: I will update to 200 for now, but we probably want to investigate how the size of aggregations affect performance

jamesgorrie: FYI 173

PR #240, opened 2019-11-06
Filter works just by from date

warrd:
## Issues

https://github.com/wellcometrust/platform/issues/4027

## Description

Update filtering so that just uses `from` date. This is to preserve consistency between filtering / sorting / aggregation.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
@@ -6,15 +6,8 @@ sealed trait WorkFilter
 
 case class ItemLocationTypeFilter(locationTypeIds: Seq[String])
     extends WorkFilter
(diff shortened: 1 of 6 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
```diff
@@ -23,18 +16,7 @@ case class DateRangeFilter(fromDate: Option[LocalDate],
...
 
 case class SubjectFilter(subjectQuery: String) extends WorkFilter
-
-trait ApplyCommaSeperated[T] {
-
-  protected val fromSeq: Seq[String] => T
-
-  def apply(str: String): T =
-    fromSeq(str.split(',').map(_.trim))
-}
```
warrd: :point_up: not really needed to do all this since akka-http refactor, query parameters are parsed using circe

PR #243, opened 2019-11-07
Swagger docs fixes

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/4041

## Description

* Changes `ontologyType` outputs to `type`
* Change `MultipleWorksResponse` to `ResultList`
* Update swagger model resolver so that `List*` schemas are not  generated, and so that the code is simpler to understand.

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayWorkV2.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayWorkV2.scala
@@ -90,12 +90,7 @@ case class DisplayWorkV2(
     `type` = "Integer",
     description = "The playing time for audiovisual works, in seconds."
   ) duration: Option[Int] = None,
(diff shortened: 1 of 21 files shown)


common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayWorkV2.scala
```diff
@@ -90,12 +90,7 @@ case class DisplayWorkV2(
     `type` = "Integer",
     description = "The playing time for audiovisual works, in seconds."
   ) duration: Option[Int] = None,
-  @Schema(
-    readOnly = true,
-    description =
-      "A broad, top-level description of the form of a work: namely, whether it is a printed book, archive, painting, photograph, moving image, etc."
```
alexwlchan: 🤦‍♀ 

Nice catch.

PR #244, opened 2019-11-08
use HTTP for wellcome deps in S3

jamesgorrie:
It's annoying to have to keep commenting / uncommenting the hack, not sure why we don't just use HTTP?

I was going to setup a cloudfront distribution to make the URL, but it felt like it would be costing us money for not much real gain?

Couple of `terraform fmt` things slipped in - not sure why they weren't dealt with before.


warrd:
I don't believe this is working on Travis :crying_cat_face: 

This only builds when I revert the change (thanks @alicefuzier for the suggestion), which indicates maybe it is only working for the other builds as is loading libs from the cache?

jamesgorrie:
Hmm. Seems suspicious that it works else where, I'll put a PR back in and try figure it out.

warrd:
I have merged #248 which includes the reversion. Would be great to get this back in if possible :thinking:

PR #245, opened 2019-11-08
Fix swagger aggregation names

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/4041

## Description

Generates the correct names of aggregation models for Swagger.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
@@ -46,7 +46,8 @@ class SwaggerScalaModelConverter extends AbstractModelConverter(Json.mapper()) {
       if (cls == classOf[DisplayAggregation[_]])
         schema.name(s"${getAggregationClassName(annotatedType)}Aggregation")
       else if (cls == classOf[DisplayAggregationBucket[_]])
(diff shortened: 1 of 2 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala
```diff
@@ -56,6 +70,19 @@ class SwaggerScalaModelConverter extends AbstractModelConverter(Json.mapper()) {
...
     }
+
+  private def getAggregationClassName(annotatedType: AnnotatedType): String = {
+    val cls = getClass(getContentType(annotatedType))
+    if (cls == classOf[DisplayGenre]) "Genre"
+    else if (cls == classOf[DisplayWorkType]) "WorkType"
+    else if (cls == classOf[DisplayPeriod]) "Period"
+    else if (cls == classOf[DisplaySubject]) "Subject"
+    else if (cls == classOf[DisplayLanguage]) "Language"
+    else throw new IllegalArgumentException(s"Unknown class $cls")
```
jamesgorrie: I know generally explicit is better - but would a check on `statsWith("Display")` and just replacing that work?

warrd: Not just explicitness here, imo it much more robust working with the types themselves rather than strings (using strings will include other models that we don't want to include etc)

PR #246, opened 2019-11-08
Add query with scoring tiers

jamesgorrie:
Started with writing some intention & expectation documents, which made me realise, even though some things you might think are just "logical" you till need to find a way to codify that across a very interesting set of data (see ordering of concepts when searching).

I'd probably like to make these more tabular and structured.

I've purposefully left the queries without much more functionality as it seems like we get quite a lot from this, and I would prefer to understand exactly what we're asking of ES when adding things like `fuzziness` and `minimum_should_match` so that we can explain it and make sure it fulfils an expectation.

A super rudimentary experiment to see how much better even a simpler version is [is here](https://gist.github.com/jamesgorrie/0365f942c522d348eed8057e045699a8).

The basic idea is that instead of choosing precision over recall, which the `AND` op test is doing, you use scoring to bring those things to the top of the results.

i.e. Search for the title with the and, and if you match, boost the boots off of it 🚀, but still return other things that _might_ be related. 

We're adding 

__TODO__
- [x] Flattening out QueryType / WorkType / Query logic so we keep less state between the params and Elastic
- [x] ~Write tests for ScoringTiersQuery~ TODO after


kenoir:
The explanation and requirement for these PR isn't as clear to me as your explanation on Friday.

Would something like this make sense?

```
In order to improve search results for users, this change introduces "scoring tiers" 

Scoring tiers provide a differing query metric at each tier with the score provided by each tier for a particular query weighted in priority order.

For example:

- Tier 1: ...
- Tier 2: ...
- Tier 3: ...

```

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchQueryBuilder.scala
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  val baseQuery = MultiMatchQuery(
+    text = q,
+    fields = fields,
+    minimumShouldMatch = Some(defaultMSM),
+    `type` = Some(MultiMatchQueryBuilderType.CROSS_FIELDS)
+  )
+
+  lazy val elasticQuery = BoolQuery(
```
kenoir: The code here makes the functionality really clear - nice.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchQuery.scala
```diff
@@ -0,0 +1,11 @@
+package uk.ac.wellcome.platform.api.models
+
+case class SearchQuery(query: String, queryType: SearchQueryType)
+object SearchQuery {
+  def apply(query: String,
+            maybeQueryType: Option[SearchQueryType]): SearchQuery =
+    SearchQuery(query, maybeQueryType.getOrElse(SearchQueryType.default))
+
+  def apply(query: String): SearchQuery =
+    SearchQuery(query, None)
```
warrd: Is this overloaded method necessary? `maybeQueryType` could be set to default of `None` above

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchQueryBuilder.scala
```diff
@@ -0,0 +1,132 @@
...
+    ("data.edition", None),
+    // Identifiers
+    ("canonicalId", None),
+    ("sourceIdentifier.value", None),
+    ("data.otherIdentifiers.value", None),
+    ("data.items.canonicalId", None),
+    ("data.items.sourceIdentifier.value", None),
+    ("data.items.otherIdentifiers.value", None),
+  )
+}
```
warrd: :+1:

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchQuery.scala
warrd: Or could just use `SearchQueryType.default` as the default instead of an `Option[_]`

jamesgorrie:
> The explanation and requirement for these PR isn't as clear to me as your explanation on Friday.
> 
> Would something like this make sense?

@kenoir will move to the docs

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchQuery.scala
jamesgorrie: I overloaded this for the tests, but didn't want to make it available to the application. Will have another look to see what exactly I am trying to do here.

PR #247, opened 2019-11-11
Improved API deployment / infra

jamieparkinson:
Resolves wellcometrust/platform#3859
Resolves wellcometrust/platform#4043
Resolves wellcometrust/platform#4044


api/terraform/catalogue_api/main.tf
```diff
@@ -1,33 +1,61 @@
...
+  variables = {
+    port = "${var.listener_port}"
+  }
+
+  depends_on = "${var.gateway_depends}"
+}
+
+resource "aws_security_group" "service_egress_security_group" {
+  name        = "${var.namespace}-${var.environment}-service_egress_security_group"
+  description = "Allow traffic from service"
```
alexwlchan: "Allow traffic from service" to do what?

Most of these security group descriptions have been copy/pasted around and aren't actually accurate, and they're a PITA to change later. If we're changing things I'd like to make them useful now, or just ditch them.

api/terraform/main.tf
```diff
@@ -1,28 +1,59 @@
...
+  subnets = ["${local.private_subnets}"]
+
+  api_container_image = "${module.staging_images.services["api"]}"
+  listener_port       = "${local.staging_listener_port}"
+  task_desired_count  = "${local.staging_task_number}"
+
+  nginx_container_image = "${module.staging_images.services["nginx_api-gw"]}"
+
+  cluster_name = "${aws_ecs_cluster.cluster.name}"
+  es_config    = "${local.staging_es_config}"
```
alexwlchan: Very small thing: when writing prod/staging modules like this, I've been gradually trying to separate the config that is the same/different. So the stuff to change is all at the top, and the equal bits are at the bottom. Makes them easier to compare.

api/terraform/catalogue_api/main.tf
```diff
@@ -1,33 +1,48 @@
...
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  tags {
+    Name = "${var.namespace}-${var.environment}-egress"
```
jamesgorrie: This is a silly tag
Bikeshed - we should probably be tagging this stuff with `Service = api`

api/terraform/catalogue_api/service/main.tf
```diff
@@ -22,11 +22,15 @@ module "service" {
...
-  launch_type           = "FARGATE"
-  target_group_protocol = "TCP"
+  launch_type = "FARGATE"
 
   listener_port = "${var.listener_port}"
   lb_arn        = "${var.lb_arn}"
+
+  # The default deregistration delay is 5 minutes, which means that ECS takes around 5-7 mins
+  # to fully drain connections to and deregister the old task in the course of its blue/green
+  # deployment of an updated service. Reducing this parameter to 90s hence makes deployments faster.
```
jamesgorrie: 👍

PR #248, opened 2019-11-11
Skeleton METS adaptor

warrd:
This includes the (slightly modified) `BagRetriever` and  `TokenService` classes from the transformer, as it now beyond the scope of the transformer to call `storage-service`.

The `MetsAdapterWorkerService` is unfinished and includes stubs, another PR will finish off this  work and  add tests.


mets_adapter/src/main/scala/uk/ac/wellcome/mets/Main.scala
```diff
@@ -0,0 +1,46 @@
+package uk.ac.wellcome.mets
```
alicefuzier: Slightly being annoying here, but wouldn't it be better if the package contained the word adapter i.e. something like `uk.ac.wellcome.adapter.mets`? We probably don't do it in the sierra adapter cause it's made of so many bits and bobs but I don't feel `mets` is enough to namespace as there is a mets transformer as well

warrd: :+1: will rename to `mets_adaptor` so it is the same as the directory name. I think using `adaptor` would be confusing unless we do for our other adaptors too.

alicefuzier: Sounds good to me 👍

mets_adapter/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+SBT_APPS = mets_adapter
+
+SBT_DOCKER_LIBRARIES =
+SBT_NO_DOCKER_LIBRARIES =
+
+PYTHON_APPS =
+LAMBDAS	=
+
+TF_NAME = mets_adapter
```
alicefuzier: Is there going to be a mets_adapter terraform stack?

warrd: I would expect so? This was just copied from another `Makefile`  so happy to remove this for now

mets_adapter/src/main/scala/uk/ac/wellcome/mets/Main.scala
warrd: Just realised forgot to do this, doing now

mets_adapter/Makefile
alicefuzier: Nope, was just wondering where the infrastucture is going to live... It's fine to keep it if it's going to be useful

mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/IngestUpdate.scala
```diff
@@ -0,0 +1,18 @@
...
+case class StubBag(info: BagInfo) extends BaseBag
+case class Bag(info: BagInfo, manifest: BagManifest, location: BagLocation)
+    extends BaseBag
+case class BagInfo(externalIdentifier: String)
+
+case class BagManifest(files: List[BagFile])
+
+case class BagLocation(bucket: String, path: String)
+
+case class BagFile(name: String, path: String)
```
jamesgorrie: Should these be shared between the adapter and transformer?

mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/Main.scala
```diff
@@ -0,0 +1,45 @@
...
+    )
+  }
+
+  private def buildBagRetriever(config: Config)(
+    implicit
+    actorSystem: ActorSystem,
+    materializer: ActorMaterializer,
+    ec: ExecutionContext): BagRetriever =
+    new BagRetriever(
+      "URL??",
```
jamesgorrie: Assuming this is because it's a stub?

mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/TokenService.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  def getToken: Future[OAuth2BearerToken] = {
+    val maybeToken = Option(token.get())
+    maybeToken match {
+      case Some(token) => Future.successful(token)
+      case None        => refreshToken()
+    }
+  }
+
+}
```
jamesgorrie: Is there not an oath service that can just do this for us and we we could reply on their tests?
https://index.scala-lang.org/dakatsuka/akka-http-oauth2-client/akka-http-oauth2-client/0.2.0?target=_2.12?

mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/IngestUpdate.scala
alicefuzier: I don't think the transformer should know anything about Bags, it should only understand METS XML

mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/TokenService.scala
alicefuzier: That's what the token service uses internally. The token service is taking care of updating the token in a thread safe way and making it available to other threads

warrd:
This build is [failing](https://travis-ci.org/wellcometrust/catalogue/jobs/611485747?utm_medium=notification&utm_source=github_status) as can't resolve internal dependencies:

```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: uk.ac.wellcome#messaging_2.12;5.3.1: not found
[warn] 	:: uk.ac.wellcome#messaging_typesafe_2.12;5.3.1: not found
[warn] 	:: uk.ac.wellcome#monitoring_2.12;2.3.0: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

A bit unsure what else needs to be done (works fine locally), any ideas?

PR #249, opened 2019-11-11
Use {id} not :id in Swagger path parameters

kenoir:
Path parameters should be curl brackets and not colons, see: https://swagger.io/docs/specification/paths-and-operations/

This change allow generated clients to function as expected.

For https://github.com/wellcomecollection/stacks-service


kenoir:
@warrd @jamesgorrie are we in a position to release this directly to prod?

warrd:
@kenoir yep you can release directly, `prod` is currently up to date with master so this would be the only thing going in

PR #251, opened 2019-11-11
return empty notes if you've asked for them

jamesgorrie:
With lists we return empty if it is empty, otherwise clients permanently have to null check.


warrd:
This appears to have broken the tests
https://travis-ci.org/wellcometrust/catalogue/jobs/610442393?utm_medium=notification

Always worth waiting for all tests to complete before merging :wink:

warrd:
https://github.com/wellcometrust/catalogue/pull/252

PR #252, opened 2019-11-12
Fix snapshot tests

warrd:


jamesgorrie:
Apologies - got tired of build time - solution is to fix build times, not ignore builds.

jamieparkinson:
We should probably think about why we're fixing tests on a service that hasn't worked since February https://github.com/wellcometrust/platform/issues/3446

PR #253, opened 2019-11-12
Make InvisibleWork carry data and title optional

alicefuzier:
So that the mets transformer can output invisible works with items and merge candidates

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayWorkV2.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayWorkV2.scala
@@ -16,7 +16,6 @@ case class DisplayWorkV2(
     readOnly = true,
     description = "The canonical identifier given to a thing.") id: String,
   @Schema(
-    `type` = "String",
     description =
       "The title or other short label of a work, including labels not present in the actual work or item but applied by the cataloguer for the purposes of search or description."
   ) title: Option[String],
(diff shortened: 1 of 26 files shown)


common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayWorkV2.scala
```diff
@@ -16,9 +16,10 @@ case class DisplayWorkV2(
     readOnly = true,
     description = "The canonical identifier given to a thing.") id: String,
   @Schema(
+    `type` = "String",
```
warrd: I don't think we need explicitly pass this here, it is derived automatically in the current `swagger.json` as far as I can see

alicefuzier: I remember we needed to do this when we wrap things into `Option` because it shows a weird type otherwise. But I haven't checked swagger output admittedly. If it doesn't anymore then, all the better

warrd: It shouldn't be necessary anymore since update to v3 (there is custom `Option[_]` handling [here](https://github.com/wellcometrust/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerScalaModelConverter.scala#L38-L39))

alicefuzier: Nice! thanks

PR #255, opened 2019-11-12
add identity cognito

jamesgorrie:
See tin for details.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -110,7 +110,8 @@ object ExternalDependencies {
     val scribeJava = "6.8.1"
   }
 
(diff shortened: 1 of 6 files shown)


project/Dependencies.scala
```diff
@@ -110,7 +110,8 @@ object ExternalDependencies {
     val scribeJava = "6.8.1"
   }
 
-  val scribeJavaDependencies = Seq("com.github.dakatsuka" %% "akka-http-oauth2-client" % "0.2.0")
+  val scribeJavaDependencies = Seq(
+    "com.github.dakatsuka" %% "akka-http-oauth2-client" % "0.2.0")
```
jamesgorrie: @alicefuzier do we know how this got here? Git annotate says it was you?

PR #257, opened 2019-11-13
Matcher does not ignore invisible works

alicefuzier:
Part of https://github.com/wellcometrust/platform/issues/3921
It won't have any effect until we start sending mets through the pipeline, but it's a simple enough change and it has simplified the code a bit, so I figured it was worth doing it

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
@@ -11,7 +11,9 @@ sealed trait IdentifiedBaseWork extends BaseWork {
   val canonicalId: String
 }
 
-sealed trait TransformedBaseWork extends BaseWork
+sealed trait TransformedBaseWork extends BaseWork {
+  val data: WorkData[MaybeDisplayable]
+}
 
 sealed trait InvisibleWork extends BaseWork
 
--- a/pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/matcher/WorkMatcherTest.scala
+++ b/pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/matcher/WorkMatcherTest.scala
@@ -10,7 +10,12 @@ import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.mockito.MockitoSugar
 import org.scalatest.{EitherValues, FunSpec, Matchers}
 import org.scanamo.syntax._
-import uk.ac.wellcome.models.matcher.{MatchedIdentifiers, MatcherResult, WorkIdentifier, WorkNode}
+import uk.ac.wellcome.models.matcher.{
+  MatchedIdentifiers,
+  MatcherResult,
+  WorkIdentifier,
+  WorkNode
+}
 import uk.ac.wellcome.models.work.generators.WorksGenerators
 import uk.ac.wellcome.models.work.internal.MergeCandidate
 import uk.ac.wellcome.platform.matcher.exceptions.MatcherException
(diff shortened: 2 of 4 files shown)


pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/matcher/WorkMatcherTest.scala
```diff
@@ -75,7 +75,9 @@ class WorkMatcherTest
               matcherResult shouldBe
                 MatcherResult(
                   Set(MatchedIdentifiers(Set(WorkIdentifier(workId, 1)))))
-              get[WorkNode](dynamoClient, graphTable.name)('id -> workId) shouldBe None
+              get[WorkNode](dynamoClient, graphTable.name)('id -> workId)
+                .map(_.right.get) shouldBe Some(
```
alexwlchan: FWIW, we've been starting to use EitherValues and `_.right.value` instead of `_.right.get` in tests while you were away; one of Robert's discoveries, I think.

I don't see a massive difference, but you might want to consider it: http://www.scalatest.org/user_guide/using_EitherValues

alicefuzier: Ah nice thanks

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -11,7 +11,9 @@ sealed trait IdentifiedBaseWork extends BaseWork {
   val canonicalId: String
 }
 
-sealed trait TransformedBaseWork extends BaseWork
+sealed trait TransformedBaseWork extends BaseWork {
+  val data: WorkData[MaybeDisplayable]
+}
```
warrd: :+1:

pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/matcher/WorkMatcherTest.scala
```diff
@@ -64,7 +64,7 @@ class WorkMatcherTest
     }
   }
 
-  it("doesn't store an invisible work and sends the work id") {
+  it("stores an invisible work and sends the work id") {
```
warrd: :thinking: do we  want to store all invisible works? What are the repercussions  of that?

alicefuzier: I think we have to. The matcher stores the nodes of the work graph in the database and then it calculates the connected components on that graph.
All works created by the mets transformer will be invisible works. For the matcher to be able to group them to the physical sierra work it has to store them as all the other works. 
We will end up storing all the works that are currently invicible for other reasons, but I don't think this will have a major impact. 
The merger won't link them and the output in the api won't change.
Performance and price wise I don't think we might end up spending a bit more because we increase operations but I don't think it will be much

warrd: :thinking: Just a thought, do you think it would make sense for there to be another type separate to `InvisibleWork`? Me and @jamesgorrie discussed at one point whether `PartialWork` or something like that would be a better description for this use case.

Beyond performance / price, the thing I am more worried about is introducing subtle bugs some how (as we are potentially using invisible to represent two quite separate things).

alicefuzier: I'm not sure it's a separate use case though. Non mets Invisible works are still partial and they are invisible because they don't have enough information to be shown in the api. 
I'm not sure whether adding another type would simplify our life or complicate it

warrd: :+1: ok let's just stick with this for now, we can always split off later

PR #259, opened 2019-11-13
Mets transformer pulls mets from S3 and creates invisible work

alicefuzier:
Part of https://github.com/wellcometrust/platform/issues/3921
Supersedes #214 
Resolves Transform METS files to works


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/MetsData.scala
```diff
@@ -0,0 +1,45 @@
+package uk.ac.wellcome.platform.transformer.mets
+
+case class MetsData(
+  recordIdentifier: String,
+  accessCondition: Option[String]
+) {
+
+  def toWork(version: Int) = ???
```
warrd: Didn't know you could do that!

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/Mets.scala
```diff
@@ -0,0 +1,62 @@
...
+        sourceIdentifier = sourceIdentifier,
+        workData(unidentifiableItem)
+      )
+
+  private def workData(unidentifiableItem: MaybeDisplayable[Item]) =
+    WorkData(
+      items = List(unidentifiableItem),
+      mergeCandidates = List(mergeCandidate))
+
+  private def mergeCandidate = MergeCandidate(
```
jamesgorrie: Does it make a difference if things are `val`s or `def`s?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -0,0 +1,50 @@
...
+      )
+  }
+
+  private def recordIdentifier(root: Elem): Try[String] = {
+    val identifierNodes =
+      (root \\ "dmdSec" \ "mdWrap" \\ "recordInfo" \ "recordIdentifier").toList
+    identifierNodes match {
+      case List(identifierNode) => Success(identifierNode.text)
+      case _ =>
+        Failure(new Exception("Could not parse recordIdentifier from METS XML"))
```
jamesgorrie: When these errors are throw in `process` - they won't have any identifiable information with them will there?

Could we add some? I can't remember what we get from the adapter - it's the S3 path?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/Mets.scala
warrd: In these sort of cases when no mutability / side effects:

`val` = Non lazy, stored in memory
`lazy val` = Lazy, stored in memory
`def` = Lazy, not stored in memory (recalculated each time needed)

Stylistically I quite like `def` here but don't think it matters much

pipeline/transformer/transformer_mets/src/test/scala/uk/ac/wellcome/platform/transformer/mets/fixtures/MetsGenerators.scala
```diff
@@ -0,0 +1,23 @@
...
+       |          <mods:recordInfo>
+       |            <mods:recordIdentifier source="gbv-ppn">$recordIdentifier</mods:recordIdentifier>
+       |          </mods:recordInfo>
+       |          <mods:accessCondition type="dz">${license.id.toUpperCase}</mods:accessCondition>
+       |        </mods:mods>
+       |      </mets:xmlData>
+       |    </mets:mdWrap>
+       |  </mets:dmdSec>
+       |</mets:mets>
+       |""".stripMargin
```
warrd: Scala has XML literals, so could use that here. Going to be removed at some time in the future in Scala 3 AFAIK.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  private def accessCondition(root: Elem): Try[Option[String]] = {
+    val licenseNodes = (root \\ "dmdSec" \ "mdWrap" \\ "accessCondition")
+      .filter(_ \@ "type" == "dz")
+      .toList
+    licenseNodes match {
+      case Nil               => Success(None)
+      case List(licenseNode) => Success(Some(licenseNode.text))
+      case _ =>
+        Failure(new Exception("Found multiple accessCondtions in METS XML"))
```
warrd: In most of the codebase it seems like `Either[_, _]` are used in preference to `Try`. Do you reckon worth being consistent? (at some point might be great to have a style guide for these sort of decisions)

common/internal_model/src/main/resources/identifier-schemes.csv
```diff
@@ -14,3 +14,4 @@ preservica-guid,Perservica GUID
 calm-authority-code,CALM authority code
 isbn,International Standard Book Number
 issn,ISSN
+mets,METS
```
warrd: I remember some talk about whether adding a new identifier type was the correct thing. Has @jtweed got any opinion on this?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/Mets.scala
jamesgorrie: I suppose my concern would be that if it's a `def` it feels like we've said `There's a reason why this recalculates each time` is all.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
jamesgorrie: > have a style guide for these sort of decisions

👍 

This function is actually getting it - so we could use `getAccessCondition` which will either get an optional license or a known error state of `multiple access conditions in METS`.

This error feels like it would be good to have somewhere associated with the b number so if asked, "why isn't this b number getting an access statement" we could point to our imaginary Success / Failure dashboard and it will be there.

Saying this - we might have (excuse the verbosity)
```scala
case class MetsAdapterFailure
case class KnownFailure extends MetsAdapterFailure
case class UnknownFailure extends MetsAdapterFailure // Probably similar to Try.Failure(err)
case class MultipleAccessConditionsInMets extends KnownFailure
def getAccessCondition(root: Elem): Either[MultipleAccessConditionsInMets, Option[AccessCondition]]
```

And then at the edge of the application you would have
```
// MetsAdapter.scala
// ...
def process: Either[MetsAdapter, UnidentifiableWork]

// Main.scala
// ...
val id = getId(input)
process(input) match {
  case Right(work) => complete(work) // continue
  case Left(err: KnownFailure) => info(err) // continue
  case Left(err: UnknownFailure) => error(err) // retry
}
```

There's some states I am missing and logic that doesn't fit right (`UnknownFailure`?).

In short I think I am saying if we know what the problem is, is it an error, or part of the application state that we know how to handle, which `Either` feels better at semantically capturing?

If this makes sense in trying to capture I can have a go at implementing it on one of the services?

Thanks to the [Scala docs](https://www.scala-lang.org/api/2.11.4/index.html#scala.util.Try) for providing clear guidance.

> The Try type represents a computation that may either result in an exception, or return a successfully computed value. It's similar to, but semantically different from the scala.util.Either type.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/Mets.scala
alicefuzier: No, no reason. I tried replacing with val and it still works. The only thing is that, if they are vals, they need to be in a specific order so that each one is initialised when it's referenced by another one which is slightly annoying

common/internal_model/src/main/resources/identifier-schemes.csv
alicefuzier: We discussed this with @jamesgorrie and I think we sort of reached the conclusion that just adding this here was OK?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
alicefuzier: Yep, I'm going to switch to `Either`.

As for the Known vs Unknown failure thing, I'm not sure... There is already a [`RecognisedFailure`](https://github.com/wellcometrust/scala-messaging/blob/master/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/RecognisedFailure.scala) class which is basically used in [`SQSStream`](https://github.com/wellcometrust/scala-messaging/blob/master/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala) to log and send metrics differently, so we can extend that if we just want to change logging.

I wouldn't base the logic around retries on whether an error is known or unknown though. I would say that anything that needs action from us and that we want to be notified about should end up in the dlq eventually,  so it shouldn't be deleted from the queue, which means it should be retried. 
I.e. the multiple access conditions in mets error: it's a data error so a retry won't change the outcome, but we make it retry so that we can see it in the dlq. We can then notify an error in the data to whoever is responsible for fixing it.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
alicefuzier: How about 

> ERROR u.a.w.p.t.m.s.MetsTransformerWorkerService - There was an error processing MetsData(mets.xml,6): blah blah blah..

PR #261, opened 2019-11-13
METS adapter stages

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/4059

## Description

Includes all stages of the METS adapter:

* Retrieving the bag
* Parsing the bag
* Filter out METS data stored in dynamo (unless newer version / file path changed)
* Publishing message to SNS
* Storing METS data in dynamo


mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsStore.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+class MetsStore(store: VersionedStore[String, Int, MetsData]) {
+
+  def storeMetsData(bagId: String, data: MetsData): Either[Throwable, Option[MetsData]] =
+    shouldUpdate(bagId, data)
+      .right.flatMap {
+        case true => store.putLatest(bagId)(data).right.map(_ => Some(data))
+        case false => Right(None)
+      }
+      .left.map(_.e)
```
warrd: I have thought of an obscure failure which could occur. If we successfully store the `MetsData`, but then encounter an error when publishing the message to SNS, the message will go back on the queue. However, the next time round when we call `shouldUpdate` it will look like we have already dealt with this data, so we don't bother storing it and publishing the message (leading to the data being lost and never sent to the transformer).

The only ways I can think of a way to getting around this apart are:
* We still publish the message when we receive another one containing the same version
* Introduce some complicated transaction logic so the dynamo table is rolled back if publishing the message to SNS fails (pretty sure the storage libs don't support such a thing out of the box)

:thinking: any thoughts @alicefuzier ?

warrd: Another option:

* First check whether the data is stored within dynamo, then publish to SNS, then actually store the data in dynamo: this could lead to the same message being published to SNS multiiple times storing in dynamo fails, but I assume this is preferable to having data never make it into the pipeline.

warrd: I have implemented the bottom option above :point_up:

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsAdapterWorkerService.scala
```diff
@@ -14,43 +15,116 @@ import uk.ac.wellcome.mets_adapter.models._
...
+          .via(retrieveBag)
+          .via(parseMetsData)
+          .via(filterMetsData)
+          .via(publishMetsData)
+          .via(storeMetsData)
+          .map { case (Context(msg, _), _) => msg }
+      }
+    )
+
+  def retrieveBag =
```
jamesgorrie: Maybe it's me just getting familiar with flows, but given they receive a type from the previous flow, could we add return types here?

Or do you think the `via` methods above imply that enough?

Whatever is more readable, I might just need to get more familiar with the syntax!

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsAdapterWorkerService.scala
```diff
@@ -14,43 +15,116 @@ import uk.ac.wellcome.mets_adapter.models._
...
-  def publishMetsLocation
-    : Flow[(SQSMessage, MetsLocation), SQSMessage, NotUsed] =
-    Flow[(SQSMessage, MetsLocation)]
+  /** Allows mapping a flow with a function, where:
+    *  - Context is passed through.
+    *  - None values are ignored and passed to the next stage: this is required
+    *    for the SQS delete action to trigger at the end of the stream
+    *  - Any errors are caught and the message prevented from propagating downstream,
+    *    resulting in the message being put back on the queue / on the dlq.
+    */
```
jamesgorrie: Do you think this is relevant here?
https://github.com/wellcometrust/catalogue/pull/259#discussion_r348358717

jamesgorrie: These comments are superb!

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsAdapterWorkerService.scala
warrd: I had included the full type signature for the flows, but at some point decided to remove them as a bit of a mouthful so added a lot of noise.

Without explicit types:

```scala
def parseMetsData =
  Flow[(Context, Option[Bag])]
  ...
```

With explicit types:

```scala
def parseMetsData: Flow[(Context, Option[Bag]), (Context, Option[MetsData]), NotUsed] =
  Flow[(Context, Option[Bag])]
  ...
```

I personally think it's pretty readable without, as it is easy to see the output from the input to the next flow, but happy to be explicit if people would prefer.

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/Main.scala
```diff
@@ -25,8 +31,9 @@ object Main extends WellcomeTypesafeApp {
 
     new MetsAdapterWorkerService(
       SQSBuilder.buildSQSStream(config),
-      SNSBuilder.buildSNSMessageSender(config, subject = "?"),
+      SNSBuilder.buildSNSMessageSender(config, subject = ???),
```
alicefuzier: I'm assuming the config will be added in a different PR?

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/BagRetriever.scala
```diff
@@ -9,18 +9,23 @@ import akka.stream.ActorMaterializer
 import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
 import grizzled.slf4j.Logging
 import io.circe.generic.auto._
-import uk.ac.wellcome.mets_adapter.models.Bag
+import uk.ac.wellcome.mets_adapter.models._
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class BagRetriever(url: String, tokenService: TokenService)(
+trait BagRetriever {
```
alicefuzier: I think I'm missing something here, but what's the reason for having a trait and a class implementing it?

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/IngestUpdate.scala
```diff
@@ -1,14 +1,43 @@
 package uk.ac.wellcome.mets_adapter.models
 
-case class IngestUpdate(space: Space, bag: StubBag)
+case class IngestUpdate(space: String, bagId: String)
```
alicefuzier: Have we agreed with @kenoir or @alexwlchan that the storage service is going to provide updates in this form? 
As far as I know, the storage service exposes two models: `Ingest` and `Bag`. The `Ingest` format is a superset of the `IngestUpdate` class with the `StubBag` as it was before. So with the slightly more convoluted `StubBag` thingy, the adapter would be able to read `Ingest`s, meaning that the storage service wouldn't need to send information in a format specific to the adapter (I should have written this somewhere, sorry). 
It might be worth checking if this matters

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsStore.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import uk.ac.wellcome.mets_adapter.models._
+import uk.ac.wellcome.storage.store.VersionedStore
+import uk.ac.wellcome.storage.{Identified, NoVersionExistsError}
+
+/** Stores METS data. The data should first be filtered to ensure storing is
+  *  performed only if it is a newer version and the location is pointing to a
+  *  different XML file.
+  */
+class MetsStore(store: VersionedStore[String, Int, MetsData]) {
```
alicefuzier: Stupid question: is this a full VHS or just dynamo?

warrd: Assumed was just a dynamo, as what we are storing is pretty small at the moment (the `MetsData` case class).

Ideally the types would be set up in such a way so we could pass a VHS in here as the argument if we wanted, with the `MetsStore` being agnostic to the exact implementation details of whether VHS or not, but the VHS currently would require a`VersionedStore[String, Int, HybridStoreEntry[MetsData, EmptyMetadata]]`, so that wouldn't be possible at the moment.

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/BagRetriever.scala
warrd: I did this for 2 (related) reasons:

* For unit testing, it gets messy to always have to mock HTTP calls: so for example when testing the `MetsAdapterService` we don't care about the details of how the `BagRetriever` works, just that it is something that returns bags. By having it as a trait we easily [create a dummy bag retriever](https://github.com/wellcometrust/catalogue/pull/261/files#diff-f84d662997f5ad022e6813a899bdee93R41-R44) without having to resort to mocking frameworks etc

* Beyond just testing I think it's better to program against interfaces rather than specific implementations (I'm sure you understand this, but to state the obvious it makes it easier to swap things out in future, and also helps clearly separate any implementation details from the core functionality of a class)

warrd:
Thanks for the reviews! Merging this now, we can finalise the message format and  add configuration in a future PR.

PR #266, opened 2019-11-15
mets_adapter terraform stack

warrd:
Just ECR for now.

### DIFF
--- a/mets_adapter/Makefile
+++ b/mets_adapter/Makefile
@@ -11,7 +11,7 @@ SBT_NO_DOCKER_LIBRARIES =
 PYTHON_APPS =
 LAMBDAS =
 
-TF_NAME =
-TF_PATH =
+TF_NAME = mets_adapter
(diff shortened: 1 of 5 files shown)


jamieparkinson:
Perhaps, given this is a new stack, it would be worth trying to do it with TF 0.12? Haven't checked compatibility of the existing ECR module but Alex's work might be a good reference

jamesgorrie:
[The ECR module does nothing](https://github.com/wellcometrust/terraform-modules/tree/master/ecr).

The issue with putting this into `tf0.12` is that when it comes to adding the service (and everything that comes with ECS) you will have to keep using `12`, and thus convert those modules.

Not saying don't do it, but be aware of the work that awaits.

alicefuzier:
Have we agreed a strategy to migrate to 12?

warrd:
:+1: reckon going to stick to 11 because of the above, and the fact that I'll have to install 2 versions of Terraform on my machine or have to use docker to run `terraform apply`.

PR #270, opened 2019-11-20
Use simple string query with AND operator for genre / subject

warrd:
https://github.com/wellcometrust/platform/issues/3971


warrd:
We are testing for `AND` operator: the previous tests were for `OR`, and they have been updated for the  new behaviour.

PR #272, opened 2019-11-20
update takedown with new accounts and work model

jamesgorrie:
I'm not too fussed about how annoying this is as I think we should have a discussion about what takedown is given the conversations we've had with D&I and collections.

Mainly around what is takedown? How do we store that we're taking things down? And how do we surface to the public that we have made that decision (and context for that decision).

### DIFF
--- a/scripts/remove_work/run.py
+++ b/scripts/remove_work/run.py
@@ -21,13 +21,11 @@ def sha256(bs):
     h.update(bs)
     return h.hexdigest()
 
+
 def aws_client(service_name, role_arn):
     sts = boto3.client("sts")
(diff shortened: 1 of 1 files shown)


jamieparkinson:
It seems totally reasonable to me that there would be different "severities" of takedown, ranging from complete obliteration, to (publicly explained) obfuscation of images or other fields, to stuff that is still accessible but via a warning or something.

This is definitely not something we can or should answer/solve alone; who should we talk to about the best practices for this?

jamesgorrie:
@jamieparkinson [yes](https://github.com/wellcometrust/platform/issues/4063)

PR #281, opened 2019-11-25
Make mets adapter read NotificationMessage

alicefuzier:


mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsAdapterWorkerService.scala
```diff
@@ -56,6 +55,10 @@ class MetsAdapterWorkerService(
       className,
       source => {
         source
+          .map {
+            case (msg, NotificationMessage(body)) =>
+              (msg, fromJson[IngestUpdate](body).get)
```
warrd: Just wondering if this is desirable behaviour in the case where it cannot parse the message.

We could do this to go through the error handling code:

```scala
.map {
   case (msg, NotificationMessage(body)) =>
    (msg, fromJson[IngestUpdate](body).toEither)
}
.via(catchErrors)
```

warrd: Actually would need to a bit of modification so doesn't just accept a `Context` object..

alicefuzier: You're definitely right about the error handling. I'm not sure I understand the second bit about the `Context` object

warrd: I mean  the `catchErrors` method  would have  to look something like this instead:

```scala
  def catchErrors[T, C] =
    Flow[(C, Result[T])]
      .map {
        case (ctx, result) =>
          result.left.map { err =>
            error(
              s"Error encountered processing SQS message. [Error]: ${err.getMessage} [Context]: $ctx",
              err)
          }
          (ctx, result)
      }
      .collect {
        case (ctx, Right(data)) => (ctx, data)
      }

```

alicefuzier: thanks, makes sense

mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsAdapterWorkerService.scala
```diff
@@ -66,6 +65,15 @@ class MetsAdapterWorkerService(
...
     )
 
+  def unwrapMessage =
+    Flow[(SQSMessage, NotificationMessage)]
+      .map {
+        case (msg, NotificationMessage(body)) =>
+          (msg, fromJson[IngestUpdate](body).toEither)
+      }
+      .via(catchErrors)
+      .map { case (msg, update) => (Context(msg, update.bagId), update) }
```
warrd: :+1:

PR #283, opened 2019-11-26
License filtering and aggregation

warrd:
## Issues

https://github.com/wellcometrust/platform/issues/3927
https://github.com/wellcometrust/platform/issues/3928

## Descriptions

Adds filtering and aggregation for digital item license.


api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
```diff
@@ -189,6 +189,12 @@ trait MultipleWorksSwagger {
         description = "Filter the work by subject.",
         required = false
       ),
+      new Parameter(
+        name = "license",
+        in = ParameterIn.QUERY,
+        description = "Filter the work by license.",
+        required = false
```
jamesgorrie: Could we add valid values here?

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorksGenerators.scala
```diff
@@ -246,4 +246,13 @@ trait WorksGenerators extends ItemsGenerators with ProductionEventGenerators {
       canonicalId = canonicalId,
       production = List(createProductionEventWith(dateLabel = Some(dateLabel)))
     )
+
+  def createLicensedWork(canonicalId: String,
+                         licenses: License*): IdentifiedWork =
```
jamesgorrie: @kenoir was asking about these.

From what I can read it's saying that we generally have 1, so the syntax `createLicensedWork("id", LICENSE)` is most common.

But we support more than 1.

But why not just make it a list? Is it to avoid having `createLicensedWork("id", List(LICENSE))` where there is only 1 license?

api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
warrd: Done

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorksGenerators.scala
warrd: No reason really probably just to make it terser, but both signatures are fine to me so have changed it to accept `List`.

PR #289, opened 2019-11-27
Add APM

jamieparkinson:
See https://github.com/wellcometrust/platform/issues/4058


api/api/src/main/scala/uk/ac/wellcome/platform/api/Tracing.scala
```diff
@@ -0,0 +1,76 @@
...
+        "application_packages" -> "uk.ac.wellcome",
+        "service_name" -> config.getOrElse("apm.service.name")("catalogue-api"),
+        "server_urls" -> config.getOrElse("apm.server.url")(
+          "http://localhost:9200"),
+        "secret_token" -> config.getOrElse[String]("apm.secret")("")
+      ).asJava)
+  }
+}
+
+trait AsyncTracing {
```
kenoir: Yarp, nice.

project/Common.scala
```diff
@@ -21,6 +21,6 @@ object Common {
       "-Xcheckinit"
     ),
     parallelExecution in Test := false,
-    libraryDependencies ++= CatalogueDependencies.commmonDependencies
```
kenoir: mmm

api/api/src/main/scala/uk/ac/wellcome/platform/api/Tracing.scala
```diff
@@ -0,0 +1,76 @@
...
+                    spanType: String = "",
+                    subType: String = "",
+                    action: String = "")(wrappedFunction: => Future[T])(
+    implicit ec: ExecutionContext): Future[T] = {
+    val span =
+      ElasticApm
+        .currentTransaction()
+        .startSpan(spanType, subType, action)
+        .setName(name)
+    try {
```
warrd: Minor stylistic things:

I generally prefer using `Try(...).recover(...)` in scala to this more imperative way of handling exceptions.

Also this could potentially return a `Try` / `Either` to make it clear to the caller of the function that it can error, and forces them to handle it explicitly in some way (whether that is just propagating it up the stack or otherwise). Not sure whether returning a `Try[Future[_]]` makes sense to you here though, so feel free to ignore.

jamieparkinson: Agreed re the first point, re the 2nd this method is intended to return the exact thing the wrapped function returns (having snuck in some tracing) - but that does make me realise that I shouldn't be rethrowing the error!

jamieparkinson: Realised it could be made lots nicer in the course of doing this

PR #290, opened 2019-11-28
add default to _queryType param

jamesgorrie:
We need to be able to allow invalid value through as the client might have an older version of the paramters that are valid, and we wouldn't want to error.

This is a transient every changing value that isn't documented so this hopefully is expected behaviour.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
@@ -88,7 +88,7 @@ case class MultipleWorksParams(
       license
     ).flatten
 
-  private def dateFilter =
+  private def dateFilter: Option[DateRangeFilter] =
     (`production.dates.from`, `production.dates.to`) match {
       case (None, None)       => None
       case (dateFrom, dateTo) => Some(DateRangeFilter(dateFrom, dateTo))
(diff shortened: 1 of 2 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -219,6 +220,14 @@ trait QueryParamsUtils extends Directives with TimeInstances {
         .getOrElse(Left(invalidValuesMsg(List(str), values.map(_._1).toList)))
     }
 
+  def decodeOneWithDefaultOf[T](default: T, values: (String, T)*): Decoder[T] =
```
jamieparkinson: I think this could be written more succinctly as:
```
Decoder.decodeString.map { values.toMap.getOrElse(_, default) }
```
since we are always giving `emap` a `Right`

warrd: Maybe the function signature should be like this, so you can explicitly request the default value also:

```scala
def decodeOneWithDefaultOf[T](default: (String, T), otherValues: (String, T)*): Decoder[T]
```

jamesgorrie: @jamieparkinson 👍 

@warrd we don't actually have a string for requesting the default. None or wrong = default in this case.

warrd: I think that might confuse me as a user of an API, where you cannot specify the default explicitly. I know this is not a public API, but maybe worth reconsidering in the case that this functionality ends up being used elsewhere.

PR #295, opened 2019-11-28
Populate METS adapter script

warrd:
https://github.com/wellcometrust/platform/issues/4083

### DIFF
--- a/mets_adapter/populate_mets/populate_mets.py
+++ b/mets_adapter/populate_mets/populate_mets.py
@@ -1,0 +1,76 @@
+#!/bin/env python3
+
+from sys import maxsize
+import click
+import boto3
+from wellcome_aws_utils.sns_utils import publish_sns_message
+
+
+def aws_resource(name, role_arn):
+    role = boto3.client("sts").assume_role(
+        RoleArn=role_arn, RoleSessionName="AssumeRoleSession1"
+    )
+    credentials = role["Credentials"]
+    return boto3.resource(
+        name,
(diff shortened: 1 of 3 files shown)


mets_adapter/populate_mets/populate_mets.py
```diff
@@ -0,0 +1,76 @@
...
+    def scan(self):
+        for page in self.paginator.paginate(TableName=self.vhs):
+            for item in page["Items"]:
+                space, id = item["id"].split(":")
+                yield (space, id)
+
+
+class MessagePublisher:
+
+    role = "arn:aws:iam::760097843905:role/platform-dev"
```
warrd: I can't assume this role for some reason when running locally, anyone got an idea why (`read_only` roles appear to be fine) :thinking:?

```
botocore.exceptions.ClientError:
An error occurred (AccessDenied) when calling the AssumeRole operation: User:
arn:aws:sts::760097843905:assumed-role/platform-superdev/Ni.Ward@wellcome.ac.uk
is not authorized to perform: sts:AssumeRole on resource:
arn:aws:iam::760097843905:role/platform-dev
```

alicefuzier: the role arn is `arn:aws:iam::760097843905:role/platform-developer` I think

warrd: That worked! But there is also a `platform-dev` role in the IAM, which is maybe a bit confusing

PR #297, opened 2019-11-29
Make WorkType an Enum

alicefuzier:


jamesgorrie:
This is looking promising!

jamesgorrie:
This is looking promising!

project/Dependencies.scala
```diff
@@ -109,7 +109,11 @@ object ExternalDependencies {
     val logstashLogback = "6.1"
     val scribeJava = "6.8.1"
     val apm = "1.11.0"
+    val enumeratumVersion = "1.5.13"
   }
+  val enumeratumDependencies = Seq(
+    "com.beachape" %% "enumeratum" % versions.enumeratumVersion
```
warrd: Looking at the docs for this [here](https://github.com/lloydmeta/enumeratum).. It's not clear to me what this gives on top of using plain old sealed traits and case objects

e.g.

```scala
sealed trait WorkType

object WorkType {
  case object Audio extends WorkType
  case object EBook extends WorkType
  ...
}
```

All of the things it mentions (dependencies, all compile time, performance etc) are just as good (or even better) without this, is there sometime I am missing?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/WorkType.scala
```diff
@@ -1,5 +1,160 @@
...
+  def fromCode(id: String): Option[WorkType] = {
+    values.find(workType => workType.id == id)
+  }
+
+  sealed trait UnlinkedWorkType extends WorkType
+  sealed trait LinkedWorkType extends WorkType {
+    val linksTo: UnlinkedWorkType
+  }
+
+  case object BooksWorkType extends UnlinkedWorkType {
```
warrd: As this is already within the `WorkType` object (and thus already namespaced), in my opinion it is easier to read if we don't include "WorkType" again in the name. e.g. 

`WorkType.Books` vs `WorkType.BooksWorkType`

project/Dependencies.scala
alicefuzier: It precomputes the possible `values`. So that there doesn't need to be an enumeration of all the objects somewhere to be able to write the `fromCode` function (or at least I don't know how to do that with objects and sealed traits)

warrd: :+1: that makes sense

jamesgorrie: This is a great bit of functionality for JSON deserialising!

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/WorkType.scala
```diff
@@ -1,5 +1,160 @@
...
+      } yield {
+        fromCode(id).getOrElse(throw new Exception(s"Invalid WorkType id: $id"))
+      }
+  }
+
+  def fromCode(id: String): Option[WorkType] = {
+    values.find(workType => workType.id == id)
+  }
+
+  sealed trait Unlinked extends WorkType
```
kenoir: So this is both an Enum _and_ an ADT?

warrd: Just speculating here, but it's likely IMO that the classic concept of an enum was only invented due to lack of ADT support in certain languages, since ADTs give the same functionality but in more type safe manner (i.e. with pattern matching). Worth mentioning that an `enum` [in Rust](https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html) provides pretty much the same thing as `sealed trait` / `case class` in Scala.

PR #298, opened 2019-12-02
METS thumbnail

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/3945

## Description

Parse the thumbnail location from the METS XML.

NOTE: We are using the first item in the physical struct map to extract a thumbnail location. I am not sure that this is the correct thing to do, and it may be more correct to additionally make use of the logical struct map. However it is not clear to me from looking the the METS structure [here](https://github.com/wellcometrust/catalogue/blob/master/pipeline/transformer/transformer_mets/src/test/resources/b30246039.xml) how exactly that would work..

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
@@ -1,9 +1,10 @@
 package uk.ac.wellcome.platform.transformer.mets.parsers
 
-import uk.ac.wellcome.platform.transformer.mets.transformer.Mets
-
+import scala.collection.immutable.ListMap
 import scala.util.Try
 import scala.xml.{Elem, XML}
+
+import uk.ac.wellcome.platform.transformer.mets.transformer.Mets
 
 object MetsXmlParser {
 
@@ -17,12 +18,11 @@ object MetsXmlParser {
     for {
       id <- recordIdentifier(root)
       maybeAccessCondition <- accessCondition(root)
-      thumbnailUrl <- thumbnailUrl(root)
     } yield
       Mets(
         recordIdentifier = id,
         accessCondition = maybeAccessCondition,
-        thumbnailUrl = thumbnailUrl,
+        thumbnailLocation = thumbnailLocation(root),
       )
   }
 
@@ -51,6 +51,46 @@ object MetsXmlParser {
     }
   }
 
-  private def thumbnailUrl(root: Elem): Either[Exception, Option[String]] =
-    Right(None)
+  private def thumbnailLocation(root: Elem): Option[String] =
+    physicalStructMap(root)
+      .headOption
+      .flatMap { case (_, fileId) =>
+        fileObjects(root).get(fileId)
+      }
+      .map(_.stripPrefix("objects/"))
+
+  private def fileObjects(root: Elem): Map[String, String] =
+    (root \ "fileSec" \ "fileGrp")
+      .toList
+      .filter(_ \@ "USE" == "OBJECTS")
+      .flatMap(_ \ "file")
+      .map { node =>
+        val id = node \@ "ID"
+        val location = (node  \ "FLocat")
+          .toList
+          .headOption
+          .map(_ \@ "{http://www.w3.org/1999/xlink}href")
+        (id, location)
+      }
+      .collect {
+        case (id, Some(location)) if location.nonEmpty => (id, location)
+      }
+      .toMap
+
+  private def physicalStructMap(root: Elem): ListMap[String, String] = {
+    val physicalMappings =
+      (root \ "structMap")
+        .toList
+        .filter(_ \@ "TYPE" == "PHYSICAL")
+        .flatMap(_ \\ "div")
+        .map { node =>
+          val id = node \@ "ID"
+          val fileId = (node \ "fptr").toList.headOption.map(_ \@ "FILEID")
+          (id, fileId)
+        }
+        .collect {
+          case (id, Some(fileId)) if fileId.nonEmpty => (id, fileId)
+        }
+    ListMap(physicalMappings:_*)
+  }
 }
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
@@ -17,10 +17,12 @@ object MetsXmlParser {
     for {
       id <- recordIdentifier(root)
       maybeAccessCondition <- accessCondition(root)
+      thumbnailUrl <- thumbnailUrl(root)
     } yield
       Mets(
         recordIdentifier = id,
         accessCondition = maybeAccessCondition,
+        thumbnailUrl = thumbnailUrl,
       )
   }
 
@@ -48,4 +50,7 @@ object MetsXmlParser {
           new Exception("Found multiple accessCondtions in METS XML"))
     }
   }
+
+  private def thumbnailUrl(root: Elem): Either[Exception, Option[String]] =
+    Right(None)
 }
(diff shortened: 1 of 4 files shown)


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -48,4 +50,96 @@ object MetsXmlParser {
...
+    (root \ "fileSec" \ "fileGrp")
+      .filterByAttribute("USE", "OBJECTS")
+      .childrenWithTag("file")
+      .toMapping(
+        keyAttrib = "ID",
+        valueNode = "FLocat",
+        valueAttrib = "{http://www.w3.org/1999/xlink}href"
+      )
+
+  /** Valid METS documents should a physicalStructMap section, with the bottom
```
alexwlchan: nit: “should a” is missing a verb.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -48,4 +50,96 @@ object MetsXmlParser {
...
+    */
+  private def physicalStructMap(root: Elem): ListMap[String, String] =
+    (root \ "structMap")
+      .filterByAttribute("TYPE", "PHYSICAL")
+      .descendentsWithTag("div")
+      .toMapping(
+        keyAttrib = "ID",
+        valueNode = "fptr",
+        valueAttrib = "FILEID"
+      )
```
alexwlchan: It’s not clear to me which of the `mets:ftpr` tags this will retrieve. Do you get:

```scala
Map("PHYS_0001" -> "FILE_0001_OBJECTS", "PHYS_0002" -> "FILE_0002_OBJECTS")
```

or

```scala
Map("PHYS_0001" -> "FILE_0001_ALTO", "PHYS_0002" -> "FILE_0002_ALTO")
```

And does it matter which? What if these attributes appear in a different order?

I had a look at the docs ListMap; it wasn't clear to me what it's getting beyond `Map[String, String]`. What am I missing?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -21,6 +22,7 @@ object MetsXmlParser {
       Mets(
         recordIdentifier = id,
```
alexwlchan: It’s not entirely in scope for this PR, but could you go back and add these nice comments to `recordIdentifier` and `accessCondition` soon please? :)

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -48,4 +50,96 @@ object MetsXmlParser {
...
+    *      <mets:file ID="FILE_0002_OBJECTS" MIMETYPE="image/jp2">
+    *        <mets:FLocat LOCTYPE="URL" xlink:href="objects/b30246039_0002.jp2" />
+    *      </mets:file>
+    *    </mets:fileGrp>
+    *  </mets:fileSec>
+    *
+    *  We need to look up the physicalStructMap to ascertain what each of these
+    *  files refer to.
+    */
+  private def fileObjects(root: Elem): Map[String, String] =
```
alexwlchan: So I think this would give me a mapping

```scala
Map("FILE_0001_OBJECTS" -> "objects/b30246039_0001.jp2", "FILE_0002_OBJECTS" -> "objects/b30246039_0002.jp2")
```

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
warrd: `ListMap` is an implementation of `Map` which preserves ordering: relevant as we want to retrieve the first defined item in the `thumbnailLocation` method above.

We get the first `fptr` node (line 135 in the `toMapping` method).

I will add some comments to make things more clear

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
warrd: Yes correct :+1:

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
warrd: > And does it matter which? What if these attributes appear in a different order?

Good question. It appears like the first one is what we want, but I don't know that we can guarantee that is always correct. Unfortunately, there doesn't seem to be other information which we can use (the only attribute being the ID). I suppose we could choose the one which exists in the `OBJECTS` map, which seems a bit crazy to me as indicates recipricol relationships between each of these structures, but maybe that is how it is.

[This METS specification](https://www.loc.gov/standards/mets/profiles/00000014.html) says the following:

> Each div element representing a page of the item(s) to be displayed should have one or more child fptr elements. One fptr element should appear for each of the different-sized images in the fileSec, referencing the ID of for the page described in the parent div in the FILEID attribute. Other fptr children may be present; however, METS Navigator will not use these in its display or functionality.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/Mets.scala
```diff
@@ -59,4 +62,14 @@ case class Mets(
       ontologyType = "Work",
       value = recordIdentifier)
   }
+
+  private val thumbnailHeight = "200"
+
+  private def thumbnail =
+    thumbnailLocation.map { location =>
+      DigitalLocation(
+        s"https://dlcs.io/iiif-img/wellcome/5/$location/full/,$thumbnailHeight/0/default.jpg",
```
alicefuzier: I thought we said we were going to use the boxed syntax?
ie `https://dlcs.io/iiif-img/wellcome/5/$location/full/!$thumbnailWeight,$thumbnailHeight/0/default.jpg`

pipeline/transformer/transformer_mets/src/test/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsTest.scala
```diff
@@ -87,4 +87,19 @@ class MetsTest extends FunSpec with RandomStrings with Matchers {
...
+    val metsData = Mets(
+      recordIdentifier = randomAlphanumeric(10),
+      accessCondition = Some("CC-BY-NC"),
+      thumbnailLocation = Some("location.png")
+    )
+    val result = metsData.toWork(1)
+    result shouldBe a[Right[_, _]]
+    result.right.get.data.thumbnail shouldBe Some(
+      DigitalLocation(
+        s"https://dlcs.io/iiif-img/wellcome/5/location.png/full/,200/0/default.jpg",
```
alicefuzier: Does this test pass now?

warrd: Probably not :see_no_evil:

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -48,4 +50,103 @@ object MetsXmlParser {
...
+    *      <mets:file ID="FILE_0002_OBJECTS" MIMETYPE="image/jp2">
+    *        <mets:FLocat LOCTYPE="URL" xlink:href="objects/b30246039_0002.jp2" />
+    *      </mets:file>
+    *    </mets:fileGrp>
+    *  </mets:fileSec>
+    *
+    *  For this input we would expect the following output:
+    *
+    *  Map("FILE_0001_OBJECTS" -> "objects/b30246039_0001.jp2",
+    *      "FILE_0002_OBJECTS" -> "objects/b30246039_0002.jp2")
```
alexwlchan: 👍 nice

PR #299, opened 2019-12-03
Fix sierra physical/digital merger rule

alicefuzier:
For https://github.com/wellcometrust/platform/issues/4081

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/physicaldigital/SierraPhysicalDigitalPartitioner.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/physicaldigital/SierraPhysicalDigitalPartitioner.scala
@@ -1,10 +1,6 @@
 package uk.ac.wellcome.platform.merger.rules.physicaldigital
 
-import uk.ac.wellcome.models.work.internal.{
-  BaseWork,
-  IdentifierType,
-  UnidentifiedWork
-}
+import uk.ac.wellcome.models.work.internal.{BaseWork, DigitalLocation, Identifiable, IdentifierType, Item, PhysicalLocation, Unidentifiable, UnidentifiedWork}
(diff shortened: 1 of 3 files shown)


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/physicaldigital/SierraPhysicalDigitalPartitioner.scala
```diff
@@ -21,14 +17,20 @@ trait SierraPhysicalDigitalPartitioner extends WorkTagPartitioner {
...
-      case Some(t) => t.id == "v" && t.label == "E-books"
+    work.data.items match {
+      case List(Unidentifiable(Item(List(_:DigitalLocation), _))) => true
+      case _ => false
+    }
+
+  private def isPhysicalWork(work: UnidentifiedWork): Boolean =
+    work.data.items match {
+      case List(Identifiable(Item(List(_:PhysicalLocation), _), _, _,_)) => true
+      case _ => false
```
warrd: What happens in cases where there is more than one item? It seems like `isDigitalWork` and `isPhysicalWork` would both be `false`, is this by design?

alicefuzier: Yes. In the use case where we definitely want to merge we will have one item in each. This will make the merger more conservative but it is intentional, to avoid catching any of the other use cases where we might have a digitital item but we don't want to merge.

warrd: That makes sense. It might be worth having a comment explaining this in the file, especially as the logic for this is split across both the partioner and the merger rule.

PR #300, opened 2019-12-04
write first test of scoringTiers query

jamesgorrie:
I'd like to write some more, but would also like to get this out so we can start testing it as the reporting on the queries are more important than the tests on fake data being sort of right.

This though does make sure it isn't broken, which is great.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchQueryBuilder.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchQueryBuilder.scala
@@ -42,8 +42,8 @@ object QueryDefaults {
     // TODO: (jamesgorrie) think of a more sustainable way of doing this
     // maybe having a just a list of terms that we use terms queries to query against,
     // and then have more structured data underlying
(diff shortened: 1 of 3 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchQueryBuilder.scala
```diff
@@ -42,8 +42,8 @@ object QueryDefaults {
     // TODO: (jamesgorrie) think of a more sustainable way of doing this
     // maybe having a just a list of terms that we use terms queries to query against,
     // and then have more structured data underlying
-    ("data.subjects.*", Some(8.0)),
-    ("data.genres.label", Some(8.0)),
+    ("subjects.agent.concepts.agent.label", Some(8.0)),
+    ("genres.concepts.agent.label", Some(8.0)),
```
warrd: Is removing `data` a mistake?

jamesgorrie: Thank you - added a test to make sure we do actually search, and rank highly, these fields.

PR #301, opened 2019-12-04
Use `ORDER` attribute for retrieving first physical thumbnail

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/3945

## Description

If present we should  use the `ORDER` attribute in the physical `structMap` to retrieve the first thumbnail.

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
@@ -153,6 +153,7 @@ object MetsXmlParser {
     (root \ "structMap")
       .filterByAttribute("TYPE", "PHYSICAL")
       .descendentsWithTag("div")
+      .sortByAttribute("ORDER")
       .toMapping(
         keyAttrib = "ID",
         valueNode = "fptr",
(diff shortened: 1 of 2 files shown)


jamesgorrie:
How is the order applied to the XML in the first place?

warrd:
> How is the order applied to the XML in the first place?

As far as I know it is added automatically in Goobi

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -37,6 +54,24 @@ object MetsXmlParser {
...
+    *          <mods:accessCondition type="player">63</mods:accessCondition>
+    *          <mods:accessCondition type="status">Open</mods:accessCondition>
+    *          ...
+    *        </mods:mods>
+    *      </mets:xmlData>
+    *    </mets:mdWrap>
+    *  </mets:dmdSec>
+    *
+    *  The expected output would be: "CC-BY-NC"
+    */
```
alexwlchan: Thanks for adding these!

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/parsers/MetsXmlParser.scala
```diff
@@ -118,6 +153,7 @@ object MetsXmlParser {
     (root \ "structMap")
       .filterByAttribute("TYPE", "PHYSICAL")
       .descendentsWithTag("div")
+      .sortByAttribute("ORDER")
```
alexwlchan: Maybe worth a comment about how ListMap preserves ordering, otherwise good.

PR #307, opened 2019-12-09
Fix APM threading issue

jamieparkinson:
### Context

APM auto-instruments a variety of libraries and frameworks: unfortunately, Akka HTTP is not one of them, so I used the [public API](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html) that APM provides in order to instrument the catalogue API. Unfortunately, APM has not been designed with Scala in mind and some standard Scala patterns present some issues when used with the public API.

APM has been logging out 2 main types of warning:
```
Tried to start a transaction with a non-zero reference count
```
and
```
Deactivating a span which is not the currently active span
```
The first of these is indicative of transactions not being properly deactivated, and thus being reintroduced to APM's internal transaction pool whilst retaining references, and the second in some senses foreshadows the former in that the span being deactivated may not be the correct one to deactivate.

As well as these warnings, I noticed that only about 20% of transactions included the `elasticTook` label - all of this points towards an issue with obtaining the current (activated) transaction and ending/deactivating/augmenting it.

### The root problem
 
APM stores activated transactions and spans in a [`ThreadLocal` pool](https://github.com/elastic/apm-agent-java/blob/c0d561735997b528e7f0bd964e27c66562f86a05/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java#L103), which is why it [requires](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-transaction-activate) that `Scope`s are closed on the same thread on which they are started.

Scala's `Future`s are run in a [`ForkJoinPool`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html) to which new `Future`s are enqueued: they can either execute on the thread from which they were created, a new thread that is forked by the pool, or be 'stolen' by an existing thread that has become available.

`map`ping on `Future`s creates new ones which, at low volume, are likely to be run on the same thread thus preserving the integrity of APM's `ThreadLocal`s. However, at high request volume, the forking and work-stealing is much more likely to occur (due to the asynchronicity of requests), and so _we cannot guarantee that the `map`ping function is run in the same thread as the `Future` which it maps_.

This results in closing `Scope`s on new threads, and referring to "current" transactions which are either null-transactions or are unrelated, causing the problems described above.

### The fix
This PR introduces a custom Akka dispatcher to be used within the `Tracing` trait. The trait's companion object contains a `DynamicVariable[WeakReference[Transaction]]` which the custom dispatcher copies across from the current thread when executing a new `Runnable` (ie a `Future`). This is inspired by [this post](http://stevenskelton.ca/threadlocal-variables-scala-futures/) and [this post](http://yanns.github.io/blog/2014/05/04/slf4j-mapped-diagnostic-context-mdc-with-play-framework/).

This singleton `DynamicVariable` is private, and a getter/setter are exposed to interact with the underlying `currentTransaction` transparently. The `WeakReference` ensures that APM controls the memory in which the `Transaction` lives (we never truly acquire the object) and so prevents the possibility of memory leaks.

Additionally, a new wrapper function `withActiveTrace` is provided in order to `activate` the `currentTransaction` for cases where we want APM's auto-instrumentation to kick in: for us, that's for it to instrument the calls to the Elastic REST API.

In order to test this fix, this PR also provides a test tool `stress_testing` which uses [Artillery](https://artillery.io/) to make a high volume of requests to the API, thus hoping to reproduce the warnings/bugs which had previously been invisible in testing.

Resolves https://github.com/wellcometrust/platform/issues/4098


api/api/src/main/scala/uk/ac/wellcome/platform/api/Tracing.scala
```diff
@@ -19,20 +30,41 @@ object Tracing {
...
+    actorSystem.dispatchers.registerConfigurator(
+      "tracing-dispatcher",
+      new TraceableDispatcherConfigurator(
+        actorSystem.dispatchers.defaultDispatcherConfig,
+        actorSystem.dispatchers.prerequisites))
+    ec = actorSystem.dispatchers.lookup("tracing-dispatcher")
+  }
+
+  implicit var ec: ExecutionContext =
+    scala.concurrent.ExecutionContext.Implicits.global
```
alexwlchan: I don't think we use the global execution context everywhere; usually it gets passed in as an implicit alongside `actorSystem`. Is there a reason we're using it here?

api/stress_testing/README.md
```diff
@@ -0,0 +1,10 @@
+# Stress testing script
+
+This script runs `/works` and `/works/{id}` queries with a selection of query strings and work IDs, at a rate that is similar to a high-ish level of usage of the API.
```
alexwlchan: How were the work IDs selected? Is it random, or are they meant to be some “interesting” cross-sample of the data (e.g. a Miro record, a Sierra record, lots of items, no items)?

jamieparkinson: Similar to the diff tool - just a human-random selection!

alexwlchan: Can you write that in this README please? :)

api/api/src/main/scala/uk/ac/wellcome/platform/api/Tracing.scala
jamieparkinson: Yeah, this was a bit of a pain - it's here so the tests work, in reality it gets mutated by `init` to use the execution context from the actor system (ie the default dispatcher). It is true that if you were to forget to call `init`, this would be "wrong", so if you have any thoughts on how to do it better please say!

alexwlchan: I don't, but please write that in the code 😉

PR #309, opened 2019-12-09
Allow METS transformer to read directly from storage service S3 bucket

alicefuzier:


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/MetsTransformerWorkerService.scala
```diff
@@ -1,24 +1,30 @@
...
 
 import scala.concurrent.Future
 
 class MetsTransformerWorkerService(
   msgStream: SQSStream[NotificationMessage],
   messageSender: BigMessageSender[SNSConfig, TransformedBaseWork],
-  store: VersionedStore[String, Int, HybridStoreEntry[String, EmptyMetadata]])
-    extends Runnable
+  adapterStore: VersionedStore[String, Int, MetsData],
+  mestStoreAssumeRoleClient: AssumeRoleClientProvider[AmazonS3]
```
warrd: In my opinion it would be better for this code to just receive a METS xml store as an argument, rather than receiving the special client and creating the store itself: in this way it is more isolated from implementation details (the fact that it is stored in S3 and needs a special client that assumes the role etc), making the code easier to understand, more easily refactored etc

alicefuzier: I agree. It's just that, with the credentials being temporary, we need to get a client with every request. But I'm going to extract that bit of functionality away so we don't deal with clients here. 

I wrote the simplest thing that could possibly work because I'm still trying to figure out how to test this thing. Moto, as always, doesn't seem to work

warrd: Makes sense :+1:

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/client/AssumeRoleClientProvider.scala
```diff
@@ -0,0 +1,46 @@
...
+  roleArn: String,
+  interval: FiniteDuration = 30 seconds)(clientFactory: ClientFactory[T])(
+  implicit actorSystem: ActorSystem,
+  ec: ExecutionContext) {
+  val client = new AtomicReference[T]
+
+  actorSystem.scheduler.schedule(0 milliseconds, interval)(refreshClient())
+
+  def getClient: Either[Throwable, T] = {
+    Option(client.get()) match {
```
kenoir: Does this throw if the client has expired?

alicefuzier: it would throw a permission error I assume... 
Unfortunately, moto does not allow to test this. I can't find a way to recreate conditions like this (or, what happens if you try to assume a role that you don't have permissions to assume etc)

warrd: Is it possible to override the client class, with the method in question throwing an error? Or more complicated than that?

alicefuzier: Actually, no, that wouldn't throw any error. It returns `null` if the `AtomicReference` is not initialised yet (which reminds me `client` should be private, I'll fix that). But because it's wrapped in `Option` it will always be `None` or `Some`.
If the client has expired, it would only show when we actually try to use it I think, so in `TemporaryCredentialsStore`

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/MetsTransformerWorkerService.scala
```diff
@@ -40,19 +43,23 @@ class MetsTransformerWorkerService(
...
+      metsString <- getFromMetsStore(metsData)
       mets <- MetsXmlParser(metsString)
       work <- mets.toWork(key.version)
       _ <- messageSender.sendT(work).toEither
     } yield ()
   }
 
-  private def fetchMets(key: Version[String, Int]) = {
-    store.get(key) match {
+  private def getKey(key: Version[String, Int]) = {
```
warrd: The naming of this is a bit confusing to me, as it returns `MetsData` and not a key to any table. Can we change this to `getMetsData` or something like that? And maybe include return type to be extra explicit.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/MetsTransformerWorkerService.scala
```diff
@@ -2,23 +2,26 @@ package uk.ac.wellcome.platform.transformer.mets.service
...
 
 import scala.concurrent.Future
 
 class MetsTransformerWorkerService(
   msgStream: SQSStream[NotificationMessage],
   messageSender: BigMessageSender[SNSConfig, TransformedBaseWork],
-  store: VersionedStore[String, Int, HybridStoreEntry[String, EmptyMetadata]])
-    extends Runnable
+  adapterStore: VersionedStore[String, Int, MetsData],
+  temporaryCredentialsStore: TemporaryCredentialsStore[String]
```
warrd: I think it would be more descriptive to name this variable according to what is stored inside it (e.g. `metsXmlStore`), rather than named after how the store accesses the data, as the data is the primary concern for this class rather than the implementation.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/client/AssumeRoleClientProvider.scala
```diff
@@ -0,0 +1,46 @@
...
+          .withRoleSessionName("transformer"))
+      val temporaryCredentials = new BasicSessionCredentials(
+        assumeRoleResult.getCredentials.getAccessKeyId,
+        assumeRoleResult.getCredentials.getSecretAccessKey,
+        assumeRoleResult.getCredentials.getSessionToken);
+
+      val t = clientFactory.buildClient(temporaryCredentials)
+      client.set(t)
+      t
+    }.toEither
```
warrd: :+1: nice work on the abstraction of all this

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/store/TemporaryCredentialsStore.scala
```diff
@@ -0,0 +1,21 @@
+package uk.ac.wellcome.platform.transformer.mets.store
+
+import com.amazonaws.services.s3.AmazonS3
+import uk.ac.wellcome.platform.transformer.mets.client.AssumeRoleClientProvider
+import uk.ac.wellcome.storage.ObjectLocation
+import uk.ac.wellcome.storage.store.s3.S3TypedStore
+import uk.ac.wellcome.storage.streaming.Codec
+
+class TemporaryCredentialsStore[T](
+  assumeRoleClientProvider: AssumeRoleClientProvider[AmazonS3])(
```
warrd: Maybe this could just be a `ClientFactory[AmazonS3]` to abstract this away further? Would mean the knowledge of credentials being temporary were not exposed here, and within `MetsTransformerWorkerService`

warrd: Oops misunderstood the type hierarchy, so I don't  think that is possible

alicefuzier: How would it abstract further though? if It only takes a `ClientFactory` it will need to instantiate a `AssumeRoleClientProvider` itself, but to do that it will need the `stsClient` and the `roleArn` as additional parameters. It feels like more stuff about assuming the role would be leaking into this class rather than being isolated in  `AssumeRoleClientProvider`.
But perhaps I'm not getting what you mean?

warrd: Yup I misread it, ignore me

PR #310, opened 2019-12-10
Title field on item model

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/4101

## Description

As part of the METS work we are bringing across more data for individual items that make up a work. Here we include a new `title` field on `Item`, which is populated from the Sierra data. This will be needed to properly display the individual items on the frontend (similarly to how items are displayed on the [library site](http://search.wellcomelibrary.org/iii/encore/record/C__Rb2975324?lang=eng))

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItems.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItems.scala
@@ -8,13 +8,15 @@ import uk.ac.wellcome.models.transformable.sierra.{
 import uk.ac.wellcome.models.work.internal._
 import uk.ac.wellcome.platform.transformer.sierra.source.{
   SierraBibData,
-  SierraItemData
+  SierraItemData,
+  SierraQueryOps
 }
 
 case class SierraItems(itemDataMap: Map[SierraItemNumber, SierraItemData])
     extends SierraTransformer
     with Logging
-    with SierraLocation {
+    with SierraLocation
+    with SierraQueryOps {
 
   type Output = List[MaybeDisplayable[Item]]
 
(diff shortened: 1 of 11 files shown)


jamesgorrie:
@warrd where are we getting this from, and where is it being used?

warrd:
@jamesgorrie still working on this, the sierra transformer will populate it. Will add a description when to the PR when ready.

jamesgorrie:
Thanks @warrd, sorry to jump the 🔫

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItems.scala
```diff
@@ -114,4 +93,31 @@ case class SierraItems(itemDataMap: Map[SierraItemNumber, SierraItemData])
       None
     }
   }
+
+  private def transformItemData(
+    itemId: SierraItemNumber,
+    itemData: SierraItemData): Identifiable[Item] = {
+    debug(s"Attempting to transform $itemId")
```
jamesgorrie: Want to keep this?

alicefuzier:
@jamesgorrie It's the "Envelope", "0", "1", etc after the shelfmark.
As we start bringing in more item information with METS, this helps to distinguish which item is which from a user point of view

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItems.scala
warrd: This was from the original code (which was copy / pasted), I'm not that bothered about it myself but could be useful

jamesgorrie:
@alicefuzier completely missed that - thanks!

Do you know where the holdings information comes from?
> Complete set of 16 prints and original envelope

holdings

kenoir:
This'll be handy in the requesting service too 👍

PR #311, opened 2019-12-10
add search parameters to API response

jamesgorrie:
# Why

## Reason 1 - default settings not requested by client 

There is some information that might not be set by the client that is relevant to the request. 

A simple example of this would be page parameters, where we are actually searching page `1` with `pageSize` of `25`, but the client isn't aware of that state - so we send it down the pipes (we don't actually send current page in the current implementation).

This saves a tiny amount of code on the client i.e. `page = query.page ? query.page : 1`.

Another default, which we can't infer, and what this was setting up for was to get the `queryType` `name`. i.e. `ScoringTiers`, `MSMBoostUsingAndOperator`. We currently have a `default` and `n` amount of tests. The client needs to report on what query is being used, but can't as if it hasn't set the `_queryType` explicitly, it has no way of knowing which is being used. Going forward we might want to take this even more out of the client's code, which would be impossible in this state.

__example__
```JS
// GET /works
{ "queryType": "ScoringTiers" }

// on client
track({ queryType: response.queryType })
```

## Reason  2 - Rendering the UI to reflect current filters that are complex objects, but requested via querystring.

We currently render part of the interface that reflects what filters have been applied:
filters

Currently we:
- Render aggregation
- Take `query.workType` and do a lookup on the ID to render the filter

```JSX
aggregations = [{ id: 'a', label: 'Books' }]

// Current implementation when this is clicked
// Change URL: workType=a
filters = { workType: ['a'] }
// Render 
filters.workType.map(id => <div>{lookup(id).label}</div>)
```

We could store the aggregation in a client side state, but when working with rendering from the URL (which we have to 💯 support) we would still need to maintain the lookup.

The only solution I can think of is to do something like this in the URL `workType=(label:Books,id:a)` <= but, well, no.

So if we returned from the API:
```JS
{ "filters": { "workType": [{ "id": "a", "label": Books }] } }
```

We'd get to remove the hardcoded lookup.

---  

Adds the parameters to the response - similar to the [HATEOAS](https://restfulapi.net/hateoas/) paradigm.

This was some preliminary work I wanted to do as a <abbr title="proof of concept">POC</abbr> and to move onto adding returning of the `workType` filters to help the UI respond to the request and the `queryType` for reporting.

I am still not 100% sure about the shape of this.

## calculated fields, properties, filters

I was thinking you could split the data up into these pieces:
```JS
{
  "@context": "https://api.wellcomecollection.org/catalogue/v2/context.json",
  "type": "ResultList",
  "pageSize": 10,
  "totalPages": 132784,
  "totalResults": 1327838,
  "results": [
  ],
  "parameters": {
    "sort": [
      "production.dates"
    ],
    "sortOrder": "asc",
    "page": 2,
    "searchQuery": {
      "query": "dogs bogs and hogs",
      "queryType": "ScoringTiers"
    }
  },
  "filters": {
    "workType": [
      {
        "type": "WorkType",
        "label": "Books",
        "id": "a"
      },
      {
        "type": "WorkType",
        "label": "Video",
        "id": "v"
      }
    ]
  }
}
```

Problem is we use `pageSize` already at the root level, which is actually a property.

Because of this, we could flatten the properties, and leave the `"filters"` property there too?

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
@@ -1,0 +1,22 @@
+package uk.ac.wellcome.display.json
+
+import io.circe.Encoder
+import uk.ac.wellcome.display.models.SortingOrder.{Ascending, Descending}
+import uk.ac.wellcome.display.models.{
+  ProductionDateSortRequest,
+  SortRequest,
+  SortingOrder
+}
+
+trait DisplayJsonSerializers {
+  implicit val encodeSortingOrder: Encoder[SortingOrder] =
+    Encoder[String].contramap {
+      case Ascending  => "asc"
+      case Descending => "desc"
+    }
+
+  implicit val encodeSortRequest: Encoder[SortRequest] =
+    Encoder[String].contramap {
+      case ProductionDateSortRequest => "production.dates"
+    }
+}
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/Responses.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/Responses.scala
@@ -5,9 +5,9 @@ import io.circe.generic.extras.JsonKey
 import io.circe.{Encoder, Json}
 import akka.http.scaladsl.model.Uri
 import io.swagger.v3.oas.annotations.media.Schema
-
 import uk.ac.wellcome.platform.api.models._
 import uk.ac.wellcome.display.models._
+import uk.ac.wellcome.display.json.DisplayJsonSerializers
 import uk.ac.wellcome.display.models.v2._
 import uk.ac.wellcome.display.models.Implicits._
 import uk.ac.wellcome.platform.api.services.WorksSearchOptions
@@ -46,13 +46,16 @@ case class MultipleWorksResponse(
   pageSize: Int,
   totalPages: Int,
   totalResults: Int,
+  page: Int,
+  sort: List[SortRequest],
+  sortOrder: SortingOrder,
   results: List[DisplayWorkV2],
   prevPage: Option[String] = None,
   nextPage: Option[String] = None,
   aggregations: Option[DisplayAggregations] = None
 )
 
-object MultipleWorksResponse {
+object MultipleWorksResponse extends DisplayJsonSerializers {
 
   import DisplayAggregations.{encoder => aggsEncoder}
   implicit val encoder: Encoder[MultipleWorksResponse] = deriveEncoder
@@ -71,13 +74,17 @@ object MultipleWorksResponse {
       ),
       currentPage = searchOptions.pageNumber,
       requestUri = requestUri,
-      contextUri = contextUri
+      contextUri = contextUri,
+      sort = searchOptions.sortBy,
+      sortOrder = searchOptions.sortOrder
     )
 
   def apply(resultList: DisplayResultList[DisplayWorkV2],
             currentPage: Int,
             requestUri: Uri,
-            contextUri: String): MultipleWorksResponse =
+            contextUri: String,
+            sort: List[SortRequest],
+            sortOrder: SortingOrder): MultipleWorksResponse =
     MultipleWorksResponse(
       context = contextUri,
       ontologyType = resultList.ontologyType,
@@ -88,6 +95,9 @@ object MultipleWorksResponse {
       prevPage = pageLink(currentPage - 1, resultList.totalPages, requestUri),
       nextPage = pageLink(currentPage + 1, resultList.totalPages, requestUri),
       aggregations = resultList.aggregations,
+      page = currentPage,
+      sort = sort,
+      sortOrder = sortOrder
     )
 
   private def pageLink(page: Int,
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
@@ -65,14 +65,22 @@ trait ApiWorksTestBase
   def resultList(apiPrefix: String,
                  pageSize: Int = 10,
                  totalPages: Int = 1,
-                 totalResults: Int = 1) =
+                 totalResults: Int = 1,
+                 page: Int = 1,
+                 sort: List[String] = Nil,
+                 sortOrder: String = "asc") = {
+
     s"""
       "@context": "${contextUrl(apiPrefix)}",
       "type": "ResultList",
       "pageSize": $pageSize,
       "totalPages": $totalPages,
-      "totalResults": $totalResults
+      "totalResults": $totalResults,
+      "sort": ${toJson(sort).get},
+      "sortOrder": "$sortOrder",
+      "page": $page
     """
+  }
 
   def singleWorkResult(apiPrefix: String): String =
     s"""
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksTest.scala
@@ -99,7 +99,8 @@ class ApiV2WorksTest extends ApiV2WorksTestBase {
             apiPrefix,
             pageSize = 1,
             totalPages = 3,
-            totalResults = 3)},
+            totalResults = 3,
+            page = 2)},
               "prevPage": "$apiScheme://$apiHost/$apiPrefix/works?page=1&pageSize=1",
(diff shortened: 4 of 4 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
```diff
@@ -4,7 +4,7 @@ import java.time.LocalDate
 
 sealed trait WorkFilter
 
-case class ItemLocationTypeFilter(locationTypeIds: Seq[String])
+case class ItemsLocationLocationTypeFilter(locationTypeIds: Seq[String])
```
warrd: :open_mouth: what is the reason for this name change? I'm sure there is a better alternative, but not sure quite sure what you want to signify with it.

common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import io.circe.Encoder
+import uk.ac.wellcome.display.models.SortingOrder.{Ascending, Descending}
+import uk.ac.wellcome.display.models.{
+  ProductionDateSortRequest,
+  SortRequest,
+  SortingOrder
+}
+
+trait DisplayJsonSerializers {
```
warrd: Maybe `DisplayParameterSerializers`, or something like that, would be a more appropriate name? The current name doesn't really differentiate these from all the serializers defined elsewhere, and communicate that they for the purpose of serializing input query parameters.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
jamesgorrie: Ah, it was part of something else, happy to pull it out of this PR - basically to follow suit with all other namings of having it the same as the parameter, but `camelCased`

common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
jamesgorrie: What do you think about this method over having the `SourceData => InternalModel => DisplayModel` and rather replacing `DisplayModel` with serialisers?

[There's some examples and discussion here](https://github.com/wellcomecollection/stacks-service/pull/49#issuecomment-558740047).

TL;DR - it feels like what we've invented as `DisplayModels` is actually just serialisers.
- They should live on the edge of the application
- we should not really use them internally
- we can do it consistently without having to remember referenced models in `DisplayModels`
- we can create different serialiser contexts (for now we just have `ApiDisplaySerialisers`or something like that)

The reason this was called this was to be more generic and see how we felt about it - I can split them up similar to the `DisplayModels` though.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
warrd: Ahhh right! But aren't the parameter `period.seperated` paths rather than `camelCased`?

Either way, personally I don't reckon it makes the code easier to read having the case class names follow this convention too, and prefer the more descriptive `ItemLocationTypeFilter`. Even better IMO we could do what we do elsewhere and have `WorkFilter.ItemLocationType` etc.

If you think having the classes name reflect the query parameters is an improvement then feel 
 free to go ahead though.

jamesgorrie: The path is `item.locations.locationType` <- I am just replicating this to leave no room for guessing if it's referencing the same thing... personally I prefer consistency over making it slightly easier to read, as easy to read doesn't always help with new people coming to the code base.

I'll remove it from the PR, and do it universally.

warrd: Well I meant the code was easier to read in the broader sense than just reading out the name (maybe should have said easier to understand). It is not clear when encountering a filter in the code that it is named in such a way that represents a path, and also means more detail is encoded in the name than it need to (we can't move things around in the model without  remembering to change the name of the class, introducing unnecessary coupling etc)

common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
kenoir: For me the display models provide a way to understand what will be serialised. Skipping that steps makes it less easy to see in the code what the serialised version will look like. It also provides a serialisation free model of what we are making available: for example if you wanted to serialise the same model to XML & JSON an intermediate display abstraction would be useful.

I don't think we need to apply the "display model" approach in all cases, especially where there are small simple models being serialised.  **To further the experiment** trying a complex model would be informative as to whether removing display models extensively is beneficial.

kenoir:
It's not clear what benefit this provides from the PR description. I'm guessing this is to make the API more self-describing and therefore easier for clients to implement against?

common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
warrd: :+1: I agree the display models introduces an extra layer of complexity which could maybe be avoided.

I need to think more about this question in general, and probably worth us all having a conversation about it at some point. One thing that springs to mind is that we will likely still need lots of these models defined somewhere, even if they are pushed further to the edge of the application, due to the fact that using case classes and circe derivation is a really useful and declaritive method to define serializers / deserializers.

warrd:
Yep, also see my comment on the PR in general :point_up: in case you have missed it.

jamesgorrie:
@kenoir @warrd 

Sorry for the murky waters, updated the description

warrd:
:+1: thanks for the context

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
alicefuzier: For what is worth I agree with @warrd on this. Especially on the coupling part of it. There is no way that we can guarantee that the name will always match the path in the JSON because there are definitely going to be instances where we will make changes in the model and we will forget to change the class. 
I think making a claim that this is always going to match the json path could make things more difficult for people newer to the codebase, rather than easier.

common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
alicefuzier: I think I prefer having the DisplayModel intermediate steps. I agree that they add complexity but, as @kenoir is saying, they make it much easier to understand what gets serialised. Also, they allow for (relatively) easy swagger documentation. 
I also disagree that they are just serialisation. The serialisation is done by circe, and we could (relatively) easily swap the circe for another json library or add another type serialisation with minimal changes to the display models. I think the DisplayModels are models and if we skip this step, we _might_ end up with less lines of code (although not convinced about that because of swagger and circe etc) but we collapse two purposes - serialisation and model representation - making our life a bit harder.

alicefuzier:
Have you checked how the generated swagger looks like? I don't think it will document `SortingOrder` and `SortRequest` properly (personally I think an extra reason for using DisplayModels)

common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonSerializers.scala
jamesgorrie: @alicefuzier I'll swap it out.

jamesgorrie:
@alicefuzier @warrd @jtweed 

Bar the serialisers part of this - what do people think about the structure in the description?

Using the case of `queryType` - how might we add that to the return value of the response?

Maybe I have overcomplicated this and should just add that to the root?

alicefuzier:
Personally I think I prefer the `parameters` element instead of everything flattened in the root. It makes it easier to separate things that relate to the context of the request from things that are part of the response proper. I'm not sure if I would call it `parameters` though

warrd:
Yes  agree with @alicefuzier better to nest under `parameters` (or some other appropriate name)

jtweed:
I think it's better to group everything together and have no problem with the `parameters` name. I don't think there's much point in separating out kinds of parameter.

I also think it's ok to duplicate `pageSize` under `parameters`, but that `parameters` should probably be an `include`.

jamesgorrie:
Thanks all for the comments, I'll continue with this and make the relevant changes.

@alicefuzier parameters was taken from swagger - but that's on the request, I'll have a think of a name saying that this was what we used to create these results.

kenoir:
Does this need to stay open?

jamesgorrie:
Closing as there's too much in here that is going to be dealt with in other PRs.
Thanks for all the super valuable feedback!

PR #316, opened 2019-12-12
Fix compilation error in merger

warrd:
Due to a merged PR not being against latest changes on master.


warrd:
@jamesgorrie not to do with default arguments, its pattern matching against a case class which had an extra field added on master, but the merged PR did not have the latest changes

warrd:
There is something to be said about requiring all latest changes from `master` on the branch that is being merged in (github has an option for this), but due to slowness of the builds I think probably better to live with things as is

PR #318, opened 2019-12-12
Merge multiple items Sierra works with METS works

alicefuzier:
Resolves https://github.com/wellcometrust/platform/issues/4104


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/physicaldigital/SierraPhysicalDigitalMergeRule.scala
```diff
@@ -8,37 +8,31 @@ import uk.ac.wellcome.platform.merger.rules.{MergerRule, WorkPairMerger}
 
 /** If we have a pair of Sierra records:
   *
-  *   - One of which has a single Physical location ("physical work")
+  *   - One of which has a at least one item with at least a Physical location ("physical work")
```
warrd: Would read better like this IMO:

```suggestion
  *   - One of which has a at least one item containing a Physical location ("physical work")
```

PR #321, opened 2019-12-12
Separate aggregation-dependent and independent filters

jamieparkinson:
Closes https://github.com/wellcometrust/platform/issues/4061


jamieparkinson:
@warrd my changes cause [this test](https://github.com/wellcometrust/catalogue/blob/fully-functioning-filter-facets/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/AggregationsTest.scala#L45) to fail, which looks like it might be something to do with us wanting date filtering to also affect the aggregation - I remember you mentioned there might be an issue relating to dates, could you help at all?

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
```diff
@@ -69,6 +69,31 @@ object Aggregations extends Logging {
       Try(License.createLicense(str)).toEither.left
         .map(err => err.getMessage)
     }
+
+  implicit class EnhancedTransformable(transformable: Transformable) {
+    def toAgg[T: Decoder]: Option[Aggregation[T]] = {
+      transformable
+        .safeTo[EsAggregation[T]]
+        .toOption
```
warrd: I believe this will swallow errors which we would probably like to be alerted to? Maybe you are aware of this as still a draft, but just noticed it.

jamieparkinson: Hmmm, kinda agree about the errors but we've been swallowing them [all along](https://github.com/wellcometrust/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala#L24) and an error in parsing the JSON here would indicate that ES has broken its API response contract, which seems like something we can assume won't happen...

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/FilteredAggregationBuilder.scala
```diff
@@ -0,0 +1,97 @@
...
+  * constituents of the ES query, this class exposes:
+  *
+  * - `aggregations`: a list of all the ES query aggregations, where those that need to be filtered
+  *   now have a sub-aggregation of the filter aggregation type, named "filtered".
+  * - `independentFilters`: a list of all of the given filters which are not paired to any of
+  *   the given aggregations. These can be used as the ES query filters.
+  * - `aggregationDependentFilters`: a list of all of the given filters which are paired to one
+  *   of the given aggregations. These can be used as the ES post-query filters.
+  */
+class FilteredAggregationBuilder(
```
jamesgorrie: Nit as I know the naming here is fiddly: Calling this `FiltersAndAggregationsBuilder` helped me understand why they're coupled in the implementing code.

jamesgorrie:
```scala
class FiltersAndAggregationsBuilder {
  val unpairedFilters: List[WorkTypeFilter]
  val filteredAggregations: Aggregations
  val pairedFilters: List[WorkTypeFilter]
}
```

Having the names reflected in all places helped too - and made the mapping of:
```scala
request
.filters { filtersAndAggregations.unpairedFilters }
.aggregations { filtersAndAggregations.unpairedFilters }
.postFilters { filtersAndAggregations.pairedFilters }
```

feel easier to understand.

Subject to a high level of subjectivitiy, take it or leave it the functionality is the same and great.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
warrd: Interesting that already has this behaviour. I would at the very least log this as an error (even if we still decide to return the response with the field as `null`), as probably something we would be keen to be alerted to

PR #324, opened 2019-12-16
Subscribe to storage service topic

alicefuzier:


mets_adapter/terraform/locals.tf
```diff
@@ -1,6 +1,6 @@
 locals {
  namespace = "mets-adapter"
- storage_notifications_topic = ""
+ storage_notifications_topic_arn = "arn:aws:sns:eu-west-1:975596993436:storage_prod_bag_register_output"
```
jamesgorrie: We're happy for this not to come from the storage state?

alicefuzier: you're right, it should come from there

PR #325, opened 2019-12-16
Transform multiple METS manifestations

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/4103

## Description

Updates METS transformer so that it uses correct data when there are manifestations:

* Thumbnail from first manifesatation
* Access condition from first manifestation


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/service/MetsTransformerWorkerService.scala
```diff
@@ -1,33 +1,34 @@
...
-  metsXmlStore: TemporaryCredentialsStore[String]
+  metsXmlStore: Readable[ObjectLocation, String]
 ) extends Runnable
     with Logging {
 
   type Result[T] = Either[Throwable, T]
 
   val className = this.getClass.getSimpleName
 
+  val xmlTransformer = new MetsXmlTransformer(metsXmlStore)
```
alicefuzier: Nitpicking here but why don't we inject the `xmlTransformer ` instead of `metsXmlStore`?

warrd: I thought about this, and chose the the former as it means the creator `MetsTransformerWorkerService` only needs to know about the store (as opposed to both the store and the transformer). It's not a major issue for me either, but think it is slightly cleaner personally keeping the transformer as something internal to this class.

PR #326, opened 2019-12-17
Matcher ignores messages for older versions

alicefuzier:


pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/storage/WorkStore.scala
```diff
@@ -0,0 +1,30 @@
...
+  def getWork(
+    key: Version[String, Int]): Either[Throwable, TransformedBaseWork] =
+    store.getLatest(key.id) match {
+      case Left(err) =>
+        error(s"Error fetching $key from VHS")
+        Left(err.e)
+      case Right(Identified(id, entry)) if id.version == key.version =>
+        Right(entry.t)
+      case Right(Identified(id, _)) if id.version > key.version =>
+        Left(MatcherException(VersionExpectedConflictException()))
```
warrd: It's not clear from this code that the message is ignored and not sent to the DLQ in this case. Where is the logic for this?

alicefuzier: It already exists [here](https://github.com/wellcometrust/catalogue/blob/master/pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/services/MatcherWorkerService.scala#L43). 
We can end up in the same situation where the matcher internal graph store already has the work, but for a newer version, in which case we ignore the message. Now that we read the work from the recorder vhs it can happen with the vhs too, but the logic was already implemented

alicefuzier: BTW the test for the already existing case is [this one](https://github.com/wellcometrust/catalogue/pull/326/files#diff-5d8681b632a877510a4b0b8ca2b557f4R60)

warrd: Probably worth adding a comment here explaining this here, as that context is pretty important here.

PR #331, opened 2019-12-19
Add InEnglishWithContributors query

jamesgorrie:
✨ new query - 3 additions 

## Boost concepts before title
Working with @taceybadgerbrook and @harrisonpim it felt these should be boosted higher as it fits how people use our service more closely

## With contributors
Matches contributors and boosts highly.

## In english
I noticed when search for "Vlad the impaler" using the ScoringTiers base query, were returning loads of stuff, nothing to do with Vlads or impaling.

As we're using the OR operator on the base query, the `the` becomes part of the query - albeit with super low score due to TF x IDF - but it still matches.

It shouldn't and this solves that.

## Next up
Looking at matching concepts more accurately using ngrams and a few other tools at our disposal.

### DIFF
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchSearchQueryTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchSearchQueryTest.scala
@@ -10,7 +10,7 @@ import uk.ac.wellcome.models.work.generators.{
   SubjectGenerators,
   WorksGenerators
 }
-import uk.ac.wellcome.models.work.internal.{IdentifiedBaseWork}
+import uk.ac.wellcome.models.work.internal.IdentifiedBaseWork
 import uk.ac.wellcome.platform.api.generators.SearchOptionsGenerators
 import uk.ac.wellcome.platform.api.models.{SearchQuery, SearchQueryType}
 import uk.ac.wellcome.platform.api.services.{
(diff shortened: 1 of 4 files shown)


api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchSearchQueryTest.scala
```diff
@@ -113,6 +113,36 @@ class ElasticsearchQueryTest
...
+          "Vlad the impaler",
+          "Dad the impala",
+        ).map { t =>
+          createIdentifiedWorkWith(title = Some(t))
+        }
+
+        insertIntoElasticsearch(index, works: _*)
+
+        // If we search the non-english analysed fields with the base query
+        // `the` would in the search as we're using the `OR` operator
```
warrd:
```suggestion
        // `the` would be in the search as we're using the `OR` operator
```

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchSearchQueryTest.scala
```diff
@@ -113,6 +113,36 @@ class ElasticsearchQueryTest
...
+        // and would be matched in both examples above as the root field
+        // (not `field` rather than `field.english`, see `WorksIndex.scala`)
+        // does not use the english analyser.
+
+        // We wouldn't want to use the english analyser at query time though
+        // as we would lose detail used in other where we use exact matching
```
warrd:
```suggestion
        // as we would lose detail used in other places where we use exact matching
```

warrd:
:thinking: I thought we had moved to using `AND` as the default query recently?

PR #333, opened 2019-12-19
Add additional note fields

warrd:
https://github.com/wellcometrust/platform/issues/4114


common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayNote.scala
```diff
@@ -68,5 +68,19 @@ object DisplayNoteType {
         DisplayNoteType("location-of-original", "Location of original")
       case BindingInformation(_) =>
         DisplayNoteType("binding-detail", "Binding detail")
+      case BiographicalNote(_) =>
```
jamesgorrie: What's the difference between this and `BibliographicalInformation `?

warrd: bibliographical vs biographical

jamesgorrie: Sorry, Christmas laziness has come early!

warrd: :christmas_tree:

PR #334, opened 2019-12-19
Some improvements to the populate_mets script

alicefuzier:
- [x] Make it able to restart at an arbitrary key in the table
- [x] Make it less verbose
- [x] Make it pause to allow to refresh the token if the credentials expire

### DIFF
--- a/mets_adapter/populate_mets/populate_mets.py
+++ b/mets_adapter/populate_mets/populate_mets.py
@@ -44,7 +44,7 @@ class StorageManifestScanner:
 
     def scan(self):
         for page in self._paginate():
-            click.echo(click.style(f"Processing {len(page['Items'])} records", fg='yellow'))
+            click.echo(click.style(f"Processing {len(page['Items'])} records", fg='green'))
             with click.progressbar(page["Items"]) as items:
                 for item in items:
                     space, id = item["id"].split("/")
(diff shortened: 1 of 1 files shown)


mets_adapter/populate_mets/populate_mets.py
```diff
@@ -60,16 +77,31 @@ def publish(self, space, id):
...
         num_records = maxsize
-    scanner = StorageManifestScanner()
+
+    scanner = StorageManifestScanner(start_record, num_records)
     publisher = MessagePublisher()
-    for i, (space, id) in zip(range(1, num_records + 1), scanner.scan()):
-        click.echo(f"{click.style(str(i), fg='blue')}: {space}/{id}")
-        publisher.publish(space, id)
+    try:
+        for i, (space, id) in zip(range(1, num_records + 1), scanner.scan()):
```
warrd: Thought I could have spotted a bug here with it overflowing if `num_records == sys.maxsize`, but apparently not:

```
In [3]: sys.maxsize                                                                                                                                                                                                                                                              
Out[3]: 9223372036854775807

In [4]: sys.maxsize + 1                                                                                                                                                                                                                                                          
Out[4]: 9223372036854775808

In [5]: type(sys.maxsize)                                                                                                                                                                                                                                                        
Out[5]: int

In [6]: type(sys.maxsize + 1)                                                                                                                                                                                                                                                    
Out[6]: int
```

Turns out `sys.maxsize` is not the biggest `int` after all!

alicefuzier: 🙄 of course...

warrd: https://stackoverflow.com/a/48138759

kenoir:
@alicefuzier why is this still open?

alicefuzier:
I don't think this script is needed anymore, so I'm going to close this PR and open another one to delete it

PR #335, opened 2019-12-19
Access conditions in model

warrd:
Prerequisite for:

https://github.com/wellcometrust/platform/issues/3946
https://github.com/wellcometrust/platform/issues/4125

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
@@ -3,7 +3,6 @@ package uk.ac.wellcome.display.json
 import io.circe.generic.extras.{AutoDerivation, Configuration}
 import io.circe.{Encoder, Printer}
 import io.circe.syntax._
-import io.circe.java8.time.TimeInstances
 
 import uk.ac.wellcome.display.models.DisplayWork
 import uk.ac.wellcome.display.models.v2._
@@ -18,7 +17,7 @@ import uk.ac.wellcome.display.models.v2._
   *     render the complete value for disambiguation.
   *
   */
-trait DisplayJsonUtil extends AutoDerivation with TimeInstances {
+trait DisplayJsonUtil extends AutoDerivation {
   val printer = Printer.noSpaces.copy(
     dropNullValues = true
   )
--- a/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
@@ -3,6 +3,8 @@ package uk.ac.wellcome.display.json
 import io.circe.generic.extras.{AutoDerivation, Configuration}
 import io.circe.{Encoder, Printer}
 import io.circe.syntax._
+import io.circe.java8.time.TimeInstances
+
 import uk.ac.wellcome.display.models.DisplayWork
 import uk.ac.wellcome.display.models.v2._
 
@@ -16,7 +18,7 @@ import uk.ac.wellcome.display.models.v2._
   *     render the complete value for disambiguation.
   *
   */
-trait DisplayJsonUtil extends AutoDerivation {
+trait DisplayJsonUtil extends AutoDerivation with TimeInstances {
   val printer = Printer.noSpaces.copy(
     dropNullValues = true
   )
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
@@ -1,6 +1,6 @@
 package uk.ac.wellcome.display.models
 
-import java.time.LocalDate
+import java.time.format.DateTimeFormatter.ISO_LOCAL_DATE
 import io.circe.generic.extras.JsonKey
 import io.swagger.v3.oas.annotations.media.Schema
 import uk.ac.wellcome.models.work.internal.{AccessCondition, AccessStatus}
@@ -11,7 +11,7 @@ import uk.ac.wellcome.models.work.internal.{AccessCondition, AccessStatus}
 case class DisplayAccessCondition(
   status: DisplayAccessStatus,
   terms: Option[String],
-  to: Option[LocalDate],
+  to: Option[String],
   @JsonKey("type") @Schema(name = "type") ontologyType: String =
     "AccessCondition"
 )
@@ -22,7 +22,7 @@ object DisplayAccessCondition {
     DisplayAccessCondition(
       status = DisplayAccessStatus(accessCondition.status),
       terms = accessCondition.terms,
-      to = accessCondition.to,
+      to = accessCondition.to.map(_.format(ISO_LOCAL_DATE)),
     )
 }
 
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
@@ -1,0 +1,53 @@
+package uk.ac.wellcome.display.models
+
+import java.time.LocalDate
+import io.circe.generic.extras.JsonKey
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.ac.wellcome.models.work.internal.{AccessCondition, AccessStatus}
+
+@Schema(
+  name = "AccessCondition"
+)
+case class DisplayAccessCondition(
+  status: DisplayAccessStatus,
+  terms: Option[String],
+  to: Option[LocalDate],
+  @JsonKey("type") @Schema(name = "type") ontologyType: String =
+    "AccessCondition"
+)
+
+object DisplayAccessCondition {
+
+  def apply(accessCondition: AccessCondition): DisplayAccessCondition =
+    DisplayAccessCondition(
+      status = DisplayAccessStatus(accessCondition.status),
+      terms = accessCondition.terms,
+      to = accessCondition.to,
+    )
+}
+
+@Schema(
+  name = "AccessStatus"
+)
+case class DisplayAccessStatus(
+  id: String,
+  label: String,
+  @JsonKey("type") @Schema(name = "type") ontologyType: String = "AccessStatus"
+)
+
+object DisplayAccessStatus {
+
+  def apply(accessStatus: AccessStatus): DisplayAccessStatus =
+    accessStatus match {
+      case AccessStatus.Open =>
+        DisplayAccessStatus("open", "Open")
+      case AccessStatus.OpenWithAdvisory =>
+        DisplayAccessStatus("open-with-advisory", "Open with Advisory")
+      case AccessStatus.Restricted =>
+        DisplayAccessStatus("restricted", "Restricted")
+      case AccessStatus.Closed =>
+        DisplayAccessStatus("closed", "Closed")
+      case AccessStatus.LicensedResources =>
+        DisplayAccessStatus("licensed-resources", "Licensed Resources")
+    }
+}
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
@@ -1,5 +1,6 @@
 package uk.ac.wellcome.display.models
 
+import java.time.{LocalDateTime, ZoneOffset}
 import java.time.format.DateTimeFormatter.ISO_LOCAL_DATE
 import io.circe.generic.extras.JsonKey
 import io.swagger.v3.oas.annotations.media.Schema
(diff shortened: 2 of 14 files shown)


common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
```diff
@@ -16,7 +18,7 @@ import uk.ac.wellcome.display.models.v2._
   *     render the complete value for disambiguation.
   *
   */
-trait DisplayJsonUtil extends AutoDerivation {
+trait DisplayJsonUtil extends AutoDerivation with TimeInstances {
```
alexwlchan: In general we've kept the “interesting” types out of the Display models.

They only contain primitive JSON types, like `String`, `Int` and `List`, and then their `apply()` method contains all the logic for serialising a Scala type (like LocalDate) into a String. It’s all kept together.

Could we push the LocalDate ~> String serialisation into `DisplayAccessCondition.apply()`? And be explicit about how we're doing it?

(I'm not completely opposed to changing it, but I'd like to at least discuss it.)

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
```diff
@@ -0,0 +1,53 @@
...
+      status = DisplayAccessStatus(accessCondition.status),
+      terms = accessCondition.terms,
+      to = accessCondition.to,
+    )
+}
+
+@Schema(
+  name = "AccessStatus"
+)
+case class DisplayAccessStatus(
```
alexwlchan: minor: should this be in a separate file?

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
```diff
@@ -0,0 +1,53 @@
...
+      case AccessStatus.Open =>
+        DisplayAccessStatus("open", "Open")
+      case AccessStatus.OpenWithAdvisory =>
+        DisplayAccessStatus("open-with-advisory", "Open with Advisory")
+      case AccessStatus.Restricted =>
+        DisplayAccessStatus("restricted", "Restricted")
+      case AccessStatus.Closed =>
+        DisplayAccessStatus("closed", "Closed")
+      case AccessStatus.LicensedResources =>
+        DisplayAccessStatus("licensed-resources", "Licensed Resources")
```
alexwlchan: Is keeping this information in the *display* model the right place for it?

I would lean towards putting it all on the core model, and reading those fields out here, a la License: https://github.com/wellcometrust/catalogue/blob/385c9bf87cd33edcd9ab2b8ee9d8e721bb2f8637/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/License.scala

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayLocationV2.scala
```diff
@@ -54,6 +57,9 @@ case class DisplayDigitalLocationV2(
     description =
       "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made."
   ) license: Option[DisplayLicenseV2] = None,
+  @Schema(
+    description = "The access conditions"
```
alexwlchan: Do we have a better description we can use?

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayLocationV2.scala
```diff
@@ -19,17 +20,19 @@ sealed trait DisplayLocationV2
...
+        accessCondition = accessCondition.map(DisplayAccessCondition(_))
       )
-    case l: PhysicalLocation =>
+    case PhysicalLocation(locationType, label, _) =>
       DisplayPhysicalLocationV2(
-        locationType = DisplayLocationType(l.locationType),
-        label = l.label)
+        locationType = DisplayLocationType(locationType),
+        label = label
+      )
```
alexwlchan: No access conditions on the display physical locations? Sounds like that's what https://github.com/wellcometrust/platform/issues/4125 is describing.

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayLocationV2.scala
```diff
@@ -54,6 +57,9 @@ case class DisplayDigitalLocationV2(
     description =
       "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made."
   ) license: Option[DisplayLicenseV2] = None,
+  @Schema(
+    description = "The access conditions"
+  ) accessCondition: Option[DisplayAccessCondition] = None,
```
alexwlchan: Jonathan’s examples in https://github.com/wellcometrust/platform/issues/4125 and https://github.com/wellcometrust/platform/issues/3946 are both lists – should this be

```scala
) accessCondition: Option[List[DisplayAccessCondition]] = None
```

?

warrd:
Thanks for the reviews! I'm a bit hazy on all the details of this after it not being on my mind for a few weeks, so will refamiliarise myself address @alexwlchan's questions

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
```diff
@@ -0,0 +1,24 @@
+package uk.ac.wellcome.models.work.internal
+
+import java.time.LocalDate
+
+case class AccessCondition(
+  status: AccessStatus,
+  terms: Option[String],
+  to: Option[LocalDate]
```
jamesgorrie: Should we keep this as an Instant and render it the same as all other dates that we're using? I'm not sure why we would need to introduce another Date type.

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
alicefuzier: We did have a discussion about this with @warrd. This is closer to how [notes are implemented](https://github.com/wellcometrust/catalogue/blob/master/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayNote.scala#L49). @warrd and I agreed to go in this direction rather because it makes it easier to change dispplay labels, ie we don't need a reindex. 
Do you have any particular concerns about it?

alicefuzier: (If I remember correctly, we also discussed migrating Licenses, WorkType etc to use this model. The idea is to push display stuff into the displaymodels as much as possible)

warrd: Yeh as @alicefuzier points out this is connected to other discussions we have been having regarding models vs display models.

A big match statement with lots of case clauses isn't the prettiest / most declaritive way to represent the data, but in my opinion things like labels should gradually be removed from the internal model, and should just be part of the display. As mentioned it allows anything related to display to be modified independently to the internal model. Also, there are areas where logic would be much simpler if we had a "minimal" internal model, rather than storing lots of data on it (such as complications in the aggregations code).

It feels like quite a big discussion to be had within this, so maybe worth having a chat about things with everyone at some point.

common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
warrd: :+1:  happy to keep to basic JSON types within the display if that is what we decided previously

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
alicefuzier: One problem with this approach is that, if it's only in the display models, it's not in es, so it's not as easy to search/aggregate. I tried to do something similar for WorkTypes and realised it would be a pain with aggregation, so worth keeping that in mind

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
warrd: The reasons I would introduce an extra type is that `LocalDate` seems to better represent the data:

1) We are just interested in a day rather than hours, minutes, seconds, ms, microseconds etc
2) There doesn't appear to be timezone information encoded within the date

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
jamieparkinson: fwiw I'm very much in favour of moving labels to the display models from the perspective of search: the Terms aggregations that we have to use in order to fetch both the ID and label have been a bit of a pain point in (eg) https://github.com/wellcometrust/catalogue/pull/321

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
warrd: We are not adhering to class per file or anything elsewhere, so think it makes sense to have this here as it is just a sub field within `DisplayAccessCondition`

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
warrd: @jamesgorrie any thoughts on above? :point_up:

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
jamesgorrie: If we think in terms of multi serialisation / deserialisation, we could have:

```scala
// I have used `ApiThing` to represent something similar to `ApiDisplayThing`
case class Work(workType: WorkType)
case class ApiDisplayWork(workType: ApiDisplayWorkType)
case class ElasticWork(workType: ElasticWorkType)

trait WorkType
case object Books extends WorkType
case object Pictures extends WorkType

case class ApiDisplayWorkType(label: String, id: String)
case object ApiDisplayWork {
  def serialize(workType: WorkType): ApiDisplayWorkType = {
    workType match {
      case Books => ApiDisplayWorkType("Books", "a")
      case Pictures => ApiDisplayWorkType("Pictures", "p")
    }
  }
}

case class ElasticWorkType(`type`: String) {
  def serialize(workType: WorkType): ElasticWorkType = {
    workType match {
      case Books => ElasticWorkType("books")
      case Pictures => ElasticWorkType("pictures")
    }
  }

  def deserialize(elasticWorkType: ElasticWorkType): WorkType = elasticWorkType.`type` match {
    case "books" => Books
    case "pictures" => Pictures
  }
}
```

There's other implementation of this approach discussed [here](https://github.com/wellcomecollection/stacks-service/pull/49#issuecomment-558740047) and [here](https://github.com/wellcometrust/catalogue/pull/311/files#r356175213) (sorry to keep ringing the same bell but actually though it was relevant).

I've left the implementation a little shoddy to get the gist across.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
jamesgorrie: > We are just interested in a day rather than hours, minutes, seconds, ms, microseconds etc 

For now until we need more resolution.

> There doesn't appear to be timezone information encoded within the date

It will almost definitely be GMT / UTC.

It represents the `Display` logic of the date, I'd opt for `Instant`s where we can unless we can really explain why we'd use something else except displaying, which we have other ways of dealing with that (store as instant, display as whatever).

warrd: Hmmm personally I would err on the side of modelling with the more accurate representation without any redundancy (according to our current knowledge), but will convert to `Instant` to be consistent with the dates elsewhere.

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayAccessCondition.scala
jamesgorrie: Moving this [here](https://github.com/wellcometrust/platform/issues/3930#issuecomment-571501887) as it bikeshedding here.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
warrd: I have updated it to use `Instant` instead of `LocalDate`, but feels a bit wonky to me if we are serialising to the front end as solely a date / expecting the incoming data to the transformer to be a date rather than an instant.

https://github.com/wellcometrust/catalogue/pull/335/commits/6dc1edc4817d663cda6eb3908161f2d852159afb

warrd:
Merging this finally.. We can return to some of the above during upcoming modelling conversations.

PR #336, opened 2020-01-06
GitBook: [search_relevance_test_6] 2 pages and one asset modified

harrisonpim:
**tl;dr** 
Looks like the new query isn't performing as well as we'd hoped - we should step back and think about why that might be the case. My hunch is that it has more to do with the new english analyser than the inclusion of contributors, but I could be wrong.


jamesgorrie:
> we should step back and think about why that might be the case.

Read - put in a new test.

docs/search_relevance/tests/test-6-english-tokeniser-and-contributors.md
```diff
@@ -0,0 +1,31 @@
...
+
+![download-1](https://user-images.githubusercontent.com/11006680/71817901-408a3280-307f-11ea-95ee-f8aaea9e2f56.png)
+
+## Conclusions
+
+The data indicates that the new variant query performs worse than the default, particularly at the top end of the list of results \(where worse performance is indicated by fewer clicks\). 
+
+The difference isn't substantial, and both queries seem to perform better than some of our previous candidates. However, we should still step back and evaluate why performance is lower before proposing a new set of changes.
+
+##
```
jamesgorrie: To be continued?

docs/search_relevance/tests/test-6-english-tokeniser-and-contributors.md
```diff
@@ -0,0 +1,31 @@
...
+
+**Click distribution**
+
+![download-1](https://user-images.githubusercontent.com/11006680/71817901-408a3280-307f-11ea-95ee-f8aaea9e2f56.png)
+
+## Conclusions
+
+The data indicates that the new variant query performs worse than the default, particularly at the top end of the list of results \(where worse performance is indicated by fewer clicks\). 
+
+The difference isn't substantial, and both queries seem to perform better than some of our previous candidates. However, we should still step back and evaluate why performance is lower before proposing a new set of changes.
```
jamesgorrie: I don't think we should suggest rolling back, but things we can do moving forward.

Maybe removing the english analyser as this is the hunch. 

Thinking of this, while adding features e.g. adding contributors, is there an amount of negative vatriation we will permit in the name of adding something we want?

Another option is saying:
> Intention: Find things made by a contributor

And then work on that until it's better? I like this more and should maybe be our next step.

I imagine the workflow to be:
- `ScoringTiers` as `A`
- `ScoringTiersWithContributorsV1` as `B`
- Keep updating `ScoringTiersWithContributorsVn` until it beats `A`
- Make `ScoringTiersWithContributorsVn` default
- Create new intention, iterate on that

How does that sound? I can capture this and share it with the team if we think it's good?

harrisonpim: that sounds very sensible to me - might be worth adding to the query design page in the docs?

docs/search_relevance/tests/test-6-english-tokeniser-and-contributors.md
harrisonpim: 👍

jamesgorrie:
@harrisonpim what should we do with this? It feels incomplete?

harrisonpim:
Sorry pal, don't know how I missed that ping the other day... 

We should be clear/open about how we deal with tests that don't show a significant improvement - maybe adding a generic version of the process you've described above as another section in https://app.gitbook.com/@wellcomecollection/s/catalogue/search_relevance, and an explicit acknowledgement at the end of the failed test's own page

PR #339, opened 2020-01-08
add parameters to search response with queryType

jamesgorrie:
Sends `queryType` back to the client.

This will help with accurate reporting as to what is going on from the client perspective. e.g.
```js
track("searched", { query: "harbinger", queryType: apiResponse.parameters.queryType })
```

We might move other things onto this to be able to rely on it completely for things like that ☝️ (see [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS)).

I've kept it at the edge as it is basically just a representation of `searchOptions` and isn't needed anywhere in the application.

`"options"` wouldn't read right I don't think as that is what you might send on the display model as things the client can chose from.


jamesgorrie:
Closing in favour of https://github.com/wellcometrust/platform/issues/4155

PR #342, opened 2020-01-09
[WIP]Rename agent field in identity state

alicefuzier:
Trying to address the whole `agent` silliness in the model

### DIFF
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/AggregationsTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/AggregationsTest.scala
@@ -127,7 +127,7 @@ class AggregationsTest
         whenReady(aggregationQuery(index, searchOptions)) { aggs =>
           val buckets = aggs.workType.get.buckets
           val expectedWorkTypes = works
-            .filter { _.data.subjects.head.agent.label == subjectQuery }
+            .filter { _.data.subjects.head.thing.label == subjectQuery }
             .map { _.data.workType.get }
           buckets.length shouldBe expectedWorkTypes.length
           buckets.map(_.data) should contain theSameElementsAs expectedWorkTypes
@@ -153,7 +153,7 @@ class AggregationsTest
         whenReady(worksService.listWorks(index, searchOptions)) { res =>
           val results = res.right.get.results
           results.map(_.data.workType.get) should contain only WorkType.Books
-          results.map(_.data.subjects.head.agent.label) should contain only subjectQuery
+          results.map(_.data.subjects.head.thing.label) should contain only subjectQuery
         }
       }
     }
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksAggregationsTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksAggregationsTest.scala
@@ -97,7 +97,7 @@ class ApiV2WorksAggregationsTest extends ApiV2WorksTestBase {
           sourceIdentifier = createSourceIdentifierWith(
             ontologyType = "Period"
           ),
-          agent = Period("periodLabel")
+          thing = Period("periodLabel")
         )
 
         val genre = Genre(
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayItemV2.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayItemV2.scala
@@ -41,15 +41,15 @@ object DisplayItemV2 {
             if (includesIdentifiers)
               Some(identifiedItem.identifiers.map { DisplayIdentifierV2(_) })
             else None,
-          title = identifiedItem.agent.title,
-          locations = displayLocations(identifiedItem.agent)
+          title = identifiedItem.thing.title,
+          locations = displayLocations(identifiedItem.thing)
         )
       case unidentifiableItem: Unidentifiable[Item] =>
         DisplayItemV2(
           id = None,
           identifiers = None,
-          title = unidentifiableItem.agent.title,
-          locations = displayLocations(unidentifiableItem.agent)
+          title = unidentifiableItem.thing.title,
+          locations = displayLocations(unidentifiableItem.thing)
         )
     }
 
--- a/common/display/src/test/scala/uk/ac/wellcome/display/models/DisplaySerialisationTestBase.scala
+++ b/common/display/src/test/scala/uk/ac/wellcome/display/models/DisplaySerialisationTestBase.scala
@@ -45,9 +45,9 @@ trait DisplaySerialisationTestBase { this: Suite =>
   def unidentifiableItem(it: Unidentifiable[Item]) =
     s"""
       {
-        "type": "${it.agent.ontologyType}",
-        ${itemTitle(it.agent)}
-        "locations": [${locations(it.agent.locations)}]
+        "type": "${it.thing.ontologyType}",
+        ${itemTitle(it.thing)}
+        "locations": [${locations(it.thing.locations)}]
       }
     """
 
@@ -55,9 +55,9 @@ trait DisplaySerialisationTestBase { this: Suite =>
     s"""
       {
         "id": "${it.canonicalId}",
-        "type": "${it.agent.ontologyType}",
-        ${itemTitle(it.agent)}
-        "locations": [${locations(it.agent.locations)}]
+        "type": "${it.thing.ontologyType}",
+        ${itemTitle(it.thing)}
+        "locations": [${locations(it.thing.locations)}]
       }
     """
 
@@ -113,9 +113,9 @@ trait DisplaySerialisationTestBase { this: Suite =>
   def identifiedOrUnidentifiable[T](displayable: Displayable[T],
                                     serialise: T => String) =
     displayable match {
-      case ag: Unidentifiable[T] => serialise(ag.agent)
+      case ag: Unidentifiable[T] => serialise(ag.thing)
       case disp: Identified[T] =>
-        val agent = parse(serialise(disp.agent)).right.get.asObject.get
+        val agent = parse(serialise(disp.thing)).right.get.asObject.get
         val identifiersJson = disp.identifiers.map { sourceIdentifier =>
           parse(identifier(sourceIdentifier)).right.get
         }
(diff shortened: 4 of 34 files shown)


jamesgorrie:
If we start from the work model, and look at the root parameters that use this model i.e. `production`, `contributors`, `genres`, `subjects`, it feels like we could simplify this significantly.

At the risk of over simplifying - what would be wrong with:

```scala
case class Identifier(source: String, id: String)

sealed trait ProductionTrait
sealed trait ConceptTrait {
  val label: String
  val id: Option[String] = None
  val sourceIdentifier: Option[Identifier] = None
  val otherIdentifiers: List[Identifier] = List()
}

case class Concept(label: String) extends ConceptTrait
case class Period(label: String) extends ConceptTrait with ProductionTrait
case class Place(label: String) extends ConceptTrait with ProductionTrait

sealed trait AgentTrait extends ConceptTrait
case class Agent(label: String) extends AgentTrait with ProductionTrait
case class Person(label: String) extends AgentTrait with ProductionTrait
case class Organisation(label: String) extends AgentTrait with ProductionTrait

sealed trait RoleTrait
case object Role extends RoleTrait
case object Contributor extends RoleTrait
case object Publisher extends RoleTrait
case object Blacksmith extends RoleTrait
case object Typesetter extends RoleTrait

case class Contribution(role: RoleTrait, agent: AgentTrait)
case class CompoundSubject(label: String, concepts: List[ConceptTrait]) extends ConceptTrait
case class Subject(label: String) extends ConceptTrait


trait GenreTrait extends ConceptTrait
case class Genre(label: String) extends GenreTrait
case class Medium(label: String) extends GenreTrait

case class ProductionEvent(label: String, concepts: List[ProductionTrait])

case class Work(
  production: List[ProductionEvent],
  contributors: List[Contribution],
  subjects: List[CompoundSubject],
  genres: List[GenreTrait]
)

object WorkApp {
  def run(work: Work): Unit = {
    val productionEventTypes = work.production.map(_.concepts.map {
        case _: Period => "Period"
        case _: Place  => "Place"
        case _: Agent  => "Agent"
        case _: Person  => "Person"
        case _: Organisation  => "Organisation"
    })

    val contributorTypes = work.contributors.map(_.agent match {
      case _: Agent => "Agent"
      case _: Person => "Person"
      case _: Organisation=> "Organisation"
    })

    val conceptTypes = work.subjects.map(_.concepts.map {
      case _: Concept => "Concept"
      case _: Period => "Period"
      case _: Place => "Place"
      case _: Agent => "Agent"
      case _: Person => "Person"
      case _: Organisation => "Organisation"
      case _: CompoundSubject => "CompoundSubject"
      case _: Subject => "Subject"
      case _: Genre => "Genre"
      case _: Medium => "Medium"
      case _: GenreTrait => "GenreTrait"
    })

    val genreTypes = work.genres.map(_ => "Genre")

    println(productionEventTypes)
    println(contributorTypes)
    println(conceptTypes)
    println(genreTypes)
  }
}


WorkApp.run(
  Work(
    production = List(
      ProductionEvent(
        label = "21st century, Tamale, Mr. Razak.",
        concepts = List(
          Period("21st century"),
          Place("Tamale"),
          Person("Mr. Razak")
        )
      ),
      ProductionEvent(
        label = "21st century, Accra, Ghanatrase.",
        concepts = List(
          Period("21st century"),
          Place("Accra"),
          Organisation("Ghanatrase")
       )
      ),
      ProductionEvent(
        label = "Just an agent",
        concepts = List(
          Agent("Just an agent")
        )
      )
    ),
    contributors = List(
      Contribution(role = Blacksmith, agent = Person("Bex Simon")),
      Contribution(role = Typesetter, agent = Person("Marian Bantjes")),
      Contribution(role = Blacksmith, agent = Person("Tubal-cain")),
      Contribution(role = Publisher, agent = Person("Unbound")),
      Contribution(role = Contributor, agent = Agent("Rumpelstiltskin"))
    ),
    subjects = List(
      CompoundSubject(
        label = "20th century, Music, Kojo Antwi",
        concepts = List(
          Subject("20th century"),
          Subject("Music"),
          Person("Kojo Antwi")
        )
      ),
      CompoundSubject(
        label = "France, Medicine",
        concepts = List(
          Place("France"),
          Subject("Medicine")
        )
      )
    ),
    genres = List(
      Medium("Music"),
      Genre("Hip-life")
    )
  )
)
// List(List(Period, Place, Person), List(Period, Place, Organisation), List(Agent))
// List(Person, Person, Person, Person, Agent)
// List(List(Subject, Subject, Person), List(Place, Subject))
// List(Genre, Genre)
```

There's some weirdness in there, but it seems to cover all our use cases?

There's a little bit of weirdness in there - but it seems easier to fathom?

alicefuzier:
There are various states that have to be codified, not just one: `Unidentifiable` which becomes `Unidentified` after the idminter and `Identifiable` which becomes `Identified` after the id minter. The reason for having the `Unidentifable` `Unidentified` ones, which admittedly don't add information, is to be able to express things like contributors or subjects which are lists of things, some of which can have identifiers and some not. 

What I like about the current model is that it codifies the constraints around how things are identified in the pipeline (I think @warrd described them as state machine at one stage, which I think is a nice way of thinking about it). Ie. if something does not have a `sourceIdentifier`, it will never have a `canonicalId` or if it doesn't have a `sourceIdentifier` it also doesn't have `otherIdentifiers` or things before the idminter will never have a `canonicalId`, etc. 

Using `Option`s and `List`s everywhere works, but loses the ability to represent this.
Your model would simplify things, but it would lose these constraints and I think it's nice to have them encoded in the type system. 

I'm definitely biased, cause this modeling is my doing, but I llike how it encodes in the type system the states of identity throught the pipeline. It means we never have to write `.get` where we know there is an identifier or handle None cases that we know are impossible, etc

warrd:
I definitely agree with @alicefuzier that having the constraints codified in a typesafe manner is a really good thing.

I also think the current modelling could be improved: it seems to be a source of confusion for new people coming to the project (it definitely was for me for quite a while), and has a few idiosyncrasies which I think could be avoided:

* I think the complex nesting of types can be avoided (e.g. `List[IdState[Subject[IdState[AbstractRootConcept]]]]` :confounded:).

* The fact that we have the actual object of interest nested under `agent` (or `thing`) is pretty annoying / confusing, and could also probably be avoided with a different implementation.

* I think the use of a string (`identifiedType`) to implement the transitions of the "FSM" is not ideal, and goes against the typesafe nature of the rest of the implementation, so this could be improved (I am probably missing lots of context on why it is this way)

* The `Displayable` / `MaybeDisplayable` types complicate the hierarchy further, and whilst these might still be required to codify certain constraints, I think the implementation could be easier to parse.

I have tried implementing a quick sketch of what an alternative implementation might look which solves some of the above:

```scala
sealed trait HasIdState[State <: IdState] {
  val state: State
}

sealed trait IdState

sealed trait MaybeDisplayable extends IdState

sealed trait Displayable extends IdState

case object Unidentifiable extends IdState with MaybeDisplayable with Displayable

case class Unidentified(
  otherIdentifiers: List[SourceIdentifier] = List(),
) extends IdState with MaybeDisplayable

case class Identified(
  canonicalId: String,
  otherIdentifiers: List[SourceIdentifier] = List(),
) extends IdState with Displayable
```

Which could then be used like this:

```scala
case class UnidentifiedWork(
  data: WorkData[MaybeDisplayable]
)

case class WorkData[State <: IdState](
  // previously List[IdState[Subject[IdState[AbstractRootConcept]]]]
  subjects: List[Subject[State]]
)

case class Subject[State <: IdState](
  state: State,
  label: String,
  concepts: List[AbstractRootConcept[State]] = Nil,
  ontologyType: String = "Subject"
) extends HasIdState[State]

sealed trait AbstractRootConcept[State <: IdState] extends HasIdState[State] {
  val label: String
}

case class Concept[State <: IdState](
  label: String,
  state: State
) extends AbstractRootConcept[State]

// etc...
```

Not suggesting we try and implement this yet: this is just a quick look at the problem and probably has downsides of it's own, not to mention changing the model is probably a very big task. But might be useful for future consideration.

alicefuzier:
>I think the use of a string (`identifiedType`) to implement the transitions of the "FSM" is not ideal, and goes against the typesafe nature of the rest of the implementation, so this could be improved (I am probably missing lots of context on why it is this way)

Yeah, this is definitely true. Unfortunately, this is required by the idminter. When it adds the `canonicalId` it needs to change whatever it added the type to the new state (ie `Identifiable` => `Identified`). 
If the idminter wasn't work-model agnostic, we wouldn't need that. Although I think the fact that it is work-model agnostic is quite handy...

>The `Displayable` / `MaybeDisplayable` types complicate the hierarchy further, and whilst these might still be required to codify certain constraints, I think the implementation could be easier to parse.

Yes, and they are also very badly named, especially since we also have Invisible works, they don't really make sense. What they really mean is 
 - `Displayable`: after the id minter
 - `MaybeDisplayable`: before the id minter
It's a bit hard to come up with good names ad `Identified` is already taken.
I am not sure if they need to exist, I would assume so, but they definitely need to be renamed

I really like your solution of inverting the wrapping. It definitely makes it more readable and it is a very simple solution. I think we should try to give it a go

warrd:
> If the idminter wasn't work-model agnostic, we wouldn't need that. Although I think the fact that it is work-model agnostic is quite handy...

Have hardly touched the ID minter, so can you provide some context about why being model agnostic is a good feature? The rest of the pipeline is tied to a specific model, and we seem to always deploy it together, so can't think of the benefits of the top of my head.

> Yes, and they are also very badly named, especially since we also have Invisible works, they don't really make sense. 

Hmmm how about `Unminted` / `Minted`? At least that is clearer regarding what exactly they are, and less likely to be confused with invisible etc

alicefuzier:
I like `Unminted` / `Minted`

> can you provide some context about why being model agnostic is a good feature?

In general the fact that the idminter is model agnostic means that we don't have to change it every time there is a new field that can be identified. 
If we were to make it tied to the work model we would still have to find a way of making it able to add identifiers to anything that is `Identifiable` in the structure, without us having to manually make code changes for every model change, cause that would be annoying.  I'm sure there are ways around that though.
Part of the reasoning at the time to make it model agnostic was that it could be a faily generic service potentially useful outside wellcome. At the time we had some aspiration of making this stuff not, just public but also packaged and release to other people. @kenoir had strong opinions about that

warrd:
:+1: thanks for letting me know all this. I would probably feel safer if the ID minter made use our types (either by being tied to our particular models, or by being made model agnostic but via use of traits or something like that), but also things are working as is and seems unnecessary to modify that aspect of things at the time being.

If you are happy with it, when I get a chance I will have a go trying to implement the solution I suggested above.

alicefuzier:
Yep, It seems a really elegant and clever solution to some of this ugliness. 
I'm going to close this PR, we can carry on this conversation in a new PR.

PR #344, opened 2020-01-09
Remove the SQL store from the id_minter

kenoir:
In order that we have predictable, sensible, scalable infrastructure the id_minter should no longer use a SQL database.

There have been issues with the id_minter running out of db connections in the past - RDS is an unnecessary fixed cost and is a scaling bottleneck. It is also an unusual and unexpected architectural choice in the context of the the rest of the pipeline architecture.

This PR will provide parallel storage in DynamoDB (while continuing to retrieve and store in RDS) to ensure continuity of service, testing and a safe and easy rollback migration path in the event of DynamoDB turns out to be a bad choice.

**Note:**

This is a _work in progress_, and will require a migration from SQL to DynamoDB of existing identifiers (this PR allows decoupling of that process from deployment of code changes).


alicefuzier:
😱

alexwlchan:
Is this ready for detailed review yet, or just here for visibility?

kenoir:
@alexwlchan just here for discussion right now. We should have done this ages ago.

alicefuzier:
You are definitely right that there's no reason other than historical reasons to have mysql and that it could be switched to dynamo and we would achieve all of the things you say. 
I'm not sure about how valuable those things are. 
I don't think dynamodb would be that much less expensive, I think it would be in the same cost range. 

As for mysql is a bottleneck, I think in practice that doesn't show. A reindex takes 4/5 hours because the transformer takes 4/5 hours , everything seems backed up right at the start of the pipeline. But most importantly, we don't really care, we don't need it to be faster, which is why we aren't speeding up the transformer.

It's true that it would make the service more consistent with the rest, but it's a service that we never change so I'm not sure whether that would make any future development easier.

It's true that it causes issues when scaling up the idminter and those are annoying, but I feel that is the only real tangible advantage we would get.

I'm not sure if that on its own is worth a big refactoring like this one, especially given that it needs to be taken into account how to migrate the data and how potentially to store the ids into two databases while we migrate. 
I feel that migrating the idminter adds lots of room for bugs and lots of room for error in an otherwise very stable code.

TLDR: I don't think it's worth the time/effort/risk..

kenoir:
> TLDR: Yes, we should have done it, but now I don't think it's worth the time/effort/risk

Damn your facts.

PR #348, opened 2020-01-13
Extract access conditions from METS

warrd:
https://github.com/wellcometrust/platform/issues/3946

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
@@ -4,8 +4,8 @@ import java.time.Instant
 
 case class AccessCondition(
   status: AccessStatus,
-  terms: Option[String],
-  to: Option[Instant]
+  terms: Option[String] = None,
+  to: Option[Instant] = None
 )
 
 sealed trait AccessStatus
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
@@ -74,8 +74,10 @@ case class MetsData(
       .map(_.toLowerCase)
       .map {
         case "open" => Right(AccessStatus.Open)
-        case "requires registation" => Right(AccessStatus.OpenWithAdvisory)
+        case "requires registration" => Right(AccessStatus.OpenWithAdvisory)
         case "restricted" => Right(AccessStatus.Restricted)
+        case "clinical images" => Right(AccessStatus.Restricted)
+        case "closed" => Right(AccessStatus.Closed)
         case "in copyright" => Right(AccessStatus.LicensedResources)
         case status => Left(new Exception(s"Unrecognised access status: $status"))
       }
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
@@ -7,7 +7,7 @@ import org.apache.commons.lang3.StringUtils.equalsIgnoreCase
 
 case class MetsData(
   recordIdentifier: String,
-  accessCondition: Option[String],
+  accessConditionDz: Option[String],
   thumbnailLocation: Option[String] = None
 ) {
 
@@ -52,7 +52,7 @@ case class MetsData(
   // sometimes the label (ie "in copyright")
   // and sometimes the url of the license
   private def parseLicense = {
-    accessCondition.map { accessCondition =>
+    accessConditionDz.map { accessCondition =>
       License.values.find { license =>
         equalsIgnoreCase(license.id, accessCondition) || equalsIgnoreCase(
           license.label,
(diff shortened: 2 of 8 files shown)


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
```diff
@@ -52,16 +53,68 @@ case class MetsXml(root: Elem) {
...
+    *          <mods:accessCondition type="player">63</mods:accessCondition>
+    *          <mods:accessCondition type="status">Open</mods:accessCondition>
+    *          ...
+    *        </mods:mods>
+    *      </mets:xmlData>
+    *    </mets:mdWrap>
+    *  </mets:dmdSec>
+    *
+    *  The expected output would be: "Open"
+    */
```
alexwlchan: 🥰

pipeline/transformer/transformer_mets/src/test/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsDataTest.scala
```diff
@@ -146,4 +147,50 @@ class MetsDataTest
...
+            AccessCondition(
+              status = AccessStatus.Restricted,
+              terms = Some("Please ask nicely")
+            )
+          )
+        )
+    }
+  }
+
+  it("fails createing a work when unknown AccessStatus") {
```
alexwlchan:
```suggestion
  it("fails creating a work when unknown AccessStatus") {
```

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
```diff
@@ -4,8 +4,8 @@ import java.time.Instant
 
 case class AccessCondition(
   status: AccessStatus,
-  terms: Option[String],
-  to: Option[Instant]
+  terms: Option[String] = None,
+  to: Option[Instant] = None
```
jamesgorrie: I'm not a massive fan of having these in the actual internal model. It feels like being explicit in the implementing is better and easier to understand?

Why hide it? It's caught us up before, especially with options where `None` would work for either one of these. So if by accident I put `AccessCondition(status, None)` that's valid and doesn't say very much.

The extra code written I feel adds more clarity which I value over succinctness.

Saying that for tests and methods they seem fine, just not on internal models.

There's a good article here:
https://blog.ssanj.net/posts/2019-05-01-why-are-default-parameter-values-considered-bad-in-scala.html

@alicefuzier thoughts?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
```diff
@@ -40,19 +43,22 @@ case class MetsData(
     reason = Some("METS work")
   )
 
-  private def digitalLocation(maybeLicense: Option[License]) = {
-    val url = s"https://wellcomelibrary.org/iiif/$recordIdentifier/manifest"
+  private def digitalLocation(license: Option[License],
```
jamesgorrie: Why change the name? It's a convention used elsewhere?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
```diff
@@ -63,14 +69,27 @@ case class MetsData(
           Left(new Exception(s"Couldn't match $accessCondition to a license"))
       }
     }.sequence
-  }
 
-  private def sourceIdentifier = {
+  private val parseAccessStatus: Either[Exception, Option[AccessStatus]] =
+    accessConditionStatus
+      .map(_.toLowerCase)
```
jamesgorrie: Do we know that we have different versions of the casing in the data? This implies that.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
warrd: We use a mixture of both all over the place. Here the optional `license` field on `DigitalLocation` which we are assigning to isn't prefixed with a `maybe`, so naming it like this is consistent with that.

I'm not a big fan of using the `maybe` prefix in general, as that information is encoded within the type. Similarly, I wouldn't name all iterators or list types like `listWorks` / `iteratorWorks`, and just use `works`.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
warrd: Nope have no idea what form the data takes tbh, I just did this as the data is typically pretty messy. Happy to change if you have a strong aversion to it (that recent bug with pattern matching and lower casing is still fresh in my mind, so that could be a reason to not have this).

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
jamesgorrie: > as that information is encoded within the type

I'm more than happy with that.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
warrd: There are some valid points, but due to the nature of our models (large case classes with lots of optional args) I think not having any default values would have a substantial impact on the ergonomics of using them (e.g. we can currently use constructs like `AccessCondition(status = _)` to make anonymous functions in a really clear way etc).

It would be a bit of a code smell to me if I saw `AccessCondition(status, None)`, especially as there is no context here as to what `None` references. For that reason it might be good for us to have a convention of always initialising case classes with named arguments, rather than positional ones (doing this would avoid pretty much all the problems alluded to).

PR #355, opened 2020-01-16
Simpler modelling of IDs

warrd:
## Issue

https://github.com/wellcometrust/platform/issues/4179

## Description

This work came out of conversations in #342.

The way IDs are modelled currently has a number of issues, and is a source of confusion for new comers to the codebase:

* Types that can have IDs attached need to be wrapped in one of the `Unidentifable` / `Identifiable` / `Identified` wrapper types, meaning to access the actual object we need to look under the confusingly named `agent` value

* To be able to abstract away the work model in various states (i.e. `Unminted`, `Minted`), with the  current method we needed to use higher kind types, which adds additional complexity

* Due to the above two points there multiple places where the type definitions become very long and cumbersome (e.g. `List[IdState[Subject[IdState[AbstractRootConcept]]]]`)

* Pattern matching on ADTs is made more verbose in many places, as we have to take into account the different sorts of ID types something can be wrapped in.

This PR attempts to address the above. Instead of having a wrapper type paremeterised with the data type in question, we allow the data type to be parameterised by an `Id`:

```scala
case class Concept[Id](
  id: Id,
  label: String 
)
```

This `Id` will be one of the `Unidentifiable` / `Identifiable`, `Identified` types, depending on whether we are pre or post ID minter.


jamesgorrie:
I did this a while back too, might be relevant:
https://gist.github.com/jamesgorrie/aebfa6c4ebcca1793a4ccb181e6ab524

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentityState.scala
```diff
@@ -1,41 +1,39 @@
...
-    with MultipleSourceIdentifiers {
-  def withAgent[A](f: T => A) = this.copy(agent = f(agent))
+case class Identifiable(
+  sourceIdentifier: SourceIdentifier,
+  otherIdentifiers: List[SourceIdentifier] = Nil,
+  identifiedType: String = classOf[Identified].getSimpleName,
+) extends IdState
+    with Unminted {
+  def id = None
+  def otherIds = sourceIdentifier +: otherIdentifiers
```
jamesgorrie: Is there a reason we don't differentiate between source and other here, but do elsewhere?

Just trying to get around  why it's useful to sometimes have source / other IDs and other times just to have the minted ID and others.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/InstantRange.scala
```diff
@@ -0,0 +1,39 @@
...
+    InstantRange(
+      from.toInstant(ZoneOffset.UTC),
+      to.toInstant(ZoneOffset.UTC),
+      label
+    )
+
+  def parse(label: String)(
+    implicit parser: Parser[InstantRange]): Option[InstantRange] =
+    parser(label)
+}
```
alexwlchan: Meta-comment: given the overall size of this PR, it'd be helpful if changes like this (e1d30dd, splitting out InstantRange, which are much smaller and simpler to review) were split out into a separate PR.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentityState.scala
```diff
@@ -1,41 +1,39 @@
 package uk.ac.wellcome.models.work.internal
 
-sealed trait IdentityState[+T]
-
-sealed trait Unminted[+T] extends IdentityState[T] {
-  val agent: T
-  def withAgent[A](f: T => A): Unminted[A]
+sealed trait IdState {
```
alexwlchan: Some comments explaining how you expect these traits/case classes to be used would be helpful, e.g. looking at this file I'm not clear on the distinction between "Minted" and "Unminted".

warrd: :+1: none of these traits / case classes are actually new, just that the structure has been modified a bit. Definitely agree now is a good time for some documentation, so will add some in.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentityState.scala
warrd: This is equivilent to the `identifiers` method in the `MultipleSourceIdentifiers` trait that already exists: https://github.com/wellcometrust/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/MultipleSourceIdentifiers.scala#L7

Agree the naming isn't ideal, any suggestions  that would help make things a bit clearer?

warrd: That works for me

kenoir:
Holy 💩this is an epic PR. 

In principle this is a **very good idea**, but it's practically un-reviewable. I'd say it is not mergeable, but it will (obviously) need thorough testing after deployment to a staging environment.

I've a couple of questions better delivered in person.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentityState.scala
warrd: Done this :)

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentityState.scala
warrd: Have added some doc strings to this file

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -711,7 +712,7 @@ class ElasticsearchServiceTest
   }
 
   private def createItemWithLocationType(
-    locationType: LocationType): Identified[Item] =
+    locationType: LocationType): Item[Minted] =
```
alicefuzier: Why `Minted` and not `Identified` here?

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2FiltersTest.scala
```diff
@@ -38,8 +38,7 @@ class ApiV2FiltersTest extends ApiV2WorksTestBase {
   }
 
   describe("filtering works by item LocationType") {
-    def createItemWithLocationType(
-      locationType: LocationType): Identified[Item] =
+    def createItemWithLocationType(locationType: LocationType): Item[Minted] =
```
alicefuzier: Same here

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayAbstractRootConcept.scala
```diff
@@ -14,26 +14,13 @@ sealed trait DisplayAbstractRootConcept {
...
-          includesIdentifiers)
-      case concept @ Unidentifiable(_: AbstractConcept) =>
-        DisplayAbstractConcept(
-          concept.asInstanceOf[Minted[AbstractConcept]],
-          includesIdentifiers)
-      case concept @ Identified(_: AbstractConcept, _, _, _) =>
-        DisplayAbstractConcept(
-          concept.asInstanceOf[Minted[AbstractConcept]],
-          includesIdentifiers)
+      case agent: AbstractAgent[Minted] =>
```
alicefuzier: So much nicer!

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentityState.scala
```diff
@@ -1,41 +1,50 @@
...
+  def allSourceIdentifiers: List[SourceIdentifier]
 }
 
-sealed trait Minted[+T] extends IdentityState[T] {
-  val agent: T
-  def withAgent[A](f: T => A): Minted[A]
+/** Parent trait for an ID of an object that is pre minter. */
+sealed trait Unminted extends IdState
+
+/** Parent trait for an ID of an object that is post minter. */
```
jamesgorrie: My only confusing, which predates this PR, is that `Minted` is the state that it has passed through the minter, rather than being minted. PreMinter / PostMinter, as the comment says might make more sense, but, bah.

jamesgorrie: I still don't see though where this is useful?

warrd: It is useful as anywhere we have an object which is parameterised by `Minted` (or vica versa with `Unminted`), we have made a guarantee that the object has either had a `canonicalId` assigned, or is an object that is unidentifiable. This helps from both the standpoint about reasoning about our code and ensuring it is correct, and reducing the verbosity (for example not having to pattern match against a not yet identified object in such cases, which is something which should never occur within certain parts of the application)

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
warrd: I think I did it like this as we are generally parameterising by `Minted` / `Unminted` in other declarations, rather than the concrete types. We never use these as an `Item[Identified]` (only as `Item[Minted]`), so being consistent should hopefully make the code more predictable on the whole.

PR #356, opened 2020-01-16
add ID search to API

jamesgorrie:
Extends on the ability to search IDs fields more affectively.

Previously
- searched exact, case sensitive single IDs

Now
- Multiple ID search
- Case insensitive
- ID search can be combined with other search e.g. "Standing on the wrong side of a horse 123abc" will return 2 results, the ID and the [standing on the wrong side of a horse](https://wellcomecollection.org/works/pczdvhdg)

Ping @jennpb 

**Requires reindex**


harrisonpim:
shouldn't make a significant difference with the `IdQuery` in there, but would we be safer wrt future changes by leaving the identifiers in the `BaseQuery`?

jamesgorrie:
> shouldn't make a significant difference with the IdQuery in there, but would we be safer wrt future changes by leaving the identifiers in the BaseQuery?

Yes - but I was balancing improving search code and testing, I'll be more mindful that we're testing in future. 👩‍💻  as though it will all be 🗑️ed

PR #357, opened 2020-01-16
Write a better description of how the Sierra adapter works

alexwlchan:


sierra_adapter/README.md
```diff
@@ -0,0 +1,288 @@
...
+*   An *item* is a retrievable object – for example, a physical copy of a book, or an electronic book that you can download.
+
+There is a many-to-many mapping between bibs and items:
+
+*   A single bib can be associated with multiple items: for example, we might have multiple physical copies (the items) of a book (the bib).
+*   A single item can be associated with multiple bibs: for example, i15699833 is a single physical item (a bound volume of patents), and each of the individual patents has its own bib number.
+
+<img style="width: 400px" src="images/shape_of_the_data.svg">
+
+When we create a Work from the Sierra data, we take a single bib and all the items it is associated with.
```
jamieparkinson: I think it would be clearer to put this under a different heading (maybe even as an addendum), with some context that this creation of a work, and indeed the relationship illustrated in `work_from_sierra.svg`, are not things that are really performed or represented in the adapter.

Perhaps you could even remove this bit entirely / put it in a different document (eg a Sierra transformer document).

sierra_adapter/README.md
```diff
@@ -0,0 +1,288 @@
...
+We turn the bib into a Work, and the items into Items.
```
jamieparkinson: Similar note - not clear without context what the difference between `items` and `Items` are / that `Items` are not something to do with Sierra.

sierra_adapter/README.md
```diff
@@ -0,0 +1,288 @@
...
+## How the adapter stores/processes data
+
+This is the architecture of the adapter:
+
+<img src="images/adapter_architecture.svg">
+
+We have the Sierra API on the left, and eventually we end up with a Sierra source data record in the VHS on the right.
+(The VHS is a data store used by the pipeline.)
+
+A "source data record" is a bib with all its associated items – the necessary pieces for a catalogue Work.
```
jamieparkinson: Maybe this is all you need to say re `Work`s <> `Item`s? I think this is clearer than the above and communicates what's needed.

alexwlchan:
> why isn't this in GitBook (/docs)? I think it should be; there's some related stuff in there already.

Is that GitBook rendered anywhere? I couldn't find it at https://docs.wellcomecollection.org.

My preference is to keep the docs with the code it describes, although if there is a pretty rendered version somewhere I'd link to it from a README in this folder.

(What is the state of our GitBook deployment? I feel this has rather passed me by.)

jamieparkinson:
It's rendered [here](https://docs.wellcomecollection.org/catalogue/) but agreed that this isn't particularly well advertised - pinging @jamesgorrie for thoughts on that + more info re the state of it.

fwiw, in the ongoing process of getting up to speed with all our stuff, I've found colocated docs/code to be a bit of an antipattern in terms of discoverability and have missed stuff and wasted time looking for it: you can only find the docs if you know where stuff lives in the codebase, so if you're new to the codebase you struggle to find the docs. Linking to a rendered version from a README sounds like the perfect solution!

sierra_adapter/README.md
alexwlchan: I think it does belong here (at least part of it) – the Sierra adapter needs to produce the information that the transformer needs to create a Work. I've pulled it into a separate section.

alexwlchan:
@jamieparkinson Is the new edition any better?

alexwlchan:
> It's rendered here but agreed that this isn't particularly well advertised - pinging @jamesgorrie for thoughts on that + more info re the state of it.

I don't actually care where we put the docs, but currently we have at least *four* places where they could live:

* Colocated with the code (where I put this README and would expect to find it)
* In the root of the associated monorepo (where there's another Sierra adapter doc)
* In the Gitbook (where there are two documents about Sierra, but in different places and not linked to each other)
* In the standalone docs repo

My preference is very much for (1), which is why this PR is laid out the way it is, but frankly that's less important than **picking a single place to put them**. Which we have thus far singularly failed to do.

PR #360, opened 2020-01-17
Reorganise the API filter tests

alexwlchan:
Some initial refactoring ahead of https://github.com/wellcometrust/platform/issues/4180

The API filter tests were organised inconsistently – some by the filter type, some by listing/searching works. This PR moves the tests around so they're better organised (in particular, they match the order in which filters are defined in WorkFilter.scala), so I can add new tests to resolve that issue.

TODO: Do we test that multiple filters can be combined?

### DIFF
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
@@ -65,7 +65,7 @@ trait ApiWorksTestBase
   def resultList(apiPrefix: String,
                  pageSize: Int = 10,
                  totalPages: Int = 1,
-                 totalResults: Int = 1) =
+                 totalResults: Int) =
     s"""
       "@context": "${contextUrl(apiPrefix)}",
       "type": "ResultList",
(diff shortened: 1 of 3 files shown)


jamesgorrie:
@alexwlchan looking at it - some tests are structured

```scala
it ("filters by {FilterName}") {
  describe("when listing works")
  describe("when searching works")
}
```

As much as this is great _if_ the filter can be proven via a test to be true for listing and searching, it would negate the need for this.

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
```diff
@@ -65,7 +65,7 @@ trait ApiWorksTestBase
   def resultList(apiPrefix: String,
                  pageSize: Int = 10,
                  totalPages: Int = 1,
-                 totalResults: Int = 1) =
+                 totalResults: Int) =
```
jamesgorrie: 👍

alexwlchan:
> As much as this is great, if the filter can be proven via a test to be true for listing and searching, it would negate the need for this.

Isn't that what this test is doing? We need to test the filter is picked up on both listing and searching.

I'm assuming the mechanism is broadly similar for language/subject/genre/whatever queries, so it's enough to check a single filter that we get the filters for both search and list – but since those are different code paths (or they used to be), we need a test for both options.

alexwlchan:
@jamesgorrie Any objections to me merging this?

jamesgorrie:
@alexwlchan apologies, I thought the intention might be to extend that sub test to all the filter tests, I was just wondering about making it a more explicit "The list / query endpoints use the same filters".

PR #362, opened 2020-01-20
Delete records in the Sierra adapter, even if they were modified on the same day

alexwlchan:
Fix for the first half of https://github.com/wellcometrust/platform/issues/4173

Waiting for somebody to confirm if records can ever be un-deleted.


sierra_adapter/sierra_reader/src/test/scala/uk/ac/wellcome/platform/sierra_reader/flow/SierraRecordWrapperFlowTest.scala
```diff
@@ -39,7 +39,7 @@ class SierraRecordWrapperFlowTest
     testWith(wrapperFlow)
   }
 
-  it("creates a SierraRecord from a bib") {
+  it("pares a bib record from the stream") {
```
alicefuzier:
```suggestion
  it("parses a bib record from the stream") {
```

PR #363, opened 2020-01-21
Split out InstantRange

warrd:
Originally from #355

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
@@ -1,8 +1,6 @@
 package uk.ac.wellcome.models.work.internal
 
-import java.time.{Instant, LocalDate, LocalDateTime, ZoneOffset}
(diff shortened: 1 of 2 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -18,45 +16,6 @@ object Concept {
...
-// might extend AbstractConcept
-case class InstantRange(from: Instant,
-                        to: Instant,
-                        label: String = "",
-                        inferred: Boolean = false) {
-
-  def withInferred(inferred: Boolean): InstantRange =
-    InstantRange(from, to, label, inferred)
-
-  // TODO: is label necessary? appears to always be same as Period.label
```
alexwlchan: You've lost this comment and the one above it.

warrd: Have put the first comment back in, but left the second out (it was me when first coming to the code base, and don't think it's an issue now that I have had some more time understanding how we are modelling the data etc)

warrd:
@alexwlchan happy with this now?

PR #364, opened 2020-01-21
Add failing access conditions to license parsing

alicefuzier:
Part of https://github.com/wellcometrust/platform/issues/4191

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
@@ -62,21 +62,24 @@ case class MetsData(
       // A lot of METS record have "Copyright not cleared"
       // or "rightsstatements.org/page/InC/1.0/?language=en" as dz access condition.
       // They both need to be mapped to a InCopyright license so hardcoding here
-      case s if s.toLowerCase() == "copyright not cleared" => Right(License.InCopyright)
-      case s if s == "rightsstatements.org/page/InC/1.0/?language=en" => Right(License.InCopyright)
+      case s if s.toLowerCase() == "copyright not cleared" =>
+        Right(License.InCopyright)
+      case s if s == "rightsstatements.org/page/InC/1.0/?language=en" =>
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
```diff
@@ -57,19 +57,30 @@ case class MetsData(
...
-          accessCondition) || license.url.equals(accessCondition)
-      } match {
-        case Some(license) => Right(license)
-        case None =>
-          Left(new Exception(s"Couldn't match $accessCondition to a license"))
+      accessCondition.toLowerCase() match {
+        // A lot of METS record have "Copyright not cleared"
+        // or "rightsstatements.org/page/InC/1.0/?language=en" as dz access condition.
+        // They both need to be mapped to a InCopyright license so hardcoding here
+        case s if s == "copyright not cleared" => Right(License.InCopyright)
```
jamesgorrie: Is there a way to get a list of these to correct in the source data?

Or does the system already allow them to do that? It's definitely worth trying to fix in the source data too.

alicefuzier: I think the "Copyright not cleared" wording is not used anymore so these are old digitised records. If we hade a dashboard that links these logs to the bnumber we could easily supply them to the digitisation team but at the moment it involves trailing the logs and the queues and it's messy and complicated.
To be honest, I think this is incorrect data and these hardcoded fixes are not nice but they are the quickest way to get a clean run right now.
As I said yesterday, I think we should start working on proper visibility into these things, preferably before we add CALM

jamesgorrie: Yip- 💯 agreed that this is the right way to do it, just thinking of helping our pals in collections.

We should probably have the talk of observability within the pipeline soonish.

alicefuzier:
Fixed the tests BTW

alicefuzier:
I'm going to merge this as it a small change and has already been deployed in the new pipeline

PR #365, opened 2020-01-22
Add RefNo to API (via Sierra)

jamesgorrie:
Adding the `RefNo` and `AltRefNo` via the Sierra data as it's pulled across in the Calm sync.

Allows us to start exploring the shape of the response and elastic indexing.

We'll be adding it to the internal model first, we can then work on the display.

So far we have:
```JSON
{
  "collection": {
    "label": "SA/MWF/J/1/15",
    "path": "SAMWF/J/1/15"
  }
}
```

~I might be able to get away with not adding those properties to elastic explicitly, but rather have them computed via analysers.~

All the other information is stored on the collection for querying at index time i.e. The rest of the paths and depth is calculated from the `path_hierarchy` analysed field.

We can calculate these values at render time from the path if / when we need to display them via the API => hence the `paths`& `depth` `val`s on the `Collection` case class.


jamieparkinson:
How come git thinks you're weco-bot? Wrong email address on the commits maybe?

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
```diff
@@ -1,9 +1,33 @@
...
+  val mapping: MappingDefinition
+  val analysis: Analysis
+}
+
+case object WorksIndexConfig extends IndexConfig {
+  // Analysis
+  val pathTokenizer = PathHierarchyTokenizer("path_hierarchy_tokenizer")
+  val pathAnalyzer =
+    CustomAnalyzer("path_hierarchy_analyzer", pathTokenizer.name, Nil, Nil)
+  val analysis = Analysis(
```
alicefuzier: I vaguely remember from the elasticsearch week that we didn't even need to use this and we could just use wildcards given the size of our index. Am I remembering correctly?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Collection.scala
```diff
@@ -0,0 +1,10 @@
+package uk.ac.wellcome.models.work.internal
+
+case class Collection(
+  label: Option[String],
+  path: String,
+) {
+  val paths: List[String] =
+    path.split("/").inits.toList.reverse.drop(0).map(_.mkString("/"))
```
warrd: :thinking: it seems to me like it would be more logical modelling `path` as a `List[String]` (which is serialised / deserialised as a single `/` separated string in Elasticsearch for purposes of display, search etc). But maybe not necessary due to the fact that it comes that way from Sierra, and we want it like that for the token analyser to do it's magic etc. What do other people think?

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
alicefuzier: Also, it would be nice to have this mapping change code in the PR with api code to query the collection field. This way, it's not clear what this analyzer is for

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Collection.scala
jamieparkinson: I have a feeling this doesn't do what you intended it to do - `drop(0)` is a noop and it looks like you'd want `drop(1)` instead. 

It doesn't look like you use `paths` anywhere else and really it's there to calculate `depth` - how about just doing
```
val depth: Int = 1 + path.count(_ == "/")
```

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
jamesgorrie: > api code to query the collection field

Left this out specifically as we're not sure and feels like a separate place to explore - but I wanted to see the data first to explore it.

jamesgorrie: > I vaguely remember from the elasticsearch week that we didn't even need to use this and we could just use wildcards given the size of our index. Am I remembering correctly?

We'd get _some_ query ability, and lose some.

To be clear this PR is to move us in the direction of seeing what we need, as we don't really know yet from the query perspective.

jamesgorrie:
@warrd could you explain any benfits to making each `path` a `List`?

Would you then expect `paths` to be `List[List[String]]`?

jamesgorrie:
I've removed paths for now as it's started conversations of lots of hyperthetical futures, which this PR is meant to start exploring, not implementing.

The purpose of this PR is to __start__ on the journey of rendering simple Collections from a works page. 

@warrd I still have the question of _why_ we should make path `List[String]`?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Collection.scala
jamesgorrie: Correct, I've removed it, when we need it I'll add tests

warrd: > @warrd could you explain any benfits to making each path a List?
> Would you then expect paths to be List[List[String]]?

No about the `List[List[String]]`!!

What I mean is, instead of representing a path such as `"SA/MWF/J/1"` as a `String`, when in Scala land it would be a `List[String]`: i.e. `List("SA", "MWF", "J", "1")`.

The reason for doing this is that it is makes use of Scala's type system to more accurately depict what we are modelling, and removes unnecessary details (such as the fact that `/` is a seperator). Things like the depth of the path or what is the root come out very easily (i.e. `path.length` or `path.head`), and it is easy to manipulate the path independently of the encoding (i.e. can concatenate to the end without having to worry about what separator is used etc).

However, as I'm not sure we are going to be manipulating these paths very much, it might be unnecessary. I guess I tend to err on the side of using types to provide extra detail when possible, but a `String` might be fine in this case. Was just throwing it out there to see what others thought.

jamesgorrie: It makes perfect sense - if we need to start doing this we can always change the model.

The model - as it's currently used for index information as well, will need to maintain the string with the separator and other information that the tokenizer uses. We should probably have that config central if that's how we want to do it.

I am erring on the side of, I will do that when we have the requirement to as it's a small, non-breaking change.

jamesgorrie: > No about the `List[List[String]]`!!

I meant if
```
type Path = List[String]
type Paths = List[Path]

/*
where:
path = List("things", "to", "things")
paths = List(
  List("things"),
  List("things", "to"),
  List("things", "to", "things")
)
*/
```

PR #366, opened 2020-01-22
Increase ingestor batchSize and make it easier to change batchSize

alicefuzier:
Sometimes during reindexes, we see a lot of stuff ending up in the ingestor dlq. No errors in the logs and no failure metrics being sent to cloudwatch. Also, some of them do get ingested and some successes are notified.

My working theory is that the ingestor picks up messages quicker than it's able to ingest them so some of them stay in memory for longer than the visibility timeout of the queue, which means some of them end up redriven in the DLQ even though they haven't actually failed. 
This might be related to the default batchSIze of 100, which seems a bit small, so increasing to see if the problem goes away

### DIFF
--- a/pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/config/builders/IngestorConfigBuilder.scala
+++ b/pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/config/builders/IngestorConfigBuilder.scala
@@ -13,7 +13,7 @@ object IngestorConfigBuilder {
     // TODO: Work out how to get a Duration from a Typesafe flag.
     val flushInterval = 1 minute
 
-    val batchSize = config.getOrElse[Int]("es.ingest.batchSize")(default = 100)
+    val batchSize = config.required[Int]("es.ingest.batchSize")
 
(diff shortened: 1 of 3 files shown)


pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/config/builders/IngestorConfigBuilder.scala
```diff
@@ -23,4 +24,23 @@ object IngestorConfigBuilder {
...
+          throw new RuntimeException(
+            s"$path is invalid type: got $obj (type ${obj.getClass}), expected Int")
+      }
+      .recover {
+        case exc: ConfigException.Missing =>
+          default getOrElse {
+            throw new RuntimeException(s"${path} not defined in Config")
+          }
+      }
+      .get
```
jamesgorrie: 🎸

alicefuzier: Shamelessly copied. Maybe that should be extracted somewhere common but for now, I'm going to merge as is

PR #367, opened 2020-01-22
Remove composite aggregations

jamieparkinson:


api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -52,7 +52,7 @@ class Router(elasticClient: ElasticClient,
                 }
               },
               path("context.json") {
-                getFromFile(context.getPath)
```
jamieparkinson: I put this in here because it's speculative: it might fix https://github.com/wellcometrust/platform/issues/4133. We have a test for whether `context.json` resolves, which doesn't fail, so I am guessing that the issue is related to how the resource is resolved in the running application... Not sure how to test it other than deploying it, and either way this is more succinct (even if it's still broken)

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksAggregationsTest.scala
```diff
@@ -55,22 +55,22 @@ class ApiV2WorksAggregationsTest extends ApiV2WorksTestBase {
                     },
                     {
                       "data" : {
-                        "id" : "d",
-                        "label" : "Journals",
+                        "id" : "k",
```
jamieparkinson: The ordering of these aggregations changed because the terms aggregation sorts by frequency

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
```diff
@@ -69,87 +59,112 @@ object Aggregations extends Logging {
...
+//
+// ## Key-only vs sample document
+// For some aggregation types `T` (in `Aggregation[T]`) we can decode `T` directly from the bucket key -
+// for example with WorkType, we can use `fromCode` to get the label of the WorkType given its id.
+// However, sometimes we need more information from the index, which is achieved using a top hits
+// aggregation of size 1 - see [[uk.ac.wellcome.platform.api.services.ElasticsearchQueryBuilder]].
+//
+// In this case there is a `sample_doc` subaggregation for which we can provide a
+// `path` to the relevant data inside it. For example, the Language type uses this to obtain both the
+// label and id at the path `data.language` within `sample_doc`.
```
warrd: (BIKESHED WARNING!) I tend to use multiline comments for documentation purposes, and inline comments for describing implementation details. Also free text is IMO hard to read when over 80-100 chars, so might be good to limit the line length here.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
```diff
@@ -69,87 +59,112 @@ object Aggregations extends Logging {
...
+  import io.circe.optics.JsonPath._
+
+  private val buckets = root.buckets.arr
+  private val bucketFilteredCount = root.filtered.doc_count.int
+  private val bucketRootCount = root.doc_count.int
+  private val bucketKey = root.key.json
+  private val bucketSampleDoc =
+    root.sample_doc.hits.hits.index(0)._source.json
+
+  def aggregationParser[T: Decoder](
```
warrd: I think this method could be split up into a few smaller sub methods to help with readability. 

Also, have just done a grep and surprisingly this is the only place in the whole repo where there are any explicit `return` statements (this might be another indicator that the method is too long). I tend to prefer a coding style in functional languages where everything is just a simple expression which evaluates to a value, but up to you if you think it helps with readability.

jamieparkinson:
Resolves https://github.com/wellcometrust/platform/issues/4149

PR #370, opened 2020-01-22
Skeleton CALM adapter

warrd:
https://github.com/wellcometrust/platform/issues/4194


warrd:
:+1: yep will need that, but can do in a future PR

PR #371, opened 2020-01-22
Upgrade the Sierra terraform stack to terraform0.12, fix the Sierra reader

alexwlchan:


alicefuzier:
This PR isn't passing @alexwlchan. The sierra_reader is failing the build, something to do with the applications.conf. [This commit](https://github.com/wellcometrust/catalogue/pull/366/commits/3c74dd449400558cb695ba32217d141b6eaa10b9) in the ingestor might be relevant. I copied it from [here](https://github.com/wellcometrust/catalogue/blob/master/pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/config/builders/RDSBuilder.scala#L45) so possibly should be extracted somewhere common, but for the sake of getting this PR merged it would be ok to just copy I think

alexwlchan:
Thanks for flagging, I've pushed a fix.

PR #373, opened 2020-01-24
Log when works fail ingesting

alicefuzier:


pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/services/WorkIndexer.scala
```diff
@@ -63,8 +63,10 @@ class WorkIndexer(
           }
 
           if (actualFailures.nonEmpty) {
-            val failedIds = actualFailures.map(_.id)
-            debug(s"Failed indexing works $failedIds")
+            val failedIds = actualFailures.map { failure =>
+              error(s"Failed ingesting ${failure.id}: ${failure.error}")
```
jamieparkinson: Assuming this change is mostly related to `debug` log level being insufficient - where does the visible log level get set, out of interest?

alicefuzier: It's here https://github.com/wellcometrust/catalogue/blob/master/pipeline/ingestor/src/main/resources/logback.xml#L19 
It's set to INFO by default but it can be overridden with an environment variable

PR #379, opened 2020-01-27
Document shingle intentions

harrisonpim:


docs/search_relevance/query_design.md
```diff
@@ -24,10 +24,35 @@ Each query has:
...
 
+### Shingled titles
+
+| Status | Query name | Ranking evaluation test |
+| :--- | :--- | :--- |
+| TODO | Shingles | TBD |
+
+#### Intentions
+
+Searching for works with structurally important features in the query. For example: a query for `east london` should give higher scores to instances of the phrase with tokens in that order than works which contain the tokens without the given structure. This should be particularly useful for users looking for specific titles, where word-order matters most.
```
jamesgorrie: Could we keep the research intentions and examples / expectations separate? 

I quite like the structure of "This is what I am trying to do, and this is what I expect when trying to do it". Or does it feel unclear when you separate it that way?

docs/search_relevance/query_design.md
```diff
@@ -78,6 +103,7 @@ Searching for a work based on local and external identifiers e.g. Catalogue API
 * Searching for a list of identifiers, I get all the results back
 * Searches should be case insensitive
 * If the search query contains and ID and other input, we should match
+
```
jamesgorrie: ?

harrisonpim: Not sure how this crept in... I assume it's gitbook autoformatting?

docs/search_relevance/query_design.md
harrisonpim: Happy to take out the middle sentence if you'd rather keep examples separate. Just so we're clear though,

> "This is what I am trying to do, and this is what I expect when trying to do it"

Do you mean "I, the user" or "I, the query designer"?

jamesgorrie: What is the intention  of the person searching.
And what is their expectation.

I prefer to not use I or them, and just write the intention, but that's my style rather than it being useful.

harrisonpim: 👍

PR #381, opened 2020-01-28
Pull out ContextFlowOps from mets_adapter

warrd:
The `ContextFlowOps` class in the `mets_adapter` contains some utility methods that will also be useful within the `calm_adapter`, so here it is pulled out into the `big_messaging` common library.

Side note: beyond just this, `big_messaging` has a fair amount of other functionality not related to solely sending "big" messages between services, including VHS and other inter-service messaging and storage related stuff. Maybe worth changing the name of it if someone has any ideas?

### DIFF
--- a/common/big_messaging/src/main/scala/uk/ac/wellcome/bigmessaging/FlowOps.scala
+++ b/common/big_messaging/src/main/scala/uk/ac/wellcome/bigmessaging/FlowOps.scala
@@ -1,7 +1,7 @@
 package uk.ac.wellcome.bigmessaging
 
 import grizzled.slf4j.Logging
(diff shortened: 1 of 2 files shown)


jamieparkinson:
Q: what is the difference between `big_messaging` and `big_messaging_typesafe`? And also for `elasticsearch`/`elasticsearch_typesafe`

warrd:
@jamieparkinson I think it refers to the [typesafe config](https://github.com/lightbend/config) library (now lightbend config :confused:). Basically it requires that extra dependency, and is generally related to configuring something for actually running it on a server, rather than the core functionality.

PR #382, opened 2020-01-28
Logstash transit on API

### LINKED ISSUE #377: Catalogue API logging to Logstash is not working
In order to save on running costs and enable opportunities for handling logs more effectively we want to log to an ELK stack.

There is no logstash service running in the Catalogue API cluster and consequently the API cannot log to elasticsearch through it.

jamesgorrie:
Fixes #377


api/terraform/main.tf
```diff
@@ -21,13 +21,17 @@ module "catalogue_api_prod" {
...
+
+  interservice_sg_id = "${aws_security_group.interservice.id}"
 }
 
 module "catalogue_api_staging" {
   source = "./catalogue_api"
 
   environment           = "${local.staging_name}"
-  api_container_image   = "${module.staging_images.services["api"]}"
+  api_container_image   = "760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/api:a379a2d9ad75cfc91b74ce992f9cd93b5b488f5b"
```
jamieparkinson: Suspect this was a change just for dev?

PR #388, opened 2020-01-29
Update workType labels for asian manuscripts

jamieparkinson:
Resolves https://github.com/wellcometrust/platform/issues/4218

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/WorkType.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/WorkType.scala
@@ -111,7 +111,7 @@ object WorkType extends Enum[WorkType] {
 
   case object ManuscriptsAsian extends Unlinked {
     override val id: String = "b"
-    override val label: String = "Manuscripts, Asian"
+    override val label: String = "Manuscripts"
   }
 
   case object Music extends Unlinked {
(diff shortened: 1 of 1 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/WorkType.scala
```diff
@@ -155,7 +155,7 @@ object WorkType extends Enum[WorkType] {
 
   case object EManuscriptsAsian extends Linked {
     override val id: String = "x"
-    override val label: String = "E-manuscripts, Asian"
+    override val label: String = "Manuscripts"
```
jamesgorrie: Is it worth leaving this here - and maybe leaving a note above saying what the Sierra label is?

jamesgorrie:
Any reason not to have the `cinefilm` in here from wellcometrust/platform#4218?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/WorkType.scala
jamieparkinson: The Sierra label is captured in the name of the object imo

jamieparkinson:
Re cinefilm: `n` is already called "Film". I guess that the request came from someone looking at Sierra rather than the API.

jamesgorrie:
@jamieparkinson I was wondering if we keep what we're changing from in Sierra somewhere in this transformer.

PR #392, opened 2020-01-30
Add image sources to works

jamieparkinson:
As discussed, seeing what this might look like - would be good to get some thoughts


common/internal_model/src/main/resources/identifier-schemes.csv
```diff
@@ -1,17 +1,18 @@
...
-lc-subjects,Library of Congress Subject Headings (LCSH)
-lc-names,Library of Congress Name authority records
-nlm-mesh,Medical Subject Headings (MESH) identifier
-calm-record-id,CALM record identifier
-preservica-guid,Perservica GUID
-calm-authority-code,CALM authority code
-isbn,International Standard Book Number
-issn,ISSN
-mets,METS
+miro-image-number,Miro image number
```
jamieparkinson: CRLF stuff, sorry

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
```diff
@@ -108,17 +117,30 @@ case class MetsData(
...
+                        bnumber: String,
+                        maybeFileId: Option[String]) =
     for {
       location <- thumbnailLocation
       url <- buildThumbnailUrl(location, bnumber)
     } yield
       DigitalLocation(
         url = url,
         locationType = LocationType("thumbnail-image"),
+        imageSourceIds = maybeFileId
```
jamieparkinson: Not completely convinced whether this needs to be here on `thumbnail` but decided it was more consistent

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location.scala
```diff
@@ -10,6 +10,7 @@ case class DigitalLocation(
   license: Option[License] = None,
   credit: Option[String] = None,
   accessConditions: List[AccessCondition] = Nil,
+  imageSourceIds: List[SourceIdentifier] = Nil,
```
jamieparkinson: After some discussion with @warrd, decided not to make this a (parametrised) `Id` type or a wrapped type that implemented `HasId`, given that it will always be `Identifiable` and that this is not the same entity that will later become a fully-formed `Image`

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
alicefuzier: I'm starting to think that we might need to create _special_ digital locations that don't get displayed in the API so that we are not coupling to what gets extracted as a thumbnail or as an item for api display purposes. 
For example, for MIRO images we extract the `thumbnail` as a `DigitalLocation` with the url of a 300x300 image and we also extract the `DigitalLocation` of the `iiif-image` api of the image, and I think this is what we want. 
But for METS we don't currently extract a `iiif-image` url even though we definitely can. So I'm starting to think that we should just extract a field `image` with a `sourceIdentifier` and a `DigitalLocation` with a `url` pointing to a `iiif-image` api. This way we could make the merger always look at the `image` field

jamieparkinson: If we extract just one `image` then how do we deal with multiple images on a source work (eg for METS)? Would that be as simple as making it an `images` field with a list?

I had been thinking that these `imageSourceIds` might get discarded when displayed in the API but agreed that a special location for them is probably better.

alicefuzier: Yep, sorry a field like `images: List[Image]]` where `Image(sourceIdentifier:SourceIdentifier, location:DigitalLocation)`

jamieparkinson:
@alicefuzier had a go at sketching out what an `Image` type that could be used pre- and post-merging would look like, what do you think of this?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
```diff
@@ -173,18 +192,27 @@ case class MetsXml(root: Elem) {
...
+  private def fileObjects: ListMap[String, FileObject] =
     (root \ "fileSec" \ "fileGrp")
       .filterByAttribute("USE", "OBJECTS")
       .childrenWithTag("file")
-      .toMapping(
-        keyAttrib = "ID",
-        valueNode = "FLocat",
-        valueAttrib = "{http://www.w3.org/1999/xlink}href"
-      )
+      .map { file =>
```
alicefuzier: I think you may need to rebase from master. I've done something very similar to this in https://github.com/wellcometrust/catalogue/pull/398

jamieparkinson: Ha, at least we're thinking along the same lines. Will do.

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
```diff
@@ -110,26 +110,49 @@ case class MetsData(
...
+      _.mimeType match {
+        case Some(m) => m == "application/pdf" || m.startsWith("image")
+        case None    => true
+      }
+    }
+    .map { fileObject =>
+      UnmergedImage(
+        getImageSourceId(recordIdentifier, fileObject.id),
+        DigitalLocation(
+          url = buildImageUrl(recordIdentifier, fileObject).getOrElse(""),
```
alicefuzier: I think image urls need to be slightly different. This builds the url to the thumbnail which has 200X200 fixed size. I think it should be something like `https://dlcs.io/iiif-img/wellcome/${space}/${filename}/info.json` so you have access to the full size image or you can request whichever size you want

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -0,0 +1,23 @@
+package uk.ac.wellcome.models.work.internal
+
+sealed trait BaseImage[+Id] extends HasIdState[Id] {
```
alicefuzier: 👍 Looks good to me! I take it that `MergedImage` will be the output of the merger?

pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
jamieparkinson: What do you think we should do for PDFs? We could use the thumbnails, or just return the whole PDF (which seems to be in DLCS for examples I've looked at, is this universal?). If the latter, then I'm not sure what `LocationType` the `DigitalLocation` should have - it doesn't seem like it should be `iiif-image`.

jamesgorrie:
@jamieparkinson @alicefuzier the PR looks good but what's the rest of the plan?

Do we then take the images from the work further down the pipe and spit out works and images?

pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroImage.scala
```diff
@@ -0,0 +1,41 @@
...
+  DigitalLocation,
+  IdentifierType,
+  LocationType,
+  SourceIdentifier,
+  UnmergedImage,
+  Unminted
+}
+import uk.ac.wellcome.platform.transformer.miro.source.MiroRecord
+
+trait MiroImage {
```
warrd: Personally don't like the mixin pattern in scala, as can lead to cases where we extend from something we no longer actually use, and the compiler doesn't catch it (lots of our tests suffer for this). If a trait is just a collection of functions, and not a contract to be implemented, I would generally just pull them out into an `object` or something.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -0,0 +1,29 @@
...
+  val id: Id
+  val location: DigitalLocation
+}
+
+sealed trait ImageState
+sealed trait Merged extends ImageState
+sealed trait Unmerged extends ImageState
```
warrd: A bit unsure what having these `Merged` / `Unmerged` types give us, given that only `UnmergedImage` and `MergedImage` subtype them, so we can also use those types in place of these currently. Is there a plan for them in a future PR?

pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroImage.scala
warrd: (also harder for people who don't use an IDE to understand where a function is coming from :wink:)

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamieparkinson: Yeah, they seemed like they were going to be used but I'll remove them for now, can always add them back later

warrd: Cool, suspected it might have been something like that :+1:

pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroImage.scala
jamieparkinson: I agree and I'll change `MetsDataImageUtils` accordingly, but here in `MiroImage` I was copying the existing convention where it's all traits:
image

I would lean towards making the minimal change but the other options would be to make this an odd-one-out object, or changing all of them to be objects etc  - what do you think?

warrd: Ah yep, I reckon just make the minimal change in that case

warrd: (I mean follow the existing convention of using a trait, not having a one off object)

jamesgorrie: Convention might have happened because others were just following.... 🚶‍♀ 🚶 🚎

warrd: > Convention might have happened because others were just following.... :walking_woman: :walking_man: :trolleybus:

 :yin_yang:  :open_mouth: :yin_yang:

jamieparkinson:
@jamesgorrie exactly - that's what the merger will be extended to do. Some of the high-level things that we think we'll need to do are linked to [this issue](https://github.com/wellcometrust/platform/issues/4176)

common/internal_model/src/main/scala/uk/ac/wellcome/models/Implicits.scala
```diff
@@ -118,4 +125,11 @@ object Implicits extends TimeInstances {
   implicit val _enc48: Encoder[IdentifiedBaseWork] = deriveEncoder
   implicit val _enc49: Encoder[BaseWork] = deriveEncoder
   implicit val _enc50: Encoder[SierraTransformable] = deriveEncoder
+  implicit val _enc51: Encoder[UnmergedImage[Unminted]] = deriveEncoder
+  implicit val _enc52: Encoder[UnmergedImage[Minted]] = deriveEncoder
+  implicit val _enc53: Encoder[MergedImage[Unminted]] = deriveEncoder
+  implicit val _enc54: Encoder[MergedImage[Minted]] = deriveEncoder
+  implicit val _enc55: Encoder[ImageData] = deriveEncoder
+  implicit val _enc56: Encoder[BaseImage[Unminted]] = deriveEncoder
+  implicit val _enc57: Encoder[BaseImage[Minted]] = deriveEncoder
```
warrd: :+1:

alicefuzier:
Not for this PR but do we need to extract image fields in the sierra transformer too? We extract links to related miro works, which will have an image as part of this change, so I'm guessing we don't need to do anything in the sierra transformer cause the merger will have enough to do its thing?

jamieparkinson:
@alicefuzier yes, that was my logic - think this PR sets a precedent that would make it easy to add the images in the Sierra transformer too, if that becomes necessary

PR #393, opened 2020-01-31
Coupling index mapping and internal model

jamesgorrie:
This PR creates a tight coupling between the index mapping and internal model.

I would read "internal model" as "index model" as that's what we serialise to and from on the Elastic search index. 

### Internal model changes
__Ingestor:__ needs updating to serialise model to right JSON

__API:__ needs updating to deserialise model to right JSON

### New mapping settings

__Ingestor:__ Needs updating to create new index with new settings.

__API:__ Needs updating to run new queries / aggs / etc on new data

We create an index which has a human readable prefix with hash of the JSON used to create that index. 

This also means, that if we wanted to tear absolutely everything down, we should be able to checkout at a certain stage and deploy it all.

### Transformers change
TBD

---

## Alternatives
* Version and publish models separately so we can have the API and ingestor / pipeline running on different versions. This seems like an artificial decoupling of services to me. It also is slightly complicated in the sense of dealing with mulltiple versions of packages with SBT


jamesgorrie:
Closing in favour of #395

PR #394, opened 2020-01-31
Retrieve archive data from CALM API

warrd:
### Issue

### Description

Includes logic for fetching CALM data from the API in the adapter.


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmQuery.scala
```diff
@@ -0,0 +1,18 @@
+package uk.ac.wellcome.calm_adapter
+
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+
+sealed trait CalmQuery {
+  def key: String
+  def value: String
+  def queryString = s"$key=$value"
```
alexwlchan: Since you're using the Akka HTTP DSL, you might want to look at the Query parameter type that already exists, which:

1. Can be added to a Uri before constructing a request
2. Will correctly URL-encode values for you

See also: https://github.com/wellcomecollection/stacks-service/pull/64

warrd: This is not used in a URI, but templated into the XML body. Maybe `queryString` is a bit of a misleading name, could call it `queryExpression` to differentiate it from a query string.

warrd:
Thanks @alexwlchan, have addressed your comments

PR #395, opened 2020-02-03
hardcode elastic index name for API

jamesgorrie:
After a few different tries at how to do this, having the index hardcoded seems like a way that makes sense.

This does mean we need to build to have a release, but means fewer steps in deployment.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
@@ -5,6 +5,7 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.ActorMaterializer
 import com.typesafe.config.Config
+import uk.ac.wellcome.elasticsearch.ElasticConfig
 import uk.ac.wellcome.elasticsearch.typesafe.ElasticBuilder
 import uk.ac.wellcome.platform.api.models.ApiConfig
 import uk.ac.wellcome.typesafe.config.builders.AkkaBuilder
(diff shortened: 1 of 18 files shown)


jamesgorrie:
We will definitely have to think about how we deploy the first version of this - basically I think we need to get it into master, and deploy to prod.

We can test this changing the current way of deploying to look at that index - which is currently how things are:

[`api-staging => 20200131`](https://api-stage.wellcomecollection.org/catalogue/v2/works)
[`api => 20200107`](https://api.wellcomecollection.org/catalogue/v2/works)

jamieparkinson:
One thought - is there a way of doing the hardcoding "higher up" in the repo? Feels a little bit hidden if it's deep down in `common/elasticsearch_typesafe/src/main/scala/uk/ac/wellcome/elasticsearch/typesafe/ElasticBuilder.scala`

jamesgorrie:
@jamieparkinson The reason it's there currently is to be shared between the snapshot generator and API. 

Where might you put it?

api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
```diff
@@ -29,7 +27,7 @@ object Main extends WellcomeTypesafeApp {
     val elasticClient = ElasticBuilder.buildElasticClient(config)
 
     val elasticConfig =
-      DisplayElasticConfig(indexV2 = Index(config.required("es.index.v2")))
+      ElasticBuilder.buildElasticConfig()
```
alicefuzier: Any reason to not hardcode it here? Something like 
```
 val elasticConfig = ElasticConfig( indexV2 = Index("v2-20200131"))	 
```
As @jamieparkinson was saying it is a bit hidden in `ElasticBuilder`. Also, it feels like it `ElasticBuilder` should remain fairly flexible if we're going down the route "we're hardcoding this only for the API"

alicefuzier:
Ah, ignore me....I didn't read your reply about the snapshot generator

jamesgorrie:
@alicefuzier moved it into ElasticConfig with a note.
How's that?

alicefuzier:
👍 Merge away!

PR #400, opened 2020-02-04
Store CALM records in adapter

warrd:
## Issue

## Description

This PR adds to the adapter a store for the CALM records, which will be made accessible from the transformer.

Messages are only emitted to the transformer if either

1) We have never seen a record with that `RecordID` before
2) The data for a particular `RecordID` has changed

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
@@ -1,0 +1,32 @@
+package uk.ac.wellcome.calm_adapter
+
+import uk.ac.wellcome.storage.{Identified, Version, NoVersionExistsError}
+import uk.ac.wellcome.storage.store.VersionedStore
+
+class CalmStore(store: VersionedStore[String, Int, Map[String, String]]) {
+
+  type Key = Version[String, Int]
+
+  type Result[T] = Either[Throwable, T]
+
+  def putRecord(record: CalmRecord): Result[Option[Key]] = 
+    shouldStoreRecord(record)
+      .flatMap {
+        case false         => Right(None)
+        case true          => store
+          .putLatest(record.id)(record.data)
+          .map { case Identified(key, _) => Some(key) }
+          .left
+          .map(_.e)
+      }
+
+  def shouldStoreRecord(record: CalmRecord): Result[Boolean] =
+    store
+      .getLatest(record.id)
+      .map { case Identified(_, storedData) => record.data != storedData }
+      .left
+      .flatMap { 
+        case NoVersionExistsError(_) => Right(true)
+        case err => Left(err.e)
+      }
+}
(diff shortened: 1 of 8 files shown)


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
```diff
@@ -0,0 +1,32 @@
...
+          .putLatest(record.id)(record.data)
+          .map { case Identified(key, _) => Some(key) }
+          .left
+          .map(_.e)
+      }
+
+  def shouldStoreRecord(record: CalmRecord): Result[Boolean] =
+    store
+      .getLatest(record.id)
+      .map { case Identified(_, storedData) => record.data != storedData }
```
alexwlchan: This will always store data that’s changed – is there a risk data could arrive out-of-order, and we store a stale version?

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmRetrieverTest.scala
```diff
@@ -154,6 +166,19 @@ class CalmRetrieverTest
     }
   }
 
+  it("fails if no RecordID in the response") {
```
alexwlchan:
```suggestion
  it("fails if there is no RecordID in the response") {
```

calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
warrd: The data which is stored is from calling the API which happens immediately preceding this (during processing of the same message, not in another service or anything). The windows we are using to call the API are also very coarse (the resolution of modified date is only daily).

So for this to happen it seems to me very unlikely (if we are only generating search windows a few times a day). Also, as the data returned from the CALM API does not appear to be versioned, I don't have any suggestions about how we would avoid this, unless you have any clever ideas?

alexwlchan: That sounds reasonable, if you put a comment to this effect alongside this code, explaining why we think “has the data changed” is sufficient. 😉

warrd: One idea I've thought of is to use `Date` timestamp in the API response header. and store this in the store additionally to the data. Then we can check we never store a response further in the past. Maybe not worth it though, I dunno :man_shrugging:

warrd: Giving this more thought, I actually think the whole behaviour of checking the data is changed, and only publishing a message if that is so, is incorrect.

In the METS adapter we still emit messages even when the data is unchanged, due to the fact that the publishing of the message could fail. So we could end up in a state where the data is stored, but publishing to SNS failed, and if we swallow the message the data will never make it to the transformer.

I think we need the same behaviour here (which will of course result in the likelihood of some redundant duplicate messages within the transformer). Maybe @alicefuzier has some thoughts on this?

warrd: (the above comment is sort of orthogonal to the issue originally brought up, but got thinking about it from this thread so seems like as good as place as any for a discussion)

warrd:
Going to merge this as is for now, can continue the conversation about the above separately as will probably have to be a few changes :point_up:

PR #406, opened 2020-02-07
Add extra mappings for accessStatus

alicefuzier:
For https://github.com/wellcometrust/platform/issues/4214

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraLocation.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraLocation.scala
@@ -69,23 +69,28 @@ trait SierraLocation extends SierraQueryOps {
         .subfieldsWithTag("f")
         .contents
         .headOption
-        .flatMap {
+        .map {
           case accessStatus(status) if status == "Open" =>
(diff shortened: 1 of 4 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraLocation.scala
```diff
@@ -69,23 +69,28 @@ trait SierraLocation extends SierraQueryOps {
...
+            AccessStatus.Restricted
           case accessStatus(status) if status == "Cannot Be Produced" =>
-            Some(AccessStatus.Restricted)
+            AccessStatus.Restricted
           case accessStatus(status) if status == "Certain restrictions apply" =>
-            Some(AccessStatus.Restricted)
+            AccessStatus.Restricted
           case accessStatus(status) if status == "Closed" =>
-            Some(AccessStatus.Closed)
-          case _ => None
```
warrd: Does removing this mean this is now a partial pattern match, and will error if something unexpected?

alicefuzier: yes, I modified the corresponding test. That was a bit of a workaround to get a usable index again, but I think we want to know if there are other mappings to add

PR #407, opened 2020-02-07
Don't store CalmRecord's older than the data currently stored

warrd:
## Issue

## Description

Currently we check whether the the Calm record in the store differs to that which has just been retrieved, swallowing the message if the data is unchanged. There are 2 issues relating to this:

1) If messages arrive out of order, we could end up storing older records than what currently exists.
2) If the message fails publishing to SNS post storage, the next time round it is picked up from the queue it will appear as nothing needs to be done, as the data already exists in the store.

To fix the above we do the following:

a) Use the timestamp from the Calm API response header to make it impossible to store older data
b) Always publish the key to SNS if the retrieved data is more recent

Note that point b here will potentially result in many redundant messages to the transformer (especially as we are using a daily window, but will want updates relatively regularly), so to mitigate this in a future PR we may add an extra `published` field to `CalmRecord`, which is written after SNS message is successfully sent.

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
@@ -26,7 +26,7 @@ class CalmStore(store: VersionedStore[String, Int, CalmRecord]) {
       .getLatest(record.id)
       .map {
         case Identified(_, storedRecord) =>
-          record.retrievedAt.toEpochMilli > storedRecord.retrievedAt.toEpochMilli
+          record.retrievedAt.isAfter(storedRecord.retrievedAt)
       }
       .left
       .flatMap {
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
@@ -30,7 +30,8 @@ class CalmStore(store: VersionedStore[String, Int, CalmRecord]) {
           val differingData = record.data != storedRecord.data
           if (sameTimestamp && differingData)
             Left(
-              new Exception("Cannot resolve latest data as timestamps are equal")
+              new Exception(
+                "Cannot resolve latest data as timestamps are equal")
             )
           else
             Right(record.retrievedAt.isAfter(storedRecord.retrievedAt))
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
@@ -26,11 +26,19 @@ class CalmStore(store: VersionedStore[String, Int, CalmRecord]) {
       .getLatest(record.id)
       .map {
         case Identified(_, storedRecord) =>
-          record.retrievedAt.isAfter(storedRecord.retrievedAt)
+          val sameTimestamp = record.retrievedAt == storedRecord.retrievedAt
+          val differingData = record.data != storedRecord.data
+          if (sameTimestamp && differingData)
+            Left(
+              new Exception("Cannot resolve latest data as timestamps are equal")
+            )
+          else
+            Right(record.retrievedAt.isAfter(storedRecord.retrievedAt))
       }
       .left
       .flatMap {
-        case NoVersionExistsError(_) => Right(true)
+        case NoVersionExistsError(_) => Right(Right(true))
         case err                     => Left(err.e)
       }
+      .flatMap(identity)
 }
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
@@ -3,7 +3,7 @@ package uk.ac.wellcome.calm_adapter
 import uk.ac.wellcome.storage.{Identified, NoVersionExistsError, Version}
 import uk.ac.wellcome.storage.store.VersionedStore
 
-class CalmStore(store: VersionedStore[String, Int, Map[String, String]]) {
+class CalmStore(store: VersionedStore[String, Int, CalmRecord]) {
 
   type Key = Version[String, Int]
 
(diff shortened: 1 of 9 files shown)


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmStore.scala
```diff
@@ -24,7 +24,10 @@ class CalmStore(store: VersionedStore[String, Int, Map[String, String]]) {
   def shouldStoreRecord(record: CalmRecord): Result[Boolean] =
     store
       .getLatest(record.id)
-      .map { case Identified(_, storedData) => record.data != storedData }
+      .map {
+        case Identified(_, storedRecord) =>
+          record.retrievedAt.toEpochMilli > storedRecord.retrievedAt.toEpochMilli
```
alexwlchan: Maybe use the comparison helper on `java.time.Instant`?

```suggestion
          record.retrievedAt.isAfter(storedRecord.retrievedAt)
```

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmStoreTest.scala
```diff
@@ -10,39 +11,44 @@ import uk.ac.wellcome.storage.{StoreReadError, Version}
...
-      Version("A", 0) -> Map("key" -> "value")
-    )
+    data.entries shouldBe Map(Version("A", 0) -> record)
   }
 
-  it("stores already seen CALM records when the data has changed") {
-    val data = dataStore(Map(Version("A", 1) -> Map("key" -> "old")))
-    val record = CalmRecord("A", Map("key" -> "new"))
-    calmStore(data).putRecord(record) shouldBe Right(Some(Version("A", 2)))
+  it("stores already seen CALM records when newer data") {
```
alexwlchan:
```suggestion
  it("replaces a previously stored CALM record if the retrieval date is newer") {
```

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmStoreTest.scala
```diff
@@ -10,39 +11,44 @@ import uk.ac.wellcome.storage.{StoreReadError, Version}
...
+    val oldRecord = CalmRecord("A", Map("key" -> "old"), retrievedAt)
+    val newRecord = CalmRecord(
+      "A",
+      Map("key" -> "new"),
+      Instant.ofEpochSecond(retrievedAt.getEpochSecond + 2))
```
alexwlchan: minor: just to make these two line up neatly, maybe pull the times into variables?

```scala
val oldTime = retrievedAt
val newTime = Instant.ofEpochSecond(retrievedAt.getEpochSecond + 2)

val oldRecord = CalmRecord("A", Map("key" -> "old"), oldTime)
val newRecord = CalmRecord("A", Map("key" -> "old"), newTime)
```

Same meaning, but the similarity between oldRecord/newRecord is more apparent.

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmStoreTest.scala
```diff
@@ -10,39 +11,44 @@ import uk.ac.wellcome.storage.{StoreReadError, Version}
...
   }
 
-  it("doesn't store already seen CALM records when unchanged data") {
-    val data = dataStore(Map(Version("A", 4) -> Map("key" -> "new")))
-    val record = CalmRecord("A", Map("key" -> "new"))
-    calmStore(data).putRecord(record) shouldBe Right(None)
-    data.entries shouldBe Map(
-      Version("A", 4) -> Map("key" -> "new"),
-    )
+  it("doesn't store already seen CALM records when older data") {
```
alexwlchan:
```suggestion
  it("does not replace a stored CALM record if the retrieval date on the new record is older") {
```

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmStoreTest.scala
```diff
@@ -10,39 +11,44 @@ import uk.ac.wellcome.storage.{StoreReadError, Version}
...
+    val oldRecord = CalmRecord("A", Map("key" -> "old"), retrievedAt)
+    val newRecord = CalmRecord(
+      "A",
+      Map("key" -> "new"),
+      Instant.ofEpochSecond(retrievedAt.getEpochSecond + 2))
```
alexwlchan: ditto above

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmStoreTest.scala
```diff
@@ -10,39 +11,44 @@ import uk.ac.wellcome.storage.{StoreReadError, Version}
...
+    val data = dataStore(Version("A", 4) -> newRecord)
+    calmStore(data).putRecord(oldRecord) shouldBe Right(None)
+    data.entries shouldBe Map(Version("A", 4) -> newRecord)
```
alexwlchan: There should also be a test here for the boundary case: what if the retrieval date of the stored and new record are exactly the same?

warrd: @alexwlchan I'm not actually sure what the correct behaviour is in this case (where the retrieval date is exactly the same to the second, but the data differs between the two). Maybe the most robust thing to do would be to raise an error, so that the message ends up being retried and we can hopefully resolve the latest data? Seems like such an edge case, but maybe worth it.

alexwlchan: Hmm. I would go the other way – maybe raise an error if the retrieval date matches *and* the data is different? If the data is the same, then it's just SQS double-sending a message, and we can quietly drop it.

warrd: I was suggesting exactly the same thing, sorry if my wording was unclear!

alexwlchan: 🤦‍♀ 

Yes you did, I suck at reading comprehension.

Okay, do that!

warrd: Done here: https://github.com/wellcometrust/catalogue/pull/407/commits/d4162e5b3e73deb632fbb0c10733d7d4d970edbe

PR #408, opened 2020-02-07
CalmRecord should store list of values

warrd:
## Issue

## Description

Repeated keys in Calm XML are not handled, even though they have been seen to exist in the data.

Here we update from a `Map[String, String]` to a `Map[String, List[String]]` to handle multiple values. We will leave it up to the transformer to make sense of this data.

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
@@ -1,7 +1,7 @@
 package uk.ac.wellcome.calm_adapter
 
 import scala.util.Try
-import scala.xml.{Elem, Node, XML}
+import scala.xml.{Elem, Node, NodeSeq, XML}
 import java.time.Instant
 import akka.http.scaladsl.model.headers.Cookie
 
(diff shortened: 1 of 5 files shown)


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
```diff
@@ -96,13 +96,21 @@ case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
...
-        data
-          .get("RecordID")
-          .map(id => Right(CalmRecord(id, data, retrievedAt)))
-          .getOrElse(Left(new Exception("RecordID not found")))
+      .flatMap { nodes =>
+        val data = toMapping(nodes)
+        data.get("RecordID") match {
+          case Some(List(id)) => Right(CalmRecord(id, data, retrievedAt))
+          case Some(_)        => Left(new Exception("Multiple RecordIDs found"))
+          case None           => Left(new Exception("RecordID not found"))
```
jamesgorrie: Ah OK - I was going to do this at the transformer level?

warrd: You are correct for all fields apart from `RecordID`: we need a unique identifier for the VHS key, so need to parse this here.

PR #409, opened 2020-02-10
Add retry with exponential backoff when calling Calm API

warrd:
## Issue

## Description

Due to the fact that we need to process messages in batch (as a single daily window query will return multiple results that need to be handled via separate requests within a session), the current behaviour is not very robust: if an error is encountered on a single request, the whole message will end up being retried rather than just that request.

Here we update the `CalmHttpClient` to include retry with exponential backoff for each request to the Calm API.

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmHttpClient.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmHttpClient.scala
@@ -1,8 +1,11 @@
 package uk.ac.wellcome.calm_adapter
 
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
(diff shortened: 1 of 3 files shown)


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmHttpClient.scala
```diff
@@ -1,18 +1,70 @@
...
 import akka.http.scaladsl.model._
 
 trait CalmHttpClient {
 
   def apply(request: HttpRequest): Future[HttpResponse]
 }
 
-class CalmAkkaHttpClient(implicit actorSystem: ActorSystem)
+abstract class CalmHttpClientWithBackoff(
+  minBackoff: FiniteDuration = 100 milliseconds,
```
jamieparkinson: this is such great syntax

PR #412, opened 2020-02-11
New tf 0.12 pipeline

jamesgorrie:
Moving to tf0.12.

Thought of moving to the new catalogue AWS account too - but then thought against it for timing and not creating too many moving parts.

This is necessary so we can start using the new tf ecs service that allows you to not use cloudfront logs.

### DIFF
--- a/pipeline/terraform0.12/stack/service_id_minter.tf
+++ b/pipeline/terraform0.12/stack/service_id_minter.tf
@@ -1,0 +1,71 @@
+module "id_minter_topic" {
+  source = "../modules/topic"
+
+  name       = "${local.namespace_hyphen}_id_minter"
+//  role_names = ["${module.id_minter.task_role_name}"]
+  role_names = []
+  messages_bucket_arn = aws_s3_bucket.messages.arn
+}
+
+//module "id_minter_queue" {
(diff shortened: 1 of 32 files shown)


pipeline/terraform0.12/stack/service_id_minter.tf
```diff
@@ -0,0 +1,71 @@
+module "id_minter_topic" {
+  source = "../modules/topic"
+
+  name       = "${local.namespace_hyphen}_id_minter"
+//  role_names = ["${module.id_minter.task_role_name}"]
+  role_names = []
+  messages_bucket_arn = aws_s3_bucket.messages.arn
+}
+
+//module "id_minter_queue" {
```
alexwlchan: 👀 

Why is this all commented out?

pipeline/terraform0.12/terraform.tf
```diff
@@ -0,0 +1,90 @@
+provider "aws" {
+  region = "eu-west-1"
+
+  assume_role {
+    role_arn = "arn:aws:iam::760097843905:role/platform-admin"
```
kenoir: Can we run this in the catalogue account now?

PR #413, opened 2020-02-11
Tf0.12 pipeline

jamesgorrie:
Upgrade to tf0.12 for 💸 & 🎉 

Thought about moving the AWS account as well, might do once this is done.

So far I have the sierra transformer and supporting infra, will keep ploughing through.

Related https://github.com/wellcometrust/platform/issues/4257

Supersedes #412

### DIFF
--- a/pipeline/terraform/locals.tf
+++ b/pipeline/terraform/locals.tf
@@ -11,9 +11,7 @@ locals {
   # Sierra adapter topics
   sierra_merged_items_topic_name = "${data.terraform_remote_state.sierra_adapter.merged_items_topic_name}"
   sierra_merged_bibs_topic_name  = "${data.terraform_remote_state.sierra_adapter.merged_bibs_topic_name}"
-  sierra_merged_items_topic_arn = "${data.terraform_remote_state.sierra_adapter.merged_items_topic_arn}"
-  sierra_merged_bibs_topic_arn  = "${data.terraform_remote_state.sierra_adapter.merged_bibs_topic_arn}"
-
+  
   # Mets adapter VHS
   mets_adapter_read_policy = "${data.terraform_remote_state.catalogue_infra_critical.mets_dynamo_read_policy}"
   mets_adapter_table_name  = "${data.terraform_remote_state.catalogue_infra_critical.mets_dynamo_table_name}"
--- a/pipeline/terraform/locals.tf
+++ b/pipeline/terraform/locals.tf
@@ -11,6 +11,8 @@ locals {
   # Sierra adapter topics
   sierra_merged_items_topic_name = "${data.terraform_remote_state.sierra_adapter.merged_items_topic_name}"
   sierra_merged_bibs_topic_name  = "${data.terraform_remote_state.sierra_adapter.merged_bibs_topic_name}"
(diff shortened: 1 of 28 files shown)


pipeline/terraform/locals.tf
```diff
@@ -11,7 +11,7 @@ locals {
   # Sierra adapter topics
   sierra_merged_items_topic_name = "${data.terraform_remote_state.sierra_adapter.merged_items_topic_name}"
   sierra_merged_bibs_topic_name  = "${data.terraform_remote_state.sierra_adapter.merged_bibs_topic_name}"
-
+
```
alexwlchan: boooo why is terraform formatting broken :(

pipeline/terraform0.12/modules/pipeline_service/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+variable "launch_type" {
+  default = "FARGATE"
+}
+
+variable "cpu" {
+  default = "256"
+}
+
+variable "memory" {
+  default = "512"
```
alexwlchan: type hints?

pipeline/terraform0.12/terraform.tf
```diff
@@ -0,0 +1,90 @@
...
+}
+
+terraform {
+  required_version = ">= 0.9"
+
+  backend "s3" {
+    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
+
+    bucket         = "wellcomecollection-platform-infra"
+    key            = "terraform/catalogue_pipeline_tf_0.12.tfstate"
```
alexwlchan:
```suggestion
    key            = "terraform/catalogue/pipeline.tfstate"
```

I've been trying to sort terraform states into per-repo folders in the S3 bucket, so it's a bit easier to find where any given state file is defined from. The platform-infra bucket in particular has a lot of state files now.

Also, let's not bake the terraform version into the name – that feels like a naming smell.

pipeline/terraform0.12/terraform.tf
```diff
@@ -0,0 +1,90 @@
...
+    dynamodb_table = "terraform-locktable"
+    region         = "eu-west-1"
+  }
+}
+
+data "terraform_remote_state" "catalogue_infra_critical" {
+  backend = "s3"
+
+  config = {
+    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
```
alexwlchan: Anything reading remote state should use a read-only role.

pipeline/terraform0.12/modules/pipeline_service/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+
+variable "aws_region" {}
+variable "namespace_id" {}
+variable "container_image" {}
+
+variable "secret_env_vars" {
+  type = map(string)
+}
+
+variable "secret_env_vars_length" {}
```
alexwlchan: All env_vars_length variables should be purged! Purged!

(This is one of the very nice things about terraform 0.12 that we get to ditch.)

pipeline/terraform0.12/terraform.tf
kenoir: I also think this will not work as you need the state from the stack pre update.

PR #416, opened 2020-02-11
Process Calm windows in batches

warrd:
## Issue

## Description

Currently we demultiplex all records within a window into separate messages in a stream. The unintended consequence of this is that when a single record has been processed, the message will be deleted from SQS, even if the others fail.

To prevent this, here we process a window as a single batch, only emitting the delete action when all have completed. Internally this processing of the records in a window is within a stream, in order to retain the advantages of concurrency and backpressure, but we wait for the whole stream to complete before emitting a delete action.

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmRetriever.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmRetriever.scala
@@ -2,14 +2,16 @@ package uk.ac.wellcome.calm_adapter
 
 import scala.concurrent.{ExecutionContext, Future}
 import java.time.Instant
+import akka.NotUsed
 import akka.stream.ActorMaterializer
+import akka.stream.scaladsl._
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.unmarshalling.Unmarshal
 
 trait CalmRetriever {
 
-  def apply(query: CalmQuery): Future[List[CalmRecord]]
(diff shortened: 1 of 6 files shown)


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmRetriever.scala
```diff
@@ -29,18 +34,16 @@ class HttpCalmRetriever(url: String, username: String, password: String)(
 
   type Result[T] = Either[Throwable, T]
 
-  def apply(query: CalmQuery): Future[List[CalmRecord]] =
-    callApi(CalmSearchRequest(query), searchResponseParser)
-      .flatMap {
+  def apply(query: CalmQuery): Source[CalmRecord, NotUsed] =
+    Source
+      .fromFuture(callApi(CalmSearchRequest(query), searchResponseParser))
+      .mapConcat {
```
jamesgorrie: ✨

warrd:
> What do we do with the Right(new Exception(s"Errors processing window $date: $errs"))

The next stream stage is [`.via(catchErrors)`](https://github.com/wellcometrust/catalogue/blob/9b8583cbdf3865fe7cfb0b5d8fa1fdac0d077427/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmAdapterWorkerService.scala#L83), which results in the message being collected, the error being reported, and not making it to the end of the stream (see [here](https://github.com/wellcometrust/catalogue/blob/master/common/big_messaging/src/main/scala/uk/ac/wellcome/bigmessaging/FlowOps.scala#L42-L53))

PR #419, opened 2020-02-12
Add result relevance information 

jamesgorrie:
For testing, and _probably_ for features in the future, you might want to attach some metadata to a result that contextualised it within the set of results.

An example of this would be score (for testing, I need to compare) where we might want to flag how relevant we think something is. While order is an indicator of this, if you have say 5 results, 2 might be highly relevant, while the other three might be loosely relevant.

Another piece of information might be what we matched (the title, publisher etc), the highlighted piece of content etc.

This __draft PR__ starts with the model and if it feels like that is the right direction?

The rendering might unpack this into something along the lines of:
```JS
GET /works?include=results.relevance&query=spain
{
  "results": [
    {
      "type": "Work",
      "title": "The rain in Maine is plainly not in Spain",
      "description": "This is not really a rhymme from Spain",
      "relevance": {
        "matched": ["title" ,"description"],
        "score": {
          "value": "1.6",
          "label": "Highly relevant"
        }
      }
    },
    {
      "type": "Work",
      "title": "The rain in spain",
      "description": "Fell mainly on the plain",
      "relevance": {
        "matched": ["title"],
        "score": {
          "value": "1.2",
          "label": "Relevant"
        }
      }
    }
  ]
}
```

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/ResultList.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/ResultList.scala
@@ -2,8 +2,10 @@ package uk.ac.wellcome.platform.api.models
 
 import uk.ac.wellcome.models.work.internal.IdentifiedWork
 
+case class WorkResult(work: IdentifiedWork, score: Int, queryName: String)
+
 case class ResultList(
-  results: List[IdentifiedWork],
+  results: List[WorkResult],
   totalResults: Int,
   aggregations: Option[Aggregations]
 )


harrisonpim:
Only immediate thought is that the range and distribution of scores is likely to change as we tweak the query, so setting a threshold for those labels (anything above 1.5 is "highly relevant", etc) could quickly fall out of date. Would be nice to have those thresholds inferred from some known results (eg we know that these results are highly relevant for these queries, what's a good lower limit on their scores according to the current query)

jamesgorrie:
@harrisonpim interesting problem!

The labels are mainly there for us to display more than just a score to users, so I suppose we could look at it as "Something could be 10000 relevant, and this is 9090, so it's highly relevant", but that is quite hard to codify.

Perhaps sticking to _why_ something is relevant might be more useful.

kenoir:
> Perhaps sticking to why something is relevant might be more useful.

That sounds _hard_.

jamesgorrie:
> That sounds hard.

If the user search matches a part of the query that is written to an intention, not so much.

Thinking super simple, "We've matched using our title search, so that's why we've made it relevant". [We have something similar already](https://catalogue.wellcomelibrary.org/search/a?searchtype=Y&searcharg=more&searchscope=12&SORT=D).

This combined with highlighting will help us build trust, especially as researchers have so many contexts of search. If we tell them why we've matched, they can then decided if our reason for relevance is in line with theirs.

jennpb:
This is great, I love this. Even to be able to say your search terms were found in: title, notes, etc. would be valuable, even if we can't make a 'highly relevant' vs 'less relevant' statement. Refers back to https://github.com/wellcomecollection/wellcomecollection.org/issues/4363

PR #420, opened 2020-02-12
Merging more than just pairs of works

jamieparkinson:
### Motivation

I wrote this to try to answer the following question:

> **If we want to merge > 2 works, can that be achieved by recursively merging pairs of works?**

If this is the case, we can keep pretty much the same merge logic before, only considering 2 works at once. **It looks like the answer is _yes_**.

This also surfaced some further questions that weren't apparent before - notably, how to choose a thumbnail when merging multiple Miro works into a single Sierra work.

***

### What is this code?

- The new requirement is represented by modifying `PotentialMergedWork` and `MergedWork` to a target work and a _sequence_ of redirected works, rather than a target work and a single redirected work. All the users of these classes are updated to give them sequences where appropriate.

- A new method on `MergerRule`, `foldWorkPairs`, is introduced to recursively merge the list of works to redirect into a single `Option[MergedWork]`.

- A `WorkTagPairPartitioner` is introduced, which maintains the existing behaviour by discarding `Partition`s with `PotentialMergedWork`s that have more than 1 redirected work. This is used for the existing rules.

- The single-miro merger rule is copied (to `sierramiro`) and does _not_ use the `WorkTagPairPartitioner`. This is tested but not used in the actual merger so as to preserve existing behaviour; instead the original single-page rule is updated to use the `WorkTagPairPartitioner` and has its classes renamed to clarify the distinction.

- The new `sierramiro` rule for merging multiple Miro works into a Sierra work is modified to deterministically select the thumbnail of the Miro work with the lexicographically minimal ID

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/model/MergedWork.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/model/MergedWork.scala
@@ -1,8 +1,11 @@
 package uk.ac.wellcome.platform.merger.model
 import uk.ac.wellcome.models.work.internal.{
+  TransformedBaseWork,
   UnidentifiedRedirectedWork,
   UnidentifiedWork
 }
 
 case class MergedWork(work: UnidentifiedWork,
                       redirectedWork: UnidentifiedRedirectedWork)
+case class PotentialMergedWork(target: UnidentifiedWork,
+                               redirectedWork: TransformedBaseWork)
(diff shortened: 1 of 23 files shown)


alicefuzier:
Looks good to me. Might be worth checking if our assumption that we always want to merge all miro works to the sierra work or none of them is correct. I think it is..

jamieparkinson:
Closing in favour of https://github.com/wellcomecollection/platform/issues/4283

PR #424, opened 2020-02-13
restructure testing of queries

jamesgorrie:
Starting to solve the issue that we release stuff that breaks stuff, but still wanting to maintain some flexibility.

To be combined with integration tests too as often tests in the code are insufficient due to queries using the data as a variable (tf/idf)

---

## What?

We've split the tests into core / scoring query tests. The core tests will make sure we don't break functionality like querying by ID, but doesn't really care about scoring.

The score tests are meant to outline what we're trying to achieve, with the queries and the tests with the realisation that they will be quite transient and are __highly coupled to the data that they search__ so will not cover all bases.

Because of this we should write some more integration like tests. See https://github.com/wellcomecollection/platform/issues/4143

---

## ConstScore / BoolBoost what?

These are two different ways of accumulating score on a result.

Previously "rain" would match a work with `id: 1, title: rain in spain` and `id: 2, title: rain in rain` but return them both with a score of `1000` as we were using [`constant_score`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-constant-score-query.html) to wrap the query. They would then be sorted by `score` and `id` thus returning `id: 1, id: 2` <- which seems wrong.

When using `BoolBoost`, it will work out the score, then multiply by the boost, which, after having spoken to @harrisonpim and @taceybadgerbrook is a much more expected ordering.

This way the match would return `id: 2, id: 1`.


api/api/src/test/scala/uk/ac/wellcome/platform/api/elasticsearch/VarientQueriesTest.scala
```diff
@@ -0,0 +1,158 @@
...
+  WorksGenerators
+}
+import uk.ac.wellcome.platform.api.generators.SearchOptionsGenerators
+import uk.ac.wellcome.platform.api.services.ElasticsearchService
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.Random
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.requests.searches.sort.SortOrder
+
+class VarientQueriesTest
```
jamieparkinson: `s/Varient/Variant`, might be clearer as `QueryVariantsTest` if I've understood its purpose right

PR #425, opened 2020-02-17
Fixes for Calm API

warrd:
## Issue

## Description

Running this locally against the Calm API there were a few small issues, which this addresses.

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
@@ -67,6 +67,12 @@ object CalmSearchResponse {
   def apply(str: String,
             cookie: Cookie): Either[Throwable, CalmSearchResponse] =
     Try(XML.loadString(str)).map(CalmSearchResponse(_, cookie)).toEither
+
+  def apply(bytes: Array[Byte],
+            cookie: Cookie): Either[Throwable, CalmSearchResponse] =
+    Try(XML.load(new java.io.ByteArrayInputStream(bytes)))
+      .map(CalmSearchResponse(_, cookie))
+      .toEither
 }
 
 case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
@@ -93,7 +99,13 @@ case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
   def parse: Either[Throwable, CalmRecord] =
     responseNode
       .flatMap(_.childWithTag("SummaryHeaderResult"))
-      .flatMap(_.childWithTag("SummaryList"))
+      .flatMap { node =>
+        // The response contains an inner XML document (ISO-8859-1 encoded)
+        // within the top level UTF-8 one, so we need to parse this here if
+        // it exists
+        Try(XML.loadString(node.text)).toEither.left
+          .flatMap(_ => node.childWithTag("SummaryList"))
+      }
       .flatMap(_.childWithTag("Summary"))
       .map(_ \ "_")
       .flatMap { nodes =>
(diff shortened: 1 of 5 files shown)


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
```diff
@@ -93,7 +99,15 @@ case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
   def parse: Either[Throwable, CalmRecord] =
     responseNode
       .flatMap(_.childWithTag("SummaryHeaderResult"))
-      .flatMap(_.childWithTag("SummaryList"))
+      .flatMap { node =>
+        // The response contains an inner XML document (ISO-8859-1 encoded)
+        // within the top level UTF-8 one, so we need to parse this here if
+        // it exists
```
jamesgorrie: 🤕

warrd: :see_no_evil:

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponseTest.scala
```diff
@@ -134,6 +134,30 @@ class CalmXmlResponseTest extends FunSpec with Matchers {
...
+        <soap:Envelope
+            xmlns:soap="http://www.w3.org/2003/05/soap-envelope"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+          <soap:Body>
+            <SummaryHeaderResponse xmlns="http://ds.co.uk/cs/webservices/">
+              <SummaryHeaderResult>&lt;?xml version="1.0" encoding="ISO-8859-1"?&gt;&lt;SummaryList&gt;&lt;Summary&gt;&lt;RecordType&gt;Component&lt;/RecordType&gt;\n  &lt;RecordID&gt;6e4edfee-e702-4da6-be6d-c2e60ce79728&lt;/RecordID&gt;&lt;/Summary&gt;&lt;/SummaryList&gt;</SummaryHeaderResult>
+            </SummaryHeaderResponse>
+          </soap:Body>
+        </soap:Envelope>
```
jamesgorrie: Have you thought about using the wiremock for this?

warrd: The [`CalmHttpClient`](https://github.com/wellcomecollection/catalogue/blob/master/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmHttpClient.scala#L11-L14) trait was added precisely because I didn't want to use something like wiremock: 

1) I think mocking is unnecessary and defining an interface (in scala via `trait`) with differing implementations is a nicer way to write tests, and helps separate concerns

2) Unit tests should avoid HTTP traffic (as well as having to use docker to mock out SQS etc), as is above and beyond what we are testing and results in much slower running tests

warrd: (tho worth mentioning something like wiremock would probably be required for integration testing)

jamesgorrie: I thought wiremock negated the need for HTTP in testing by storing responses from the APIs as mocks?

This seems easier than copy/pasting responses?

warrd: 1) You still need to store those responses in a separate json file
2) Those files aren't particularly readable, and are not inside the tests themselves so harder to work out what is going on when reading a test (rather than direct comparison to a scala object)
3) It runs a whole HTTP server which is overkill and unnecessary IMO
4) It goes through lots of third party code (the library that is making the request in scala, the OS) which a unit test shouldn't need to do, and encourages coupling between your code and the library code

Obviously I'm a bit opinionated about this, and I'm sure there are some nice things about doing tests that way, but in general I find that method of testing really convoluted, slow etc..

warrd: > I thought wiremock negated the need for HTTP in testing by storing responses from the APIs as mocks?

Ah so I think you are suggesting it patches any http calls within the JVM so a server is not actually running? I am not sure if that is how it works internally, I thought it was running a local server. Even if it does do this all of what I said apart from 3 above still stands

PR #432, opened 2020-02-18
Calm adapter terraform

warrd:
## Issue

## Description

Sets up infrastructure:

* Cluster
* Service / task defs etc
* VHS
* SQS queue
* SNS topic

### DIFF
--- a/calm_adapter/terraform/queue.tf
+++ b/calm_adapter/terraform/queue.tf
@@ -1,13 +1,16 @@
-# TODO: Remove this once we're hooked into getting events for Calm
-# This is used as a placeolder for where the
-resource "aws_sns_topic" "test_receive_updates_topic" {
-  name = "calm_adapter_test_receive_updates_topic"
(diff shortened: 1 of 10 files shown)


calm_adapter/terraform/queue.tf
```diff
@@ -4,10 +4,15 @@ resource "aws_sns_topic" "test_receive_updates_topic" {
   name = "calm_adapter_test_receive_updates_topic"
 }
 
-module "ingestor_queue" {
+module "calm_window_queue" {
```
warrd: I think I also need to set up an SNS topic to send windows to from the initialisation script / lambda. At the moment [the script is using the incorrect topic](https://github.com/wellcomecollection/catalogue/pull/434/files#diff-d753722952a03443ec769b6fd25b2f51R12) (pretty sure thats the one the calm adapter outputs messages to the transformer). Would someone be able to point me in the right direction about how to go about this?

warrd:
@kenoir yep, afaik everyone else is done in terms of infra

PR #434, opened 2020-02-20
Calm window generator

warrd:
## Issue

## Description

Script for sending windows to the Calm adapter, used for initialisation.


calm_adapter/window_generator/window_generator.py
```diff
@@ -0,0 +1,71 @@
+#!/bin/env python3
+
+from datetime import date, timedelta
+import click
+import json
+import boto3
+from tqdm import tqdm
```
alexwlchan: minor: Python style is to group imports into three chunks: standard library, third-party, our own code. See https://www.python.org/dev/peps/pep-0008/

```python
from datetime import date, timedelta
import json

import boto3
import click
from tqdm import tqdm
```

calm_adapter/window_generator/window_generator.py
```diff
@@ -0,0 +1,71 @@
...
+    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
+
+    def __init__(self, start, end=None):
+        self.start = start
+        self.end = end or start
+        assert self.start <= self.end, "Start window not be after end window"
+        self.client = self.sns_client
+
+    def run(self):
+        for window in tqdm([*self.windows]):
```
alexwlchan: This is slightly odd. Are you trying to work out how many windows there will be? If so, `list(self.windows)` might be a bit clearer.

calm_adapter/window_generator/window_generator.py
```diff
@@ -0,0 +1,71 @@
...
+            yield self.start
+            self.start += timedelta(days=1)
+
+    def send_window_to_sns(self, window):
+        msg = json.dumps({"date": window.isoformat()})
+        self.client.publish(
+            TopicArn=self.topic_arn, Message=msg, Subject=f"Window sent by {__file__}"
+        )
+
+    @property
```
alexwlchan: minor: the general trend is to reserve `@property` for methods that:

* Return immediately
* Will never throw an exception

So they might as well be a static attribute. We shouldn't use `@property` here, because neither of those rules hold.

```
self.sns_client  # we expect won't throw, returns immediately

self.sns_client()  # is doing something interesting, might delay, might throw
```

calm_adapter/window_generator/window_generator.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.command()
+def full_ingest():
+    start = date(2015, 9, 30)
```
alexwlchan: What's the reason for this particular start date?

calm_adapter/window_generator/window_generator.py
warrd: I prefer this syntax for "unravelling" a generator into a list personally, but happy to change it

calm_adapter/window_generator/window_generator.py
warrd: Hmm I'm not sure that is always the trend, especially as this is python and I would expect things to throw at any time. Where I have coded python previously we agreed that things which have clear side effects should be called  as a function, but properties are fine elsewhere.

Happy to change here tho to be more consistent with the rest of the code base.

warrd: (worth mentioning this does obviously have a side effect of calling an external API, but does not really impact on the application beyond exception)

calm_adapter/window_generator/window_generator.py
warrd: That's what I found to be the first date where any records are returned. Can add a comment saying that much.

calm_adapter/window_generator/window_generator.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.command()
+def full_ingest():
+    start = date(2015, 9, 30) # First date returning any records
```
jamieparkinson: I reckon it might be better to have this as a
```python3
@click.option('--start', default=date(2015, 9, 30))
```
so it could be overridden without modifying this tool. Feels more like config than code.

alexwlchan: If you're going to do this, change the name as well.

It’s either a full ingest – every record created, which is every record created after 30 September 2015, which is a static date – or it’s a window which defaults to getting the entire set, but can be restricted further.

warrd: Thanks for suggestions, I'm gonna add a new command for this and keep the full ingest

PR #436, opened 2020-02-20
Lambda function for generating Calm windows

warrd:
## Issue

## Description

Lambda which generates windows at regular intervals.


makefiles/functions.Makefile
```diff
@@ -17,8 +17,8 @@ include $(ROOT)/makefiles/terraform.Makefile
 #
 define publish_lambda
     $(ROOT)/docker_run.py --aws --root -- \
-        wellcome/publish_lambda:14 \
-        "$(1)" --key="lambdas/$(1).zip" --bucket="$(INFRA_BUCKET)" --sns-topic="arn:aws:sns:eu-west-1:760097843905:lambda_pushes"
+        wellcome/publish_lambda:latest \
```
jamesgorrie: Could we lock this to avoid it breaking without warning later?

warrd:
:+1: will need to release the latest `publish_lambda` container and then can lock it.

PR #440, opened 2020-02-24
Merger rewrite

jamieparkinson:
Obviously, this PR is (perhaps too) huge, but I've separated each component into its own commit and tried to keep the deletions and additions separate.

I've tried as best I can to infer the rules that we want from the code we had before, but I'm not confident that I've captured all of this as I don't understand all of it: suggestions and scrutiny for the test cases in particular would be appreciated.


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,77 @@
...
+ * a `rule` that has the same signature as the `merge` method, the implementor
+ * must provide predicates for valid targets and sources.
+ *
+ * Because PartialRule is a PartialFunction, the `rule` will never be called
+ * for targets/sources that don't satisfy these predicates, and we gain things
+ * like `orElse` and `andThen` for free.
+ */
+trait FieldMergeRule {
+  protected final type Params = (UnidentifiedWork, Seq[TransformedBaseWork])
+  protected type Field
```
warrd: At the risk of being pedantic: when I first saw this type I assumed that we would somehow encode the name and value into the type, rather than just the type of data. Maybe `FieldData` would be a more accurate description, as a `Field` to me indicates a mapping between two things? But maybe I'm over thinking things..

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,77 @@
...
+    case (target, _) => target
+  }
+
+  protected trait PartialRule extends PartialFunction[Params, Field] {
+    val isDefinedForTarget: WorkFilter
+    val isDefinedForSource: WorkFilter
+
+    def rule(target: UnidentifiedWork, sources: Seq[TransformedBaseWork]): Field
+
+    override def apply(x: Params): Field =
```
warrd: Can we just call this var `params` rather than `x`?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,77 @@
...
+      x match {
+        case (target, sources) =>
+          rule(target, sources.filter(isDefinedForSource))
+      }
+
+    override def isDefinedAt(x: Params): Boolean = x match {
```
warrd: Same as above

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,77 @@
...
+ * PartialRules can be used with `orElse` as a chain of fallbacks for picking a
+ * field off sources, or they can be composed.
+ *
+ * Extending the ComposedFieldMergeRule trait provides a helper, `composeRules`,
+ * which takes a list of rules, as well as a helper function to copy the new
+ * merged Fields from each step into the "running-total" target, and composes
+ * them right-to-left.
+ */
+trait ComposedFieldMergeRule extends FieldMergeRule {
+  protected final def composeRules(
```
warrd: I don't personally see the need for this to exist in a separate trait, as could just be included in the base one and only be used when required. I guess it could be argued it is more explicit, but it's at the cost of a more complex type hierarchy.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,77 @@
...
+    liftIntoTarget: UnidentifiedWork => Field => UnidentifiedWork)(
+    rules: PartialRule*)(target: UnidentifiedWork,
+                         sources: Seq[TransformedBaseWork]): UnidentifiedWork =
+    rules.foldRight(target) { (applyRule, nextTarget) =>
+      (applyRule andThen liftIntoTarget(nextTarget) orElse identityOnTarget)(
+        (nextTarget, sources))
+    }
```
warrd: :tada:

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,77 @@
+package uk.ac.wellcome.platform.merger.rules
+import uk.ac.wellcome.models.work.internal.{
+  TransformedBaseWork,
+  UnidentifiedWork
+}
+import uk.ac.wellcome.platform.merger.rules.WorkFilters.WorkFilter
+
+case class MergeResult[T](field: T, redirects: Seq[TransformedBaseWork])
```
warrd: Would be helpful having a comment documenting this type. Also I think `data` (or something like that) would be a better name than `field` (see my other comment)

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -0,0 +1,145 @@
...
+object ItemsRule extends ComposedFieldMergeRule with MergerLogging {
+  type Field = List[Item[Unminted]]
+
+  override def merge(target: UnidentifiedWork,
+                     sources: Seq[TransformedBaseWork]): MergeResult[Field] =
+    MergeResult(
+      field = composeRules(liftIntoTarget)(
+        metsItemsRule,
+        miroItemsRule,
+        physicalDigitalItemsRule)(target, sources).data.items,
```
warrd: Whenever `composeRules` is used, we just seem to fetch the field in questions data from the output `UnidentifiedWork`, and throw away the rest. Is there a reason we need to lift the data into the target if this is the case?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
jamieparkinson: Yeah, agreed, I initially called it `FieldType` but putting "type" in the name of a type felt bad - `FieldData` is good

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamieparkinson: Yeah, unfortunately it is necessary: because we're composing the functions we need to transform the return type into the input type, in this case by lifting the field into the `target`.

We can (and do) keep `sources` constant but in order for this to be a composition the `target` needs to contain the value that the last applied rule produced, rather than that of the 'initial' target.

warrd: >  the target needs to contain the value that the last applied rule produced, rather than that of the 'initial' target.

Am I right in saying the data for `target` stays the same, apart from the field data that is lifted into it? So a rule could potentially take as arguments the `target` and `sources` (which stay constant), and optionally any latest field data from previously applied rules. That seems to me would avoid having to lift data into the target unnecessarily, but could end up confusing things in other ways, I dunno :man_shrugging:

jamieparkinson: > Am I right in saying the data for target stays the same, apart from the field data that is lifted into it?

Yes

> a rule could potentially take as arguments the target and sources (which stay constant), and optionally any latest field data from previously applied rules

I think this would be more confusing, as the logic related to composition would then have leaked into the rule's implementation. I spent a while playing with rules that took just the field data as arguments, or had access to extractor functions that could get the field data without exposing how this was done, but all of it seemed to introduce more complexity than just working on targets and sources with no extra magic involved.

jamieparkinson: More generally I'm optimistic that composition like this can be done away with if we work out the "true" business logic involved in merging items

warrd: :+1:

jamieparkinson: @warrd based on the above I stopped using the composition helper in `otherIdentifiers` as it could be done more simply in that specific case - fingers crossed we can do the same for items

alicefuzier: Yep, I think the original rules are so complicated because the sierra-mets rule, which is the one applied last, had to undo what the previous rules did. I think the actual business logic we want for items is something like:
```
if target has only one item => add mets location if there is a mets work, otherwise add locations from both the digital work and the miro work if they exist
if target has more than one item => add mets item if there is a mets work, otherwise add item from digital work
```

jamieparkinson: @alicefuzier turned the above into code and the tests are still passing... I think it worked?!

alicefuzier: 🎉 🎆

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,62 @@
...
+  protected type FieldData
+
+  def merge(target: UnidentifiedWork,
+            sources: Seq[TransformedBaseWork]): MergeResult[FieldData]
+
+  protected val identityOnTarget: PartialFunction[Params, UnidentifiedWork] = {
+    case (target, _) => target
+  }
+
+  protected trait PartialRule extends PartialFunction[Params, FieldData] {
```
kenoir: Smart.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -0,0 +1,77 @@
...
+ * - Miro locations are added to single-item Sierra locations.
+ */
+object ItemsRule extends FieldMergeRule with MergerLogging {
+  type FieldData = List[Item[Unminted]]
+
+  override def merge(
+    target: UnidentifiedWork,
+    sources: Seq[TransformedBaseWork]): MergeResult[FieldData] =
+    MergeResult(
+      fieldData = (mergeMetsItems orElse mergeMiroPhysicalAndDigitalItems
```
kenoir: Wow. This is very nice.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/OtherIdentifiersRule.scala
```diff
@@ -0,0 +1,64 @@
...
+  UnidentifiedWork
+}
+import uk.ac.wellcome.platform.merger.logging.MergerLogging
+import uk.ac.wellcome.platform.merger.rules.WorkFilters.WorkFilter
+
+/*
+ * The otherIdentifiers field is made up of all of the identifiers on the sources,
+ * except for any Sierra identifiers on Miro sources, as these may be incorrect
+ */
+object OtherIdentifiersRule extends FieldMergeRule with MergerLogging {
```
kenoir: Using partial functions like this & the field level merging, makes this _so much clearer_.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ThumbnailRule.scala
```diff
@@ -0,0 +1,66 @@
...
+  Location,
+  TransformedBaseWork,
+  UnidentifiedWork
+}
+import uk.ac.wellcome.platform.merger.logging.MergerLogging
+import uk.ac.wellcome.platform.merger.rules.WorkFilters.WorkFilter
+
+import scala.util.Try
+
+/*
```
kenoir: These comments are super helpful too.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -0,0 +1,62 @@
...
+    val isDefinedForTarget: WorkFilter
+    val isDefinedForSource: WorkFilter
```
warrd: A bit unsatisfied with the naming of these vars :thinking: 

It's not whether the rule is "defined", more whether it "applies", or something. `appliesToTarget` / `appliesToSource`?

warrd: Maybe also change `WorkFilter` to `WorkPredicate`? Or is that too much of a mouthful?

warrd: (not really a filter as such, and already have `WorkFilter` type elsewhere which doesn't help)

jamieparkinson: Hmmm, the naming is very deliberately a reference to the naming used by (and the general concept of) [`PartialFunction`](https://www.scala-lang.org/files/archive/api/2.12.10/scala/PartialFunction.html), as `PartialRule`s are no more than functions over _a subset of_ the product of the set of targets and the set of lists of sources. Especially given that I use `isDefinedAt` outside of the implementation I would prefer to keep the naming of these the same so as not to obfuscate the connection.

Completely agree for `WorkPredicate` though!

alexwlchan:
(I couldn't finish my earlier review and a lot of the comments are outdated, so I don't think I can be much use here.

The code still isn't super clear to me, but I also have less experience with the partial functions and Scala-y stuff than Robert or Nick.

I do remember being dissatisfied with the last major rewrite of the merger I worked on with Mark L – we both knew there was a better version waiting out there, but couldn't make it happen ourselves. If Nick and Robert think this is an improvement, it's a 👍 from me.)

jamieparkinson:
This has actually been merged but I did it during the GitHub outage yesterday and it's gone weird - I've opened a ticket with GH support as I'd rather everything was consistent than this was closed-but-really-merged

PR #441, opened 2020-02-25
add CalmWorker with tests

jamesgorrie:
Adds an Worker that works with AKKA streams.

Might have to pull some of the worker stuff out in favour of https://github.com/wellcomecollection/scala-messaging/pull/30.

Splits out the worker to so we can write some tests without being dependant on SQS, rather just a AKKA source. I haven't written the tests as my IDE is being very annoying, but things are compiling.

---

Next up:
- error flow

---
Side notes: 

[`divertTo` is pretty nifty!](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/divertTo.html)


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/Worker.scala
```diff
@@ -1,38 +1,34 @@
...
-    fromJson[StoreKey](message.body).toEither match {
-      case Left(err) =>
-        Left(new Exception(s"Worker: Cannot read message, ${err}"))
-      case _ => _
+  private def divertEither[I, O](
+    flow: Flow[I, Result[O], NotUsed],
+    to: Sink[Result[_], Future[Done]]): Flow[I, O, NotUsed] =
+    flow.divertTo(Sink.ignore, _.isLeft).collect {
+      case Right(out) => out
     }
```
warrd: Nice. Similar to `catchErrors` [here](https://github.com/wellcomecollection/catalogue/blob/master/common/big_messaging/src/main/scala/uk/ac/wellcome/bigmessaging/FlowOps.scala#L42) fyi

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/Worker.scala
```diff
@@ -1,38 +1,34 @@
...
+
+  val workFlow: Flow[MessageData, SinkData, NotUsed] =
+    divertEither(Flow.fromFunction(work), to = errorSink)
+
+  val doneFlow: Flow[SinkData, Unit, NotUsed] =
+    divertEither(Flow.fromFunction(done), to = errorSink)
 
+  def decodeMessage(message: Message): Result[MessageData]
+  def work(messageData: MessageData): Result[SinkData]
   def done(sinkData: SinkData): Result[Unit]
```
warrd: Any reason we don't allow the implementer to specify a `Flow` for these here, instead of just a function? That would allow easier concurrency / async stuff etc, and provide more fine grained backpressure rather than for the stage as a whole

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmWorker.scala
```diff
@@ -0,0 +1,57 @@
...
+import uk.ac.wellcome.models.work.internal.TransformedBaseWork
+import uk.ac.wellcome.platform.transformer.calm.models.CalmRecord
+import uk.ac.wellcome.storage.store.VersionedStore
+import uk.ac.wellcome.storage.{Identified, Version}
+
+sealed trait CalmWorkerError extends Throwable
+object DecodeKeyError extends CalmWorkerError
+object StoreReadError extends CalmWorkerError
+object TransformerError extends CalmWorkerError
+object MessageSendError extends CalmWorkerError
```
warrd: I think we probably want some more info about these errors rather than just a type. Could something like this work instead?

```scala
sealed class CalmWorkerError(msg: String) extends Exception(msg)
class DecodeKeyError(msg: String) extends CalmWorkerError(msg)
class StoreReadError(msg: String) extends CalmWorkerError(msg)
...
```

On a side note, I think we should avoid using the `Throwable` type to represent application level exceptions, and instead use `Exception` (which is what that class is generally for in Java)

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/Worker.scala
warrd: Wait a min, I don't think you need the `divertTo` here, the `collect` would filter out any `Left(_)` anyway

jamesgorrie: It filters them out, but I want a place for them to be to use later (which I am admittedly not using, but will).

Then I could run an error flow on the error sink.

warrd: :+1: yup if the sink wasnt just ignore then makes sense

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/Worker.scala
jamesgorrie: Could do when it becomes a thing I suppose.

The reason I haven't is it feels like I have abstracted as far back as might be useful over as far back as I can.

Another reason is to move the Flow stuff out. You just need to give us the business logic. 

Maybe having a `WorkerWork` and a `WorkerFlow` that you can compose might be useful. I am thinkinng of how to split this up into more composable parts, and maybe this is another one.

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmWorker.scala
jamesgorrie: Definitely do - I was going to add that when implementing the flow on the error sink. Happy to tackle that as a next PR.

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/Worker.scala
warrd: But then I don't really see the advantage of using akka-streams at all? As it is currently it is just a sequential functions chained together, the streams stuff mainly gets useful when there is async stuff or heavier computations etc, and this interface doesn't give any control over how any of that is executed

jamesgorrie: It's not a bad point, we could just wrap them in the [for comprehension as before](https://github.com/wellcomecollection/catalogue/pull/399/files#diff-143261f81c47872cc717d20e7a22c886R47-R52).

The point of this abstraction was to be able have an interface defined on a Worker that read a message, did some work, and then did something with the output of that work, and keeping any tracing of that interface outside of the implementation.

I was going to use the graph like interface for ☝️ in terms of tracing.

It feels like each step would need to have it's own configuration, at which point you're basically re-writing the trait everytime you want to use it.

Saying that, this is a good point, and I'll have a little stab at how it might look.

warrd: Fwiw I think we should use akka streams, as it is really nice way to express the sort of data flow computations we work with (especially when it does get more complicated with calls to other services etc)..

Maybe we could define the "higher level" function interface like we have here, but also a "lower level" stream interface which the other is based on (this would be the one used by e.g. [`CalmAdapterWorkerService`](https://github.com/wellcomecollection/catalogue/blob/master/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmAdapterWorkerService.scala), [`MetsAdapterWorkerService`](https://github.com/wellcomecollection/catalogue/blob/master/mets_adapter/mets_adapter/src/main/scala/uk/ac/wellcome/mets_adapter/services/MetsAdapterWorkerService.scala), which do work on streams)

warrd: Another though: I think lots of our current worker services would have to use an interface which at least allows returning of `Future[_]`, as they do async things. Otherwise we are going to have to `Await` everywhere, which I don't think we want

jamesgorrie: Did this as a first pass?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/SimpleStreamWorker.scala
```diff
@@ -0,0 +1,25 @@
...
+    Flow.fromFunction(decodeMessage)
+  val workFlow: Flow[MessageData, Result[SinkData], NotUsed] =
+    Flow.fromFunction(work)
+  val doneFlow: Flow[SinkData, Result[Unit], NotUsed] =
+    Flow.fromFunction(done)
+
+  def decodeMessage(message: Message): Result[MessageData]
+  def work(messageData: MessageData): Result[SinkData]
+  def done(sinkData: SinkData): Result[Unit]
+}
```
warrd: :tada:

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/Worker.scala
jamesgorrie: Was thinking we could also:

```scala
/** This worker:
  * - Reads from a VHS, and decodes to `VHSDataType`
  * - Transforms it into a `TransformedBaseWork`
  * - Emits the work via `BigMessageSender`
  * */
class VhsReaderWorker[VHSDataType]
    extends StreamWorker[
      (Message, NotificationMessage),
      VHSDataType,
      TransformedBaseWork] {

  type StoreKey = Version[String, Int]
}
```

But we land up coupling in the name of abstraction...

PR #444, opened 2020-02-25
Images from the merger

jamieparkinson:
Good to see that the merger rewrite was actually useful - this is what I think creating images will look like, what do you think?

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
@@ -9,18 +9,21 @@ case class UnmergedImage[Id <: IdState](
   id: Id,
   location: DigitalLocation
 ) extends BaseImage[Id] {
-  def mergeWith(data: => ImageData[Id]): MergedImage[Id] =
+  def mergeWith(parentWork: Id,
+                fullText: Option[String] = None): MergedImage[Id] =
     MergedImage[Id](
       id = id,
       location = location,
-      data = data
+      parentWork = parentWork,
+      fullText = fullText
     )
 }
 
 case class MergedImage[Id <: IdState](
   id: Id,
   location: DigitalLocation,
-  data: ImageData[Id]
+  parentWork: Id,
+  fullText: Option[String] = None
 ) extends BaseImage[Id] {
   def toUnmerged: UnmergedImage[Id] =
     UnmergedImage[Id](
@@ -28,11 +31,6 @@ case class MergedImage[Id <: IdState](
       location = location
     )
 }
-
-case class ImageData[Id <: IdState](
-  parentWork: Id,
-  fullText: Option[String] = None
-)
 
 object UnmergedImage {
   def apply(sourceIdentifier: SourceIdentifier,
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
@@ -30,10 +30,8 @@ case class MergedImage[Id <: IdState](
 }
 
 case class ImageData[Id <: IdState](
-  caption: Option[String] = None,
-  description: Option[String] = None,
-  privateDescription: Option[String] = None,
-  parentWorks: List[Id] = Nil
+  parentWork: Id,
+  fullText: Option[String] = None
 )
 
 object UnmergedImage {
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
@@ -1,15 +1,27 @@
 package uk.ac.wellcome.models.work.internal
 
-sealed trait BaseImage[+Id] extends HasIdState[Id] {
+sealed trait BaseImage[+Id <: IdState] extends HasIdState[Id] {
   val id: Id
   val location: DigitalLocation
 }
 
-case class UnmergedImage[+Id](id: Id, location: DigitalLocation)
-    extends BaseImage[Id]
+case class UnmergedImage[Id <: IdState](
+  id: Id,
+  location: DigitalLocation
+) extends BaseImage[Id] {
+  def mergeWith(data: => ImageData[Id]): MergedImage[Id] =
+    MergedImage[Id](
+      id = id,
+      location = location,
+      data = data
+    )
+}
 
-case class MergedImage[+Id](id: Id, location: DigitalLocation, data: ImageData)
-    extends BaseImage[Id] {
+case class MergedImage[Id <: IdState](
+  id: Id,
+  location: DigitalLocation,
+  data: ImageData[Id]
+) extends BaseImage[Id] {
   def toUnmerged: UnmergedImage[Id] =
     UnmergedImage[Id](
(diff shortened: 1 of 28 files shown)


jamieparkinson:
I realise that the `fullText` field is in some ways Not Great - it's there as a first pass so there's _something_ to search.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorksGenerators.scala
```diff
@@ -200,11 +208,17 @@ trait WorksGenerators extends ItemsGenerators with ProductionEventGenerators {
 
   def createUnidentifiedInvisibleMetsWorkWith(
     sourceIdentifier: SourceIdentifier = createMetsSourceIdentifier,
-    items: List[Item[Unminted]] = List(createDigitalItem))
-    : UnidentifiedInvisibleWork =
+    items: List[Item[Unminted]] = List(createDigitalItem),
+    nImages: Int = 1): UnidentifiedInvisibleWork =
```
kenoir: maxImages?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ImagesRule.scala
```diff
@@ -0,0 +1,92 @@
...
+import uk.ac.wellcome.platform.merger.models.FieldMergeResult
+import uk.ac.wellcome.platform.merger.rules.WorkPredicates.{
+  not,
+  WorkPredicate,
+  WorkPredicateOps
+}
+
+import scala.Function.const
+
+object ImagesRule extends FieldMergeRule {
```
kenoir: ✨

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -1,17 +1,38 @@
...
+      id = id,
+      location = location
+    )
+}
 
-case class ImageData(title: Option[String], parentWorks: List[String])
+case class ImageData[Id <: IdState](
+  parentWork: Id,
+  fullText: Option[String] = None
+)
```
warrd: Is having a seperate `ImageData` class necessary? The reason we have `WorkData` is that multiple work types need the same data, but here it appears like only one does

pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/rules/ImagesRuleTest.scala
```diff
@@ -0,0 +1,159 @@
...
+      val miroWorks = (1 to n).map(_ => createMiroWork).toList
+      val sierraWork = createSierraDigitalWork
+      val result = ImagesRule.merge(sierraWork, miroWorks :+ metsWork).fieldData
+
+      result should have length n
+      result.map(_.location) should contain theSameElementsAs
+        miroWorks.map(_.data.images.head.location)
+      result.flatMap(_.data.parentWork.allSourceIdentifiers) should contain only sierraWork.sourceIdentifier
+    }
+  }
```
kenoir: Lovely clear test cases.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamesgorrie: Yea, I'd say of we can avoid this nesting in the model, it would be nice.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorksGenerators.scala
jamieparkinson: It creates exactly this number, to me `max` implies that there may be fewer...

warrd: (`numImages` sounds more descriptive / cuter to me)

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamieparkinson: Yeah, I'll have a go at removing it - don't think it is necessary

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ImagesRule.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  private def createFulltext(works: Seq[TransformedBaseWork]): Option[String] =
+    works
+      .map(_.data)
+      .flatMap { data =>
+        List(
+          data.title,
+          data.description,
+          data.physicalDescription,
+          data.lettering
```
jamesgorrie: Would you want to put IDs of miro / sierra etc in here?

jamesgorrie: Sorry, I know you said to not think of the shape that much...

jamieparkinson: Quite possibly/probably - I think the contents of `Image`s will be a big ongoing discussion that will mostly be informed by search performance.

It also seems likely that `fullText` will turn out to be too limited, or at least that it shouldn't be constructed in the merger. Would prefer to leave this for now whilst being very clear that we'll come back and improve it.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/models/MergeResult.scala
```diff
@@ -0,0 +1,14 @@
...
+  MergedImage,
+  UnidentifiedWork,
+  Unminted
+}
+
+/*
+ * MergeResult holds the resultant target after all fields have been merged,
+ * and the images that were created in the process
+ */
+case class MergeResult(target: UnidentifiedWork,
```
jamesgorrie: Could this be more explicit and be called `mergedTarget` or `result`? (`target` to me meaning it's still... well the `target`.)

jamieparkinson:
@jamesgorrie yes, that's pretty much it (I'd add that the merger doesn't initially know which work is the target and which are the sources, it infers that itself).

The concept is that the `MergerOutcome` is created from some combination of `MergeResults`, as you've described - seems quite possible that it might want to do something to the images (like filtering some out, perhaps) at some point in the future, but as you say it's just passed straight through now.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ImagesRule.scala
jamesgorrie: Just as a point so it doesn't get lost, having someone know why an image has come up is very important.

Searching for something that could be ambivalently neutral or derogatory is a major issue i.e. surfacing images when someone searches for animals (a case that happens). If this information isn't there to show why it was surfaced it might be assumed we are saying 👈 is true, rather than having been catalogued historically like that.

Makes me wonder if we should remove this and leave it TBD until the discussion can happen more thoroughly.

Ping @harrisonpim

jamesgorrie:
@jamieparkinson thanks! It's really clear, and easy to fathom.

PR #451, opened 2020-03-02
Transform CalmRecord to Work

warrd:
https://github.com/wellcomecollection/platform/issues/4211


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -1,28 +1,174 @@
...
+    workData(record) map { data =>
       UnidentifiedWork(
-        sourceIdentifier = SourceIdentifier(
-          value = record.id,
-          identifierType = CalmIdentifierTypes.recordId,
-          ontologyType = "IdentifierType"),
-        version = 1,
-        data = WorkData()
+        sourceIdentifier = sourceIdentifier(record),
+        version = 0,
```
jamesgorrie: Is this how all the transformers work?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -1,28 +1,174 @@
...
       )
     )
+
+  def physicalLocation(record: CalmRecord,
+                       status: Option[AccessStatus]): PhysicalLocation =
+    PhysicalLocation(
+      locationType = LocationType("scmac"),
+      label = "Closed stores Arch. & MSS",
+      accessConditions = List(accessCondition(record, status))
+    )
```
jamesgorrie: Is this right?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -1,28 +1,174 @@
...
+
+  def accessCondition(record: CalmRecord,
+                      status: Option[AccessStatus]): AccessCondition =
+    AccessCondition(
+      status = status,
+      terms = record.getJoined("AccessConditions"),
```
jamesgorrie: Would this potentially error?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
warrd: Nope (if you mean setting `version = 0`), other transformers initialise this to the source version, but we don't have a version field returned from Calm.

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
warrd: yes, according to @jtweed

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
warrd: no, doesn't error

warrd:
@jamesgorrie needs some tests, so will do that before merging

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
warrd: I lied! It actually uses the adapter version, good thing you made me check

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
jamesgorrie: Any chance of whacking a comment in here for our future selves about this being a temp measure so whern I see it in the future I don't question why it's like this?

As far as I know it's because we're working on locations so this is OK for now.

warrd: Is it a temporary measure, I'm not sure :thinking:

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
```diff
@@ -23,4 +31,28 @@ object AccessStatus {
...
+        Right(AccessStatus.Restricted)
+      case status if status.startsWith("closed") =>
+        Right(AccessStatus.Closed)
+      case status if status.startsWith("missing") =>
+        Right(AccessStatus.Unavailable)
+      case status if status.startsWith("temporarily unavailable") =>
+        Right(AccessStatus.Unavailable)
+      case status if status.startsWith("permission required") =>
+        Right(AccessStatus.PermissionRequired)
+      case status =>
```
jamesgorrie: Just wondering, do we want to throw in the model, or throw in the transformer? I can see why it works in both, but feels strange to do it in the model. Maybe returning an `Option[AccessStatus]` and then when empty throwing an error in the transformer?

warrd: This functionality is also used by the Sierra transformer, hence why shared between both. Having a `Result` vs `Option` doesn't feel any different to me personally, and a `Result` is more descriptive about what went wrong.

PR #452, opened 2020-03-02
Speed up METS transformer by a factor of 30

jamieparkinson:
In https://github.com/wellcomecollection/catalogue/pull/450, we had to stop the METS transformer from extracting all of the images from METS files because it was too slow, and because having that many images caused issues in the ID minter. _This PR solves the first problem_.

Using the [new profiler in IntelliJ](https://blog.jetbrains.com/idea/2019/06/intellij-idea-2019-2-eap-4-profiling-tools-structural-search-preview-and-more/), I saw the construction of a `ListMap` for `fileReferences` was the bottleneck: addition to a `ListMap` has linear time complexity, and it's constructed by adding each element in turn (I think that means the construction goes as `O(n log n)`? Not sure).

This PR removes that `ListMap` and instead looks up `FileReference`s lazily when it parallel-maps over the (ordered) list of physical IDs, the generation of which I also simplified.

***

The figure of 30x improvement comes from running `MetsXmlTransformer.transform(..).toWork` on the enormous METS file, `b21465071`. This has 640 assets in its first manifestation and 1094 in its second.

Over a sample of 100 transformations, the transformation averaged 4.36s before this PR, and 0.140s after 🎉.

***

**This PR should not be merged until we've worked out how to get the ID minter to handle this many images**.

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
@@ -134,15 +134,44 @@ case class MetsXml(root: Elem) {
     }
   }
 
-  /** The keys of the physicalStructMap are the METS file IDs.
-    * We require nothing of them here other than that they are unique. */
-  def physicalFileIds: List[String] = physicalStructMap.values.toList
+  /** Valid METS documents should contain a physicalStructMap section, with the
+    *  bottom most divs each representing a physical page, and linking to files
+    *  in the corresponding fileSec structures:
+    *
+    *  <mets:structMap TYPE="PHYSICAL">
+    *    <mets:div DMDID="DMDPHYS_0000" ID="PHYS_0000" TYPE="physSequence">
+    *      <mets:div ADMID="AMD_0001" ID="PHYS_0001" ORDER="1" TYPE="page">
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
```diff
@@ -193,62 +220,23 @@ case class MetsXml(root: Elem) {
...
-        valueAttrib = "FILEID"
+  private def getFileReference(id: String) =
+    for {
+      fileGrp <- root \ "fileSec" \ "fileGrp"
+      objects <- fileGrp.find(_ \@ "USE" == "OBJECTS")
+      listedFiles = objects \ "file"
+      file <- listedFiles.find(_ \@ "ID" == id)
+      objectHref <- (file \ "FLocat").headOption
+        .map(_ \@ "{http://www.w3.org/1999/xlink}href")
+      if objectHref.nonEmpty
```
warrd: Finding this block pretty tricky to get my head around, but probably mainly due to fact I don't like xpath, and it's not like the original was much easier.

jamieparkinson: XML parsing is hell

kenoir:
@jamieparkinson nice application of brain to code.

PR #453, opened 2020-03-03
Use TransformerWorker

jamesgorrie:
Had a few different abstractions.

I think I got caught on trying to abstract away from the initial stream as it was harder to test as it was tied to the SQS implementation.

This moves all the abstractions into a `TransformerWorker` that takes a `stream`, `store`, `transformer` and `sender` and follows that flow.

We could probably abstract a few of the pieces out but I want to wait for the work @alicefuzier is doing on the Worker before I do.

Writing tests for the generic TransformerWorker should mean we can just use it in all the palces we want to without having to rewrite it the whole time.


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -0,0 +1,94 @@
...
+    case Left(err)                   => Left(StoreReadError(err.toString))
+    case Right(Identified(_, entry)) => Right(entry)
+  }
+
+  def run(): Future[Done] = {
+    stream.runStream(
+      "CalmTransformerWorkerService",
+      source => {
+        val end = source.via(Flow.fromFunction(message => message._1))
+        val processed = withSource(source)
```
jamesgorrie: @warrd was there something here you mentioned about `AKKA` kicking up another process that I should be wary of?

warrd: Hmmm not sure what you are referring to. Akka shouldn't kick up another process: has a thread pool in the same process it distributes jobs to as far as I know, but don't remember discussing anything round that.

jamieparkinson:
If this isn't used elsewhere, what do we gain by having a generic `TransformerWorker` as opposed to the same code but with the specialised parameters and 1 less layer of abstraction?

warrd:
@jamieparkinson I think the plan was to make something generic which could be used within other transformers (as they all work pretty much the same way apart from the Mets transformer)

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformerWorker.scala
```diff
@@ -1,55 +1,22 @@
...
+
+class CalmTransformerWorker(
+  val stream: SQSStream[NotificationMessage],
+  val sender: BigMessageSender[SNSConfig, TransformedBaseWork],
+  val store: VersionedStore[String, Int, CalmRecord],
+  val transformer: Transformer[CalmRecord]
+)(implicit val ec: ExecutionContext)
+    extends Runnable
+    with TransformerWorker[CalmRecord, SNSConfig] {
+  val name = "CalmTransformerWorker"
```
warrd: I would remove the `transformer` argument, as this will end up being `CalmTransformer` in #451 rather than being passed in some arbitrary `Transformer[CalmRecord]`:

```suggestion
)(implicit val ec: ExecutionContext)
    extends Runnable
    with TransformerWorker[CalmRecord, SNSConfig] {
  val transformer: Transformer[CalmRecord] = ???
  val name = "CalmTransformerWorker"
```

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -0,0 +1,95 @@
...
+  * - Gets the record of type `In`
+  * - Runs it through a transformer and transforms the `In` to `TransformedBaseWork`
+  * - Emits the message via `BigMessageSender` to an SNS topic
+  */
+trait TransformerWorker[In, SenderDest] {
+  type StreamMessage = (Message, NotificationMessage)
+  type Result[T] = Either[Throwable, T]
+  type StoreKey = Version[String, Int]
+
+  val name: String
```
warrd: How about this? It generally ends up being the class name in all our workers anyway (and could override with something different if required):

```suggestion
  def name: String = this.getClass.getSimpleName
```

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -0,0 +1,95 @@
...
+  val stream: SQSStream[NotificationMessage]
+  val sender: BigMessageSender[SenderDest, TransformedBaseWork]
+  val store: VersionedStore[String, Int, In]
+  val transformer: Transformer[In]
+
+  val errorSink: Sink[Result[_], Future[Done]] = Sink.ignore
```
warrd: What do we want to happen to these errors? Maybe at lease log them?

```suggestion
  def errorSink: Sink[Result[_], Future[Done]] =
    Flow[Result[_], NotUsed].collect { case Left(err) => logError(err) }.to(Sink.ignore)
  
  def logError(err: Throwable) =
    error(s"Exception in worker $name", err)
```

jamieparkinson:
@warrd @jamesgorrie I guessed that was the case (although I don't think this'll work as-is for the Sierra transformer, either), but I feel like we should only be creating complex abstractions like this at the time that we use them - otherwise we're just making this service more complex.

If we're to stick with this and commit to re-using it, then should it be in `/common`?

warrd:
@jamieparkinson yeh agree there is a balance to be had over how much abstraction and when to introduce it (I maybe tend towards the over abstraction at times). In this case I don't think it really adds much complexity / extra cognitive load to understand the code, and helps keeps the transformation of data and processing of the message nicely separated.

I think the Sierra transformer could be updated to fit this pattern without too much work (e.g. [this class](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformableTransformer.scala) is basically a `Transformer[SierraTransformable]`, and the [`HybridRecord`](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/services/HybridRecordReceiver.scala#L21-L25) message it receives  is key into some store)

jamesgorrie:
@jamieparkinson the reason to do this was to start moving towards a common way of logging, which I would argue hasn't happened as we haven't created the abstraction. 

I am working to the rule of threes, I was creating the fourth one, so thought I would abstract it out.

It will make abstracting to the new Worker more easily, and also make creating a new one easier and get rid of the need to think about the shape of a transformer, rather on the transformation logic.

I hate abstracting things as copy/paste generally works, but it feels like, as a small team, this work is taking too long, and becoming unsustainable to maintain 4 transformers, the cognitive load of doing them slightly differently, so this felt like it would free up human resource to be doing more interesting things.

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformerWorker.scala
jamesgorrie: This would remove the ability to move this to the other transformers though.

jamesgorrie:
To be clear this is the first step in moving towards having _a_ way to do transformers, I just prefer to do PRs in 1 massive sweep.

Do we not feel that that aim is worth abstracting?

kenoir:
@jamieparkinson on the surface it seems obvious that we should have an abstraction for a transformer as we have 4 of them and they all perform the same basic function - how different that is from the function of any pipeline service is a good question though, i.e. thing in -> do work -> thing out. 

I'm tempted to conclude that it is a useful abstraction as transformers also pull records for transformation and have a common set of failure modes worth describing.

kenoir:
> To be clear this is the first step in moving towards having _a_ way to do transformers, I just prefer to do PRs in 1 massive sweep.
> 
> Do we not feel that that aim is worth abstracting?

I think this is worth abstracting but that large PRs that affect existing code are not a good idea (though they are tempting as it can reduce cognitive load on the author, something I can relate to).

kenoir:
> @warrd @jamesgorrie I guessed that was the case (although I don't think this'll work as-is for the Sierra transformer, either), but I feel like we should only be creating complex abstractions like this at the time that we use them - otherwise we're just making this service more complex.
> 
> If we're to stick with this and commit to re-using it, then should it be in `/common`?

That these seems like a "complex abstraction" is a red flag - the underlying commonality between the transformers is not complex and the code should represent that.

I would prefer to see the common behaviours expressed as traits and as you suggest remove these to `/common`. That might also provide a more effective place to experiment, rather than introducing this to in progress work (of course doing that increases the risk that useful work will be ignored, or never make it to a mergeable state as time pressure pushes on other direct "product work").

kenoir:
Ultimately this is a small change and probably just a thing that we should merge as an experiment (having the abstraction confined to a single transformer is a benefit here).

I think that if we are interested in this approach there must be more tests (IMHO they might have been useful here to drive out a generic transformer) but that it's not a deal breaker to merging this.

jamesgorrie:
To be clear on this, it is probably my own workflow I wasn't clear about.

- This is in the direction of an abstraction
- I wanted to try and do it on the Calm Transformer as it felt weird to have to setup all the testing / code etc again
- I did that in the transformer
- I have ambitions to move this out to the other transformers, but didn't want that to become massive PRs

PR #454, opened 2020-03-04
Batch SQL queries in the idminter

alicefuzier:


pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
```diff
@@ -0,0 +1,49 @@
...
+import scala.util.{Failure, Success, Try}
+
+object SourceIdentifierScanner extends Logging {
+  def scan(inputJson: Json): Try[List[SourceIdentifier]] =
+    Try(
+      iterate(
+        root.each.json.getAll(inputJson),
+        findIdentifier(root.sourceIdentifier.json.getOption(inputJson)).toList))
+
+  private def iterate(
```
jamieparkinson: I wonder if there's a way to make this tail recursive as the stack is going to get huuuuge for big documents

alicefuzier: yeh,  I couldn't find a way to do that. I'll poke around a bit more but we might need to give it more memory

alexwlchan:
Will review shortly.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -1,82 +1,145 @@
...
+
+        val i = identifiers.i
+        val query = withSQL {
+
+          select
+            .from(identifiers as i)
+            .where
+            .in(
+              (i.OntologyType, i.SourceSystem, i.SourceId),
+              sqlParametersToSourceIdentifier.keys.toList)
```
alexwlchan: Personal opinion: the functions make this a bit less readable: if you’d created the list of SQL parameters right next to these column headers, their use would be obvious.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -1,82 +1,145 @@
...
+
+        }.map(rs => {
+            val foundParameter = buildSqlParametersFromResult(i, rs)
+            sqlParametersToSourceIdentifier.get(foundParameter) match {
+              case Some(sourceIdentifier) =>
+                (sourceIdentifier, Identifier(i)(rs))
+              case None =>
+                // this should be impossible in practice
+                throw new RuntimeException(
+                  "The values returned by the query could not be matched to a sourceIdentifier")
```
alexwlchan: Minor: if I get this “impossible” error, I’m not going to know why. Include `foundParameter` in the error message to make it easier to debug.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -1,82 +1,145 @@
...
+
+          })
+          .list
+        debug(s"Executing:'${query.statement}'")
+        val foundIdentifiers = query.apply().toMap
+        val notFoundIdentifiers =
+          sqlParametersToSourceIdentifier.values.toSet -- foundIdentifiers.keySet
+        LookupResult(
+          foundIdentifiers,
+          notFoundIdentifiers.toList
```
alexwlchan: Minor: "found" / "not found" describes these in terms of the database concept; would names that describe what this means at a higher level be useful?

Say "existingIdentifiers" and "newIdentifiers"?

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -1,82 +1,145 @@
...
+    val failedIdentifiers =
+      groupedResult.getOrElse(Statement.EXECUTE_FAILED, Nil).map {
+        case (identifier, _) => identifier
+      }
+    val succeededIdentifiers =
+      groupedResult.getOrElse(Statement.SUCCESS_NO_INFO, Nil).map {
+        case (identifier, _) => identifier
+      }
+    InsertError(failedIdentifiers, exception, succeededIdentifiers)
+  }
```
alexwlchan: FYI, I know very little about SQL errors or the like, so I’m taking it on faith this works correctly. (Comments would be nice!)

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
```diff
@@ -0,0 +1,108 @@
...
+import scala.annotation.tailrec
+import scala.util.{Failure, Success, Try}
+
+/**
+  * SourceIdentifierScanner provides 2 methods:
+  *
+  * - `scan` takes Json and returns all of the sourceIdentifiers that are in it
+  * - `update` takes Json and a map of (SourceIdentifier -> Identifier) and adds
+  *   a canonicalId field next to sourceIdentifiers, as well as replacing
+  *   `identifiedType` fields with `type` fields of the same value.
```
alexwlchan: tiny nit: listing the methods in the order they're described would have been easier to scan

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
```diff
@@ -0,0 +1,108 @@
...
+            .map(parseSourceIdentifier)
+        )
+    }
+
+  private def parseSourceIdentifier(
+    sourceIdentifierJson: Json): SourceIdentifier = {
+    sourceIdentifierJson.as[SourceIdentifier].toTry match {
+      case Success(sourceIdentifier) => sourceIdentifier
+      case Failure(exception) =>
+        error(s"Error parsing source identifier: $exception")
```
alexwlchan: Is it worth logging the value of `sourceIdentifierJson`? My experience of Circe error messages is that they’re only helpful if you can see the JSON it was looking at.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
```diff
@@ -0,0 +1,108 @@
...
+    root.identifiedType.json
+      .getOption(node)
+      .flatMap(_.asString)
+      .map { identifiedType =>
+        root.obj.modify { obj =>
+          ("type", Json.fromString(identifiedType)) +:
+            obj.remove("identifiedType")
+        }(node)
+      }
+      .getOrElse(node)
```
alexwlchan: Should we throw an exception here if we can’t find a field `"identifiedType"`? Is that ever something that could happen in practice?

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
```diff
@@ -0,0 +1,108 @@
...
+  *
+  */
+object SourceIdentifierScanner extends Logging {
+  def update(inputJson: Json,
+             identifiers: Map[SourceIdentifier, Identifier]): Try[Json] =
+    Try {
+      val updateNode =
+        (updateNodeType _) compose addCanonicalIdToNode(identifiers)
+      // Plated transforms operate on self-similar *children*
+      // so we need to update the root separately
```
alexwlchan: Confession: I have no idea what this comment means.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
```diff
@@ -1,41 +1,48 @@
...
-
-  private def generateAndSaveCanonicalId(
-    sourceIdentifier: SourceIdentifier
-  ): Try[String] = {
-    val canonicalId = Identifiable.generate
+class IdentifierGenerator(identifiersDao: IdentifiersDao) extends Logging {
+  // This function fetches canonicalIds for sourceIdentifiers, and generates
+  // and saves canonicalIds where it can't find existing ones. For this reason
+  // it's important not to run it in situations where there might be a race
+  // condition (IDs being inserted in between lookup and update).
```
alexwlchan: What can I do to prevent race conditions?

Isn't that what MySQL integrity checks are for? If two processes try to create and save identifiers at the same time, only one of them will win, the other will throw an exception and the message will either retry or be dropped from the queue.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
```diff
@@ -1,41 +1,48 @@
...
-      .map { _ =>
-        canonicalId
+      .lookupIds(sourceIdentifiers)
+      .flatMap {
+        case IdentifiersDao.LookupResult(foundMap, notFoundIdentifiers) =>
+          generateAndSaveCanonicalIds(notFoundIdentifiers).map {
+            newIdentifiers =>
+              val newMap: Map[SourceIdentifier, Identifier] =
+                (notFoundIdentifiers zip newIdentifiers)(breakOut)
+              foundMap ++ newMap
```
alexwlchan: See comment earlier; picking better names than "found" or "not found" would make the path to `newIdentifiers` clearer.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
alexwlchan: Also, what is a Scanner doing updating JSON?

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/services/IdMinterWorkerService.scala
```diff
@@ -33,12 +41,37 @@ class IdMinterWorkerService[Destination](
...
 
-  def processMessage(json: Json): Future[Unit] =
-    for {
-      identifiedJson <- idEmbedder.embedId(json)
-      _ <- Future.fromTry { sender.sendT(identifiedJson) }
-    } yield ()
+  private def extractIdentifiers =
+    Flow[(Message, Json)].map {
+      case (msg, json) =>
+        (msg, json, SourceIdentifierScanner.scan(json).get)
```
alexwlchan: Why are we doing a `.get` here? I thought that was bad practice. (Ditto below.)

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -35,56 +37,124 @@ class IdentifiersDaoTest
       testWith((identifiersDao, identifiersTable))
     }
 
-  describe("lookupID") {
-    it("gets an Identifier if it finds a matching SourceSystem and SourceId") {
+  describe("lookupIds") {
+
+    it("retrieves a single id from the identifiers database") {
```
alexwlchan: minor: more accurately, what this tests is the stronger property that it can store and then retrieve an identifier from the DB (and the old description was wrong too)

The pattern we have in the storage service/scala-storage library would be closer to

```scala
withIdentifiersDao(existingEntries) { ... }
```

and when the caller defines that fixture, they're responsible for putting those entries in place – in this case that would be SQL statements, like there are for testing the "does it store a thing correctly" code below.

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -35,56 +37,124 @@ class IdentifiersDaoTest
...
+          triedLookup shouldBe a[Success[_]]
+          triedLookup.get.found shouldBe identifiersMap
+          triedLookup.get.notFound shouldBe empty
       }
     }
 
     it(
-      "does not get an identifier if there is no matching SourceSystem and SourceId") {
-      val identifier = createSQLIdentifier
+      "matches correctly the ids to the sourceIdentifier when the sourceIdentifiers have overlapping values") {
```
alexwlchan: Maybe clearer:

```suggestion
      "matches ids based on all fields on the SourceIdentifier (same value, different ontology type)") {
```

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -35,56 +37,124 @@ class IdentifiersDaoTest
...
+      val sourceIdentifier = createSourceIdentifier
+      val overlappingSourceIdentifier =
+        sourceIdentifier.copy(ontologyType = "Item")
```
alexwlchan: This works as long as `createSourceIdentifier` picks an ontology that isn't `Item`.

Make the difference more explicit, and get names that line up nicely to boot:

```suggestion
      val workSourceIdentifier = createSourceIdentifierWith(ontologyType = "Work")
      val itemSourceIdentifier = workSourceIdentifier(ontologyType = "Item")
```

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -35,56 +37,124 @@ class IdentifiersDaoTest
...
+          triedLookup.get.notFound shouldBe empty
+      }
+    }
 
-          val unknownSourceIdentifier = createSourceIdentifierWith(
-            ontologyType = identifier.OntologyType,
-            value = "not_an_existing_value"
-          )
+    it(
+      "retrieves multiple ids and returns the ones it doesn't have in notFound") {
```
alexwlchan:
```suggestion
      "stores the IDs that aren't in the dao in the notFound field") {
```

(still not in love with the name)

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -35,56 +37,124 @@ class IdentifiersDaoTest
...
+      val identifiersMap = (1 to 3).map { _ =>
+        val sourceIdentifier = createSourceIdentifier
+        val identifier =
+          createSQLIdentifierWith(sourceIdentifier = sourceIdentifier)
+        (sourceIdentifier, identifier)
+      }.toMap
+      val nonExistingSourceIdentifiers =
```
alexwlchan: nit: be consistent

```suggestion
      val notFoundSourceIdentifiers =
```

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -99,55 +169,84 @@ class IdentifiersDaoTest
...
+    it("fails saving the identifiers if the canonical id is already present") {
+      val ids = (1 to 5).map { _ =>
+        val sourceIdentifier = createSourceIdentifier
+        (sourceIdentifier, Identifier(createCanonicalId, sourceIdentifier))
+      }.toMap
+      val identifiers = ids.values.toList
+      val duplicatedIdentifier1 =
+        createSQLIdentifierWith(canonicalId = identifiers.apply(1).CanonicalId)
+      val duplicatedIdentifier2 =
+        createSQLIdentifierWith(canonicalId = identifiers.apply(3).CanonicalId)
```
alexwlchan: Why `3` and not `2`?

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -99,55 +169,84 @@ class IdentifiersDaoTest
...
-      )
-
-      val identifier1 = createSQLIdentifierWith(
-        sourceIdentifier = sourceIdentifier1
-      )
-
-      val identifier2 = createSQLIdentifierWith(
-        sourceIdentifier = sourceIdentifier2
-      )
+      "fails saving the identifiers if the same OntologyType,SourceName, and SourceId are already present") {
```
alexwlchan:
```suggestion
      "fails saving the identifiers if the same OntologyType, SourceName, and SourceId are already present") {
```

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -99,55 +169,84 @@ class IdentifiersDaoTest
...
 
     it(
-      "saves records with different SourceId but the same OntologyType and SourceSystem") {
-      val sourceIdentifier1 = createSourceIdentifierWith(
-        value = "1234"
-      )
-      val sourceIdentifier2 = createSourceIdentifierWith(
-        value = "5678"
-      )
+      "saves records with the same SourceSystem and SourceId but different OntologyType") {
```
alexwlchan: Is it the SourceSystem or the SourceName? This and your previous test disagree.

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -99,55 +169,84 @@ class IdentifiersDaoTest
...
+      val sourceIdentifier1 = createSourceIdentifier
+      val sourceIdentifier2 = sourceIdentifier1.copy(ontologyType = "Foo")
```
alexwlchan: An ontologyType of `"Foo"` is probably less likely from `createSourceIdentifier`, but would still be good to be explicit that they're different.

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/fixtures/SqlIdentifiersGenerators.scala
```diff
@@ -0,0 +1,18 @@
...
+    canonicalId: String = createCanonicalId,
+    sourceIdentifier: SourceIdentifier = createSourceIdentifier
+  ): Identifier =
+    Identifier(
+      canonicalId = canonicalId,
+      sourceIdentifier = sourceIdentifier
+    )
+
+  def createSQLIdentifier: Identifier = createSQLIdentifierWith()
+}
```
alexwlchan: 👍

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScannerTest.scala
```diff
@@ -0,0 +1,190 @@
...
+    `type`: String,
+    classWithIdentifiers: List[ClassWithIdentifierTypeAndCanonicalId] = Nil
+  )
+  describe("scan") {
+    it("retrieves a sourceIdentifiers at the root of the json") {
+      val sourceIdentifier = createSourceIdentifier
+      val objectWithIdentifier = ClassWithIdentifier(sourceIdentifier)
+      SourceIdentifierScanner
+        .scan(objectWithIdentifier.asJson)
+        .get shouldBe List(sourceIdentifier)
```
alexwlchan: It's more lines, but IMO these tests would be way more readable if I could see the JSON strings that were being manipulated directly, rather than having to memorise the case class hierarchy. For example:

```suggestion
      val jsonString = """
        |{
        |  "sourceIdentifier": {
        |    "identifierType": "miro-image-number",
        |    "value": "A0001234",
        |    "ontologyType": "Work"
        |  }
        |}
      """.stripMargin

      val sourceIdentifier = SourceIdentifier(
        identifierType = "miro-image-number",
        value = "A0001234",
        ontologyType = "Work"
      )
      
      val json = parse(jsonString).get

      SourceIdentifierScanner
        .scan(json)
        .get shouldBe List(sourceIdentifier)
```

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScannerTest.scala
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    it("retrieves a sourceIdentifiers nested in the json") {
```
alexwlchan: Is this testing that (1) it can find multiple instances of SourceIdentifier or (2) that it can find nested instances or (3) both? The test description is throwing a parsing error in my brain.

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScannerTest.scala
```diff
@@ -0,0 +1,190 @@
...
+      val identified = SourceIdentifierScanner.update(
+        objectWithIdentifier.asJson,
+        Map(sourceIdentifier -> identifier))
+      identified shouldBe a[Success[_]]
+      identified.get
+        .as[ClassWithIdentifierAndCanonicalId]
+        .right
+        .get shouldBe ClassWithIdentifierAndCanonicalId(
+        sourceIdentifier,
+        canonicalId = identifier.CanonicalId)
```
alexwlchan: As above, I'd find the JSON easier to follow than the case classes:

```suggestion
      val jsonString = """
        |{
        |  "sourceIdentifier": {
        |    "identifierType": "miro-image-number",
        |    "value": "A0001234",
        |    "ontologyType": "Work"
        |  },
        |  "identifiedType": "IdentifiedWork"
        |}
      """.stripMargin
      
      val sourceIdentifier = SourceIdentifier(
        identifierType = "miro-image-number",
        value = "A0001234",
        ontologyType = "Work"
      )
      
      val identifier = Identifier(
        canonicalId = "single-canonical-id-test",
        sourceIdentifier = sourceIdentifier
      )
            
      val identified = SourceIdentifierScanner.update(
        parse(jsonString).get,
        identifiers = Map(sourceIdentifier -> identifier)
      )
      identified shouldBe a[Success[_]]
      
      val updatedJsonString = 
      
      assertJsonStringsEqual(
        identified.get.noSpaces,
        """
        {
          "sourceIdentifier": {
            "identifierType": "miro-image-number",
            "value": "A0001234",
            "ontologyType": "Work"
          },
          "canonicalId": "single-canonical-id-test",
          "type": "IdentifiedWork"
        }
        """
      )
```

(Note: as written, I don't think these case classes have an `identifiedType` field. Should they?)

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
jamieparkinson: To be honest I'm doing the same - @alicefuzier did the difficult stuff (ie the SQL) in this PR! If it proves to be a problem we can add some comments in future

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/services/IdMinterWorkerService.scala
jamieparkinson: Because if there's an exception we want to throw it so akka-streams can do it's thing.

I was going to move all of the exception logging to here, but then I found that there's so much context that's useful to have (as you've noticed in other comments) in the logs that didn't work very well. 

When this was a for comprehension (which you can see in the commit history) it was all rather nice but it was necessary to move to streams to deal with concurrency issues.

I might add an error-handling convenience method to tack on the end of each flow and strip off the `Try` instead, what do you think?

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
jamieparkinson: Welcome to the weird world of circe-optics - that's not something that could happen in practice, because `root.identifiedType.json.getOption` is ensuring that it exists

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
jamieparkinson: Looks like something changed while these were being written last time round! (I've mostly just copied and pasted them). Good spot.

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScannerTest.scala
jamieparkinson: both, will clarify

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScannerTest.scala
jamieparkinson: re `identifiedType` - there are separate types/classes for with and without because it needs to work for both situations

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
alexwlchan: I think we’re talking at cross-purposes – `getOption` ensures you either find it or you don’t, so there won’t be an unexpected runtime error – but what if we’re passed a blob of JSON where `"identifiedType"` isn't present? That could surely happen?

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
jamieparkinson: That is true - whilst I believe it is still technically a race condition, it would cause issues 100% of the time. I'll try to explain...

Consider `n` messages containing identical and unminted `sourceIdentifiers` arriving in the minter at the same time.

Because we are good developers, we know we should be able mint everything concurrently, so we use a `mapAsyncUnordered` on the message stream, creating `n` threads - one for each message. Let's simplify and say we only have one processing core, so they'll never _actually_ be running at the same time.

JDBC calls are inherently blocking, so when a thread gets to the `lookup` operation it will block and the scheduler will switch to one of the other `n - 1` threads, where the same thing will happen, and before we know it all `n` threads are doing their lookup at once and finding nothing for the `sourceIdentifier`.

Each thread will then generate a new identifier and try to save it. One of them will get there first, and the other `n - 1` will fail with certainty and need to be redriven, every time! 😞

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
jamieparkinson: Yes absolutely! That's what the `getOrElse` at the end is doing - allowing `node`s without an `identifiedType` to pass straight through.

alexwlchan: I’m struggling to explain what I mean, I think.

When would the ID minter be asked to mint identifiers on a JSON map that *didn't* have an `"identifiedType"` field? I thought we had that everywhere?

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
```diff
@@ -50,19 +50,24 @@ object SourceIdentifierScanner extends Logging {
...
+      .map { canonicalId =>
+        root.obj.modify { obj =>
+          ("canonicalId", Json.fromString(canonicalId)) +: obj
+        }(node)
       }
       .getOrElse(node)
 
+  private def getCanonicalId(identifiers: Map[SourceIdentifier, Identifier])(
+    sourceIdentifier: SourceIdentifier): String =
+    identifiers
```
kenoir: This is definitely clearer.  You could avoid the exception here by dealing with `Either` everywhere  but that might be overcomplicating things.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierScanner.scala
jamieparkinson: Ah, I get you now - I don't know whether we have it everywhere, but I do know that the old ID minter code explicitly handled (and didn't throw an exception for) the case where a thing didn't have an `identifiedType`, so I chose to maintain existing behaviour here.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
jamieparkinson: `newIdentifiers` doesn't work great in other contexts as they are not yet new - the "new"ness happens after they're generated. I'm going to opt for `unmintedIdentifiers` instead.

alexwlchan:
Re-reviewing now.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -1,82 +1,134 @@
...
-      debug(s"Executing:'${query.statement}'")
-      query.apply()
+  def lookupIds(sourceIdentifiers: Seq[SourceIdentifier]): Try[LookupResult] = {
+    Try {
+      debug(s"Matching ($sourceIdentifiers)")
+      val identifierRows = sourceIdentifiers.map { id =>
+        (id.ontologyType, id.identifierType.id, id.value)
+      }
+      val identifierRowsMap: Map[(String, String, String), SourceIdentifier] =
+        (identifierRows zip sourceIdentifiers).toMap
```
alexwlchan: Personally I'd be tempted to skip the intermediate variable and the `zip`, but I don't feel strongly:

```suggestion
      val identifierRowsMap: Map[(String, String, String), SourceIdentifier] =
        sourceIdentifiers
          .map { id => (id.ontologyType, id.identifierType.id, id.value) -> id }
          .toMap
```

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -1,82 +1,134 @@
...
+
+        }.map(rs => {
+            val row = getRowFromResult(i, rs)
+            identifierRowsMap.get(row) match {
+              case Some(sourceIdentifier) =>
+                (sourceIdentifier, Identifier(i)(rs))
+              case None =>
+                // this should be impossible in practice
+                throw new RuntimeException(
+                  s"The row $row returned by the query could not be matched to a sourceIdentifier")
```
alexwlchan: 👍

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
```diff
@@ -1,41 +1,53 @@
...
-          canonicalId = canonicalId,
-          sourceIdentifier = sourceIdentifier
-        )
-      )
-      .map { _ =>
-        canonicalId
+      .lookupIds(sourceIdentifiers)
+      .flatMap {
+        case IdentifiersDao
+              .LookupResult(existingIdentifiersMap, unmintedIdentifiers) =>
```
alexwlchan: Minor thing: the dot syntax made me think you were calling a method here for a moment; I didn't realise this was unpacking a case class. If you added `import IdentifiersDao._` and made this `case LookupResult(…)` it would have read slightly cleaner to me.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
```diff
@@ -1,41 +1,53 @@
...
+            newIdentifiers =>
+              val newIdentifiersMap: Map[SourceIdentifier, Identifier] =
+                (unmintedIdentifiers zip newIdentifiers).toMap
+              existingIdentifiersMap ++ newIdentifiersMap
+          }
       }
-  }
+
+  private def generateAndSaveCanonicalIds(
+    sourceIdentifiers: List[SourceIdentifier]): Try[List[Identifier]] =
```
alexwlchan: Let's be clear about what sort of identifiers we expect here, especially as this is a private method:

```suggestion
    unmintedIdentifiers: List[SourceIdentifier]): Try[List[Identifier]] =
```

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierEmbedder.scala
```diff
@@ -0,0 +1,109 @@
...
+  def scan(inputJson: Json): Try[List[SourceIdentifier]] =
+    Try(
+      iterate(
+        root.each.json.getAll(inputJson),
+        root.sourceIdentifier.json
+          .getOption(inputJson)
+          .map(parseSourceIdentifier)
+          .toList
+      )
+    )
```
alexwlchan: ordering! :D

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -32,59 +34,137 @@ class IdentifiersDaoTest
...
     it(
-      "does not get an identifier if there is no matching SourceSystem and SourceId") {
-      val identifier = createSQLIdentifier
+      "matches ids based on all fields on the SourceIdentifier (same value, different ontology type)") {
+      val workSourceIdentifier = createSourceIdentifierWith(
+        ontologyType = "Work"
+      )
+      val itemSourceIdentifier = workSourceIdentifier.copy(
+        ontologyType = "Item"
+      )
```
alexwlchan: 🎉

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGeneratorTest.scala
```diff
@@ -45,98 +48,107 @@ class IdentifierGeneratorTest
...
-              identifiersTable.column.SourceSystem -> sourceIdentifier.identifierType.id,
-              identifiersTable.column.SourceId -> sourceIdentifier.value,
-              identifiersTable.column.OntologyType -> sourceIdentifier.ontologyType
+              identifiersTable.column.CanonicalId -> sqls.?,
+              identifiersTable.column.SourceSystem -> sqls.?,
+              identifiersTable.column.SourceId -> sqls.?,
+              identifiersTable.column.OntologyType -> sqls.?
             )
-        }.update().apply()
+        }.batch(insertParams: _*).apply()
```
alexwlchan: Worth pushing this into the fixture as well, as you've done above?

pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/SourceIdentifierEmbedderTest.scala
```diff
@@ -0,0 +1,388 @@
...
+        |    "ontologyType": "${sourceIdentifier.ontologyType}"
+        |  }
+        |}
+      """.stripMargin
+      val json = parse(jsonString).right.get
+      val identifier =
+        createSQLIdentifierWith(sourceIdentifier = otherSourceIdentifier)
+      val identified = SourceIdentifierEmbedder.update(
+        json,
+        Map(otherSourceIdentifier -> identifier))
```
alexwlchan: Do we need the `otherSourceIdentifier` here? AIUI, it’s walking the JSON and trying to look up the `sourceIdentifier` in the map of identifiers it receives – would `Map.empty` work too?

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
jamieparkinson: Used 2 variables here as the `identifierRows` is used later

PR #457, opened 2020-03-04
Calm adapter fixes

warrd:


calm_adapter/calm_window_generator/src/lambda.py
```diff
@@ -19,5 +20,7 @@ def main(event=None, _ctxt=None):
     start = (datetime.now() - timedelta(hours=crossover_hours)).date()
     end = date.today()
 
+    sns_client = boto3.client("sns")
```
jamesgorrie: Does this need to use `STS` too?

jamesgorrie: Wait, no, this iexists in AWS.

PR #460, opened 2020-03-05
Fix duplicate locations when merging METS and Sierra

jamieparkinson:
It turns out that we missed a subtlety when simplifying the METS merging logic - if the Sierra work has an item location with the same URL as the METS item, that should be filtered out (see [here](https://github.com/wellcomecollection/catalogue/blob/3e3ccbc3134b2aaff0b3f906335f26ad97ba60a2/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/sierramets/SierraMetsWorkPairMerger.scala#L85)).

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
@@ -48,7 +48,7 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
       val metsUrls = metsItems.flatMap(_.locations).collect {
         case DigitalLocation(url, _, _, _, _, _) => url
       }
-      info(s"Merging METS items from ${describeWorks(sources)}")
+      debug(s"Merging METS items from ${describeWorks(sources)}")
       sierraItems match {
         case List(sierraItem) =>
           List(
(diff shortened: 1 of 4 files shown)


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -44,15 +46,21 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
              sources: Seq[TransformedBaseWork]): FieldData = {
       val sierraItems = target.data.items
       val metsItems = sources.flatMap(_.data.items)
-      info(s"Merging METS items from ${describeWorks(sources)}")
+      val metsUrls = metsItems.flatMap(_.locations).collect {
+        case DigitalLocation(url, _, _, _, _, _) => url
+      }
+      debug(s"Merging METS items from ${describeWorks(sources)}")
```
jamesgorrie: Why debug?

jamieparkinson:
@jamesgorrie yep, specifically this - https://github.com/wellcomecollection/catalogue/blob/3e3ccbc3134b2aaff0b3f906335f26ad97ba60a2/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/sierramets/SierraMetsWorkPairMerger.scala#L85

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamieparkinson: It was logging so much per merge I thought I'd quiet it down a bit

warrd: :+1: for `DEBUG`, tests especially are so noisy (can always temporarily set log level to `DEBUG` in config if need to investigate anything further)

PR #462, opened 2020-03-09
make transformer work a for comprehension

jamesgorrie:
Combining the streams from the worker and the work that's needed in a transformer was create a bit of a spaghetti mess as I think I was conflating responsibilities.

This defers the stream logic back to the messaging library when dealing with the logic about failing / non-failing / dlq etc logic.

The transformer logic is then a simple for comprehension.

When we update the messaging library we should be able to update this abstraction pretty easily as the shape of it won't change much.


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -37,60 +43,64 @@ trait TransformerWorker[In, SenderDest] {
...
+      case Right(Identified(key, entry)) => Right((entry, key))
+    }
 
   def run(): Future[Done] =
     stream.runStream(
       name,
       source => {
-        val end = source.via(Flow.fromFunction(message => message._1))
-        val processed = withSource(source)
+        source.mapAsync(2) {
```
warrd: Probably want this as a parameter on the trait?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -37,60 +43,64 @@ trait TransformerWorker[In, SenderDest] {
...
+                // We do some slightly nicer logging here to give context to the errors
+                err match {
+                  case DecodeKeyError(_, message) =>
+                    error(s"$name: DecodeKeyError from $message")
+                  case StoreReadError(_, key) =>
+                    error(s"$name: StoreReadError on $key")
+                  case TransformerError(_, sourceData, key) =>
+                    println(s"$name: TransformerError on $sourceData with $key")
+                  case MessageSendError(_, work, key) =>
+                    println(s"$name: MessageSendError on $work with $key")
```
warrd: Can we use debug or info instead of println for these two?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
jamesgorrie: Why? We do that elsewhere but never use it?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
jamesgorrie: Sorry - bad debugging on my behelf - thanks!

jamesgorrie: Using `error` though - does that make sense?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
warrd: It feels presumptuous to assume 2 concurrent transformations will always be the ideal value for any transformer. And this isn't always the case anyway (at least in the METS transformer this isn't true).

How about we set this as to 2 by default and then can be overriden if required?

```scala
val concurrentTransformations: Int = 2
```

jamesgorrie: Was playing it way to much in the non-abstract thinking way forgetting this will be used elsewhere.

jamesgorrie: After having done that though, I wonder if it's a parameter sent to the `run` method?

warrd: I would probably keep it at trait level, as the caller shouldn't be so concerned about it (seems to be an implementation detail the worker would know more about than whoever is calling the worker). But up to you, would work as a default arg too.

PR #464, opened 2020-03-10
add collection.depth filter

jamesgorrie:
Adding a filter that allows you to get works with a specified collection depth.

This is not intended to cover all bases of what we _might_ want to do, but allow us to start surfacing stuff to the `www` so we can expose it to people who can tell us how they expect to use it.

## 👁 forward
This _might_ be more useful as a range, and I imagine that we might use something like `collection.depth.from` & `collection.depth.till`.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
@@ -155,7 +155,7 @@ object MultipleWorksParams extends QueryParamsUtils {
     decodeCommaSeparated.emap(strs => Right(LicenseFilter(strs)))
 
   implicit val collectionsDepthFilter: Decoder[CollectionDepthFilter] =
-    decodeInt map CollectionDepthFilter
+    decodeInt map (CollectionDepthFilter(_))
 
   implicit val aggregationsDecoder: Decoder[List[AggregationRequest]] =
     decodeOneOfCommaSeparated(
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
@@ -50,6 +50,7 @@ case class MultipleWorksParams(
   sort: Option[List[SortRequest]],
   sortOrder: Option[SortingOrder],
   query: Option[String],
+  `collection.depth`: Option[CollectionDepthFilter],
   _queryType: Option[SearchQueryType],
   _index: Option[String],
 ) extends QueryParams {
@@ -85,6 +86,7 @@ case class MultipleWorksParams(
       language,
       `genres.label`,
       `subjects.label`,
+      `collection.depth`,
       license
     ).flatten
 
@@ -120,6 +122,7 @@ object MultipleWorksParams extends QueryParamsUtils {
         "sort".as[List[SortRequest]].?,
         "sortOrder".as[SortingOrder].?,
         "query".as[String].?,
+        "collection.depth".as[CollectionDepthFilter].?,
         "_queryType".as[SearchQueryType].?,
         "_index".as[String].?,
       )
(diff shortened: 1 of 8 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -151,6 +154,9 @@ object MultipleWorksParams extends QueryParamsUtils {
   implicit val licenseFilter: Decoder[LicenseFilter] =
     decodeCommaSeparated.emap(strs => Right(LicenseFilter(strs)))
 
+  implicit val collectionsDepthFilter: Decoder[CollectionDepthFilter] =
+    decodeInt map CollectionDepthFilter
```
warrd: I'm surprised this compiles, as used to getting errors when doing this and have to do `CollectionDepthFilter(_)` instead. Just checked apparently it only complains when you have a companion object defined as well:

```
[E]      type mismatch;
[E]       found   : uk.ac.wellcome.platform.api.models.CollectionDepthFilter.type
[E]       required: Int => uk.ac.wellcome.platform.api.models.CollectionDepthFilter
```

api/api/src/test/scala/uk/ac/wellcome/platform/api/QueryParamsTest.scala
```diff
@@ -0,0 +1,5 @@
+package uk.ac.wellcome.platform.api
+
+import org.scalatest.FunSuite
+
+class QueryParamsTest extends FunSuite {}
```
warrd: Were you planning to add any tests here?

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/FiltersAndAggregationsBuilder.scala
```diff
@@ -87,6 +87,7 @@ class FiltersAndAggregationsBuilder(
     case _: GenreFilter            => Some(AggregationRequest.Genre)
     case _: SubjectFilter          => Some(AggregationRequest.Subject)
     case _: LicenseFilter          => Some(AggregationRequest.License)
+    case _: CollectionDepthFilter  => None
```
warrd: Maybe this pattern matching could just have a single case for all patterns returning `None`, rather than having to define them all individually?

```
case _ => None
```

jamesgorrie: I quite like having it there as it errors if you don't, and make you remember to think about it.

api/api/src/test/scala/uk/ac/wellcome/platform/api/QueryParamsTest.scala
jamesgorrie: Nope, I was going to but we had similar tests elsewhere, I'dlike to have a strategy for what, where and why we tests things as there's a few different bits all over the place.

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
jamesgorrie: Not 100% sure, I'll have a play around

api/api/src/test/scala/uk/ac/wellcome/platform/api/QueryParamsTest.scala
jamesgorrie: I think it is just that the compiler infers that you mean the case object rather than the class, exposes a function taking an `Int`.


Interestingly you can get the compiler into a tangle like so:
```
case class IntTaker(i: Int)
case object IntTaker {
  def apply(i: Int): IntTaker = IntTaker(i)
}
val intMaker = List(1,2,3)


intMaker map IntTaker.apply
```

warrd: Yup I understood that's what was going on, just posted cos caught me out at first :)

PR #465, opened 2020-03-11
add collection filter

jamesgorrie:
Second part allowing us to get all works in a collection.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
@@ -50,7 +50,7 @@ case class MultipleWorksParams(
   sort: Option[List[SortRequest]],
   sortOrder: Option[SortingOrder],
   query: Option[String],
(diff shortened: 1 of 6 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -50,6 +50,7 @@ case class MultipleWorksParams(
   sort: Option[List[SortRequest]],
   sortOrder: Option[SortingOrder],
   query: Option[String],
+  `collection`: Option[CollectionPathFilter],
```
warrd: no need for this to be in quotes

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -86,6 +87,7 @@ case class MultipleWorksParams(
       language,
       `genres.label`,
       `subjects.label`,
+      `collection`,
```
warrd: again

PR #469, opened 2020-03-13
Rename calm sourcedata VHS to be consistant

jamesgorrie:
Sorry for the revert nonsense, committed straight into master.

Changing the name as it is annoying to have to dig around thinking why it was named differently and realising it doesn't matter.


jamesgorrie:
Scrapping this as I am going to use the VHS in the source data.

PR #463, opened 2020-03-10
Merge calm works

warrd:
I have pulled out the docs stuff into #470
Also the calm items merging is commented out until @jtweed has rewritten the new rules.

PR #470, opened 2020-03-13
[WIP] Merger docs

warrd:
Pulled out from #463 

This is still a WIP, waiting for update from @jtweed


jamieparkinson:
@warrd can this be merged / can it include an update of the items rule block comment?

warrd:
Not ready yet as the rules may be incorrect currently

warrd:
Merging rules from @jtweed:

The works coming into the merger are as follows:
- Sierra physical work (one or more items)
- Sierra digital work (one item)
- METS work (one or more items)
- Calm work (one item)
- Miro work (one item)

When there are physical and digital Sierra works:
- Take all work data from the physical work
- Add the digital Sierra identifiers to the physical work
- Suppress the digital work

When there is a METS work:
- If the Sierra work has a single item, add the METS digital location to the Sierra item
- If the Sierra work has multiple items, add the METS item to the Sierra work
- Suppress the METS work

When there is a Calm work:
- Take all work and item data from the Calm work
- Add the Sierra identifiers to the Calm work and item
- If there is a digital Sierra work, add the identifiers to the Calm work and item
- If there is a METS work, add the digital location to the Calm item
- Suppress the Sierra work

When there is a Miro work:
- The image pipeline means that any Miro images linked to Sierra should added as images and suppressed
- This means that any remaining standalone Miro images should not have any merge candidates

pipeline/merger/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+  3. If METS data doesn't exist, we fall back to using digital locations from Miro / other matched Sierra works, and merge these onto the item
+
+When there are multiple items on the Sierra work, the following rules apply:
+
+  1. If there is a Calm work we add the physical item from Calm in addition to the Sierra items
+
+  2. If there is a METS work we add the digital item from METS in addition to the Sierra items
+
+  3. If METS data doesn't exist, we fall back to using digital locations from other matched Sierra works, and add these items in addition to the existing one, but ignore any Miro data
```
kenoir, line 25: It might be worth contextualising "Calm/METS/Sierra", define somewhere that Sierra is the library and Calm the archive etc. It's very obvious to us, but won't be to a newcomer.

jamesgorrie:
Given I am taking the logic from above and creating docs for each bit and getting it PRed and then validated that the understanding is also shared with the collections team - can we close this off?

(I am still using it heavily as a base though).

warrd:
more recent docs here: https://docs.wellcomecollection.org/catalogue/pipeline/merging

PR #475, opened 2020-03-16
use 1 calm adapter

jamesgorrie:
Agreeing with @warrd that it feels better with the project, but I would rather have 1 place and naming convention for the source data for now so I don't have to remember we did it partially correctly later on.


warrd:
Don't mind so much where the tf is, but does this mean we need to run the calm adapter ingest again?

jamesgorrie:
@warrd by that do you mean repopulating the calm adapter VHS?

jamesgorrie:
Thinking about it - almost definitely as the Dynamo table, even if copied, will have the wrong data.

PR #478, opened 2020-03-17
Make the Sierra progress reporter be less spammy

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4372

The Sierra reader writes a marker to S3 to say “I got all the updates from 00:00–00:30”. The Sierra progress reporter consolidates overlapping markers, so we can see if we've got complete coverage of the Sierra updates. Recently it’s been timing out and spamming our Slack channel with errors, because there were more markers in the bucket than it could deal with.

Previously it would list *all* the markers before doing any consolidation – so if there were millions of markers, it would hit the Lambda timeout just from calling the ListObject API.

This patch changes the behaviour – for every 1000 objects it gets from the ListObject API (a single page of results), it consolidates those into a single marker. It will likely still time out if there are millions of markers, but gradually it will reduce the number over consecutive runs, so the timeouts will eventually go away.

I also added a bit of logging for every call to the ListObjects API, so it’ll be easier to see if it’s hanging, crashing or just very busy the next time it has an error.

### DIFF
--- a/sierra_adapter/sierra_progress_reporter/README.md
+++ b/sierra_adapter/sierra_progress_reporter/README.md
@@ -1,0 +1,24 @@
+# sierra_progress_reporter
+
+We harvest records from Sierra every 15 minutes, grabbing the last 30 minutes of updates.
+When the Sierra reader is done, it writes a marker into S3 to say "I got these updates":
+
+```
+s3://wc-platform-adapters-sierra
+  └── windows_bibs_complete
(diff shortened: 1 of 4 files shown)


sierra_adapter/sierra_progress_reporter/README.md
```diff
@@ -0,0 +1,24 @@
+# sierra_progress_reporter
```
kenoir: 👏 📜 👏

PR #479, opened 2020-03-17
Inference in pipeline

jamieparkinson:
This implements the _Inference Manager_ described in https://github.com/wellcomecollection/docs/tree/master/rfcs/021-data_science_in_the_pipeline

Terraforming + the inferrer to come later.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
@@ -30,7 +30,32 @@ case class MergedImage[Id <: IdState](
       id = id,
       location = location
     )
+
+  def augment(inferredData: => InferredData): AugmentedImage[Id] =
+    AugmentedImage[Id](
+      id = id,
+      location = location,
+      parentWork = parentWork,
+      fullText = fullText,
+      inferredData = Some(inferredData)
+    )
 }
+
+case class AugmentedImage[Id <: IdState](
+  id: Id,
+  location: DigitalLocation,
+  parentWork: Id,
+  fullText: Option[String] = None,
+  inferredData: Option[InferredData] = None
+) extends BaseImage[Id]
+
+case class InferredData(
+  // We split the feature vector so that it can fit into
+  // ES's dense vector type (max length 2048)
+  features1: List[Float],
+  features2: List[Float],
+  lshEncodedFeatures: List[String]
+)
 
 object UnmergedImage {
   def apply(sourceIdentifier: SourceIdentifier,
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
@@ -31,13 +31,7 @@ case class MergedImage[Id <: IdState](
       location = location
     )
 
-  def augment(inferredData: => InferredData): AugmentedImage[Id] =
-    augmentWithOption(Some(inferredData))
-
-  def augmentWithNone: AugmentedImage[Id] = augmentWithOption(None)
-
-  private def augmentWithOption(
-    inferredData: => Option[InferredData]): AugmentedImage[Id] =
+  def augment(inferredData: => Option[InferredData]): AugmentedImage[Id] =
     AugmentedImage[Id](
       id = id,
       location = location,
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
@@ -32,12 +32,18 @@ case class MergedImage[Id <: IdState](
     )
 
   def augment(inferredData: => InferredData): AugmentedImage[Id] =
+    augmentWithOption(Some(inferredData))
+
+  def augmentWithNone: AugmentedImage[Id] = augmentWithOption(None)
+
+  private def augmentWithOption(
+    inferredData: => Option[InferredData]): AugmentedImage[Id] =
     AugmentedImage[Id](
       id = id,
       location = location,
       parentWork = parentWork,
       fullText = fullText,
-      inferredData = Some(inferredData)
+      inferredData = inferredData
     )
 }
 
--- a/pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferenceManagerWorkerService.scala
+++ b/pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferenceManagerWorkerService.scala
@@ -8,23 +8,17 @@ import akka.stream.scaladsl.Flow
 import com.amazonaws.services.sqs.model.Message
 import uk.ac.wellcome.bigmessaging.BigMessageSender
 import uk.ac.wellcome.bigmessaging.message.BigMessageStream
-import uk.ac.wellcome.models.work.internal.{
-  AugmentedImage,
-  InferredData,
-  MergedImage
-}
-import uk.ac.wellcome.platform.inference_manager.models.InferrerResponse
 import uk.ac.wellcome.typesafe.Runnable
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class InferenceManagerWorkerService[Destination, Id](
-  msgStream: BigMessageStream[MergedImage[Id]],
-  msgSender: BigMessageSender[Destination, AugmentedImage[Id]],
-  inferrerAdapter: InferrerAdapter[MergedImage[Id], InferrerResponse],
-  inferrerClientFlow: Flow[(HttpRequest, (Message, MergedImage[Id])),
-                           (Try[HttpResponse], (Message, MergedImage[Id])),
+class InferenceManagerWorkerService[Destination, Input, Output](
+  msgStream: BigMessageStream[Input],
+  msgSender: BigMessageSender[Destination, Output],
+  inferrerAdapter: InferrerAdapter[Input, Output],
+  inferrerClientFlow: Flow[(HttpRequest, (Message, Input)),
+                           (Try[HttpResponse], (Message, Input)),
                            HostConnectionPool]
 )(implicit ec: ExecutionContext, materializer: Materializer)
     extends Runnable {
@@ -38,47 +32,39 @@ class InferenceManagerWorkerService[Destination, Id](
       _.via(createRequest)
         .via(inferrerClientFlow)
         .via(unmarshalResponse)
-        .via(augmentImage)
-        .via(sendAugmentedImage)
+        .via(augmentInput)
+        .via(sendAugmented)
         .map { case (msg, _) => msg }
     )
 
   private def createRequest =
-    Flow[(Message, MergedImage[Id])].map {
-      case (msg, image) =>
-        (inferrerAdapter.createRequest(image), (msg, image))
+    Flow[(Message, Input)].map {
+      case (msg, input) =>
+        (inferrerAdapter.createRequest(input), (msg, input))
     }
 
   private def unmarshalResponse =
-    Flow[(Try[HttpResponse], (Message, MergedImage[Id]))]
+    Flow[(Try[HttpResponse], (Message, Input))]
       .mapAsync(parallelism) {
-        case (tryResponse, (msg, image)) =>
+        case (tryResponse, (msg, input)) =>
           tryResponse match {
             case Success(response) =>
               inferrerAdapter
                 .parseResponse(response)
-                .map((msg, image, _))
+                .map((msg, input, _))
             case Failure(exception) =>
               Future.failed(exception)
           }
       }
 
-  private def augmentImage =
-    Flow[(Message, MergedImage[Id], InferrerResponse)].map {
-      case (msg, image, InferrerResponse(features, lsh_encoded_features)) =>
-        val augmentedImage = image.augment {
-          val (features1, features2) = features.splitAt(features.size / 2)
-          InferredData(
-            features1 = features1,
-            features2 = features2,
-            lshEncodedFeatures = lsh_encoded_features
-          )
-        }
-        (msg, augmentedImage)
+  private def augmentInput =
+    Flow[(Message, Input, inferrerAdapter.InferrerResponse)].map {
+      case (msg, input, response) =>
+        (msg, inferrerAdapter.augmentInput(input, response))
     }
 
-  private def sendAugmentedImage =
-    Flow[(Message, AugmentedImage[Id])].mapAsync(parallelism) {
+  private def sendAugmented =
+    Flow[(Message, Output)].mapAsync(parallelism) {
       case (msg, image) =>
         Future.fromTry { msgSender.sendT(image) }.map((msg, _))
     }
--- a/pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferenceManagerWorkerService.scala
+++ b/pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferenceManagerWorkerService.scala
@@ -58,7 +58,7 @@ class InferenceManagerWorkerService[Destination, Input, Output](
       }
 
   private def augmentInput =
-    Flow[(Message, Input, inferrerAdapter.InferrerResponse)].map {
+    Flow[(Message, Input, Option[inferrerAdapter.InferrerResponse])].map {
       case (msg, input, response) =>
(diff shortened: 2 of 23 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -30,8 +30,39 @@ case class MergedImage[Id <: IdState](
...
+  location: DigitalLocation,
+  parentWork: Id,
+  fullText: Option[String] = None,
+  inferredData: Option[InferredData] = None
+) extends BaseImage[Id]
+
+case class InferredData(
+  // We split the feature vector so that it can fit into
+  // ES's dense vector type (max length 2048)
+  features1: List[Float],
```
jamesgorrie: Would it be worth making this a 2048 sized list type?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -30,8 +30,39 @@ case class MergedImage[Id <: IdState](
       id = id,
       location = location
     )
+
+  def augment(inferredData: => InferredData): AugmentedImage[Id] =
+    augmentWithOption(Some(inferredData))
+
+  def augmentWithNone: AugmentedImage[Id] = augmentWithOption(None)
```
jamesgorrie: Seems like not a massively convenient convenience method?

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
```diff
@@ -0,0 +1,90 @@
...
+// The InferrerAdaptor for feature vectors, consuming MergedImages and
+// augmenting them into AugmentedImages
+object FeatureVectorInferrerAdapter
+    extends InferrerAdapter[MergedImage[Minted], AugmentedImage[Minted]] {
+  type InferrerResponse = FeatureVectorInferrerResponse
+
+  def createRequest(image: MergedImage[Minted]): HttpRequest =
+    HttpRequest(
+      method = HttpMethods.GET,
+      uri = Uri("/feature-vectors/").withQuery(
```
jamesgorrie: I'm not sure if this not implemented fully, or I'm not getting it, but where does this API live?

jamesgorrie:
Have I got this right 
DSC_1011

Similar to the RFC, but just the Inferrer in that document is now a combo of the `InferrerAdapter` and the service on the other side of that?

Saying that, you've abstracted the `InferrerAdapter` and implemented it with `FeatureVectorInferrerAdapter`. What is that API?

As this has been abstracted I assume we are going to implement it in different places? Do you see there being multiple inferrer services, or one that runs multiple InferrerAdapters?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
warrd: Maybe a `Vector[Float]`?

jamieparkinson: I can't find a sized list type, where does it exist?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamieparkinson: It is useful (when we fail to infer data) but do you have a suggestion of a better name?

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
jamieparkinson: Have a look in here for the design https://github.com/wellcomecollection/docs/tree/master/rfcs/021-data_science_in_the_pipeline - it doesn't exist in this repo yet

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
warrd: There isn't size list afaik (thats an `Array`), but from your comment it is "max size" rather than "size"?

jamieparkinson: The ES limitation is a max but this data is a definite size. I can change it to a `Vector` but I'm not sure what advantages we gain from this other than a more fitting name...

jamieparkinson:
I think that's right although `InferrerAdapter` is, to me, an implementation detail of the inference manager, as the boundary is defined by the 2 separate tasks. The manager and the inferrer have to be coupled, and that coupling lives in the implementation of the `InferrerAdapter`.

> Do you see there being multiple inferrer services, or one that runs multiple InferrerAdapters?

That is the question, which was covered in the RFC questions/problems section. As such the RFC left open some choices that weren't clear, namely whether we should have:
- 1 manager-inferrer pair per inferrer (n pairs)
- 1 manager to n inferrers
- 1 manager to 1 super-inferrer that performs n inferences

The abstractions I chose to make here minimise the changes required between those options. At the moment, with 1 inferrer, all options are equivalent.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
warrd: List docs:

> This class is optimal for last-in-first-out (LIFO), stack-like access patterns. If you need another access pattern, for example, random access or FIFO, consider using a collection more suited to this than List.

Vector docs:

> Vector is a general-purpose, immutable data structure. It provides random access and updates in effectively constant time, as well as very fast append and prepend. Because vectors strike a good balance between fast random selections and fast random functional updates, they are currently the default implementation of immutable indexed sequences.

jamieparkinson: Exactly - we never access these List elements except for mapping and slicing

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferenceManagerWorkerService.scala
```diff
@@ -0,0 +1,72 @@
...
+    Flow[(Message, Input)].map {
+      case (msg, input) =>
+        (inferrerAdapter.createRequest(input), (msg, input))
+    }
+
+  private def unmarshalResponse =
+    Flow[(Try[HttpResponse], (Message, Input))]
+      .mapAsync(parallelism) {
+        case (tryResponse, (msg, input)) =>
+          tryResponse match {
```
kenoir: I find nested cases hard to read, can we extract this?

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
```diff
@@ -0,0 +1,90 @@
...
+  AugmentedImage,
+  InferredData,
+  MergedImage,
+  Minted
+}
+import uk.ac.wellcome.platform.inference_manager.models.FeatureVectorInferrerResponse
+
+import scala.concurrent.Future
+
+/*
```
kenoir: Nice use of comments.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
warrd: Ah ok, I get you. Might still use vector anyway because of the name, and the fact that it shows intent of it being a indexable data structure but :man_shrugging:

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamesgorrie: Not sure why just having `augment(data: Option[Data])` and `augment(None)` isn't as clear as what you have here?

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
kenoir: So the plan is get the inference manager into the pipeline now and passthrough messages (as nothing will be interesting to it yet)?

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
```diff
@@ -0,0 +1,90 @@
...
+ * An InferrerAdapter is specific to the inferrer and the data that is being augmented.
+ * Implementors must provide:
```
kenoir: Is a model where the inferrer request/response is standardised feasible (and desirable)?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamesgorrie: Yea, you can fix the size of an array, but this is a max size.
Would it just be a matter of having something like:
```
case class DenseVector(features: List[Float]) {
  def apply(features: List[Float]) = {
    if (features.length) { /* blow up */ }
  }
}
```

Just so that we can error on that instead of erroring on ingest

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
jamieparkinson: I hadn't actually thought of that! Maybe, but probably not - I haven't done the infra for it yet nor have we integrated the inferrer (although that is done in the data-science repo now). We'll see what order the work gets done in, I guess.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamesgorrie: You could probably have that case class have a list of lists that never exceed 2048 negating the need for the `features1` and `features2` and just splits accordingly. The index can also allow for a dynamic set of the features, but not sure if that will make it hard to use.

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
jamesgorrie: Ah, so it's going to be an API that does magic? Cool.

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
```diff
@@ -0,0 +1,90 @@
...
+
+trait InferrerAdapter[Input, Output] extends Logging {
+  type InferrerResponse
+
+  implicit val responseDecoder: Decoder[InferrerResponse]
+  def createRequest(input: Input): HttpRequest
+  def augmentInput(input: Input,
+                   inferrerResponse: Option[InferrerResponse]): Output
+
+  def parseResponse(response: HttpResponse)(
```
kenoir: What will happen in a time-out situation?

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
jamieparkinson: We did discuss some kind of standardised response in which the inferrer also instructs the manager where to put the data in the work/image, or that the data always goes in a fixed place, but we decided that the inferrer shouldn't know what works/images look like.

fwiw that's also why the manager does the work of manipulating the data from the inferrer (in this case, splitting the feature vector), as we also didn't want the inferrer to know about (for example) ElasticSearch's dense vector limitations.

Another thing is that it would also be nice to have other, and potentially open, deployments of the inferrer somewhere, in which case a descriptive response format is probably preferable.

jamieparkinson: Basically I agree it could be nice but marginally fall on the side of thinking it isn't desirable, up for debate though

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
kenoir: Will we end up retrying?

pipeline/inferrer/inference_manager/src/test/scala/uk/ac/wellcome/platform/inference_manager/services/InferenceManagerWorkerServiceTest.scala
```diff
@@ -0,0 +1,128 @@
...
+        sendMessage(queue, image404)
+        sendMessage(queue, image400)
+        eventually {
+          assertQueueEmpty(queue)
+          assertQueueHasSize(dlq, 2)
+        }
+    }
+  }
+
+  it("allows images that fail inference nondeterministically to pass through") {
```
kenoir: 👍

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
jamieparkinson: Akka-HTTP makes this really confusing - I _think_ that it will retry with exponential backoff [if it can't connect](https://doc.akka.io/docs/akka-http/current/client-side/host-level.html#retrying-a-request), and I think requests... don't timeout (!!). See eg this issue https://github.com/akka/akka-http/issues/42.

I was sufficiently confused by this that I left it as is, do you have any thoughts?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamieparkinson: I don't think there would ever be a dynamic set of feature vectors; they're being stored in the index for the convenience of the model builder (rather than as something to query).

I'll add a check on size in `augmentInput`, but whilst we can't express the fixed size in the type system, I'd rather not go full defensive programming on it. We trust the inferrer, it's coupled not only to the inference manager but to the `InferredData` itself, and it [does contain](https://github.com/wellcomecollection/data-science/blob/master/apis/feature_extraction/inference/app/src/feature_extraction.py) the guarantee on the size of this list.

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
kenoir: Interested in a high level opinion - not an issue blocking this PR.

kenoir: > Basically I agree it could be nice but marginally fall on the side of thinking it isn't desirable, up for debate though

👍 Probably trying for a standard request/response now would be over-engineering the solution - as we add more inferrers it'll focus the problem.

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
kenoir: Basically I think if it can't get a response for whatever reason this future will fail which will result in a received message being retried until it ends up on the DLQ. I'm fine with that.

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferrerAdapter.scala
kenoir: We should try to organise a situation where we can deploy this service as soon as possible into a production environment even if it does nothing so we can kind of soak test it, and of course it makes it will make it easier to add the actual functionality later on.

PR #485, opened 2020-03-21
Update the reindexer to terraform 0.12

alexwlchan:
I've rearranged the remote state so this should be good to plan/apply; I haven't actually done it because the reindexer seems to be running over the weekend and I didn't want to interrupt it.

Look at how much less code we have to look after now! 💃 

- [x] Merge and release https://github.com/wellcomecollection/terraform-aws-ecs-service/pull/8
- [x] Refer to the released version in this patch


reindexer/terraform/reindex_worker/iam_policy_document.tf
```diff
@@ -40,27 +36,27 @@ data "aws_iam_policy_document" "sns_publish_policy" {
...
 }
 
 # This block of interpolation syntax gets a list of all the topic ARNs that the
 # reindexer is configured to be able to publish to.
 #
 
 data "template_file" "topic_arn" {
-  count    = "${length(var.reindexer_jobs)}"
-  template = "${lookup(var.reindexer_jobs[count.index], "topic")}"
+  count    = length(var.reindexer_jobs)
```
kenoir: ✨

PR #487, opened 2020-03-23
Fix inference manager Makefile magic

jamieparkinson:
I swear I'll get this right at some point...

### DIFF
--- a/pipeline/inferrer/Makefile
+++ b/pipeline/inferrer/Makefile
@@ -1,11 +1,11 @@
 ROOT = $(shell git rev-parse --show-toplevel)
 include $(ROOT)/makefiles/functions.Makefile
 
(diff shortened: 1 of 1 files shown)


pipeline/Makefile
```diff
@@ -18,4 +18,5 @@ TF_PATH = $(STACK_ROOT)/terraform
 
 $(val $(call stack_setup))
 
-include $(ROOT)/$(STACK_ROOT)/transformer/Makefile
+include $(ROOT)/pipeline/transformer/Makefile
```
jamieparkinson: fyi; necessary to hardcode what was previously `$(STACK_ROOT)` here because the `include` overwrites it. There are better ways to do this (eg with secondary expansion) but I think that's out of scope.

PR #489, opened 2020-03-23
Collection tree in API

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4378

## Description

Displays the collection tree when requested under the `collectionTree` field in the `works/{id}` endpoint.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
@@ -15,7 +15,7 @@ sealed trait QueryParams
 
 case class SingleWorkParams(
   include: Option[V2WorksIncludes],
-  expandPaths: Option[List[String]],
+  _expandPaths: Option[List[String]],
   _index: Option[String],
 ) extends QueryParams
 
@@ -30,7 +30,7 @@ object SingleWorkParams extends QueryParamsUtils {
     parameter(
       (
         "include".as[V2WorksIncludes].?,
-        "expandPaths".as[List[String]].?,
+        "_expandPaths".as[List[String]].?,
         "_index".as[String].?
       )
     ).tmap((SingleWorkParams.apply _).tupled(_))
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
@@ -15,6 +15,7 @@ sealed trait QueryParams
 
 case class SingleWorkParams(
   include: Option[V2WorksIncludes],
+  expandPaths: Option[List[String]],
   _index: Option[String],
 ) extends QueryParams
 
@@ -29,9 +30,13 @@ object SingleWorkParams extends QueryParamsUtils {
     parameter(
       (
         "include".as[V2WorksIncludes].?,
+        "expandPaths".as[List[String]].?,
         "_index".as[String].?
       )
     ).tmap((SingleWorkParams.apply _).tupled(_))
+
+  implicit val decodePaths: Decoder[List[String]] =
+    decodeCommaSeparated
 }
 
 case class MultipleWorksParams(
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
@@ -247,8 +247,8 @@ class Router(elasticClient: ElasticClient,
     work.data.collection
       .map {
         case Collection(_, path) =>
-          val paths = path :: expandedPaths
-          collectionService.retrieveTree(index, paths).map {
+          val allPaths = path :: expandedPaths
+          collectionService.retrieveTree(index, allPaths).map {
             case Left(err) =>
               // We just log this here rather than raising so as not to bring down
               // the work API when tree retrieval fails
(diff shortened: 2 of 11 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -29,9 +30,13 @@ object SingleWorkParams extends QueryParamsUtils {
     parameter(
       (
         "include".as[V2WorksIncludes].?,
+        "expandPaths".as[List[String]].?,
```
jamesgorrie: Not sure this parameter name makes very much sense to me?

Why not just have it in the includes?

warrd: It's different to includes:

* `includes=collection` means we display the tree, with the current work expanded in the tree

* `expandPaths=a/b/c,a/x/y` means we also want to expand `a/b/c` and `a/x/y` in the returned tree

alexwlchan:
Reviewing now.

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
```diff
@@ -29,9 +30,13 @@ object SingleWorkParams extends QueryParamsUtils {
...
         "_index".as[String].?
       )
     ).tmap((SingleWorkParams.apply _).tupled(_))
+
+  implicit val decodePaths: Decoder[List[String]] =
+    decodeCommaSeparated
```
alexwlchan: Nice, I didn't know this was a thing!

api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -108,20 +114,32 @@ class Router(elasticClient: ElasticClient,
...
-          case Right(Some(work: IdentifiedRedirectedWork)) => workRedirect(work)
-          case Right(Some(_))                              => workGone
-          case Right(None)                                 => workNotFound(id)
-          case Left(err)                                   => elasticError(err)
+        .flatMap {
+          case Right(Some(work: IdentifiedWork)) =>
+            val expandedPaths = params.expandPaths.getOrElse(Nil)
+            retrieveTree(index, work, expandedPaths).map {
+              workFound(work, _, includes)
+            }
```
alexwlchan: Would slightly prefer, for clarity:

```suggestion
            retrieveTree(index, work, expandedPaths).map { tree =>
              workFound(work, tree, includes)
            }
```

Also, can we flatMap here to avoid having to Future up all the other branches?

api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -222,6 +240,30 @@ class Router(elasticClient: ElasticClient,
...
     }
 
+  def retrieveTree(index: Index,
+                   work: IdentifiedWork,
+                   expandedPaths: List[String])
+    : Future[Option[(CollectionTree, List[String])]] =
+    work.data.collection
+      .map {
+        case Collection(_, path) =>
+          val paths = path :: expandedPaths
```
alexwlchan: Call this `allPaths` to increase the difference from `path`? I like to keep variable names different.

api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -222,6 +240,30 @@ class Router(elasticClient: ElasticClient,
...
+          collectionService.retrieveTree(index, paths).map {
+            case Left(err) =>
+              // We just log this here rather than raising so as not to bring down
+              // the work API when tree retrieval fails
```
alexwlchan: 👍

api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -246,5 +288,10 @@ class Router(elasticClient: ElasticClient,
   lazy val worksService =
     new WorksService(new ElasticsearchService(elasticClient))
 
+  lazy val collectionService =
+    new CollectionService(elasticClient)
+
+  lazy val logger = Logger(this.getClass.getName)
```
alexwlchan: The way we normally add logging is to extend from `grizzled.sl4j.Logging`; is there a reason you didn't do that here?

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/CollectionRequestBuilder.scala
```diff
@@ -12,21 +12,30 @@ case class CollectionRequestBuilder(index: Index,
...
 
-  val excludeFields = List(
-    "data.items",
-    "data.notes",
-    "data.production",
-    "data.subjects",
-    "data.genres",
-    "data.images",
+  // To reduce response size and improve Elasticsearch performance we only
+  // return core fields for works in the tree.
```
alexwlchan: 👍 

smart

api/api/src/main/scala/uk/ac/wellcome/platform/api/QueryParams.scala
warrd: It's defined below, not library functionality

api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
warrd: `error` is defined already is this class

PR #491, opened 2020-03-23
Split out the "update API docs" terraform; bump it to terraform 0.12

alexwlchan:
A first step towards https://github.com/wellcomecollection/platform/issues/4250 and https://github.com/wellcomecollection/platform/issues/4386

This now becomes a standalone stack that we can plan/apply/destroy independently of the live catalogue API.

Getting the rest of the catalogue API to terraform 0.12 is an enormous task, and trying to do it in one go is proving to be more than I can manage.


api/terraform/update_api_docs/iam.tf
```diff
@@ -0,0 +1,4 @@
+module "ecs_update_api_docs_iam" {
+  source    = "github.com/wellcomecollection/terraform-aws-ecs-service.git//task_definition/modules/iam_role?ref=v1.5.0"
```
jamesgorrie: Bikeshed - why do we have more than just the singe modules in this repo. It feels like we're using something that is _piece_ of that module, rather than just implementing the module, creating a potential coupling situation like before?

alexwlchan: Not a bikeshed! The answer is that actually we shouldn't and I should remove this; I ended up consolidating it into the `single_container` module and this bit is redundant.

kenoir: Agree with @alexwlchan, this is a good point not a bikeshed @jamesgorrie.

PR #492, opened 2020-03-24
Split the data_api into its own terraform stack

alexwlchan:
This means we can make changes to the snapshot service with less risk of affecting the production catalogue API. (In particular, in a later pull request we can upgrade it to terraform 0.12!)

Follows https://github.com/wellcomecollection/catalogue/pull/491; another step towards https://github.com/wellcomecollection/platform/issues/4386

The eventual plan is to have the following folder structure in the `terraform` directory:

- catalogue_api_shared (API Gateway, ECR repositories, ECS cluster, etc.)
- catalogue_api_prod (ECS services for the prod API, alarms)
- catalogue_api_staging (ECS services for the staging API)
- data_api (snapshots service)
- update_api_docs
- modules (for shared pieces)

And then each of those can be rolled to terraform 0.12 independently (and in particular, the first time I hit apply I won't be affecting the production API!).

### DIFF
--- a/api/terraform/data_api/terraform.tf
+++ b/api/terraform/data_api/terraform.tf
@@ -11,23 +11,11 @@ terraform {
   }
 }
 
-data "terraform_remote_state" "catalogue_pipeline_data" {
-  backend = "s3"
-
-  config {
-    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
-
-    bucket = "wellcomecollection-platform-infra"
-    key    = "terraform/catalogue_pipeline_data.tfstate"
-    region = "eu-west-1"
-  }
-}
-
 data "terraform_remote_state" "shared_infra" {
   backend = "s3"
 
   config {
-    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
+    role_arn = "arn:aws:iam::760097843905:role/platform-read_only"
 
     bucket = "wellcomecollection-platform-infra"
     key    = "terraform/platform-infrastructure/shared.tfstate"
--- a/api/terraform/data_api/terraform.tf
+++ b/api/terraform/data_api/terraform.tf
@@ -2,10 +2,10 @@ terraform {
   required_version = ">= 0.9"
 
   backend "s3" {
-    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
+    role_arn = "arn:aws:iam::756629837203:role/catalogue-developer"
 
-    bucket         = "wellcomecollection-platform-infra"
-    key            = "terraform/data_api.tfstate"
+    bucket         = "wellcomecollection-catalogue-infra-delta"
+    key            = "terraform/catalogue/api/data_api.tfstate"
     dynamodb_table = "terraform-locktable"
     region         = "eu-west-1"
   }
--- a/api/terraform/data_api/terraform.tf
+++ b/api/terraform/data_api/terraform.tf
@@ -1,6 +1,4 @@
 terraform {
-  required_version = ">= 0.9"
-
   backend "s3" {
     role_arn = "arn:aws:iam::756629837203:role/catalogue-developer"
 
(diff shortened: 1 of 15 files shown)


jamesgorrie:
Just questions from the description rather than the code:

1. Why does API gateway have to be shared? Each env has it's own gateway?
1. ECR repos, I think, are all in platform. Do you have a separate ECR registry in storage? And should we? It might complicate deployment tooling unnecessarily.
1. Why no alarms on staging? Rather get them there before we prodify. Maybe round two when looking at monitoring of services, where we should monitor both.

Otherwise this feels like it'll give us way more confidence!

api/terraform/data_api/terraform.tf
```diff
@@ -2,10 +2,10 @@ terraform {
   required_version = ">= 0.9"
 
   backend "s3" {
-    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
+    role_arn = "arn:aws:iam::756629837203:role/catalogue-developer"
 
-    bucket         = "wellcomecollection-platform-infra"
-    key            = "terraform/data_api.tfstate"
+    bucket         = "wellcomecollection-catalogue-infra-delta"
+    key            = "terraform/catalogue/api/data_api.tfstate"
```
jamesgorrie: Just something I am picking up, we are categorising via folder structure, and I am unsure how, as we do it differently elsewhere.

is it `terraform/{repo}/{project}`? Is so should it not just be `terraform/catalogue/data_api`?

api/terraform/data_api/variables.tf
```diff
@@ -2,24 +2,14 @@ variable "aws_region" {
...
-variable "snapshot_generator_release_uri" {}
-
-variable "critical_slack_webhook" {}
-
-variable "vpc_id" {}
-
-variable "private_subnets" {
-  type = "list"
+variable "critical_slack_webhook" {
+  default = ""
```
jamesgorrie: Critical with empty value?

alexwlchan:
1. Currently the two environments share an API Gateway. I’m not opposed to changing that, but I’m trying to change as little other stuff as possible right now.

2. We have a separate ECR registry for the storage account, yep – because everything in the storage repo publishes into that account. This repo is kinda a weird hybrid of platform/catalogue.

    (But doesn’t really affect this – we have one ECR repo for API images, prod and staging alike.)

3. That’s the way it’s always been; again, open to changing but not right now.

api/terraform/data_api/variables.tf
alexwlchan: Yeah cos the alarm isn’t currently hooked up. It’s how it was before I started. 🤷‍♀️

api/terraform/data_api/terraform.tf
alexwlchan: `terraform/{repo}/{project}` is what I’m aiming for, yeah

It’s not super-duper consistent; the aim is that somebody could look at a terraform remote state key and work out where it’s defined, rather than being completely machine-readable. (In the past it’s become tricky to know what states are actively maintained/in a renamed folder/stale.)

jamesgorrie: I'm thinking for humans when we create a new one and look how it's been done before and there's three ways, then I need to work out which one to follow.

How do you feel about keeping it we've specified above?

alexwlchan: Ah, sorry, I see the distinction now.

`terraform/{repo}/{project}` is *not* the idea in my head, it’s `terraform/{repo}/{path_within_repo}`.

I don’t feel strongly about it either way, but will note that we already have `terraform/catalogue/api/update_api_docs.tfstate` in the stack from the previous PR.

kenoir: I like the hierarchy @alexwlchan has introduced - it namespaces stack to project which is useful and necessary as we split up stacks.

PR #494, opened 2020-03-24
Update the data API to terraform 0.12

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4250

I inlined a few Lambda-related modules and increased the memory in the snapshot generator; previously 1GB, now 4GB – if you look at the logs, it keeps dropping out-of-memory exceptions.

### DIFF
--- a/api/terraform/data_api/domain.tf
+++ b/api/terraform/data_api/domain.tf
@@ -1,7 +1,11 @@
-data "aws_route53_zone" "dotorg" {
+/*data "aws_route53_zone" "dotorg" {
   provider = "aws.routemaster"
 
   name = "wellcomecollection.org."
+}*/
+
+locals {
+  route53_zone_id = "Z3THRVQ5VDYDMC"
 }
 
 resource "aws_acm_certificate" "data_page" {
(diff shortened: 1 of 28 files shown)


api/terraform/data_api/domain.tf
```diff
@@ -1,12 +1,17 @@
-data "aws_route53_zone" "dotorg" {
+/*data "aws_route53_zone" "dotorg" {
   provider = "aws.routemaster"
 
   name = "wellcomecollection.org."
+}*/
+
+locals {
+  route53_zone_id = "Z3THRVQ5VDYDMC"
 }
```
alexwlchan: This seems to have broken with the newer terraform version/provider:

> Error: Error finding Route 53 Hosted Zone: AccessDenied: User: arn:aws:sts::250790015188:assumed-role/wellcomecollection-assume_role_hosted_zone_update/1585053181121301000 is not authorized to perform: route53:ListTagsForResource on resource: arn:aws:route53:::hostedzone/Z3THRVQ5VDYDMC

😿

kenoir: It's not on the permissions for that role as of now - just looked.

PR #495, opened 2020-03-24
Upgrade the shared API stack to terraform 0.12

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4250


api/terraform/modules/interface_endpoints/main.tf
```diff
@@ -0,0 +1,52 @@
+# The full list of services that support Interface Endpoints in the docs:
```
kenoir: We should move these to the platform-infrastructure stack as part of provisioning a VPC (and extend that to all other projects).

alexwlchan: Digging into it, we do have this module in platform-infrastructure, we just aren't using it anywhere.

(I think it's slightly complicated by the fact that you need to be in the same security group.)

Will open a ticket for it.

PR #497, opened 2020-03-24
Bump the catalogue API to terraform 0.12

alexwlchan:
This has been deployed successfully, and with zero downtime in the live API. 💃 

This is (AFAIK) the last bit of terraform 0.11 in the platform, so merging this will close https://github.com/wellcomecollection/platform/issues/4250 🎉

### DIFF
--- a/api/terraform/modules/catalogue_api_stack/domains.tf
+++ b/api/terraform/modules/catalogue_api_stack/domains.tf
@@ -12,8 +12,8 @@ locals {
 }
 
 resource "aws_api_gateway_domain_name" "catalogue_api" {
-  domain_name              = "${var.domain_name}"
-  regional_certificate_arn = "${var.certificate_arn}"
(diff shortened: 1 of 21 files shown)


api/terraform/modules/catalogue_api_stack/domains.tf
```diff
@@ -1,27 +1,35 @@
-data "aws_route53_zone" "dotorg" {
-  provider = "aws.routemaster"
+/*data "aws_route53_zone" "dotorg" {
```
jamieparkinson: What's going on here?

alexwlchan: We can’t (currently) look up the hosted zone ID programmatically in tf 0.12. I’ve got a patch out to give the role the right permissions, so I’m hoping this will be back to the uncommented Code extremely soon.

PR #499, opened 2020-03-25
Don't return separate collection and collectionTree fields in API

warrd:
Only return a single `collection` field, containing the tree.

Advantages:

* Potentially simpler to understand with a single field
* No duplication of data
* Slightly smaller response size

Drawbacks:

* The work will not be at the root of the tree, and will have to traverse the tree to find the collection data for it
* When building of a tree from elasticsearch fails, the client won't have access to the works collection data
* No data willl be available on the list endpoint currently


common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayCollection.scala
```diff
@@ -1,23 +1,57 @@
...
-import io.circe.generic.extras.JsonKey
 import io.swagger.v3.oas.annotations.media.Schema
-import uk.ac.wellcome.models.work.internal.Collection
+
+import uk.ac.wellcome.models.work.internal._
+import uk.ac.wellcome.display.models.v2.DisplayWorkV2
 
 @Schema(
   name = "Collection",
-  description = "A collection that a work is part of"
```
jamesgorrie: It's still the collection this work is part of.

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayCollection.scala
```diff
@@ -1,23 +1,57 @@
...
+      children =
+        if (isExpanded(tree.path, expandedPaths))
+          Some(tree.children.map(DisplayCollection(_, expandedPaths)))
+        else
+          None
     )
+
+  private def isExpanded(path: String, expandedPaths: List[String]): Boolean =
+    expandedPaths.exists(_.startsWith(path))
 }
```
warrd: Note that this whole file is just `DisplayCollectionTree` but with the name changed to `DisplayCollection`, but git hasnt picked up on that

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayCollection.scala
warrd: See this comment: https://github.com/wellcomecollection/catalogue/pull/499#discussion_r398643348

Happy for either description /  a new one if suggested.

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayCollection.scala
```diff
@@ -1,23 +1,57 @@
...
-  @Schema(description = "The label of the collection") label: Option[String],
-  @Schema(description = "Where in the hierarchy a work is in the collection") path: String,
-  @JsonKey("type") @Schema(name = "type") ontologyType: String = "Collection"
+  @Schema(
+    description =
+      "A slash separated path containing the position within the hierarchy."
+  ) path: String,
+  @Schema(
+    description =
+      "The level of the node. Either Collection, Section, Series, SubSeries or Item"
```
jamesgorrie: Maybe not worth listing these out as they're already out of date.

warrd: :disappointed:

jamesgorrie:
> The work will not be at the root of the tree, and will have to traverse the tree to find the collection data for it

We can always add it as `collectionPath` or similar.

warrd:
> We can always add it as collectionPath or similar.

But what about the `label` and the `level`? I'm starting to think that not including this data is not a great idea, especially as won't currently be able to access any of it on the work list endpoint

jamesgorrie:
> I'm starting to think that not including this data is not a great idea

If I were a client, I'd be really annoyed that you didn't. It's an easy, but inconvenient computation that _will_ have to be done (where am I in the collection?).

Having a think about how you might structure it.

warrd:
Not sure I understand.

The data for a single work currently included under `collection`. A tree containing data for multiple works is included under `collectionTree`.

This PR replaces those two fields with just the tree under `collection`, which is fine apart from the circumstances given in the description at the top. So one solution to this is to just close the PR and leave how it is, which is what I am leaning towards.

What do u reckon?

PR #500, opened 2020-03-25
Increase size of minter RDS instance

jamieparkinson:
When I actually apply this I intend to do it with `apply_immediately = true`, so we don't have to wait until the Monday maintenance window - this means there will be a bit of downtime, is that a problem?


jamieparkinson:
I am going to leave those comments for now as the reason I need this right now is just to make testing easier - I'll change the comments when I add the image minter, which will be as soon as this issue is fixed.

PR #502, opened 2020-03-25
Extracting an ingestor_common library

alicefuzier:


pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala
```diff
@@ -0,0 +1,81 @@
...
+        .doc(document)
+    }
+
+    client
+      .execute {
+        bulk(inserts)
+      }
+      .map { response: Response[BulkResponse] =>
+        if (response.isError) {
+          warn(s"Error from Elasticsearch: $response")
```
jamesgorrie: Why `warn` over `error`?

alicefuzier: Good question. This is copied over from `WorkIndexer` so it's how it currently works and I didn't even see it. Will change, thanks.

pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala
```diff
@@ -0,0 +1,81 @@
...
+trait Indexer[T] extends Logging {
+  val client: ElasticClient
+
+  implicit val ec: ExecutionContext
+  implicit val indexable: Indexable[T]
+  implicit val id: CanonicalId[T]
+  implicit val version: Version[T]
+  val indexName: Index
+
+  final def index(documents: Seq[T]): Future[Either[Seq[T], Seq[T]]] = {
```
jamesgorrie: You could use [`EitherT[Future, Seq[T], Seq[T]]`](https://typelevel.org/cats/datatypes/eithert.html)?

pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala
```diff
@@ -0,0 +1,82 @@
...
+      }
+      .map { response: Response[BulkResponse] =>
+        if (response.isError) {
+          error(s"Error from Elasticsearch: $response")
+          Left(documents)
+        } else {
+          debug(s"Bulk response = $response")
+          val bulkResponse = response.result
+          val actualFailures = bulkResponse.failures.filterNot {
+            isVersionConflictException
```
jamesgorrie: With this - is the idea that we might get a conflicting version from a source / VHS and if that's the case, we just don't care?

Could we log these to see how often it happens?

pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala
```diff
@@ -0,0 +1,82 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+trait Indexer[T] extends Logging {
+  val client: ElasticClient
+
+  implicit val ec: ExecutionContext
+  implicit val indexable: Indexable[T]
+  implicit val id: CanonicalId[T]
+  implicit val version: Version[T]
+  val indexName: Index
```
jamesgorrie: Not just `index`?

pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala
alicefuzier: This is the current behaviour: because SQS can send messages out of order the ingestor might receive version 2 before version 1. If that happens we don't want version 1 to override version 2. In that case, a version conflict exception is what you would get in elasticsearch. Since it's an expected scenario (although not frequent) we just log and move on. Logging is done [here]()https://github.com/wellcomecollection/catalogue/blob/709fdcf4d6967072455edfe1c619342928c3a2bf/pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala#L74

pipeline/ingestor/ingestor_common/src/test/scala/uk/ac/wellcome/platform/ingestor/common/IndexerTest.scala
```diff
@@ -0,0 +1,170 @@
...
+class IndexerTest
+    extends FunSpec
+    with ScalaFutures
+    with Matchers
+    with IdentifiersGenerators
+    with ElasticsearchFixtures
+    with IngestorFixtures
+    with WorksGenerators {
+
+  it("inserts an identified Work into Elasticsearch") {
```
jamesgorrie: Assuming the test names can change to be abstract of `Work`?

alicefuzier: ah thanks.. I changed the test, not the title

pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala
alicefuzier: NIce thanks! I didn't even know that it existed. I'm going to do it in the next PR though cause I don't want to wait for another round of Travis if there are no objections

PR #507, opened 2020-03-27
Add CollectionPath

warrd:


common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayCollectionPath.scala
```diff
@@ -0,0 +1,44 @@
+package uk.ac.wellcome.display.models.v2
```
warrd: Note that this file is renamed from `DisplayCollection.scala` but github hasnt picked up on that so all looks new,

PR #511, opened 2020-03-30
ID minter for images

jamieparkinson:


pipeline/terraform/stack/service_image_id_minter.tf
```diff
@@ -0,0 +1,79 @@
...
+    db_password = "catalogue/id_minter/rds_password"
+  }
+
+  // The maximum number of connections to RDS is 90.
+  // Each id minter task is configured to have 8 connections
+  // in the connection pool (see `max_connections` parameter above).
+  // To avoid exceeding the maximum nuber of connections to RDS,
+  // the maximum capacity for *EACH* id minter  (ie the total
+  // maximum capacity of all ID minters) should be no higher than 5
+  max_capacity = 5
```
alexwlchan: You've updated the number but not taken advantage of it! Possibly I wrote a less-than helpful comment.

If each ID minter maxes its connections, we'd be using `8 * id_minter = 8 * 5 < 45` count connections. You can play with these numbers if you want more performance/to take advantage of the extra connections.

jamieparkinson: I disagree! There are 2 ID minter services now, so you are missing a factor of 2 - however I see that my `*EACH*` isn't clear that I'm talking about the services. Will clarify.

alexwlchan: I wonder if having these numbers be computed on the fly might help?

```hcl
locals {
  id_minter_count       = 2
  id_minter_concurrency = 8
  db_max_connections    = 90
}

...

max_capacity = floor(local.db_max_connection / local.id_minter_count / local.id_minter_concurrency)
```

jamieparkinson: Nice idea, will do - what a shame that terraform can't do integer linear programming

pipeline/terraform/stack/service_image_id_minter.tf
```diff
@@ -0,0 +1,78 @@
...
+    max_connections = local.id_minter_task_max_connections
+
+    logstash_host = local.logstash_host
+  }
+
+  secret_env_vars = {
+    cluster_url = "catalogue/id_minter/rds_host"
+    db_port     = "catalogue/id_minter/rds_port"
+    db_username = "catalogue/id_minter/rds_user"
+    db_password = "catalogue/id_minter/rds_password"
```
jamesgorrie: Do we want to create different users for different services?

pipeline/terraform/stack/service_work_id_minter.tf
```diff
@@ -40,40 +40,40 @@ module "id_minter" {
...
 
-  // The maximum number of connections to RDS is 45.
-  // Each id minter task is configured to have 8 connections
-  // in the connection pool (see `max_connections` parameterabove).
-  // To avoid exceeding the maximum nuber of connections to RDS,
-  // the maximum capacity for the id minter should be no higher than 5
-  max_capacity = 5
+  // The total number of connections to RDS across all tasks from all ID minter
+  // services must not exceed the maximum supported by the RDS instance.
+  max_capacity = floor(
```
jamesgorrie: 💎

pipeline/terraform/stack/service_image_id_minter.tf
jamieparkinson: Interesting point, I don't _think_ so as they're identical in every way apart from their location

PR #512, opened 2020-03-30
Add inferrer from data science repo

jamieparkinson:
This has been filter-branched but I think a squash had happened so the history is a bit sparse. It's just the contents of https://github.com/wellcomecollection/data-science/pull/14, the code has already been reviewed etc.


jamieparkinson:
Not just in the catalogue account - in the pipeline!

jamieparkinson:
I fully expect that changes will be required but they will come in another PR so as not to confuse matters

PR #513, opened 2020-03-31
add test to prove otherIdentifiers merger rule redirects calm work to Sierra

jamesgorrie:
Appends the Calm work to create as a redirect if merged.

---

## Previously

After a conversation with @warrd we were wondering if the calm merger would spit out a merged sierra work and redirected calm work.

The answer is yes, after digging through realising it the `FieldMergeRule`s that return a MergeState which is in the shape of `(Redirects, MergedWork)` from which we then [calculate the unmerged works from the remaining sources](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala#L60-L64).

I might have a go at seeing if I can make this a little clearer, or at least document the process from Transformer => Matcher => Merger => List of things to ingest

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/models/FieldMergeResult.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/models/FieldMergeResult.scala
@@ -4,9 +4,8 @@ import uk.ac.wellcome.models.work.internal.TransformedBaseWork
 
 /*
  * FieldMergeResult is the return type of a FieldMergeRule's `merge` method
- * and contains both the new (merged) value for the field but also a list
- * of sources which the rule would like the merger to redirect. It is up to
- * the merger how to handle these.
+ * and contains both the new (merged) `data` for the field but also a list
+ * of `sources` which values were used to compute the new value.
+ * It is up to the merger how to handle these.
  */
-case class FieldMergeResult[T](fieldData: T,
-                               redirects: Seq[TransformedBaseWork])
+case class FieldMergeResult[T](data: T, sources: Seq[TransformedBaseWork])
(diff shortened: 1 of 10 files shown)


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/CalmRules.scala
```diff
@@ -23,7 +23,7 @@ trait UseCalmWhenExistsRule extends FieldMergeRule with MergerLogging {
     sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData] =
     FieldMergeResult(
       fieldData = calmRule(target, sources) getOrElse getData(target),
-      redirects = Nil
+      mergedFrom = Nil
```
jamieparkinson: Given the rule about "any" data, I don't think this being `Nil` (here and in other rules) makes sense any more?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/models/FieldMergeResult.scala
```diff
@@ -9,4 +9,4 @@ import uk.ac.wellcome.models.work.internal.TransformedBaseWork
  * the merger how to handle these.
  */
 case class FieldMergeResult[T](fieldData: T,
-                               redirects: Seq[TransformedBaseWork])
+                               mergedFrom: Seq[TransformedBaseWork])
```
jamieparkinson: I'm not sure about `mergedFrom` as a name, I don't think it makes the most sense when it's used outside of these rules. I think a plural of a noun is a less confusing way of naming this, and it might make sense to be explicit about the fact that these are a subset of the sources, eg:

- `mergedSources`
- `usedSources`
- `mergedFromSources`
- `dataSources`

jamesgorrie:
Closing as I think I have a deeper understanding of redirects

PR #514, opened 2020-03-31
Ingestor images

alicefuzier:
depends on #527


warrd:
Having a read through the code now.

Just an initial question about the linked PR on elastic4s. Is it clear to you what needs to be done so it's suitable for release? I ask because it's not that clear to me, and theres a danger we'll get in a situation where we don't get to update elastic4s / have to patch our change every time.

alicefuzier:
Actually good point, I'd forgotten about that PR. I have to make some changes that might affect the code here

alicefuzier:
Actually, I've replied to the comment in the elastic4s PR but it might take a while and I'd rather not leave this open for as long, so I'd rather merge this now. I'll have to remember to keep track of that PR and make sure it gets merged soon

warrd:
Yup wasn't saying we should hold off on merging here, just wanted to make sure there is a clear route to getting that change upstream

alicefuzier:
Yeh.. not sure we can call it a clear route but I will keep bugging until something we can use gets released

PR #515, opened 2020-04-02
Add sources used to merge a field to FieldMergeResult

jamesgorrie:
Given `WorkA` and `WorkB`
If work `B` is merged in anyway into `WorkA`, `WorkB` should redirect to `WorkA`.

This does that by running checking which `PartialRule`s were applicable to the `FieldMergeRule`, and returning those sources.

Those then get returned to the `Merger` where they are turned into redirects.

There are probably more abstracted ways to ensure that this is enforced, but it feels like we would be creating code too rigid and potentially coupling concepts that don't need to be.

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
@@ -34,13 +34,23 @@ trait FieldMergeRule {
   def apply(target: UnidentifiedWork,
             sources: Seq[TransformedBaseWork]): MergeState =
     merge(target, sources) match {
-      case FieldMergeResult(field, ruleRedirects) =>
-        State(existingRedirects =>
-          (existingRedirects ++ ruleRedirects.toSet, field))
+      case FieldMergeResult(field, mergedSources) =>
+        State(existingMergedSources =>
+          (existingMergedSources ++ mergedSources.toSet, field))
     }
 
   def merge(target: UnidentifiedWork,
             sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData]
+
+  protected def getMergedSources(
+    rules: List[PartialRule],
+    target: UnidentifiedWork,
+    sources: Seq[TransformedBaseWork]): Seq[TransformedBaseWork] = {
+
+    sources.filter { source =>
+      rules.exists(_(target, source).isDefined)
+    }
+  }
 
   protected trait PartialRule {
     val isDefinedForTarget: WorkPredicate
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
@@ -42,7 +42,7 @@ trait FieldMergeRule {
   def merge(target: UnidentifiedWork,
             sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData]
 
-  protected def getMergedSources(
+  protected def getSourcesToMerge(
     rules: List[PartialRule],
     target: UnidentifiedWork,
     sources: Seq[TransformedBaseWork]): Seq[TransformedBaseWork] = {
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
@@ -42,16 +42,6 @@ trait FieldMergeRule {
   def merge(target: UnidentifiedWork,
             sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData]
 
-  protected def getSourcesToMerge(
-    rules: List[PartialRule],
-    target: UnidentifiedWork,
-    sources: Seq[TransformedBaseWork]): Seq[TransformedBaseWork] = {
-
-    sources.filter { source =>
-      rules.exists(_(target, source).isDefined)
-    }
-  }
-
   protected trait PartialRule {
     val isDefinedForTarget: WorkPredicate
     val isDefinedForSource: WorkPredicate
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
@@ -25,8 +25,8 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
     target: UnidentifiedWork,
     sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData] =
     FieldMergeResult(
-      fieldData = mergeItems(target, sources),
-      redirects = sources.filter { source =>
+      data = mergeItems(target, sources),
+      sources = sources.filter { source =>
         (mergeMetsItems(target, source) orElse mergeMiroPhysicalAndDigitalItems(
           target,
           source)).isDefined
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/OtherIdentifiersRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/OtherIdentifiersRule.scala
@@ -21,23 +21,19 @@ object OtherIdentifiersRule extends FieldMergeRule with MergerLogging {
   override def merge(
     target: UnidentifiedWork,
     sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData] = {
-    val miroIds =
-      miroIdsRule(target, sources) getOrElse Nil
-    val physicalDigitalIds =
-      physicalDigitalIdsRule(target, sources) getOrElse Nil
-    val calmIds =
-      calmIdsRule(target, sources) getOrElse Nil
+
+    val rules = List(miroIdsRule, physicalDigitalIdsRule, calmIdsRule)
+    val ids = rules.flatMap(_(target, sources)).flatten
+    val mergeSources = sources.filter { source =>
+      rules.exists(_(target, source).isDefined)
+    }
 
     FieldMergeResult(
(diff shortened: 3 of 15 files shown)


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -34,14 +34,24 @@ trait FieldMergeRule {
...
-          (existingRedirects ++ ruleRedirects.toSet, field))
+      case FieldMergeResult(field, mergedSources) =>
+        State(existingMergedSources =>
+          (existingMergedSources ++ mergedSources.toSet, field))
     }
 
   def merge(target: UnidentifiedWork,
             sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData]
 
+  protected def getSourcesToMerge(
```
jamieparkinson, line 45: I find this method pretty unclear - both to read and, more importantly, to understand what it's doing when it's called - it doesn't really explain what it's doing. I think it might be clearer to separate the logic a bit so it's explicit what `sources` is (a subset of all the sources). Maybe something like...

```scala
val result = FieldMergeResult(
  data = ...,
  sources = sources.filter(anyRuleDefinedForSource(
    target = target,
    partialRules = Seq(fooRule, barRule, bazRule)
  ))
)

def anyRuleDefinedForSource(target, partialRules: Seq[PartialRule])(
  source: TransformedBaseWork): Boolean 
  partialRules.exists { rule =>
    rule(target, source).isDefined
  }
```

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -25,8 +25,8 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
     target: UnidentifiedWork,
     sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData] =
     FieldMergeResult(
-      fieldData = mergeItems(target, sources),
-      redirects = sources.filter { source =>
+      data = mergeItems(target, sources),
+      sources = sources.filter { source =>
```
jamieparkinson, line 29: I think `getSourcesToMerge` or equivalent should be used consistently across all the rules - not sure if this is the only place where it's not yet used

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/OtherIdentifiersRule.scala
```diff
@@ -15,28 +15,24 @@ import uk.ac.wellcome.platform.merger.rules.WorkPredicates.WorkPredicate
...
-    val miroIds =
-      miroIdsRule(target, sources) getOrElse Nil
-    val physicalDigitalIds =
-      physicalDigitalIdsRule(target, sources) getOrElse Nil
-    val calmIds =
-      calmIdsRule(target, sources) getOrElse Nil
+
+    val rules = List(miroIdsRule, physicalDigitalIdsRule, calmIdsRule)
+    val mergedSources = getSourcesToMerge(rules, target, sources)
+    val ids = rules.flatMap(_(target, sources)).flatten
```
jamieparkinson, line 28: These 3 lines are quite confusing but I think that might be helped by the proposed changes to `getSourcesToMerge`, and also by replacing the `_(target, sources)` syntax with an explicit (named) function call

jamesgorrie: * You have the rules to apply
* The sources that those rules will apply to
* The FieldData from those rules concatenated

Would a comment work? Not sure if the confusion comes from the fact that I am using `rules` to get the fielddata and sources?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
jamesgorrie: I don't see what the difference is besides taking the `filter` out of this method, and running it in the code above.

I'd rather remove it completely and just copy/paste this logic. It's just a convenience function, so don't want it to be partially convenient.

jamieparkinson: Totally in favour of copy/pasting it. Taking the `filter` out adds a lot of clarity imo because it represents that `sources` is some subset of the input sources, rather than a mystery list that comes out of a function with quite disparate arguments.

jamieparkinson: In fact, rather than copy/pasting, why not keep the old syntax like in `ItemsRule`?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/OtherIdentifiersRule.scala
jamieparkinson: The confusion is probably mostly due to `getSourcesToMerge` (and from the anonymous function syntax). Again, I think lists of functions are a bit hard to reason about and wonder if being explicit/repetitive would help...

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
jamesgorrie: The main reason is I got bored of copy/pasting.

There's probably somewhere else where storing the logic of "if a rule is applied here, then add it to the result.sources", but I'll just leave that as explicit goodness for now.

jamesgorrie: Still don't understand why this is massively unclear though, and we do it more than 3 times, so seems convenient. I get the idea of explicitness, but the merger in and of itself has loads of useful magic (👀 look at those overloaded apply functions ✨ !).

Give me some rules, a target and sources (that we use everywhere) and I will return you the sources which will have the rules applied, which, incidentally we need to create redirects.

jamieparkinson: You'll have to take my word for it that I didn't understand what it did! I'm not sure this counts as magic - and I'm not the biggest fan of overloaded apply anyway - but rather, a function which doesn't explain what it does until you read its implementation.

> Give me some rules, a target and sources (that we use everywhere) and I will return you the sources which will have the rules applied, which, incidentally we need to create redirects.

There is nothing in the function signature, name, or arguments that tells you that it will return `the sources which will have the rules applied` - you know that because you wrote it. From the name I would probably have guessed that it returns a list of sources which I then need to `merge`. 

My suggestion was one way in which I thought it could be made clearer what was happening - something like "filter out the sources which don't have any rule defined for them".

jamesgorrie: Yea, I think I haven't gone deep enough to make this abstraction useful, trying to encapsulate the logic of "If a rule matches, it needs to return which source matched". Removed it all for now.

> you know that because you wrote it.

I 100% agree with trying to get to a common understanding - but this sentence forgets that the person reading the code does need to try and understand the code, sometimes by more than just looking at it (what I have been doing this week for a lot of the merger code).

To write something that will be understandable to all, especially over time, is hard.

Given the sentence:
>  one way in which I thought it could be made clearer 

and the method suggested `anyRuleDefinedForSource` is a example. 

With that implementation, in 3 months, I would bne where I was at the beginning of this PR not understanding _why_ we do that, which for me is very important as I battle to understand without context.

So being clear in the method name & signature, but also clear in tests and implementation as to why we use it.

jamieparkinson: I think we're talking about different things here - I was only concerned with the question of _"can I understand **what** this code does by looking at it?"_.

The question of _why_ it does that is, as you say, extremely tricky and I don't think either/any implementation of this logic can make that clear - naming, tests, documentation, and comments help, which you've improved in this PR.

> you know that because you wrote it.

This was unclear/unfair/inconsiderate of me to write this - what I really meant was, as someone reading the code you had to go inside the method to understand what it was doing. My preference is to surface the "what" of a method with as few layers of indirection as possible, so it's easier to ask at least the right sort of "why" question, even if you can't answer it.

Thinking about that (as in, why it was confusing to me) made me think that my suggested naming could be clarified even more to point us even closer to the right "why" question, by capturing both the idea of the rules being "defined" (confusing terminology) and the merging. Perhaps something like:
```scala
sources.filter(anyRuleMergesSource(
  target = target,
  rules = Seq(fooRule, barRule, bazRule)
))
```

jamesgorrie: > This was unclear/unfair/inconsiderate of me to write this

Not at all - I think I understood what you meant, but it did make me think about what "understandable" code means.

I'll take this thinking into the next PR which will be looking at https://github.com/wellcomecollection/catalogue/pull/470#issuecomment-604317173

The method above is nice and clear, and fits into the definition of what a merge is, which I need to write down (PR coming soon, I want to document the new implementation at the same time as writing the rules).

jamesgorrie: With the removal of this - are you happy with that? It feels clear enough, and explicit. My only concern is that it doesn't enforce "if a rule is applied, consider it a merge, and return the source" - but that's probably a good thing for now.

jamieparkinson: :+1: documentation would be great, I think a lot of the oddities in the merger (which are being gradually ironed out) are caused in part by refactoring old code, preserving exactly what it did even where that was unnecessary, because a big picture description didn't exist.

jamieparkinson: Ah I hadn't refreshed this page and didn't see your 2nd comment - yes, that's fine by me!

PR #518, opened 2020-04-07
Simplify ElasticsearchIndexCreatorTest

alexwlchan:
I was looking at this test when I ported it to the storage-service repo, and I’ve long felt it was unnecessarily complicated. I did some refactoring, and I think this version is substantially easier to read.


alexwlchan:
Ack, I can't be bothered right now.

PR #521, opened 2020-04-07
Refactor ItemsRule on merging

jamesgorrie:
Here is a digram of the current state of linking works is here:
https://excalidraw.com/#json=5756447476416512,NGMwzTAUToDrgV3jUFzYWg

I have specifically been a bit loose with the Calm record as I the next step is to make the Calm record the `target` when merging.

The tests probably need a bit of working, v2 will be in the next PR as this is getting too large.

We have also remove the need to get digital locations from sierra.

In short:

## Miro
- Single item sierra: merge location into sierra item
- Multi item sierra: Leave unmerged

## Calm
- Single item sierra: Replace physical locations with calm physical location
- Multi item sierra: add calm item

## METS
- Single item sierra: Add mets location to sierra item
- Multi item sierra: add mets item

## Sierra
When there is a Sierra candidate, which would be it's digital copy, we just redirect that to the physical target. This is because all information is now coming from more close-to-the-metal sources i.e. METS.


jamesgorrie:
Waiting on validation

pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/rules/ItemsRuleTest.scala
```diff
@@ -59,16 +59,6 @@ class ItemsRuleTest
     }
   }
 
-  it(
-    "merges non-duplicate-URL locations from METS items into single-item Sierra works") {
-    inside(ItemsRule.merge(digitalSierra, List(metsWorkWithSierraUrl))) {
```
jamieparkinson, line 64: Think you'll want to remove `metsWorkWithSierraUrl` too

pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/rules/WorkPredicatesTest.scala
```diff
@@ -85,12 +85,12 @@ class WorkPredicatesTest
...
-    }
-  }
+//  it("selects digital Sierra works") {
+//    forAll(works.filter(WorkPredicates.digitalSierra)) { work =>
+//      work.sourceIdentifier.identifierType.id shouldBe "sierra-system-number"
+//      every(work.data.items.flatMap(_.locations)) should matchPattern {
+//        case _: DigitalLocation =>
+//      }
+//    }
+//  }
```
warrd, line 95: Should either be removed / uncommented I think

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -79,51 +105,45 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
-                (sierraSources ++ miroSources).flatMap(
-                  _.data.items.flatMap(_.locations)
-                )
+            sierraItem.copy(
+              locations = sierraItem.locations ++ miroItems.flatMap(_.locations)
             )
           )
-        case (multipleSierraItems, (_, sierraSources)) =>
-          multipleSierraItems ++ sierraSources.flatMap(_.data.items)
+        // This case is unreachable due to the `singleItemSierra` predicate
```
jamieparkinson, line 143: Maybe it would be clearer to replace this block with a map in this case? Not sure, removes the redundant arm of the pattern match at least, but doesn't represent that there's only ever one item.
```
sierraItems.map { sierraItem =>
  sierraItem.copy(
    locations = sierraItem.locations ++ miroItems.flatMap(_.locations)
  )
}
```

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidates.scala
```diff
@@ -68,11 +68,9 @@ object SierraMergeCandidates
...
   /** We can merge a single-page Miro and Sierra work if:
     *
-    *   - The Sierra work has type "Picture" or "Digital images"
-    *   - There's exactly one Miro ID in MARC tag 962 subfield $u
-    *     (if there's more than one Miro ID, we can't do a merge).
-    *   - There's exactly one item on the Sierra record (if there's more
-    *     than one item, we don't know where to put the Miro location).
+    *   - The Sierra work has type "Picture", "Digital images", or "3D Object"
+    *   - There's exactly one Miro ID in MARC tag 962$u or 089$a
+    *     (if there's more than one Miro ID in either, we can't do a merge)
```
jamieparkinson, line 73: I don't think this is in the scope of this PR, but just a note that I think (??) we intend to relax this and allow merging multiple Miros into 1 Sierra? I know that the cases where we don't do that are a definite pain point at the moment (eg https://wellcomecollection.org/works?query=borax%20pet)

docs/pipeline/merger_rules.md
```diff
@@ -0,0 +1,5 @@
+# Merger Rules
+
+## Items
+![How items are selected as candidates and merged](../.gitbook/assets/merger_items_rule.png)
```
jamieparkinson, line 4: I'm afraid I find this diagram very confusing - I've been staring at it for a few minutes and still don't understand it, so it's a bit tricky to say what I think would help, but some thoughts would be:

- Could you organise it so that information flows in one direction? It's not clear what's an input or an output - are the Miro and Calm works at the bottom the same type of thing as the works above them?
- I think that different arrows represent different things - eg some on the left hand side are field names in source data referring to other sources, whereas some of the ones that go left-to-right are about the results of merging (ie the data flow). Could the entity relationships vs merges/transforms be made explicit?
- Related to that - could you clarify that this also encompasses some logic in the transformers? It seems to be about more than just the merger's items rule.
- What does 'Nothing' refer to?
- Why is a SIM rule called a SIM rule? Do we use that name anywhere else? Should we be using it in code? I might also rephrase point (2) in that description to something like "Otherwise, if there is more than one item, append the items from the source to those in the target" - I'm not sure that 'verbatim' explains what happens here.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -79,51 +105,45 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
       }
     }
   }
 
-  private val mergeMiroPhysicalAndDigitalItems = new PartialRule {
-    val isDefinedForTarget: WorkPredicate = WorkPredicates.sierraWork
-    val isDefinedForSource
-      : WorkPredicate = WorkPredicates.miroWork or WorkPredicates.digitalSierra
+  /** We merge the Miro location to the Sierra work with a single item.
+    * If there are multiple items, we don't merge.
```
jamieparkinson, line 124: It would be nice to have a note on why this is (I don't know the answer)

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidates.scala
jamieparkinson: > we can't do a merge

As in the merger - why not?

docs/pipeline/merger_rules.md
jamesgorrie: Yea, 100%. Apologies for the slapdashness of it, this was almost meant to be a place holder for the next iteration, and a way for me to keep track of the conversations I was having with JT, so not very useful for others.

I am trying to strike a balance of keeping this PR as a stepping stone to something better. Does the description of the PR help? I'll try mimic that.

jamieparkinson: The description is helpful, would be ideal to have something like a slightly fleshed out version of that next to an improved diagram. Possibly remove the docs from this PR and do that separately? Up to you

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: This wouldn't work as we only want to add the locations if there is a single sierra item, hence the pattern match.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -79,51 +105,46 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
     }
   }
 
-  private val mergeMiroPhysicalAndDigitalItems = new PartialRule {
-    val isDefinedForTarget: WorkPredicate = WorkPredicates.sierraWork
-    val isDefinedForSource
-      : WorkPredicate = WorkPredicates.miroWork or WorkPredicates.digitalSierra
+  /** We merge the Miro location to the Sierra work with a single item.
+    * If there are multiple items, we don't merge as we wouldn't know
+    * which item to attache the location to.
```
jamieparkinson, line 125: Why can't we do the same as with other types (ie, appending the items)?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamieparkinson: But we've already restricted the sierra item to be a `singleItemSierra`

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidates.scala
jamesgorrie: Not sure where this came from, happy to dig it out further as another PR.
https://github.com/wellcomecollection/catalogue/issues/536

docs/pipeline/merger_rules.md
jamesgorrie: Na, I'll try get it a little better than a brain splurge.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: We want it to be a separate work, but the question as to why, I am lost. Ping @jtweed

jamieparkinson: Feel like this might be related to https://github.com/wellcomecollection/catalogue/pull/521#discussion_r409647521

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: I've done it using head to be more explicit about what we're doing. Is that nicer?

jamieparkinson: Yeah that works too - although to be truly nitpicky we could also use `headOption.map{...}.getOrElse(...)` to stay type safe...

jamesgorrie: I prefer that but we're not using it elsewhere.

And this makes it even more safe:
https://github.com/wellcomecollection/catalogue/blob/master/pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/rules/WorkPredicatesTest.scala#L72-L77

For legibility I might keep it. And this will throw if something is up which might be better than swallowing it in a `getOrElse`.

You could have a `getOrElse(Left(PredicateFailedMe))` - but that would be a whole other refactor.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamieparkinson: Also somehow relevant, I remember I made this after a conversation with @jtweed https://github.com/wellcomecollection/platform/issues/4299

docs/pipeline/merger_rules.md
jamesgorrie: Updated - what do you think?

The merging bit I will do post a Calm target world as it will add a bit of complexity to diagramming.

jamieparkinson: Much clearer! Maybe change `work` -> `works` as otherwise it kinda looks like a Sierra work would be referring to itself?

PR #523, opened 2020-04-08
Inference services infrastructure

jamieparkinson:


pipeline/terraform/modules/service_with_manager/main.tf
```diff
@@ -0,0 +1,62 @@
+module "service" {
+  source = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//service?ref=v1.2.0"
```
kenoir, line 2: If you want cloudwatch logs don't forget to flip the switch in this module.

PR #524, opened 2020-04-08
Inference manager fixes and integration test

jamieparkinson:
The docker-compose won't work yet - that will depend upon https://github.com/wellcomecollection/catalogue/pull/525


pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/Main.scala
```diff
@@ -41,7 +41,7 @@ object Main extends WellcomeTypesafeApp {
       Http()
         .cachedHostConnectionPool[(Message, Input)](
           config.getOrElse[String]("inferrer.host")("localhost"),
-          config.getOrElse[Int]("inferrer.port")(80)
+          config.getOrElse[String]("inferrer.port")("80").toInt
```
kenoir, line 44: 😬 _know we have to do this, but yuck_

PR #525, opened 2020-04-08
Inference service updates for integration

jamieparkinson:
Optimisation to come in a later PR


pipeline/inferrer/feature_extraction/training/src/aws.py
```diff
@@ -1,21 +1,31 @@
...
-def put_object_to_s3(binary_object, key, bucket_name, session=None, profile_name=None):
-    s3_client = get_s3_client(session, profile_name)
+def put_object_to_s3(binary_object, key, bucket_name):
+    s3_client = get_s3_client()
+    print("Uploading object to S3...")
     s3_client.put_object(Bucket=bucket_name, Key=key, Body=binary_object)
+    print("Uploaded object to S3.")
+
+
+def put_ssm_parameter(path, value, description):
```
kenoir, line 20: What are we using SSM for?

kenoir: Ignore me, had to read to the end. Could we not just use S3 to also store the path though - reduce the types of data store we're using.

jamieparkinson: hmmm, we could, but the intention here is for Terraform to be able to get it and inject it as an environment variable.

The work to get the training stuff deployed still [needs to be done](https://github.com/wellcomecollection/platform/issues/4406) and I think will affect this sort of thing, so I might prefer to leave working out the optimal way to get this info until that piece of work.

pipeline/inferrer/feature_extraction/inference/app/main.py
```diff
@@ -54,3 +56,13 @@ def main(image_url: str = None, iiif_url: str = None):
...
+
+
+@app.on_event("startup")
+def on_startup():
+    http.start_persistent_client_session()
+
+
+@app.on_event("shutdown")
+def on_shutdown():
+    http.stop_persistent_client_session()
```
harrisonpim, line 68: this interface to the smart stuff is lovely 😍

harrisonpim:
https://github.com/wellcomecollection/catalogue/blob/e8f5e77ccf7f6d0492e8a592e6774cbdaa74ece7/pipeline/inferrer/feature_extraction/training/src/elastic.py#L24

not _really_ in the scope of these changes, just noting that we should only really be fetching `["feature_vector_1", "feature_vector_2"]` from the `_source` here, especially as the number of inferrers (and the size of the inferred data on each work) grows

pipeline/inferrer/feature_extraction/training/src/aws.py
kenoir: oh, sorry - i imagined this happening at runtime, what you suggest makes sense.

PR #531, opened 2020-04-16
Suppress Calm works with Transmission=No

warrd:
Use `Transmission` field to indicate whether a work should be transformed or surpressed

### DIFF
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
@@ -34,6 +34,16 @@ object CalmTransformer extends Transformer[CalmRecord] with CalmOps {
       )
     }
 
+  def shouldTransform(record: CalmRecord): Boolean =
+    record
+      .get("Transmission")
+      .map(_.toLowerCase)
+      .map {
+        case "no" => false
+        case _    => true
+      }
+      .getOrElse(true)
+
   def workData(record: CalmRecord): Result[WorkData[Unminted, Identifiable]] =
     for {
       accessStatus <- accessStatus(record)
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
@@ -42,7 +42,8 @@ object CalmTransformer extends Transformer[CalmRecord] with CalmOps {
         case "no"  => false
         case "yes" => true
         case other =>
-          info(s"Unrecognised value for Transmission field; assuming true: $other")
+          info(
+            s"Unrecognised value for Transmission field; assuming true: $other")
           true
       }
       .getOrElse(true)
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
@@ -39,8 +39,11 @@ object CalmTransformer extends Transformer[CalmRecord] with CalmOps {
       .get("Transmission")
       .map(_.toLowerCase)
       .map {
-        case "no" => false
-        case _    => true
+        case "no"  => false
+        case "yes" => true
+        case other =>
+          info(s"Unrecognised value for Transmission field; assuming true: $other")
+          true
       }
       .getOrElse(true)
 
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
@@ -37,6 +37,11 @@ trait TransformerWorker[In, SenderDest] extends Logging {
   type Result[T] = Either[TransformerWorkerError, T]
   type StoreKey = Version[String, Int]
 
+  sealed trait TransformResult;
+  case class Success(key: StoreKey, work: TransformedBaseWork) extends TransformResult
+  case class Surpress(key: StoreKey) extends TransformResult
+  case class Error(error: TransformerWorkerError) extends TransformResult
+
   def name: String = this.getClass.getSimpleName
   val stream: SQSStream[NotificationMessage]
   val sender: BigMessageSender[SenderDest, TransformedBaseWork]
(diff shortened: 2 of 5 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Language.scala
```diff
@@ -35,12 +35,15 @@ object Language {
...
       .toList
 
   lazy private val languageCodeMap =
     languageCodes.toMap
 
   lazy private val languageLabelMap =
-    languageCodes.map { case (code, label) => (label, code) }.toMap
+    languageCodes
+      .flatMap { case (code, label) =>
+          label.split(";").toList.map(label => (label.trim, code))
```
alexwlchan: Might be worth an example here of the line in the CSV you're trying to parse? The Dutch/Flemish example in the pull request comment was good!

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -34,6 +34,16 @@ object CalmTransformer extends Transformer[CalmRecord] with CalmOps {
...
     }
 
+  def shouldTransform(record: CalmRecord): Boolean =
+    record
+      .get("Transmission")
+      .map(_.toLowerCase)
+      .map {
+        case "no" => false
+        case _    => true
+      }
```
alexwlchan, line 44: I assume the other value you'd expect here would be `"yes"`?

Suggestions:

* Add a comment explaining what this field means in CALM (what have you been told it means?)
* Add a third branch like so:

```suggestion
      .map {
        case "no"  => false
        case "yes" => true
        case other =>
          info(s"Unrecognised value for Transmission field; assuming true: $other")
          true
      }
```

So if we start getting works appearing that we don't expect, we can see what happened here.

(I can imagine, for example, a record with value `"no "` that this would let through, because CALM does no checking of values.)

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -37,34 +37,42 @@ trait TransformerWorker[In, SenderDest] extends Logging {
   type Result[T] = Either[TransformerWorkerError, T]
   type StoreKey = Version[String, Int]
 
+  sealed trait TransformResult;
+  case class Success(key: StoreKey, work: TransformedBaseWork) extends TransformResult
+  case class Surpress(key: StoreKey) extends TransformResult
```
alexwlchan, line 43: Is this a typo for Suppress?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -37,34 +37,42 @@ trait TransformerWorker[In, SenderDest] extends Logging {
...
-      work <- work(recordAndKey._1, key)
-      done <- done(work, key)
-    } yield done
-  }
+      record <- getRecord(key)
+      shouldTransform = transformer.shouldTransform(record)
+      result <- if (!shouldTransform) Right(Surpress(key)) else for {
+        success <- work(record, key)
+        _ <- done(success)
+      } yield success
```
alexwlchan: This is just a bit much on one line for my taste. Could you break out the inner for comprehension to another function?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
```diff
@@ -100,10 +108,14 @@ trait TransformerWorker[In, SenderDest] extends Logging {
                 }
                 Future.failed(err)
               }
-              case Right((work, key)) => {
+              case Success(key, work) => {
                 info(s"$name: from $key transformed $work")
                 Future.successful(message)
               }
+              case Surpress(key) => {
```
jamesgorrie, line 115: I like this - but it break the rule @jamieparkinson and I have spoken about of 1 in 1 out - where there is a source version, we should always go right through to ingestion.

We have "suppressed" works previously by using `InvisibleWorks` - should we stick to this way of doing things?

pipeline/transformer/transformer_calm/src/test/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformerTest.scala
```diff
@@ -254,6 +308,32 @@ class CalmTransformerTest extends FunSpec with Matchers {
     CalmTransformer(record, version) shouldBe a[Left[_, _]]
   }
 
+  it("surpresses data when Transmission=No") {
```
alexwlchan, line 311:
```suggestion
  it("suppresses data when Transmission=No") {
```

pipeline/transformer/transformer_calm/src/test/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformerTest.scala
```diff
@@ -254,6 +308,32 @@ class CalmTransformerTest extends FunSpec with Matchers {
...
+    )
+    val c = calmRecord(
+      "Title" -> "abc",
+      "Level" -> "Collection",
+      "RefNo" -> "a/b/c",
+      "AltRefNo" -> "a.b.c",
+    )
+    CalmTransformer.shouldTransform(a) shouldBe false
+    CalmTransformer.shouldTransform(b) shouldBe true
+    CalmTransformer.shouldTransform(c) shouldBe true
```
alexwlchan, line 334: One more test case needed here: what if the Transmission field contains a completely unexpected value?

`"Transmission" -> "Manual"`, I dunno

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
warrd: Ah didn't know that rule was something we wanted.

The reason I didn't try and transform to invisbile work is because the way it appeared when discussing with @jtweed is that the work was not parsable and hence the transformer would fail. He said works with `Tranmission=No` don't even make it to Sierra, so if that is the case shouldn't be coming through the pipeline on that side. Any thoughts @jtweed?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
warrd: I guess so, looked right to me lol!

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/TransformerWorker.scala
jamesgorrie: My 2¢ - I don't really mind how we do it, but we have it one way, and would prefer to have only 1 way.

[Here's an example test with Sierra](https://github.com/wellcomecollection/catalogue/blob/2c57aec3571326ba34adbe1d61c10e4db4c67133/pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraTransformableTransformerTest.scala#L291).

jamieparkinson: I am kind of OK with suppressing things at the transformer level as at this stage I would argue that they are not works - they are source data.

However - it's spelled `Suppress` 😛

jamesgorrie: The other consideration is if someone goes from a Transmittable record, to setting it to non-transmittable, we should have that pass through the pipeline, and render the appropriate message on the site. i.e. this item is now suppressed.

warrd: I can see why we would want to do that, but don't think that is possible with the current pipeline architecture? We would have to have the record history as input to the transformer to indicate something has gone from non-suppresses to suppressed.

jamesgorrie: That's why we just transform it to an invisible work if it's suppressed, so it would override the current one, allowing us to respond correctly from the API.

e.g. A user in calm does these steps.
```
1. CalmRecord.Transmission = "No"  => InvisibleWork
2. CalmRecord.Transmission = "Yes" => TransformedWork
3. CalmRecord.Transmission = "No" => InvisibleWork
```

warrd: Ah I get you, yes would say the work is surpressed, thought you meant would display it's history somehow.

jamesgorrie: Na, just that it is _no longer_ available.

warrd: Well my point was that the user won't know whether an invisible work is no longer available or never available, but in practices thats probably fine

jamesgorrie: Yea, a way to get around having to know previous state is always send suppressed as Invisible - agreed, the type of unavailable is not currently possible or desired.

We currently just say something has been deleted:
https://api.wellcomecollection.org/catalogue/v2/works/qefk954w

warrd:
#559 instead

PR #532, opened 2020-04-16
Supress calm works with Transmission=No

warrd:
Split out from #531


warrd:
Closing this and using #531 as want to preserve current comments

PR #533, opened 2020-04-16
Calm Language and Level fixes

warrd:
Split out from #531 

* In the CSV use `;` to  indicate language label synonyms (e.g. `Dutch; Flemish` parses both "Dutch" and "Flemish" as `nl`)
* Ignore leading / trailing whitespace when parsing all fields
* Ignore case for level (e.g. lots of records marked as `Subseries`)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Language.scala
```diff
@@ -35,12 +35,17 @@ object Language {
...
 
   lazy private val languageCodeMap =
     languageCodes.toMap
 
   lazy private val languageLabelMap =
-    languageCodes.map { case (code, label) => (label, code) }.toMap
+    languageCodes.flatMap {
+      case (code, label) =>
+        // This is so that e.g. we parse both Dutch and Flemish from
+        // "Dutch; Flemish" as nl
```
alexwlchan, line 48: 🎉

PR #544, opened 2020-04-17
Making the merger only work on how to merge

jamesgorrie:
**This PR is for me to think aloud and have anyone interested to be able to contribute - if someone would rather scribble on excalidraw, let's do that**

Thinking about moving the logic to link works out of the `merger`, making it responsible for _how things merge_ rather than _should things merge_ which is already partially being defined by the `transformer`.

I'm not 100% convinced that the `transformer` is the place for this, but the service creating the links needs access to:
1. Source data to access linked IDs
1. Transformed target data
1. [Transformed sources data] ❓ 

Point 3 I am unsure of, and if it is required, we're a bit stuck on this path.

### Move "if something can merge" logic to matcher 1/5 review

Transformer just gets any available linked IDs
Matcher fetches graph of links **and applies logic as to whether they should merge**
Merger works out how to merge ☝️ 

### Move all "if something can merge" logic to matcher 1/5 review

Transformer just gets any available linked IDs
Matcher fetches graph of links **and applies logic as to whether they should merge**
Merger works out how to merge ☝️ 

... more thinking coming soon.

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
@@ -120,31 +120,35 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
     }
   }
 
-  /** We merge the Miro location to the Sierra work with a single item
-    * as we assume that this is the only item that the Miro image
-    * could be associated with.
+  /**
+    * If we have a single item, we then assume that the MIRO items MUST
+    * be associated with that item.
     *
-    * If we have multiple items, we assume it is definitely associated with
-    * one of the Sierra items, but unsure of which. We thus don't append it
-    * to the Sierra items to avoid certain duplication, and leave the works
-    * unmerged.
+    * If not, we just add the append the MIRO items to the list.
+    *
+    * Note: The Sierra transformer currently only ever send us Sierra works with single item.
     */
   private val mergeMiroItems = new PartialRule {
-    val isDefinedForTarget: WorkPredicate = WorkPredicates.singleItemSierra
+    val isDefinedForTarget: WorkPredicate = WorkPredicates.sierraWork
     val isDefinedForSource: WorkPredicate = WorkPredicates.miroWork
 
     def rule(target: UnidentifiedWork,
              sources: NonEmptyList[TransformedBaseWork]): FieldData = {
-      // This is safe due to the `singleItemSierra` predicate
-      val sierraItem = target.data.items.head
+      val sierraItems = target.data.items
       val miroItems = sources.toList.flatMap(_.data.items)
 
       debug(s"Merging Miro items from ${describeWorks(sources)}")
 
-      List(
-        sierraItem.copy(
-          locations = sierraItem.locations ++ miroItems.flatMap(_.locations)
-        ))
+      sierraItems match {
+        case List(sierraItem) =>
+          List(
+            sierraItem.copy(
+              locations = sierraItem.locations ++ miroItems.flatMap(_.locations)
+            ))
+
+        case multipleItems =>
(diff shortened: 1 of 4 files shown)


alicefuzier:
My 2 cents: perhaps the merger is too complex and some stuff should be moved out. That's quite possible and we should do that if that's the case. 


But I don't think the matcher is the place for merging rules (which decide if something needs to be merged). It's already a very complex beast as it is, and that's because what it does isn't simple. But the nice thing about it, in my opinion, is that it doesn't change. It's a bit like the idminter in that it cares about only a tiny bit of our model to do its thing so we can extend the model and the pipeline all we like, but the matcher still works with almost no change. I think it should stay that way.


Also, I don't think the transformer decides what needs to merged. It's a bit confusing that the field is called `mergedCandidates` so it makes it natural to think that way. But the way I see the transformer is just an extractor and translator. That means its role is to understand the source data and make decisions about what source data is relevant to the catalogue pipeline as a whole, not just the api. But it doesn't have enough information at that point to make decisions about merging and even if it did I think that would fall outside the transformer responsibilities.

So we might take some stuff out of the merger but that to me would look like a new service between the matcher and the merger.

jamesgorrie:
> But I don't think the matcher is the place for merging rules

Yea, I 💯 agree with that.

> I don't think the transformer decides what needs to merged.

I agree that it shouldn't but currently does. Taking the MIRO works as an example.

When we extract the MIRO works, there's logic in there to only extract single miro IDs. There is then logic to only merge with Sierra works dependant on how many items the work has in the merger.

Given this:

> But it doesn't have enough information at that point to make decisions about merging and even if it did I think that would fall outside the transformer responsibilities.

The logic of the miro IDs should sit within the merger as it stands.

But then might be moved into here:

> a new service between the matcher and the merger.

☝️ I think I dreamt about this that weekend, it sound useful, but maybe not something in priority right now.

jamieparkinson:
As I see it (I think this is what @alicefuzier is saying as well) the transformer simply extracts links to other works from source data - it shouldn't really have an opinion on whether these should be merged, which is where the filtering of these links and the naming `mergeCandidates` are confusing.

I feel like simply changing that name to (eg) `linkedWorks` and removing all filtering (ie, adding all linked works no matter what) in the transformer, and letting the merger make judgements on what it can or can't merge, would be quite an easy way of making this clearer.

In my mind, the problem is not that the merger is doing stuff other than merging, it's that merging is being done by stuff other than the merger - if the "can this be merged" and the "merge these things" were in separate services I feel like these would be so tightly coupled as to make it a bit of an arbitrary split.

Remember that the previous version of the merger did separate "partitioning" and "merging", but that this ended up being a sort of false dichotomy as the "merging" was still not guaranteed to be possible after the partitioning(!), and that these 2 operations ended up being so coupled and/or redundant as to lose their meaning a little.

Whilst it seems clear(er) to partition for items/locations, it doesn't make sense for a lot of the other fields: maybe the `ItemsRule` could be written wholly differently to the other rules if we're finding it's confusing as-is? That said, the work @jamesgorrie has done today has made it a lot clearer without diverging all that much from the existing pattern.

_TLDR;_ I am very much in favour of `mergeCandidates` -> `linkedWorks` and think it would really help with these problems.

jamesgorrie:
> the transformer simply extracts links to other works from source data

It also makes some decisions around Miro works, hence my confusion as to where people thought this should live.

> I feel like simply changing that name to (eg) linkedWorks and removing all filtering (ie, adding all linked works no matter what) in the transformer, and letting the merger make judgements on what it can or can't merge, would be quite an easy way of making this clearer.

Then we're all agreed.

jamieparkinson:
Yes I was very unclear there, I actually meant:

> the transformer **should** simply extract links to other works from source data - it shouldn't really have an opinion on whether these should be merged, which is where the filtering of these links and the naming mergeCandidates are confusing.

where the filtering is what you linked.

PR #547, opened 2020-04-20
Feature inferrer CI

jamieparkinson:
Things this does:

- Runs inference manager integration test for changes in manager *or* inferrer
- Publishes the inferrer when necessary
- "tests" the inferrer by checking that it compiles
- Adds the ability to define a custom test command for Python projects
- Adds the ability to watch changes matching a glob for certain Makefile recipes
- Upgrades the travis build scripts/environment to Python 3
- Upgrades the travis build VM to the latest (until Thursday 🙃) Ubuntu LTS

Tests are broken because they are in master - will rebase when that's fixed.

### DIFF
--- a/.travis/git_utils.py
+++ b/.travis/git_utils.py
@@ -14,7 +14,7 @@ def git(*args):
         sys.exit(err.returncode)
 
 
-def get_changed_paths(*args, globs=[]):
+def get_changed_paths(*args, globs=None):
     """
     Returns a set of changed paths in a given commit range.
 
--- a/.travis/git_utils.py
+++ b/.travis/git_utils.py
@@ -14,16 +14,15 @@ def git(*args):
         sys.exit(err.returncode)
 
 
-# Root of the Git repository
-ROOT = git("rev-parse", "--show-toplevel")
-
-
-def get_changed_paths(*args):
+def get_changed_paths(*args, globs=[]):
     """
     Returns a set of changed paths in a given commit range.
 
-    :param commit_range: Arguments to pass to ``git diff``.
+    :param args: Arguments to pass to ``git diff``.
+    :param globs: List of file globs to include in changed paths.
     """
+    if globs:
+        args = list(args) + ["--", *globs]
     diff_output = git("diff", "--name-only", *args)
 
     return set([line.strip() for line in diff_output.splitlines()])
--- a/.travis/run_job.py
+++ b/.travis/run_job.py
@@ -80,7 +80,7 @@ if __name__ == "__main__":
     travis_commit_range = os.environ["TRAVIS_COMMIT_RANGE"]
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("project_name", default=os.environ["SBT_PROJECT"])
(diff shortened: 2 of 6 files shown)


harrisonpim:
lots that i've never touched here, so not sure i have any useful comments to contribute to the code itself - the list of things done looks perfect

.travis/git_utils.py
```diff
@@ -14,16 +14,15 @@ def git(*args):
         sys.exit(err.returncode)
 
 
-# Root of the Git repository
-ROOT = git("rev-parse", "--show-toplevel")
-
-
-def get_changed_paths(*args):
+def get_changed_paths(*args, globs=[]):
```
alexwlchan, line 17: Beware! Using mutable default arguments in Python is a bad habit, because any mutation within the body of the function persists between calls. Example:

```pylon
>>> def print_names(names=[]):
...     names.append("alex")
...     print(names)
...
>>> print_names(names=["jamie", "harrison"])
['jamie', 'harrison', 'alex']

>>> print_names()
['alex']

>>> print_names()
['alex', 'Alex']

>>> print_names()
['alex', 'alex', 'Alex']
```

Preferred:

```python
def get_changed_paths(*args, globs=None):
    if globs is None:
        globs = []
```

.travis/run_job.py
```diff
@@ -86,30 +77,40 @@ def should_run_sbt_project(repo, project_name, changed_paths):
...
-        sbt_project_name = get_project_name()
+        parser = argparse.ArgumentParser()
+        parser.add_argument("project_name", default=os.environ.get("SBT_PROJECT"))
+        parser.add_argument("--changes-in", nargs="*")
+        args = parser.parse_args()
 
-        repo = Repository(".sbt_metadata")
+        task = f"{args.project_name}-test"
 
+        change_globs = args.changes_in + [".travis.yml"] if args.changes_in else None
```
alexwlchan, line 93: nit: maybe pull this on to a separate line? I don't find one-line ternaries especially used in Python.

.travis/git_utils.py
alexwlchan: Actually, you don't even need the default argument here, because you've fixed the two places where this function is called.

.travis/run_job.py
```diff
@@ -86,30 +77,40 @@ def should_run_sbt_project(repo, project_name, changed_paths):
...
-            )
-            sys.exit(0)
+        sbt_repo = Repository(".sbt_metadata")
+        try:
+            if not should_run_sbt_project(sbt_repo, args.project_name, changed_paths):
+                print(
+                    f"Nothing in this patch affects {args.project_name}, so skipping tests"
+                )
+                sys.exit(0)
+        except (FileNotFoundError, KeyError):
```
alexwlchan, line 107: When would we throw either of these errors?

jamieparkinson: If we're not in an SBT project, from [here](https://github.com/wellcomecollection/catalogue/blob/feature-inferrer-ci/.travis/sbt_dependency_tree.py#L49)

PR #548, opened 2020-04-20
Merging items into calm target

jamesgorrie:
This defines the workflow of:

* Calm item created 
* Sierra record created in Calm harvest
* METS files created for ☝️ Sierra record

Fixes: https://github.com/wellcomecollection/platform/issues/4434


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -147,4 +155,41 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
+    */
+  private val mergeIntoCalmTarget = new PartialRule {
+    val isDefinedForTarget: WorkPredicate = WorkPredicates.calmWork
+    val isDefinedForSource
+      : WorkPredicate = WorkPredicates.metsWork or WorkPredicates.sierraWork
+
+    def rule(target: UnidentifiedWork,
+             sources: NonEmptyList[TransformedBaseWork]): FieldData = {
+
+      // As the Calm transformer always adds 1 item with 1 location, this is safe
```
jamieparkinson, line 174: I'm a bit uncomfortable with relying on runtime behaviour in another service - could you add a check for the item number to `WorkPredicates.calmWork` (or make a `singleItemCalm` predicate)? Or do it with `headOption`, whatever you'd prefer but just keeping this safety within the merger.

jamesgorrie: Yea, I copied this from before, but happy to improve.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -147,4 +155,41 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
+      val calmItem = target.data.items.head
+
+      val metsSources = sources.filter(WorkPredicates.metsWork)
+      val metsDigitalLocations =
+        metsSources.flatMap(_.data.items.flatMap(_.locations.filter(_ match {
```
jamieparkinson, line 179: I _think_ that all METS locations are digital locations - do you know if that's the case?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: I almost went down the route of trying to type this., but just seemed to add way more complexity. 

> in another service

What do you mean?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: Yes.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamieparkinson: Say that something changes in the calm transformer (the other service) that means calm works can have > 1 items - that would break the merger. Tests wouldn't necessarily catch it because there would have been no changes in the merger itself.

When we're relying on runtime checks/behaviour rather than types for safety, I think it's preferable to keep the assumptions within one service.

You don't have to do this in a typesafe way if you think it makes it too complex, just do it like the single-item Sierra checks:

```scala
val singleItemCalm: WorkPredicate = satisfiesAll(calmWork, singleItem)
```

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamieparkinson: This might've been here from before due to types and/or me wanting to avoid using brain when refactoring the old code

jamesgorrie: The type system hasn't made it very convenient as you need to keep having checks everywhere. We could rewrite the works type to be something like:

```
Work[LocationsType <: List](locations)

CalmWork extends Work[SingleItemList[Location]]
SierraWork extends [List[Locations]]
METSWork extends Work[SingleItemList[Location]]
```

But it's getting very weird at that point.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: Added it to the predicate as it felt like the most explanatory way of doing it.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: METS as a source is just a digitized item.

This points that the generic work model does not necessarily help at defining our source data.

jamesgorrie: Another way would be send partial information down the pipes from the transformers, but not sure that's a change I want to make.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: This still doesn't help that much as everything will still pass if we forget to look at the merger if we change the transformer. I've just defined the `calmWork` as you have above as there is no other type of calm work.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: Removed the filter.

PR #549, opened 2020-04-21
Patch/merging items into sierra

jamesgorrie:
Forked from #548 

Moves away from differentiating on how we merge based on `sources` given that the result of those merges are often then remerged above.

I fiddled with the idea of having an `FieldData = List[Locations]` `FieldMergeRule`, but it felt like I was creating abstractions to have pure code rather than reflect what we're actually doing.

Currently We use the `sierraSingleItemWork`, `sierraMultiItemWork` and `calmWork` predicated to define how we merge items into them.

This has the added benefit that we can then include / exclude source types, making it easier to fathom which sources to return as `mergedSources`.


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -24,21 +24,20 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
     target: UnidentifiedWork,
     sources: Seq[TransformedBaseWork]): FieldMergeResult[FieldData] = {
     val rules =
-      List(mergeCalmItems, mergeMetsItems, mergeMiroItems)
+      List(
```
jamieparkinson, line 27: Does the comment above this rule need updating? Github won't let me comment outside of the changes...

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -63,96 +62,60 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
 
-  /**
-    * If there is 1 Sierra item, we add the location from the METS record
-    * else we just append METS items as we wouldn't know which item to
-    * augment.
-    */
-  private val mergeMetsItems = new PartialRule {
-    val isDefinedForTarget: WorkPredicate = WorkPredicates.sierraWork
-    val isDefinedForSource: WorkPredicate = WorkPredicates.singleItemDigitalMets
+      val metsSources = sources.filter(WorkPredicates.metsWork)
```
jamieparkinson, line 83: This filtering is redundant as you're partitioning the list of sources, performing the ssame transforms to each partition, and then joining them; you could simply do:
```
val sierraItem = target.data.items.head
val sourceLocations = sources.flatMap(_.data.items).flatMap(_.locations)
List(
  sierraItem.copy(
    locations = sierraItem.locations ++ sourceLocations
  )
)
```

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -63,96 +62,60 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
     def rule(target: UnidentifiedWork,
              sources: NonEmptyList[TransformedBaseWork]): FieldData = {
-      // This is safe due to the `singleItemSierra` predicate
-      val sierraItem = target.data.items.head
-      val miroItems = sources.toList.flatMap(_.data.items)
 
-      debug(s"Merging Miro items from ${describeWorks(sources)}")
+      val sierraItems = target.data.items
+      val metsItems =
+        sources.filter(WorkPredicates.metsWork).flatMap(_.data.items)
```
jamieparkinson, line 116: No need to filter, `isDefinedForSource` is doing that already.
```
rule(...): FieldData = target.data.items ++ sources.flatMap(_.data.items)
```

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: Could do, but it feels like simplifying the code at the expense of expressing business logic, which I have spent quite a long time gathering - from that perspective, does it feel like it's worth keeping?

jamieparkinson: I think the business logic is captured in `isDefinedForSource`, the code is just redundant rather than simplifiably complex. You could call `sourceLocations` in my suggestion (eg) `metsAndMiroLocations` to be super explicit?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamesgorrie: > Github won't let me comment outside of the changes...

You have to comment on the left side of the split.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
jamieparkinson: imo the issue with having things like this to express business logic is that they are actually confusing to read - I spent a little while reading it making sure that I had understood correctly that it was redundant, assuming that it was doing kind of filtering or transformation on top of what the rule does

jamesgorrie: 🤷 I found it easier to say "Merge the mets and miro locations onto the sierra item". Non fussed either way. It's been changed to have less redundancy.

PR #550, opened 2020-04-21
Don't include collection tree on work endpoint if not in includes

warrd:
Currently it is always attempted to be created

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
@@ -86,7 +86,7 @@ class Router(elasticClient: ElasticClient,
   }
 
   def multipleWorks(params: MultipleWorksParams): Future[Route] =
-    transactFuture("GET /works")({
+    transactFuture("GET /works") {
       val searchOptions = params.searchOptions(apiConfig)
       val index = params._index.map(Index(_)).getOrElse(elasticConfig.index)
       worksService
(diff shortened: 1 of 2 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
```diff
@@ -106,27 +106,30 @@ class Router(elasticClient: ElasticClient,
...
       val includes = params.include.getOrElse(V2WorksIncludes())
       worksService
         .findWorkById(id)(index)
         .flatMap {
           case Right(Some(work: IdentifiedWork)) =>
-            val expandedPaths = params._expandPaths.getOrElse(Nil)
-            retrieveTree(index, work, expandedPaths).map {
-              workFound(work, _, includes)
-            }
+            if (includes.collection) {
```
jamieparkinson, line 119: Would it make more sense to move this logic down to `DisplayWorkV2.apply` (or an overload thereof)? Feels uncomfortable to be doing a special case for collections here, but perhaps it's  actually simpler...

warrd: The code inside the conditional calls off to `CollectionService`, which in turn callls off to elasticsearch, so not sure how the logic would exist in `display` unless we allowed passing in callbacks etc, which seems a bit messy.

jamieparkinson: :+1: makes sense

PR #553, opened 2020-04-21
Patch/simplify work predicates

jamesgorrie:
We had separate predicates for `miro`, `singleItemMiro` etc.

☝️ That logic isn't significant to the merging logic, it was just saying "When it's from this source". I have been more specific so future people 🛰️ coming to this code, specifically from the land of transformers might easily figure out that we merge a specific shape of work coming from said transformer land.

TODO
- [x] Update tests

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/WorkPredicates.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/WorkPredicates.scala
@@ -11,27 +11,40 @@ import uk.ac.wellcome.models.work.internal.{
 object WorkPredicates {
   type WorkPredicate = TransformedBaseWork => Boolean
 
-  val sierraWork: WorkPredicate = identifierTypeId("sierra-system-number")
-  val metsWork: WorkPredicate = identifierTypeId("mets")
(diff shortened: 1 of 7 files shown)


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/WorkPredicates.scala
```diff
@@ -11,27 +11,40 @@ import uk.ac.wellcome.models.work.internal.{
 object WorkPredicates {
   type WorkPredicate = TransformedBaseWork => Boolean
 
-  val sierraWork: WorkPredicate = identifierTypeId("sierra-system-number")
-  val metsWork: WorkPredicate = identifierTypeId("mets")
-  val miroWork: WorkPredicate = identifierTypeId("miro-image-number")
-
+  private val sierraIdentified: WorkPredicate = identifierTypeId(
```
jamieparkinson, line 14: Why private?

jamesgorrie: My logic was they're not intended for use in the Rules.
But I suppose could be if need be.

PR #556, opened 2020-04-23
Filter sensitive fields from Calm data

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4444

## Description

This filters fields which contain personal information.

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmRetriever.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmRetriever.scala
@@ -25,7 +25,8 @@ trait CalmRetriever {
 class HttpCalmRetriever(url: String,
                         username: String,
                         password: String,
-                        concurrentHttpConnections: Int = 2)(
+                        concurrentHttpConnections: Int = 2,
+                        suppressedFields: Set[String] = Set.empty)(
   implicit
   ec: ExecutionContext,
   materializer: ActorMaterializer,
@@ -97,7 +98,7 @@ class HttpCalmRetriever(url: String,
   val summaryResponseParser = new CalmResponseParser[CalmRecord] {
     def parseXml(resp: HttpResponse,
                  bytes: Array[Byte]): Result[CalmXmlResponse[CalmRecord]] =
-      CalmSummaryResponse(bytes, parseTimestamp(resp))
+      CalmSummaryResponse(bytes, parseTimestamp(resp), suppressedFields)
   }
 
   def parseCookie(resp: HttpResponse): Cookie =
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
@@ -75,7 +75,10 @@ object CalmSearchResponse {
       .toEither
 }
 
-case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
+case class CalmSummaryResponse(
+  val root: Elem,
+  retrievedAt: Instant,
+  suppressedFields: Set[String] = Set.empty)
     extends CalmXmlResponse[CalmRecord] {
 
   val responseTag = "SummaryHeaderResponse"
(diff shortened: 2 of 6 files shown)


calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponseTest.scala
```diff
@@ -57,11 +57,11 @@ class CalmXmlResponseTest extends FunSpec with Matchers {
     }
   }
 
-  describe("CALM suummary response") {
+  describe("Calm suummary response") {
```
alexwlchan, line 60: nit, while you're fixing it:

```suggestion
  describe("Calm summary response") {
```

calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
```diff
@@ -80,6 +80,10 @@ case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
 
   val responseTag = "SummaryHeaderResponse"
 
+  val suppressedFields = List(
+    "UserWrapped7"
```
alexwlchan, line 84: Would be good to have a comment here explaining why we're suppressing this particular field. How would we know if we didn't have to suppress this field any more? Or if we had to suppress new fields?

calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmRetriever.scala
```diff
@@ -97,7 +98,7 @@ class HttpCalmRetriever(url: String,
   val summaryResponseParser = new CalmResponseParser[CalmRecord] {
     def parseXml(resp: HttpResponse,
                  bytes: Array[Byte]): Result[CalmXmlResponse[CalmRecord]] =
-      CalmSummaryResponse(bytes, parseTimestamp(resp))
+      CalmSummaryResponse(bytes, parseTimestamp(resp), suppressedFields)
```
alexwlchan, line 101: nit: keyword arguments for readability?

calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
```diff
@@ -75,7 +75,10 @@ object CalmSearchResponse {
       .toEither
 }
 
-case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
+case class CalmSummaryResponse(
+  val root: Elem,
+  retrievedAt: Instant,
+  suppressedFields: List[String] = Nil)
```
alexwlchan, line 81: nit: use `Set[String]` instead of `List[String]`. Better if you're doing `elem in collection` type lookups.

calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
```diff
@@ -120,6 +123,7 @@ case class CalmSummaryResponse(val root: Elem, retrievedAt: Instant)
   def toMapping(nodes: NodeSeq): Map[String, List[String]] =
     nodes
       .map(node => node.label -> node.text)
+      .filterNot { case (name, text) => suppressedFields.contains(name) }
```
alexwlchan, line 126:
```suggestion
      .filterNot { case (name, _) => suppressedFields.contains(name) }
```

calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmXmlResponse.scala
warrd: Obvs `O(n)` vs `O(1)` not gonna be an issue in this case (if there is only a single item in the list as currently I would actually expect the list implementation to be more efficient as is just an equals and doesn't involve hashing). Maybe set is more descriptive of the data we want so happy to change for that reason

alexwlchan: Yeah for sure, I feel it's more a good habit than strictly necessary in this case. I don't feel strongly either way.

warrd: :+1:

PR #560, opened 2020-04-24
searching  collectionPath.path and .label

jamesgorrie:
Adds searching the collection fields.

This is purposefully not very fully featured as it's to start working with the experience team on moving forward.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,4 @@
 import java.io.File
-import java.util.UUID
-
-import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider
 
 def setupProject(
   project: Project,
@@ -226,6 +223,8 @@ lazy val snapshot_generator = setupProject(
 /**
  * To download form the wellcome maven repo in s3 locally, uncomment this
  **/
+//import java.util.UUID
+//import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider
 //s3CredentialsProvider := { _ =>
 //  val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
 //    "arn:aws:iam::760097843905:role/platform-read_only",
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -1,4 +1,7 @@
 import java.io.File
+import java.util.UUID
+
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider
```
warrd, line 4: Can we move these imports down into that comment block too?

build.sbt
```diff
@@ -219,3 +219,16 @@ lazy val snapshot_generator = setupProject(
...
+ **/
+//import java.util.UUID
+//import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider
+//s3CredentialsProvider := { _ =>
+//  val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
+//    "arn:aws:iam::760097843905:role/platform-read_only",
+//    UUID.randomUUID().toString
+//  )
+//  builder.build()
+//}
```
warrd, line 234: Would it work for you this block being formatted like this? Unfortunately my editor doesn't seem to like blocks of line comments in sbt files, so when I select the block and press my uncomment shortcut it doesn't work as expected.

```suggestion
// To download form the wellcome maven repo in s3 locally, uncomment this:
/* import java.util.UUID */
/* import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider */
/* s3CredentialsProvider := { _ => */
/*   val builder = new STSAssumeRoleSessionCredentialsProvider.Builder( */
/*     "arn:aws:iam::760097843905:role/platform-read_only", */
/*     UUID.randomUUID().toString */
/*   ) */
/*   builder.build() */
/* } */
```

jamesgorrie: That breaks intelliJ

PR #562, opened 2020-04-27
Split up Router

jamieparkinson:
There are no meaningful code changes/additions here, but I've moved some things into a new package, `rest`:

- entity-agnostic helper methods that were previously on `Router` live in a new trait, `CustomDirectives`
- the methods that performed the searching/fetching logic for Works now live in `WorksController`
- the works-specific query param parsing classes now live in `WorksParams`

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/Router.scala
@@ -1,9 +1,7 @@
 package uk.ac.wellcome.platform.api
 
-import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success}
+import scala.concurrent.ExecutionContext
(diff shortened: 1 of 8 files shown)


jamesgorrie:
Is this with the idea of using those abstractions in the images API?

jamieparkinson:
Yes, exactly. None of the abstractions are new but it was quite clear to draw a line between the general-purpose and the works-only ones.

warrd:
Nice work, `Router.scala` was getting pretty big and was in need of a refactor.

One thing is the module name `rest` doesn't really make sense to me (the whole API is potentially RESTful, I'm not sure, but the functionality doesn't seem to me to be to do with REST). But then again, I can't really think of any other name beyond `utils` or whatever.

jamieparkinson:
Yeah, I did struggle with the package name and `rest` was as good as I could do. Think I got it from a Spring application I worked with years ago...

PR #565, opened 2020-04-28
Single image API endpoint

jamieparkinson:
Mostly just wiring with a sprinkle of refactoring, sorry about the largish diff

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/ImagesController.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/ImagesController.scala
@@ -12,7 +12,7 @@ import uk.ac.wellcome.platform.api.services.{
   ImagesService
 }
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 
 class ImagesController(
   elasticClient: ElasticClient,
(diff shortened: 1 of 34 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/ImagesController.scala
```diff
@@ -0,0 +1,50 @@
...
+                complete(
+                  ResultResponse(
+                    context = contextUri,
+                    result = DisplayImage(image)
+                  )
+                )
+              )
+            case Right(None) =>
+              Future.successful(notFound(s"Image not found for identifier $id"))
+            case Left(err) => Future.successful(elasticError(err))
```
warrd, line 43: I don't think this needs to be `flatMap` rather than just a `map`? Everything is just returning a `Future.successful(...)`.

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayLocationV2.scala
```diff
@@ -20,20 +20,10 @@ sealed trait DisplayLocationV2
...
-    case PhysicalLocation(locationType, label, accessConditions, _) =>
-      DisplayPhysicalLocationV2(
-        locationType = DisplayLocationType(locationType),
-        label = label,
-        accessConditions = accessConditions.map(DisplayAccessCondition(_))
-      )
+    case digitalLocation: DigitalLocation =>
+      DisplayDigitalLocationV2(digitalLocation)
+    case physicalLocation: PhysicalLocation =>
+      DisplayPhysicalLocationV2(physicalLocation)
```
warrd, line 26: Maybe worth removing the V2s here as well?

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/ImagesController.scala
jamieparkinson: Good point, I'd copied this mindlessly from the WorksController where it can be changed too!

common/display/src/main/scala/uk/ac/wellcome/display/models/v2/DisplayLocationV2.scala
jamieparkinson: Ha, yes, I was scared of doing the models stuff as it was a bit more complex and might be a long thread to pull... I'll give it a go in a separate PR after this?

warrd: :+1: np another PR sounds great

PR #567, opened 2020-04-30
Upgrade libs

alicefuzier:
This is upgrading the catalogue (with the exception of the sierra adapter) to use newest versions of wellcome libraries which in turn have been updated to import newer versions of third party libraries (scalatest, circe, aws, akka, alpakka).

The vast majority of the changes is just different imports due to classses being renamed in the new versions of third party libraries. The main change in this project is that the `SierraTransformable` model has been moved fron `intenral_model` to `sierra_adapter_common` in this PR. This is more in line with how the other adapters work and was necessary to isolate the sierra adapter to update it in a subsequent PR as it's very out of date.


common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/test/fixtures/ElasticsearchFixtures.scala
```diff
@@ -48,18 +48,13 @@ trait ElasticsearchFixtures
 
   // Elasticsearch takes a while to start up so check that it actually started
   // before running tests.
-  eventually {
+  eventually(Timeout(Span(40, Seconds))) {
```
warrd, line 51: Not exactly related to the upgrading of libs, but I think its better to just a sleep the thread a few seconds before sending the request to elasticsearch: by doing it this way we have to wait 40 seconds on a failing test just to find out it has failed.

jamieparkinson:
Oops, sorry about the merge conflicts

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmAdapterWorkerServiceTest.scala
```diff
@@ -43,7 +43,7 @@ class CalmAdapterWorkerServiceTest
     withCalmAdapterWorkerService(retriever, store) {
       case (calmAdapter, QueuePair(queue, dlq), topic) =>
         sendNotificationToSQS(queue, CalmWindow(queryDate))
-        Thread.sleep(500)
+        Thread.sleep(1500)
```
warrd, line 46: If this fixes the tests thats an argument for using `eventually`, but still prefer this way :see_no_evil:

alicefuzier: I'm not sure if it does.. I only see the failure in travis. I can't replicate it in any way locally it seems

alicefuzier:
@warrd yes, I think circe 0.9.0 and 0.13.0 should serialize the same way (I hope)

common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/test/fixtures/ElasticsearchFixtures.scala
alicefuzier: I don't really see the advantage. The timeout here is quite long because elasticsearch docker container is quite slow to startup especially on travis. But if you have it locally already started, this is much faster than a `Thread.sleep` because it will succeed the first time it tries. If you were to put just a `Thread.sleep` you would be forced to make it sleep 40 seconds every time you run the test locally, when in fact the long wait is only needed to cold start in CI

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmAdapterWorkerServiceTest.scala
alicefuzier: RE comment above regarding use of `eventually`, I generally prefer eventually in these cases because you can set a very long timeout for cases where things may be slow in CI, but have it run faster if things are up and responding. `Thread.sleep` always forces you to wait, while `eventually doesn't

PR #568, opened 2020-04-30
Don't display collection as expanded when paths share the same prefix 

warrd:
Currenty when expanding a path, other paths which contain the the expanded path as a prefix will also display as expanded.

For example, when expanding `PP/CRI/D/2/41`, `PP/CRI/D/2/4` will also be displayed as expanded.

This fixes this issue.


common/display/src/test/scala/uk/ac/wellcome/display/models/DisplayCollectionTest.scala
```diff
@@ -184,4 +185,38 @@ class DisplayCollectionTest extends FunSpec with Matchers with WorksGenerators {
         )
       )
   }
+
+  it("doesn't set expanded to paths sharing the same prefix as expanded") {
```
alexwlchan, line 189: I’m confused – how does this test prove the issue is fixed?

Is it because `a/b` isn't expanded (`children` is `None`)? Wouldn’t that happen with the old behaviour?

If you were looking at `a/b`, you'd previously see `a/b0` expanded as well, but I'm not sure the inverse is true.

warrd: > Is it because a/b isn't expanded (children is None)? Wouldn’t that happen with the old behaviour?

Nope the old behaviour children was be `Some(List())` rather than `None`. Does that make sense?

alexwlchan: Ah, yeah, I think I'd misunderstood the original issue. Got it!

PR #569, opened 2020-05-01
Logging RFC - worker module and calm adapter

warrd:
Includes a new `worker` module with the new infrastructure setup, which will be used in all the catalogue applications using firelens / fluent bit.

The calm adapter is the only app so far to use this (with the terraform already have been applied), but should be pretty straightforward to do for other apps once this is in.

### DIFF
--- a/calm_adapter/terraform/worker.tf
+++ b/calm_adapter/terraform/worker.tf
@@ -1,0 +1,19 @@
+module "worker" {
+  source = "../../infrastructure/modules/worker"
+
+  name = local.namespace
+
+  image              = local.calm_adapter_image
+  env_vars           = local.env_vars
+  secret_env_vars    = local.secret_env_vars
+  min_capacity       = 1
+  max_capacity       = 2
+  desired_task_count = 1
+  cpu                = 512
+  memory             = 1024
+
+  cluster_name = local.namespace
+  cluster_arn  = aws_ecs_cluster.cluster.arn
+  namespace_id = aws_service_discovery_private_dns_namespace.namespace.id
+  subnets      = local.private_subnets
+}
--- a/infrastructure/modules/worker/terraform.tf
+++ b/infrastructure/modules/worker/terraform.tf
@@ -2,34 +2,34 @@ module "service" {
   source = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/service?ref=v2.4.1"
 
   task_definition_arn            = module.task_definition.arn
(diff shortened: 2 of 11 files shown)


kenoir:
Looks good so far.

calm_adapter/terraform/worker.tf
```diff
@@ -0,0 +1,19 @@
...
+  source = "../../infrastructure/modules/worker"
+
+  name = local.namespace
+
+  image              = local.calm_adapter_image
+  env_vars           = local.env_vars
+  secret_env_vars    = local.secret_env_vars
+  min_capacity       = 1
+  max_capacity       = 2
+  desired_task_count = 1
```
alexwlchan, line 11: Why doesn't the CALM adapter scale down to 0 when it's not running?

infrastructure/modules/worker/terraform.tf
```diff
@@ -0,0 +1,79 @@
...
+  source = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/autoscaling?ref=v2.4.1"
+
+  name = var.name
+
+  cluster_name = var.cluster_name
+  service_name = var.name
+
+  min_capacity = var.min_capacity
+  max_capacity = var.max_capacity
+}
```
alexwlchan, line 23: Does the worker need to export the scale_high/scale_low actions in `outputs.tf`? Otherwise how can you attach this to a queue?

calm_adapter/terraform/worker.tf
warrd: This is what I did previously so that's how I set it up.

In practice it looks like we have `desired_task_count = 1` for pretty much all our catalogue applications:

https://github.com/wellcomecollection/catalogue/blob/master/pipeline/terraform/modules/worker/variables.tf#L29-L32
https://github.com/wellcomecollection/catalogue/blob/master/sierra_adapter/terraform/modules/scaling_worker/variables.tf#L28-L30

(we don't explicitly set any of these when using the modules).

infrastructure/modules/worker/terraform.tf
warrd: Ah ok. Am I right that we would need to attach the scaling policy to a queue for `desired_task_count = 0` to work? Otherwise I don't see how it would know work has to be performed and scale up.

alexwlchan: Yeah, they need to be attached to queues to do anything useful.

PR #573, opened 2020-05-04
add description field from calm

jamesgorrie:
Ref https://github.com/wellcomecollection/platform/issues/4177

### DIFF
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
@@ -50,18 +50,16 @@ object CalmTransformer
 
   def shouldSuppress(record: CalmRecord): Boolean =
     record
-      .get("Transmission")
-      .map { value =>
-        value.toLowerCase match {
-          case "no"  => true
-          case "yes" => false
-          case _ =>
-            info(
-              s"Unrecognised value for Transmission field; assuming 'Yes': $value")
-            false
-        }
+      .get("Transmission").exists { value =>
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -196,19 +196,34 @@ object CalmTransformer
...
   def physicalDescription(record: CalmRecord): Option[String] =
     (record.getList("Extent") ++ record.getList("UserWrapped6")) match {
       case Nil  => None
       case strs => Some(strs.mkString(" "))
     }
 
+  def production(record: CalmRecord): List[ProductionEvent[Unminted]] = {
+    record.getList("Date") match {
+      case Nil => Nil
+      case list => List(ProductionEvent(
```
warrd, line 211: nit: would be clearer to me if the variable name here was `dates` rather than naming it after the type

jamesgorrie: This function gets the production event if there is dates. I could have two function both `production` and `dates` but feel if I don't use them outside of this context that's a useless abstraction?

warrd: Not saying there should be another function just that  `list` isn't a great variable name, and `dates` is a better description of what it is

jamesgorrie: Misread, thought you meant the method name.

PR #580, opened 2020-05-06
make id optional in language

jamesgorrie:
We currently have strings coming from the source data that isn't parsable into a language code.

We still want these to come through, but as "unstructured". This makes the `code` / `id` optional.


common/internal_model/src/test/scala/uk/ac/wellcome/models/work/internal/LanguageTest.scala
```diff
@@ -0,0 +1,25 @@
+package uk.ac.wellcome.models.work.internal
+import org.scalatest.{FunSpec, Matchers}
+
+class LanguageTest extends FunSpec with Matchers {
+  it("get's labels for known code or errors") {
```
jamieparkinson, line 5:
```suggestion
  it("gets labels for known code or errors") {
```

jamesgorrie:
> What does this do to aggregations?

Nothing - the data structure is exactly the same. We just allow data to now no include `ids`. The non-existant ids should all be lumped into 1 with a subaggregation of label being created from that.

PR #581, opened 2020-05-06
Image search endpoint

jamieparkinson:
Apologies for the big diff; although quite a lot of that is moving stuff around / extracting stuff. This doesn't match the RFC quite yet; some model changes relating to that will come in my next PR.


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/Responses.scala
```diff
@@ -35,75 +39,59 @@ object ResultResponse {
 
 @Schema(
   name = "ResultList",
-  description = "A paginated list of works."
+  description = "A paginated list of documents."
```
jamesgorrie, line 42: Documents is very Elastic...
```suggestion
  description = "A paginated list of results."
```
❓

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesService.scala
```diff
@@ -31,4 +45,42 @@ class ImagesService(searchService: ElasticsearchService)(
         }
       }
 
+  def listOrSearchImages(index: Index, searchOptions: ImagesSearchOptions)
+    : Future[Either[ElasticError, ResultList[AugmentedImage, Unit]]] =
+    searchService
+      .executeSearch(
+        queryOptions = toElasticsearchQueryOptions(searchOptions),
+        index = index,
+        scored = searchOptions.searchQuery.isDefined
```
jamesgorrie, line 54: I wonder if this could just be calculated when we create the searchOptions, keeping it at the edge e.g. (bad code)
```
val searchOptions = SearchOptions(
  // ...
  sort = queryParams.sort.getOrElse(params.query match {
   case Some(a) => ScoreSort
   case None => IdSort
  })
)
```

That way we could also sort on a non-queried list. e.g. `images?sort=colour`.

jamieparkinson: Hmmm, I don't really want to "lie" about where the search options are from - they should reflect what the user asked for and no more. 

This is only a refactor of the old `listOrSearchWorks` code. Regardless of query, we always sort by ID, and when there's a query we sort by score as well - that's why those sorts live inside the request builder. 

There's nothing stopping us sorting on a list other than that we don't have anything to sort on yet - you can see an example of us doing that in the works request builder.

api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
```diff
@@ -156,6 +172,20 @@ final case class IdQuery(q: String) extends ElasticsearchPartialQuery {
     )
 }
 
+final case class ImageIdQuery(q: String) extends ElasticsearchPartialQuery {
+  lazy val idFields = Seq(
+    "id.canonicalId.text",
+    "id.sourceIdentifier.value.text",
+    "parentWork.text"
+  )
```
warrd, line 180: Having this defined as `lazy` is a little confusing , as it is currently a just constant that will be included as some bytes in the jar

jamieparkinson: Yeah, I did it to keep consistency with the other classes but I think I'll remove it from all of them 👍

jamesgorrie: No idea why I'd done this...

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesService.scala
jamesgorrie: Hmm. I thought we had the `SearchParams` which then mapped to `SearchOptions`. page is a good example of this where `Option[String] => Int`.

Yea, it's something that has bothered from the `listOrSearch` method.

I suppose my main annoyance is it's not closer to the edge, either the Elastic client of the Request client, but rather sitting in this service in the middle.

> other than that we don't have anything to sort on yet

With works we already have date search, which _should_ work on search or list.

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/ImagesParams.scala
```diff
@@ -9,3 +12,42 @@ object SingleImageParams extends QueryParamsUtils {
...
+  def searchOptions(apiConfig: ApiConfig): ImagesSearchOptions =
+    ImagesSearchOptions(
+      searchQuery = query.map(SearchQuery(_)),
+      filters = filters,
+      pageSize = pageSize.getOrElse(apiConfig.defaultPageSize),
+      pageNumber = page.getOrElse(1)
+    )
+
+  private def filters: List[ImageFilter] =
+    List(license).flatten
```
warrd, line 34: https://www.scala-lang.org/api/current/scala/Option.html#toList:List[A]

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchRequestBuilder.scala
```diff
@@ -1,159 +1,12 @@
...
-    }
-  }
+import com.sksamuel.elastic4s.requests.searches.sort.FieldSort
+import com.sksamuel.elastic4s.Index
+
+trait ElasticsearchRequestBuilder {
+  def request(queryOptions: ElasticsearchQueryOptions,
+              index: Index,
+              scored: Boolean): SearchRequest
+  val idSort: FieldSort
```
warrd, line 11: There is also a `CollectionRequestBuilder`, not sure whether fits this interface or not but potentially could extend from this too

jamieparkinson: Good point, I'll see whether that can be done

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesService.scala
jamieparkinson: > With works we already have date search, which should work on search or list.

We already do that, no?

jamesgorrie: [Yip](https://github.com/wellcomecollection/catalogue/pull/581/files#diff-6068495eb1eb5651cc269501889f73b8R26-R30). I was just looking at the image implementation that didn't allow for that (rightfully as we don't have another sort).

The main objection again is having three places where sort is configured, instead of potentially 2 or 1.

jamesgorrie: In essence maybe moving the scored into the `executeSearch` method.

Having an argument called `scored` defined by if there is a `query` that affects `sort` is a little hard to figure out at first glance.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchRequestBuilder.scala
jamieparkinson: Unfortunately this doesn't really work and I'm not sure it's worth trying to make it fit

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesService.scala
jamieparkinson: Yeah, agreed - had wanted it to be explicit to the caller that there was some extra stuff going on but to be honest it's slightly surprising that you have to ask to sort by score anyway

PR #583, opened 2020-05-07
Upgrade libraries in sierra_adapter

alicefuzier:


project/Dependencies.scala
```diff
@@ -365,8 +310,8 @@ object CatalogueDependencies {
   // Snapshots stack
 
   val snapshotGeneratorDependencies: Seq[ModuleID] =
-    ExternalDependencies.newMockitoDependencies ++
-      WellcomeDependencies.newMessagingTypesafeLibrary ++
-      WellcomeDependencies.newStorageLibrary ++
+    ExternalDependencies.mockitoDependencies ++
+      WellcomeDependencies.messagingTypesafeLibrary ++
+      WellcomeDependencies.storageLibrary ++
```
alexwlchan, line 315: 😍 

So much cleaner!

sierra_adapter/common/src/main/scala/uk/ac/wellcome/sierra_adapter/config/builders/SierraTransformableVHSBuilder.scala
```diff
@@ -1,33 +1,21 @@
...
+    config: Config): VersionedStore[String, Int, SierraTransformable] = {
 
-    new SierraVHS(
-      vhsConfig = VHSBuilder.buildVHSConfig(config),
-      objectStore = ObjectStore[SierraTransformable],
-      dynamoDbClient = DynamoBuilder.buildDynamoClient(config)
+    new VersionedStore(
+      new VHSWrapper(
+        VHSBuilder.build[SierraTransformable](config)
+      )
```
alexwlchan, line 18: FWIW, in the Calm adapter and transformer, this method is inlined in Main rather than having a single-use class – ah, but I’m guessing this is used in multiple places?

sierra_adapter/common/src/test/scala/uk/ac/wellcome/sierra_adapter/utils/SierraAdapterHelpers.scala
```diff
@@ -1,89 +1,61 @@
...
+//
+//  def assertStoredAndSent(transformable: SierraTransformable,
+//                          topic: Topic,
+//                          table: Table): Assertion =
+//    assertStoredAndSent[SierraTransformable](
+//      transformable,
+//      id = transformable.sierraId.withoutCheckDigit,
+//      topic = topic,
+//      table = table
+//    )
```
alexwlchan, line 60: Are you going to use these assertion helpers again? If not, can they be deleted rather than just commented out?

warrd:
Do you know what the payload for the message put on the transformer queue looks like? In the previous version of the adapter it puts a different sort of message to other interservice messages (due to being an older version of `scala-messaging`), hence some weird backwards compatabilty stuff in the transformer. I think this needs to be changed on the transformer side at the same time?

https://github.com/wellcomecollection/catalogue/blob/master/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/services/HybridRecordReceiver.scala

alicefuzier:
Ah yes, good point, the transformer doesn't need to use the backwards compatible thing anymore. I'll change that. The new adapter sends just the VHS key and then it would be up to the transformer to look that up

sierra_adapter/common/src/main/scala/uk/ac/wellcome/sierra_adapter/config/builders/SierraTransformableVHSBuilder.scala
alicefuzier: yes, it's used in 2 places

sierra_adapter/common/src/main/scala/uk/ac/wellcome/sierra_adapter/config/builders/SierraVHSBuilder.scala
```diff
@@ -0,0 +1,20 @@
...
+  def buildSierraVHS[T](config: Config)(
+    implicit codec: Codec[T]): VersionedStore[String, Int, T] = {
+
+    new VersionedStore(
+      new VHSWrapper(
+        VHSBuilder.build[T](config)
+      )
+    )
+  }
+}
```
warrd, line 20: :+1:

PR #584, opened 2020-05-07
Align images endpoints with RFC

jamieparkinson:


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/ImagesParams.scala
```diff
@@ -43,7 +43,7 @@ object MultipleImagesParams extends QueryParamsUtils {
         "page".as[Int].?,
         "pageSize".as[Int].?,
         "query".as[String].?,
-        "license".as[LicenseFilter].?,
+        "locations.license".as[LicenseFilter].?,
```
jamesgorrie, line 46: This should be `items.locations.license` as we try to replicate the JSON path as close as possible.

[There's a small comment in this PR talking about defaulting to `id`or `value`](https://github.com/wellcomecollection/catalogue/pull/586).

jamesgorrie: Wait, I read this wrong, I thought we were in works land, ignore.

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayImageSource.scala
```diff
@@ -0,0 +1,28 @@
...
+  name = "ImageSource",
+  description = "A description of the entity from which an image was sourced"
+)
+case class DisplayImageSource(
+  @Schema(
+    description = "Identifer of the image source"
+  ) id: String,
+  @JsonKey("type") @Schema(
+    name = "type",
+    description = "What kind of source this is"
```
jamesgorrie, line 17: Implying multiple kinds are possible? Although this is a very odd field to describe.

jamesgorrie: Yes, there could be multiple sources.... got it.

PR #586, opened 2020-05-12
add identifiers filter

jamesgorrie:
Fixes https://github.com/wellcomecollection/platform/issues/4507

Thought about the filter being `identifiers.value` but it felt wrong. @jtweed what do we think about this? If a path isn't given could we assume it means `.id` or `.value`?

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
@@ -255,6 +255,12 @@ trait MultipleWorksSwagger {
         required = false
       ),
       new Parameter(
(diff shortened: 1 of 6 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
```diff
@@ -254,6 +254,12 @@ trait MultipleWorksSwagger {
         description = "Filter the work by subject.",
         required = false
       ),
+      new Parameter(
+        name = "identifiers",
+        in = ParameterIn.QUERY,
+        description = "Filter the work by identifiers.",
```
jamesgorrie, line 260: This is copy pasted - might look at the language in the API docs separately.

PR #587, opened 2020-05-12
add reasons to InvisibleWorks

jamesgorrie:
Creates more defined application states on the Calm Transformer and uses reasons on Invisible works to identify why they are invisible.

This is pass 1 - and the next stage is to use another type of `UntransformableWork`. I attempted this but it was becoming a large refactor, which I would prefer to not do given that:
* We are trying to still deploy the pipeline and this helps with that
* This will also allow us to evaluate the usefulness of this feature


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -91,7 +92,8 @@ case class IdentifiedInvisibleWork(
   canonicalId: String,
   version: Int,
   sourceIdentifier: SourceIdentifier,
-  data: WorkData[Minted, Identified]
+  data: WorkData[Minted, Identified],
+  reasons: List[Reason] = Nil,
```
jamieparkinson, line 96: Should the field name be a bit more descriptive? Other fields make sense by themselves (ie "the version of a work", "the sourceIdentifier of a work", etc etc) but "the reasons of a work" is confusing. Maybe like "suppressionReasons" or something?

jamesgorrie: Nice point. It was "Why is this work invisible?" and made sense on `InvisibleWork` rather than `Work`.

Trying to think of a nice set of terms such as `InvalidWork` and `invalidations`, but those aren't right as it's not that they invalid.

* `InvisibleWork.reasonsInvisible`
* `InvisibleWork.invisibilityReasons`
* `InvisibleWork.invisibilityCloaks`
* `HiddenWork.hiddenReasons`
* `HiddenWork.reasonsHidden`
* `HiddenWork.hidingFrom`
* `ShadyWork.shadows`

If I want to keep refactoring down, I would probably stick to the `invisibilityReasons`, and that can be because it's suppressed by the source system.

jamesgorrie:
Tests have passed

PR #590, opened 2020-05-13
add pipeline back and make errors use JSON

jamesgorrie:
While debugging this is useful. -means I won't have to go lookup the json, just copy it from the logs.

### DIFF
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -1,3 +1,67 @@
+module "catalogue_pipeline_20200414" {
+  source = "./stack"
+
+  namespace = "catalogue-20200414"
+
+  release_label = "prod"
+
+  account_id      = data.aws_caller_identity.current.account_id
+  aws_region      = local.aws_region
+  vpc_id          = local.vpc_id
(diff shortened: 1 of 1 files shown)


warrd:
I'm not sure this can be applied, due to the fact that all the underlying infrastructure has changed?

jamesgorrie:
@warrd I've already reapplied it as I tore it down by accident.

warrd:
Ah so it's a new pipeline, just pointing to the old indices?

jamesgorrie:
[The current API points to `04`](https://github.com/wellcomecollection/catalogue/blob/master/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticConfig.scala#L15)

So it needs to stay up and running till we fix the new one and move the API over.

I tore down the old pipeline by accident - so just making sure I don't do it again until the new one is prod and the API pointing at it.

jamesgorrie:
Removed the logging as it wasn't compiling due to not having the right encoders, didn't fancy mucking around with it too much.

PR #591, opened 2020-05-13
Update API for logging RFC

warrd:
At kenoirs suggestion i have created a whole new module rather than editing the existing catalogue_api_stack module. this is a much better idea from a deployment perspective (e.g. to avoid situations like with the pipeline where we had to check out an old revision to bring back the old stack), but doesnt have a nice diff showing just what has changed, so i have made it so the first commit is just copying the original and the subsequent ones are the actual changes

### DIFF
--- a/api/terraform/catalogue_api_staging/main.tf
+++ b/api/terraform/catalogue_api_staging/main.tf
@@ -1,5 +1,5 @@
 module "catalogue_api_staging" {
-  source = "../modules/catalogue_api_stack"
+  source = "../modules/new_catalogue_api_stack"
 
   environment        = "staging"
   domain_name        = "catalogue.api-stage.wellcomecollection.org"
--- a/api/terraform/modules/new_catalogue_api_stack/domains.tf
+++ b/api/terraform/modules/new_catalogue_api_stack/domains.tf
@@ -1,0 +1,35 @@
+/*data "aws_route53_zone" "dotorg" {
+  provider = aws.routemaster
+
+  name = "wellcomecollection.org."
+}*/
+
+locals {
+  # This is the Zone ID for wellcomecollection.org in the routemaster account.
+  # We can't look this up programatically because the role we use doesn't have
+  # the right permissions in that account.
+  route53_zone_id = "Z3THRVQ5VDYDMC"
+}
+
+resource "aws_api_gateway_domain_name" "catalogue_api" {
+  domain_name              = var.domain_name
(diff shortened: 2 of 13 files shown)


api/terraform/modules/catalogue_api_stack/service/main.tf
```diff
@@ -31,18 +43,20 @@ module "task_definition" {
     apm_secret     = "catalogue/api/apm_secret"
   }
 
-  sidecar_container_image = var.nginx_container_image
-  sidecar_container_port  = var.nginx_container_port
-  sidecar_container_name  = local.sidecar_container_name
-  sidecar_cpu             = 512
-  sidecar_memory          = 1024
+  sidecar_name   = local.sidecar_container_name
+  sidecar_image  = var.nginx_container_image
```
kenoir, line 47: You should use the nginx module for this if you can, so as to keep a consistent image for API Gateway backed services across all services.

infrastructure/modules/worker_with_sidecar/main.tf
```diff
@@ -10,6 +10,9 @@ module "service" {
   desired_task_count             = var.desired_task_count
   security_group_ids             = var.security_group_ids
   use_fargate_spot               = var.use_fargate_spot
+  container_name                 = var.container_name
```
kenoir, line 13: I would create a specific "api" module following the pattern in storage: https://github.com/wellcomecollection/storage-service/tree/master/terraform/modules/service/api

warrd: My instinct was to abstract away common functionality into one of these workers so that there isn't loads of duplication, but do you think that isn't such an issue with our terraform?

Also, I don't know if creating a new module is necessary, couldn't we just stick everything in the [current catalogue API module](https://github.com/wellcomecollection/catalogue/tree/master/api/terraform/modules/catalogue_api_stack) seeing as thats the only place it is going to be used for now?

kenoir: I think using another module with some duplication makes it clearer what makes up the service. I use the term "worker" to refer to a queue backed service, so it feels wrong.

> Also, I don't know if creating a new module is necessary, couldn't we just stick everything in the current catalogue API module seeing as thats the only place it is going to be used for now?

That would work.

My preference would be to create a new module anyway as part of the process of replacing the old one: Create new module -> create an instance of the new module -> replace old running instance with new instance -> delete old instance & module.

PR #599, opened 2020-05-19
Decouple API service from routing

warrd:
This will allow us to deploy a new API stack listening on a new port, then switch traffic from `api.wellcomecollection.org/catalogue` to this new stack, thus deploying without modifying any running service in place.

The changes here are already applied, and took a fair amount of terraform state trickery to move the resources.


api/terraform/modules/routing/main.tf
```diff
@@ -33,3 +37,34 @@ resource "aws_route53_record" "catalogue_api" {
...
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_api_gateway_stage" "stage" {
+  stage_name    = var.environment
+  rest_api_id   = var.api_id
+  deployment_id = aws_api_gateway_deployment.stage.id
+  variables     = local.variables
```
alexwlchan, line 69: Any reason to refer to a local defined elsewhere rather than inline the definition?

warrd: This is how it was before. I'm guessing as this definition was used twice, but I agree I think fine to inline.

api/terraform/modules/routing/main.tf
```diff
@@ -33,3 +33,38 @@ resource "aws_route53_record" "catalogue_api" {
...
+  # If we specify the stage name here then API Gateway tries to create it even
+  # if it already exists (from below). Setting it to an empty string prevents this.
+  # See https://github.com/terraform-providers/terraform-provider-aws/issues/2918#issuecomment-356684239
+  stage_name = ""
+
+  variables = {
+    port = var.listener_port
+  }
+
+  stage_description = filemd5("${path.module}/../../shared/gateway.tf")
```
jamesgorrie, line 56: What is this useful for?

warrd: I'm not sure, copied from before

jamesgorrie: @kenoir ?

PR #600, opened 2020-05-19
Decrease maximum inferrer count

jamieparkinson:
Loris latency (particularly the p99) spikes with the addition of the last couple of workers so this should relieve the load a bit (and probably won't work out all that much slower)


jamesgorrie:
>  spikes with the addition of the last couple of workers 

Do we have any idea as to why?

alexwlchan:
Add a comment about why it's this low?

PR #601, opened 2020-05-20
Move service discovery namespace to shared API terraform

warrd:


api/terraform/catalogue_api_staging/main.tf
```diff
@@ -2,14 +2,16 @@ module "catalogue_api_staging" {
   source = "../modules/new_catalogue_api_stack"
 
   environment        = "staging"
-  listener_port      = 8080
+  listener_port      = 1234
```
jamesgorrie, line 5: Any reason to change this?

warrd: Shouldn't particularly be in this PR, but it is changed temporarily as I have spun up the new prod stack which is listening on that port. Once I am confident it is working (it will be accessible via the staging URL), it will be switched back.

kenoir: AFAIK it's temporary to allow for safe deployment of the new api infra.

PR #606, opened 2020-05-27
Filter by access status

warrd:
https://github.com/wellcomecollection/platform/issues/4529

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
@@ -8,27 +8,6 @@ sealed trait DocumentFilter
 sealed trait WorkFilter extends DocumentFilter
 sealed trait ImageFilter extends DocumentFilter
 
-trait IncludeExcludeFilter[T] {
-  val include: List[T]
-  val exclude: List[T]
-
-  import com.sksamuel.elastic4s.ElasticDsl._
-
-  def toElasticString(obj: T): String
-
-  def toElasticQuery(field: String) =
-    (include.map(toElasticString(_)), exclude.map(toElasticString(_))) match {
-      case (include, Nil) =>
-        termsQuery(field = field, values = include)
-      case (Nil, exclude) =>
-        not(termsQuery(field = field, values = exclude))
-      case (include, exclude) =>
-        must(
-          termsQuery(field = field, values = include),
-          not(termsQuery(field = field, values = exclude)))
-    }
-}
-
 case class ItemLocationTypeFilter(locationTypeIds: Seq[String])
     extends WorkFilter
 
@@ -56,10 +35,6 @@ case class CollectionPathFilter(path: String) extends WorkFilter
 
 case class CollectionDepthFilter(depth: Int) extends WorkFilter
 
-case class AccessStatusFilter(val include: List[AccessStatus],
-                              val exclude: List[AccessStatus])
-    extends WorkFilter
-    with IncludeExcludeFilter[AccessStatus] {
-  def toElasticString(status: AccessStatus) =
-    status.getClass.getSimpleName.stripSuffix("$")
-}
+case class AccessStatusFilter(includes: List[AccessStatus],
+                              excludes: List[AccessStatus])
+  extends WorkFilter
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
@@ -60,5 +60,6 @@ case class AccessStatusFilter(val include: List[AccessStatus],
                               val exclude: List[AccessStatus])
     extends WorkFilter
     with IncludeExcludeFilter[AccessStatus] {
-  def toElasticString(status: AccessStatus) = status.getClass.getSimpleName
+  def toElasticString(status: AccessStatus) =
+    status.getClass.getSimpleName.stripSuffix("$")
 }
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
@@ -1,10 +1,33 @@
 package uk.ac.wellcome.platform.api.models
 
+import uk.ac.wellcome.models.work.internal.AccessStatus
+
 import java.time.LocalDate
 
 sealed trait DocumentFilter
 sealed trait WorkFilter extends DocumentFilter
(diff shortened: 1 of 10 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
```diff
@@ -1,11 +1,34 @@
...
+      case (include, Nil) =>
+        termsQuery(field = field, values = include)
+      case (Nil, exclude) =>
+        not(termsQuery(field = field, values = exclude))
+      case (include, exclude) =>
+        must(
+          termsQuery(field = field, values = include),
+          not(termsQuery(field = field, values = exclude)))
+    }
+}
```
jamesgorrie, line 30: Generally we've kept all the Elastic stuff on the edge of where we interact with Elastic, leaving the responsibility of the conversion up to the elasticsearch code.

[`AggregationsRequest`](https://github.com/wellcomecollection/catalogue/blob/master/common/display/src/main/scala/uk/ac/wellcome/display/models/AggregationRequest.scala) and [`SortRequest`](https://github.com/wellcomecollection/catalogue/blob/master/common/display/src/main/scala/uk/ac/wellcome/display/models/SortRequest.scala) are examples of this.

Just a note that often I have found doing things in multiple ways make it hard to fathom which way to do it, or not do it, in a few months.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala
```diff
@@ -153,7 +153,9 @@ object WorksRequestBuilder extends ElasticsearchRequestBuilder {
             field = "data.otherIdentifiers.value",
             values = identifiers)
         )
-
+      case filter @ AccessStatusFilter(_, _) =>
+        filter.toElasticQuery(
```
jamesgorrie, line 157: We could just have a method here that converts the `IncludeExcludeFilter` into the query, rather than having it as part of the model to keep the separation clear, or at least consistent.

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
```diff
@@ -48,23 +48,38 @@ trait QueryParamsUtils extends Directives {
...
-          Left(invalidValuesMsg(invalidValues, values.map(_._1).toList))
-      }
+      mapStringsToValues(strs, mapping).left
+        .map { invalidStrs =>
+          invalidValuesMsg(invalidStrs, validStrs)
+        }
     }
+  }
+
+  def decodeIncludesAndExcludes[T](
```
jamesgorrie, line 62: Is it worth having a test for this as it's pretty self explanatory but has a few moving parts.

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
```diff
@@ -48,23 +48,38 @@ trait QueryParamsUtils extends Directives {
...
+    values: (String, T)*): Decoder[(List[T], List[T])] = {
+    val mapping = values.toMap
+    val validStrs = values.map(_._1).toList
+    decodeCommaSeparated
+      .emap { strs =>
+        val (excludeStrs, includeStrs) = strs.partition(_.startsWith("!"));
+        val includes = mapStringsToValues(includeStrs, mapping);
+        val excludes = mapStringsToValues(excludeStrs.map(_.tail), mapping);
```
jamesgorrie, line 70: Not too fussed but would `_.stripSuffix("!")` be more clear than `.tail`?

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
warrd: Do you mean a test of just the decoder itself? We tend to just test the actual API functionality it provides (which I have done), but having specific unit tests for the decoders isn't a bad idea.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
warrd: Ok, will see what I can do

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
jamesgorrie: Yea, for the decoder. 

If you give it rubbish, it gives you 🗑️ back etc.

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
warrd: `tail` works for me, and means it's all on one line :see_no_evil:... but can do the `stripPrefix` if you think `tail` is confusing?

warrd: `stripPrefix` will actually do unnecessary work (checking the string starts with `!` when we know that all of them do), but I don't suppose that matters

jamesgorrie: Na, not really, I was probably just reading line by line rather than the whole method, whatever works for you.

jamesgorrie: 🚀 performance!

api/api/src/test/scala/uk/ac/wellcome/platform/api/rest/QueryParamsTest.scala
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    it("should fail decoding when unrecognised values") {
+      val result = decode("b,c,!d")
+      result shouldBe a[Left[_, _]]
+      result.left.get.message shouldBe
+        "'c', 'd' are not valid values. Please choose one of: ['a', 'b']"
+    }
+  }
+}
```
jamesgorrie, line 42: nice!

warrd:
Not sure I completely understand what you mean but will give answering this a go.

The serialised representation of a `AccessStatus` such as `Open` will be

```json
{"type": "Open"}
```

There is no `id` field in question here (whether this is the correct representation in elasticsearch is a different question, if we are to change it I would probably make it just a string rather than getting an `id field involved).

So we pass in a string `open` but need to convert this to `Open` to filter on the `type` field. The confusion might be to do with the fact that we have a display model for `AccessStatus`, where we do use the `kebab-case` representation: using this representation makes sense for the query parameter as thats what the user of the API ends up seeing, but we still need to convert this to the internal model.

PR #607, opened 2020-05-28
Sierra transformer receives key

alicefuzier:
Modifies the sierra transformer to accept a VHS key instead of the whole Hybrid record and to look up the record


alicefuzier:
Closing in as changes in #583  have to go in first

PR #609, opened 2020-05-29
20200528 pipeline

alicefuzier:


jamesgorrie:
Should we add it so we don't demolish the other one?

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticConfig.scala
```diff
@@ -12,7 +12,7 @@ object ElasticConfig {
   // i.e. The API and the snapshot generator.
   def apply(): ElasticConfig =
     ElasticConfig(
-      worksIndex = Index("works-20200513"),
-      imagesIndex = Index("images-20200513")
+      worksIndex = Index("works-20200528"),
+      imagesIndex = Index("images-20200528")
```
alexwlchan, line 16: Shouldn't these be a config flag that we can update in Terraform, so changing the index doesn't require a whole new app?

jamesgorrie: This was to avoid us deploying a version of the app that might not work with the corresponding index in the config. Well, it goes a bit of the way there.

I think there is a [bigger discussion to be had here](https://github.com/wellcomecollection/platform/issues/4197) as I think this might go back to configuration but as part of a deploy process that makes sure all bases are covered.

jamieparkinson: I see this as "if we have a new index, we probably have a new mapping, so we definitely want a new app" - as @jamesgorrie says it was a deliberate decision which I think has caused fewer issues so far than the old tf variable approach.

jamesgorrie: > has caused fewer issues so far than the old tf variable approach.

💯

PR #612, opened 2020-05-29
merge available works sent from matcher

jamesgorrie:
Sometimes the graph of IDs from the matcher will have works that haven't been stored in the recorder.

This ignores those  

- [ ] Make sure we aren't throwing an error, but rather returning `None` when something doesn't exist.


jamieparkinson:
What's the status of this? I remember thinking it seemed like a very good idea...

jamesgorrie:
I was waiting to get the pipeline healthy again.

Another issue was trying to figure out why things were not being places on the DLQ when they were missing [as the code suggests it should](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/RecorderPlaybackService.scala#L57). More investigation to happen first.

PR #613, opened 2020-06-01
Suppress resource guides from Calm

warrd:
https://github.com/wellcomecollection/platform/issues/4446

### DIFF
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
@@ -85,6 +85,18 @@ object CalmTransformer
     }
 
   def shouldSuppress(record: CalmRecord): Boolean =
+    transmissionIsNo(record) match {
+      case true  => true
+      case false =>
+        // Records prefixed with AMSG (Archives and Manuscripts Resource Guides)
+        // are not actual archives but instead guides for researchers, so we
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -85,6 +85,19 @@ object CalmTransformer
     }
 
   def shouldSuppress(record: CalmRecord): Boolean =
+    transmissionIsNo(record) match {
+      case true  => true
+      case false =>
+        // Records prefixed with AMSG (Archives and Manuscripts Resource Guides)
+        // are not actual archives but instead guides for researches, so we
```
jamieparkinson, line 92:
```suggestion
        // are not actual archives but instead guides for researchers, so we
```

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -85,6 +85,19 @@ object CalmTransformer
...
+        // suppress them here.
+        record
+          .get("RefNo")
+          .map(_.startsWith("AMSG"))
+          .getOrElse(false)
```
jamieparkinson, line 97: Trivia: you can express this boolean in one statement rather than using `getOrElse`
```suggestion
          .exists(_.startsWith("AMSG"))
```

PR #615, opened 2020-06-02
Use title page for thumbnail if avaialble in METS data

warrd:
https://github.com/wellcomecollection/platform/issues/4266

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
@@ -154,6 +154,7 @@ case class MetsXml(root: Elem) {
         )
     }
 
+  /** Returns the physical ID for the TitlePage element if it exists. */
   def titlePageId: Option[String] =
     logicalStructMapForType
       .collectFirst { case (id, "TitlePage") => id }
@@ -260,7 +261,7 @@ case class MetsXml(root: Elem) {
     *
     *  For this input we would expect the following output:
     *
-    *  Map("LOG_0000" -> "b22012692_0001.xml",
+    *  Seq("LOG_0000" -> "b22012692_0001.xml",
     *      "LOG_0002" -> "b22012692_0003.xml")
     */
   private def logicalStructMapForMultipleManifestations: Seq[(String, String)] =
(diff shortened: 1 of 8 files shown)


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala
```diff
@@ -134,6 +133,37 @@ case class MetsXml(root: Elem) {
     }
   }
 
+  /** Here we use the the items defined in the physicalStructMap to look up
+    * file IDs in the (normalised) fileObjects mapping
+    */
+  def fileReferencesMapping(bnumber: String): List[(String, FileReference)] =
```
jamieparkinson, line 139: Might be missing something obvious here but why are we using these tuples/mappings rather than adding the `id` to the `FileReference` ADT?

warrd: I don't think the physical ID is to do with a file reference (which has it's own ID), so didn't make sense to me to shove it on there :thinking: 

(nit: `FileReference` isn't an ADT I don't think? just a standard data type)

jamieparkinson: Case classes are syntactic sugar for product types - they're just fancy tuples

jamieparkinson: Actually the reason I referred to it as an ADT here is because recognising it as such made it look strange to me - the type `(String, FileReference)` is the product type, `String * (Whatever * Is * In * FileReference)` which associativity makes equivalent to `String * Whatever * Is * In * FileReference`, ie extending `FileReference`

warrd: I get what you are saying about intrinsically being tuples, but here I think the choice is more to do with how we conceptualise and name things in relation to the logic of the application rather than underlying types and equivalence between them. Still interesting though. I generally use ADT here as shorthand to refer to sum types rather than product types, probably due to lack of any accepted Scala terminology for referring to a sealed trait with another of case class variants.

PR #619, opened 2020-06-03
Suppress calm records on CatalogueStatus

jamesgorrie:
> suppress everything except Catalogued / Not yet available / Partially catalogued

-- The cataloguers

We also get some funky versions:
```
CatalogueStatus,Count
Catalogued,233298
"Third-party metadata",39322
Uncatalogued,3379
"Catalogued ",282
catalogued,89
"Not yet available",37
"Closed description",34
Draft,27
"In Conservation",16
"Partially catalogued",12
Deaccessioned,5
"Box list",3
"In Progress",3
"Catalogued 	 ",1
"Partially Third-party metadata",1
```

### DIFF
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
@@ -85,7 +85,7 @@ object CalmTransformer
     }
 
   def shouldSuppress(record: CalmRecord): Boolean =
-    transmissionIsNo(record) match {
+    catalogueStatusIsSuppressible(record) match {
       case true  => true
       case false =>
         // Records prefixed with AMSG (Archives and Manuscripts Resource Guides)
@@ -96,17 +96,17 @@ object CalmTransformer
           .exists(_.startsWith("AMSG"))
     }
 
-  def transmissionIsNo(record: CalmRecord): Boolean =
+  // We suppress unless CatalogueStatus exists and is valid
+  def catalogueStatusIsSuppressible(record: CalmRecord): Boolean =
     record
-      .get("Transmission")
-      .exists { value =>
-        value.toLowerCase match {
-          case "no"  => true
-          case "yes" => false
-          case _ =>
-            info(
-              s"Unrecognised value for Transmission field; assuming 'Yes': $value")
+      .get("CatalogueStatus")
+      .forall { value =>
+        value.toLowerCase.trim match {
+          // As much as it might not look like it, these values mean it should
+          // not be suppressed.
+          case "catalogued" | "not yet available" | "partially catalogued" =>
             false
+          case _ => true
         }
       }
 
--- a/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
@@ -85,7 +85,7 @@ object CalmTransformer
     }
 
   def shouldSuppress(record: CalmRecord): Boolean =
-    catalogueStatusIsSuppressible(record) match {
+    catalogueStatusSuppressesRecord(record) match {
       case true  => true
       case false =>
         // Records prefixed with AMSG (Archives and Manuscripts Resource Guides)
@@ -97,7 +97,7 @@ object CalmTransformer
     }
 
   // We suppress unless CatalogueStatus exists and is valid
-  def catalogueStatusIsSuppressible(record: CalmRecord): Boolean =
+  def catalogueStatusSuppressesRecord(record: CalmRecord): Boolean =
     record
       .get("CatalogueStatus")
       .forall { value =>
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -96,17 +96,17 @@ object CalmTransformer
           .exists(_.startsWith("AMSG"))
     }
 
-  def transmissionIsNo(record: CalmRecord): Boolean =
+  // We suppress unless CatalogueStatus exists and is valid
+  def catalogueStatusIsSuppressible(record: CalmRecord): Boolean =
```
jamesgorrie, line 100: I bumbled around a bit as to whether we should have `catalogueStatusIsValid`, but it confuses the logic further up, so this did for now.

jamieparkinson: Maybe this could be just
```suggestion
  def catalogueStatusIsSuppressed(record: CalmRecord): Boolean =
```
to be clear about intent?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -96,17 +96,17 @@ object CalmTransformer
...
-          case "yes" => false
-          case _ =>
-            info(
-              s"Unrecognised value for Transmission field; assuming 'Yes': $value")
+      .get("CatalogueStatus")
+      .forall { value =>
+        value.toLowerCase.trim match {
+          // As much as it might not look like it, these values mean it should
+          // not be suppressed.
+          case "catalogued" | "not yet available" | "partially catalogued" =>
```
jamieparkinson, line 107: **BIKESHED / NANO-NIT** I think my preference would be to have each string as a separate case

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
jamesgorrie: That makes it sound like the status is suppressed, which it isn't.
`catalogueStatusSuppressesRecord`?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
jamesgorrie: Any idea why? I find this more readable?

jamesgorrie: I think I prefer this as this is a list of valid `CatalogueStatus`es, and that is what I am trying to reflect.

What about having a `val validCatalogueStatuses = List("catalogued", "not yet available", "partially catalogued")`  

and then `case status if validStatuses.contains(status) => true`

jamesgorrie: What I am trying to express is that these three are _the_ valid options that make this return true.

On a new line makes it feel like the terms aren't connected, which they are.

jamesgorrie: Saying that - what about

```scala
List("catalogued", "not yet available", "partially catalogued").contains(str) match {
  case true => false
  case false => true
}
```

jamesgorrie:
```scala
def catalogueStatusSuppressesRecord(record: CalmRecord): Boolean =
    record
      .get("CatalogueStatus")
      .forall(isValidCatalogueStatus)

  def isValidCatalogueStatus(str: String) =
    List("catalogued", "not yet available", "partially catalogued").contains(
      str)
```

jamesgorrie:
```scala
val valid = List("catalogued", "not yet available", "partially catalogued").contains(str)
(true, false) match {
  case (t, _) if t == valid => false
  case (_, f) if f == valid => true
}
```

warrd: My vote is for the original, I don't reckon it's getting any clearer

jamesgorrie: Not even the last attempt?

I've looked through the code, and there's no similar example [bar here](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraConcepts.scala#L92), and even that isn't really the same.

I'm going to go with the original and leave it to the next dev who touches this to then decide.

jamesgorrie:
Tests have passed, GH not updated./

PR #622, opened 2020-06-04
Do not create images from METS files with restricted access statuses

jamieparkinson:
This just extends the logic to suppress thumbnails to `Image`s as well

Closes https://github.com/wellcomecollection/platform/issues/4566

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
@@ -108,9 +108,10 @@ case class MetsData(
         }
       }
 
-  private def thumbnail(bnumber: String,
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
```diff
@@ -123,35 +115,40 @@ case class MetsData(
...
-      .flatMap { fileReference =>
-        ImageUtils.buildImageUrl(recordIdentifier, fileReference).map { url =>
-          UnmergedImage(
-            ImageUtils
-              .getImageSourceId(recordIdentifier, fileReference.id),
-            version = version,
-            DigitalLocation(
-              url = url,
-              locationType = LocationType("iiif-image")
+  private def images(version: Int, accessStatus: Option[AccessStatus]) =
```
warrd, line 133: Can we get a return type here

PR #623, opened 2020-06-04
don't redirect when we merge images

jamesgorrie:
`ImagesRule` shouldn't currently affect the way redirection of works work.

Currently we redirect the Miro work because of this rule, where it needs to exist as it's own identified work as we haven't merged the metadata that's used by the works index.

This will probably change once we start using the `work.images` and/or loosen up on the rules around item merging.


jamieparkinson:
Related-ish: https://github.com/wellcomecollection/platform/issues/4565

PR #632, opened 2020-06-10
More miro merging

jamieparkinson:
This achieves the following:

_In the transformer:_
- Removes the restriction on material type for Sierra's Miro mergeCandidates
- Remove the requirement that these mergeCandidates are 1-to-1 (ie, merge any number of linked Miro works)

_In the merger:_
- Suppresses any "historical library" (`L`/`M`) Miro works **that are not linked with a Sierra work** and does not create images from them.
- Otherwise, maintains existing behaviour by only merging in items/identifiers and redirecting Sierra/Miro works if there is **1** Miro work and the Sierra work has a `workType` of `Picture`, `Digital Image`, or `3D Object`.

Resolves https://github.com/wellcomecollection/platform/issues/4565


common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/IdentifiersGenerators.scala
```diff
@@ -52,8 +54,22 @@ trait IdentifiersGenerators extends RandomStrings {
       identifierType = IdentifierType("isbn")
     )
 
+  private val MIRO_ID_PREFIXES: Seq[Char] = Seq(
```
jamesgorrie, line 57: Why all the big letters?

jamieparkinson: Because it's a constant and I've written too much C and it has rotted my brain - I'll change it to `MiroIdPrefixes` as per https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods

warrd: Fwiw I haven't seen `ConstantsLikeThis` once in our code base. I tend to just use `standardCamelCaseForConstants` (there are loads of examples, they are all over the place), and don't find it particularly confusing / misleading, but can understand if people not into it.

jamesgorrie: I was just writing what @warrd said, if we are going to come up with style rules, we need to be able to lint them otherwise these conversations become often and boring

jamieparkinson: OK sure - I certainly can't be bothered with linting since we have scalafmt. That said this conversation is evidence that we do have style rules, just undocumented ones that are different to the ones that someone who was looking for Scala style guidance would find, no? 😛

warrd: I would probably agree with that statement, yep :see_no_evil:

warrd: But I kind of think that is unavoidable for any team of devs, as no guidelines covers everything. It might be easier to follow the official scala style guide for these sort of cases, but starting at this point would probably lead to lots of inconsistencies (not that in this case I think it particularly matters if we were slightly inconsistent tho)

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/InvisibilityReason.scala
```diff
@@ -8,4 +8,5 @@ object InvisibilityReason {
   case class CopyrightNotCleared(info: String) extends InvisibilityReason
   case class SourceFieldMissing(info: String) extends InvisibilityReason
   case class InvalidValueInSourceField(info: String) extends InvisibilityReason
+  case class UnlinkedHistoricalLibraryMiro() extends InvisibilityReason
```
warrd, line 11: Any reason this needs to be a `case class` rather than a `case object`?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -15,33 +11,34 @@ import cats.data.NonEmptyList
...
-    } ++ findFirstLinkedDigitisedSierraWorkFor(target, sources)
+    val mergedSources = (
+      List(
+        mergeIntoCalmTarget,
+        mergeMetsIntoSingleItemSierraTarget,
+        mergeSingleMiroIntoSingleItemSierraTarget,
+        mergeIntoMultiItemSierraTarget)
+        .flatMap(_.mergedSources(target, sources)) ++
+        findFirstLinkedDigitisedSierraWorkFor(target, sources)
+    ).distinct
```
warrd, line 41: nit: the way this is indented / laid out makes it pretty hard to parse for me (dunno if this is scalafmt being a nuisance or not)

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/InvisibilityReason.scala
jamieparkinson: Just for consistency with the others tbh

warrd: I don't think inconsistent, I think others are that way because have an `info` argument. Maybe we want an `info` string for this too?

jamieparkinson: Yeah, I couldn't find anything that would be useful `info` for this - the others use it to say (eg) what their source is. I'll make it an `object`

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -45,20 +46,31 @@ trait FieldMergeRule {
   protected trait PartialRule {
     val isDefinedForTarget: WorkPredicate
     val isDefinedForSource: WorkPredicate
+    val isDefinedForSourceList: Seq[TransformedBaseWork] => Boolean =
```
jamesgorrie, line 49: For a rule to be applied:
* The `target` has to return `true` with the `isDefinedForTarget` predicate
* The `sources` has to return `true` with the `isDefinedForSourceList` predicate
* Only is some `sources` that return `true` with the `isDefinedForSource` predicate

And only the `isDefinedForSource` `sources` get sent to the rule.

Is that about right?

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidates.scala
```diff
@@ -70,37 +69,15 @@ object SierraMergeCandidates
...
-                "Single page Miro/Sierra work (secondary source)")
-            )
-          case _ => Nil
-        }
-      case _ => Nil
-    }
+    (matching089Ids(bibData) ++ matching962Ids(bibData)).distinct
+      .map {
+        miroMergeCandidate(_, "Miro/Sierra work")
+      }
```
jamesgorrie, line 80: I much prefer this process, it's so much easier to tell what's going on where.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
```diff
@@ -45,20 +46,31 @@ trait FieldMergeRule {
...
+        case head +: tail =>
           Some(rule(target, NonEmptyList(head, tail.toList)))
         case _ => None
       }
 
     def apply(target: UnidentifiedWork,
               source: TransformedBaseWork): Option[FieldData] =
       apply(target, List(source))
+
+    def mergedSources(
```
jamesgorrie, line 67: Maybe a little comment as to what this method is for?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala
jamieparkinson: That is exactly right

PR #635, opened 2020-06-12
[WIP] Deploy model training service

jamieparkinson:
This is mostly untested, just wanted to push it now


pipeline/inferrer/feature_training/run.py
```diff
@@ -0,0 +1,81 @@
...
+            "awsvpcConfiguration": {
+                "subnets": [get_subnet()],
+                "assignPublicIp": "ENABLED",
+            }
+        },
+        platformVersion="1.4.0",
+    )
+    print(f"Successfully started training task [{res['tasks'][0]['taskArn']}]")
+    print("You can follow the logs at:")
+    print(get_kibana_logs_url(pipeline_name))
```
alexwlchan, line 57: Could you include the task ARN here?

pipeline/inferrer/feature_training/run.py
```diff
@@ -0,0 +1,81 @@
...
+
+@cli.command()
+@click.option("--from-label", help="The release label to deploy from", default="latest")
+@click.option("--to-label", help="The release label to deploy to", required=True)
+def deploy_model(from_label, to_label):
+    ssm_path_template = "/catalogue_pipeline/config/models/%s/lsh_model"
+    ssm_client = session.client("ssm")
+    from_value = ssm_client.get_parameter(Name=(ssm_path_template % from_label))[
+        "Parameter"
+    ]["Value"]
```
alexwlchan, line 68: There's a lot going on in this line; breaking it onto a couple of lines would make this easier to follow.

pipeline/inferrer/feature_training/run.py
```diff
@@ -0,0 +1,81 @@
...
+
+    print(f"Updating {to_label} to '{from_value}'...")
```
alexwlchan, line 70:
```suggestion
    print(f"Updating {to_label} to {from_value!r}...")
```

Adding `!r` causes Python to print the repr, i.e. unambiguous representation of an object. Bit nicer here, it handles quotes and the like for you.

pipeline/inferrer/feature_training/src/aws.py
```diff
@@ -1,28 +1,39 @@
...
 from botocore.exceptions import ClientError
+from .logging import get_logger
+
+logger = get_logger(__name__)
+
+
+def get_ecs_container_metadata():
+    endpoint = os.environ["ECS_CONTAINER_METADATA_URI_V4"]
+    req = requests.get(endpoint)
+    return req.json()
```
alexwlchan, line 13: You probably want to throw in `req.raise_for_status()` here, so if it returns a 404 or whatever, you know immediately.

alexwlchan: Also nit: prefer `resp` to `req`; what you get back is a Response object, not a Request.

pipeline/inferrer/feature_training/src/aws.py
```diff
@@ -1,28 +1,39 @@
...
     s3_client.put_object(Bucket=bucket_name, Key=key, Body=binary_object)
-    print("Uploaded object to S3.")
+    logger.info("Uploaded object to S3.")
 
 
 def put_ssm_parameter(path, value, description):
     ssm_client = boto3.client("ssm")
 
-    print(f"Updating SSM path `{path}` to `{value}`...")
+    logger.info(f"Updating SSM path `{path}` to `{value}`...")
```
alexwlchan, line 35: Minor: Python best practice is not to do string interpolation inside logging calls, but to pass the parameters as separate arguments and let them be interpolated lazily, i.e.
```suggestion
    logger.info("Updating SSM path %r to %r", path, value)
```

Python will only actually perform the string interpolation if this message is going to be sent anywhere; if it’s below the current log level you save an interpolation.

(Which granted isn't a huge benefit here, but it's a good habit to get into.)

pipeline/inferrer/feature_training/src/aws.py
```diff
@@ -1,28 +1,39 @@
...
     return s3_client
 
 
 def put_object_to_s3(binary_object, key, bucket_name):
     s3_client = get_s3_client()
-    print("Uploading object to S3...")
+    logger.info("Uploading object to S3...")
     s3_client.put_object(Bucket=bucket_name, Key=key, Body=binary_object)
-    print("Uploaded object to S3.")
+    logger.info("Uploaded object to S3.")
```
alexwlchan, line 29: Worth logging the bucket/key while you're fixing these?

pipeline/inferrer/feature_training/src/elastic.py
```diff
@@ -1,46 +1,62 @@
...
-def get_all_ids(es_client, index_name):
-    response = helpers.scan(
-        client=es_client,
-        index=index_name,
-        query={"query": {"match_all": {}}, "stored_fields": []},
-    )
-
-    return [hit["_id"] for hit in list(response)]
 
+def get_documents_from_es_in_batches(es_client, index_name, ids, batch_size=5000):
```
alexwlchan, line 10:
```suggestion
def get_all_documents_from_es(es_client, index_name, ids, batch_size=5000):
```

The caller only cares that you get all the documents; that it's happening in batches is mostly irrelevant to them.

alexwlchan: Would it be worth allowing the caller to specify extra fields on the body here? You're only using two fields on the returned documents; I don't know how many fields there are but it might be a tidy saving.

pipeline/inferrer/feature_training/run.py
jamieparkinson: See line 55 😛

jamieparkinson: oh unless you mean in the kibana filter?

pipeline/inferrer/feature_training/src/elastic.py
jamieparkinson: It actually only returns the 2 fields we want because of field-level permissions on the kibana role, but I may add this anyway

pipeline/inferrer/feature_training/run.py
alexwlchan: Yeah, I’m pretty sure the ECS task ARN is one of the fields that get sent by Logstash, so you could watch the logs from this exact task.

PR #638, opened 2020-06-16
Handle By appointment AccessStatus

alicefuzier:


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AccessCondition.scala
```diff
@@ -64,6 +66,6 @@ object AccessStatus {
       case status if status.startsWith("permission required") =>
         Right(AccessStatus.PermissionRequired)
       case status =>
-        Left(new UnknownAccessStatus(status))
+        Left(new UnknownAccessStatus(str))
```
jamesgorrie, line 69: Assuming this needs to be `status`?

alicefuzier: No, I changed that intentionally. `status` is `str` lowercased but, this being an error that gets displayed, it's more useful if it prints the actual string that caused the error.

jamesgorrie: Maybe a note?

jamesgorrie: And assuming we can `_` `status` then too.

alicefuzier: Not sure i follow the last comment...Rather than a note I'd lean towards renaming the variables to something more meaningful `status` and especially `str` are awful names so that line is quite confusing

alicefuzier:
Is this any better @jamesgorrie ?

PR #651, opened 2020-06-25
Work data images

alicefuzier:
For https://github.com/wellcomecollection/platform/issues/4584


common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
```diff
@@ -43,6 +43,10 @@ class ElasticsearchIndexCreator(
           // If/when we index more records, we should revisit this.
           //
           .shards(1)
+          // Elasticsearch has a default maximum number of fields of 1000.
+          // Because images have all of the WorkData fields defined twice in the mapping,
+          // they end up having more than 1000 fields, so we increase them to 2000
+          .settings(Map("mapping.total_fields.limit" -> 2000))
```
jamesgorrie, line 49: Another way around this is to store, but not index, non-searchable data.
https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html

I've wanted to do it for the catalogue, as we do leak loads of model specific data into elastic.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/ImageSource.scala
```diff
@@ -1,11 +1,20 @@
...
-sealed trait ImageSource[Id <: WithSourceIdentifier] {
+sealed trait ImageSource[Id <: WithSourceIdentifier, DataId <: IdState] {
   val id: Id
   val ontologyType: String
 }
 
-case class SourceWork[Id <: WithSourceIdentifier](
+case class SourceWorks[Id <: WithSourceIdentifier, DataId <: IdState](
+  canonicalWork: SourceWork[Id, DataId],
+  redirectedWork: Option[SourceWork[Id, DataId]]
```
jamesgorrie, line 10: What if we have multiple redirected works?

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
warrd: I agree that we should be keeping on top of how many fields we index. Did we work out why there was more than 1000 in the mapping?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/ImageSource.scala
jamieparkinson: I think `redirectedWork` here refers to the work that the image was actually from, but is now redirected (eg, the Miro work that was merged with a Sierra work)

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
alicefuzier: We talked about non-searchable data, but this would mean that if we want to make a field searchable we have to reindex. As the scale of our data isn't huge and ingestion times aren't an issue, we decided to index everything so we can iterate on how to query the data without needing reindexes. We can always make them non-searchable later if we find a reason. I don't think the 1000 limit is a good reason

As per the number of fields, I've logged the mapping, it's quite big. I didn't count them, but WorkData has way more than 100 fields

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/ImageSource.scala
alicefuzier: We had this discussion with @jamieparkinson but in practice, it's always going to be one or 2 works, not more. If that changes we can change the model

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
warrd: Do you have a file with the mapping that I can see?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/ImageSource.scala
jamesgorrie: Ah yes, OK, that makes sense

api/api/src/test/scala/uk/ac/wellcome/platform/api/images/ImagesTest.scala
```diff
@@ -73,6 +73,36 @@ class ImagesTest extends ApiImagesTestBase with ElasticsearchFixtures {
...
+          imageId = Identified("a", createSourceIdentifier),
+          parentWork = createIdentifiedWorkWith(title = Some("Baguette is a French bread style"))
+        )
+        val focacciaImage = createAugmentedImageWith(
+          imageId = Identified("b", createSourceIdentifier),
+          parentWork = createIdentifiedWorkWith(title = Some("A Ligurian style of bread, Focaccia"))
+        )
+        val schiacciataImage = createAugmentedImageWith(
+          imageId = Identified("c", createSourceIdentifier),
+          parentWork = createIdentifiedWorkWith(title = Some("Schiacciata is a Tuscan focaccia")),
```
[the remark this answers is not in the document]
alicefuzier: Actually I didn't mean to knock off mantou, it's just that because Ligurian focaccia was there I felt I had to add the Tuscan one. You know, regional rivalries and stuff... 😬 😁 
You're probably right about noodles

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -1,35 +1,32 @@
 package uk.ac.wellcome.models.work.internal
 
-sealed trait BaseImage[+Id <: WithSourceIdentifier] extends HasIdState[Id] {
+sealed trait BaseImage[+Id <: WithSourceIdentifier, DataId <: IdState] extends HasIdState[Id] {
```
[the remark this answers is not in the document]
alicefuzier: Yes, it's actually not too bad

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
kenoir: If we're not actually interested in searching these fields, I think @jamesgorrie's suggestion of not indexing them is a good one.

alicefuzier: Yes, I uploaded to the slack channel in the thread, can't seem to figure out how to add it here..

jamieparkinson: We're definitely interested in searching these fields - probably not all of them, but we don't know which ones we do want so we should have all of them searchable to make testing easier imo

api/api/src/test/scala/uk/ac/wellcome/platform/api/images/ImagesTest.scala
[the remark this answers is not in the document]
jamieparkinson: ha, you must be the first Italian to say that...

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
alicefuzier: I think which fields are we going to search is a decision that will need iterating and testing. It would make our life easier if adding one field to the search query didn't mean reindexing. 100% we index stuff in workdata that has no use to search but that is probably cross purposes in works _and_images, so if we fix it once in  the works mapping, we should fix it for both and remove this line. But I think, for now, it's best if the code is reused even if that means a bit of waste at indexing time, rather than having a different mapping for the workdata on images to the mapping for workdata on works

warrd: There is things in the mapping like `ontologyType` (appears 162 times) which I think are completely pointless to index as these are constants (I don't think there is any need for this to appear in the internal model at all, and should maybe just be in display, but that is another question).

I understand the arguments about indexing things we might at a future date need, but I reckon if there is anything like that we are sure we won't need I think we should not index them. Thoughts?

alicefuzier: Yes, definitely. Also, the "type" field is the circe type discriminator and I'm not even 100% sure when it's there and when it's not so definitely not going to be used for search, at least not is that many places. Might be useful on works but nowhere else. So 100% take out those that are never needed. 
I think it's something that should be done for both images _and_ works though so the approach in this PR to reuse the mapping code for WorkData mapping can be retained. So, I agree but I also think It's outside of the scope of this PR

warrd: We use a `type` discriminator in one place for search apparently:
https://github.com/wellcomecollection/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala#L158-L163

I think I have found it useful filtering for e.g. `IdentifiedInvisibleWork` or what have you in Kibana, so there might be more of a use case there.

Agree beyond scope of this PR to remove `ontologyType` mapping.

alicefuzier: >  (I don't think there is any need for this to appear in the internal model at all, and should maybe just be in display, but that is another question)

I totally agree with that, I would really like to see it disappear from the internal model

jamieparkinson: +1 to removing ontologyType!

PR #656, opened 2020-06-25
Fix id minter

jamieparkinson:
It turns out the issue was to do with MySQL struggling with certain queries, but in the course of getting there I added some logging of slow queries, started using the read replica, and increased the CPU/memory of the task as it had been saturated at 100%.

### DIFF
--- a/pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
+++ b/pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
@@ -11,8 +11,6 @@ import scala.concurrent.blocking
 import scala.util.Try
 
 object IdentifiersDao {
-  final val maxSelectSize = 50
-
   case class LookupResult(
     existingIdentifiers: Map[SourceIdentifier, Identifier],
     unmintedIdentifiers: List[SourceIdentifier])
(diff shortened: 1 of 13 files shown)


pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -25,68 +25,73 @@ object IdentifiersDao {
...
+            .flatMap { identifierBatch =>
+              blocking {
+                val i = identifiers.i
+                val matchRow = rowMatcherFromSyntaxProvider(i) _
+
+                // The query is manually constructed like this because using WHERE IN
+                // statements with multiple items causes a full-index scan on MySQL 5.6.
+                // See: https://stackoverflow.com/a/53180813
+                // Therefore, we perform the rewrites here on the client.
+                //
```
kenoir, line 50: Aren't we using Aurora, with mysql compatibility?

kenoir:
Good job getting to the bottom of this.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -25,68 +25,73 @@ object IdentifiersDao {
...
+                // Because of issues in `scalikejdbc`'s typing, it's necessary to
+                // specify the `Nothing` type parameters as done here in order for
+                // the `map` statement to work properly.
+                val query = withSQL[Nothing] {
```
alicefuzier, line 54: Eww...Not saying you're doing anything wrong here but this is where a DSL is really not helping methinks. 

`<heresy>`Also possibly scala is not helping. Maybe the idminter would be simpler in python`</heresy>`

jamieparkinson: fwiw this is old code that got indented differently... but yes, absolutely, and I remember when I originally wrote it it took _ages_ to work that out (and there were other issues which made it really hard to do with the library's `sql` and `sqls` string template things).

I agree with the heresy - in fact I'd go further and say it might be better/simpler in the language that's the J in JSON! 😱

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
jamieparkinson: Good question. The short answer is yes. The long answer after reading about it is yes, but the Aurora versions are actually forks of the corresponding MySQL, Postgres, etc versions where Amazon have replaced the underlying Things What Do Storage but have left the query, indexing, etc stuff completely intact - so the compatibility is up to and including bugs and other issues.

pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
alicefuzier: 😱

PR #657, opened 2020-06-26
Pipeline 26/06/2020

warrd:


pipeline/terraform/main.tf
```diff
@@ -19,29 +19,29 @@ module "catalogue_pipeline_20200619" {
   # reindexer topic names.
 
   sierra_adapter_topic_arns = [
-    //    local.sierra_reindexer_topic_arn,
+    local.sierra_reindexer_topic_arn,
```
jamieparkinson, line 22: If this is going to prod you probably want these commented out still

PR #659, opened 2020-06-29
Similarity search queries

jamieparkinson:
This PR adds the option to `?include=visuallySimilar` on `/images/{id}` queries

 It also refactors `ElasticsearchService` so that it can be used (to transform our data into requests, handle APM stuff, and transform responses into the form we want) across the application, as we'd previously been using it as intended for "simple" search/get requests but were not able to use it for collection requests and would not have been able to use it for image similarity searches.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
@@ -297,8 +297,7 @@ final case class BoolBoostedQuery(q: String) extends ElasticsearchPartialQuery {
 final case class ImageSimilarityQuery(q: String, index: Index)
     extends ElasticsearchQuery {
   private final val lshFields = List("inferredData.lshEncodedFeatures")
-
-  lazy val documentRef = DocumentRef(index = index, id = q)
+  private lazy val documentRef = DocumentRef(index, q)
 
   lazy val elasticQuery =
     moreLikeThisQuery(lshFields)
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
@@ -1,7 +1,8 @@
 package uk.ac.wellcome.platform.api.elasticsearch
 
-import com.sksamuel.elastic4s.ElasticDsl.{must, should}
-import com.sksamuel.elastic4s.requests.common.Operator
+import com.sksamuel.elastic4s.ElasticDsl.{moreLikeThisQuery, must, should}
+import com.sksamuel.elastic4s.Index
+import com.sksamuel.elastic4s.requests.common.{DocumentRef, Operator}
 import com.sksamuel.elastic4s.requests.searches.queries.{BoolQuery, Query}
 import com.sksamuel.elastic4s.requests.searches.queries.matches.{
   FieldWithOptionalBoost,
@@ -291,4 +292,20 @@ final case class BoolBoostedQuery(q: String) extends ElasticsearchPartialQuery {
     must(ContributorQuery(q).elasticQuery).boost(2000),
     must(TitleQuery(q).elasticQuery).boost(1000)
   )
+}
+
+final case class ImageSimilarityQuery(q: String, index: Index)
+    extends ElasticsearchQuery {
+  private final val lshFields = List("inferredData.lshEncodedFeatures")
+
+  lazy val documentRef = DocumentRef(index = index, id = q)
+
+  lazy val elasticQuery =
+    moreLikeThisQuery(lshFields)
+      .likeDocs(List(documentRef))
+      .copy(
+        minTermFreq = Some(1),
+        maxQueryTerms = Some(1000),
+        minShouldMatch = Some("1")
+      )
 }
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/CollectionService.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/CollectionService.scala
@@ -1,9 +1,8 @@
 package uk.ac.wellcome.platform.api.services
 
 import scala.concurrent.{ExecutionContext, Future}
-import com.sksamuel.elastic4s.ElasticDsl._
-import com.sksamuel.elastic4s.{ElasticClient, ElasticError, Index}
-import com.sksamuel.elastic4s.requests.searches.{SearchHit, SearchResponse}
+import com.sksamuel.elastic4s.Index
+import com.sksamuel.elastic4s.requests.searches.SearchHit
 
 import uk.ac.wellcome.models.work.internal._
 import uk.ac.wellcome.models.work.internal.result._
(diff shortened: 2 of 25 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
```diff
@@ -292,3 +293,18 @@ final case class BoolBoostedQuery(q: String) extends ElasticsearchPartialQuery {
     must(TitleQuery(q).elasticQuery).boost(1000)
   )
 }
+
+final case class ImageSimilarityQuery(q: String, index: Index)
```
warrd, line 297: I don't think we ever do inheritance with case classes, is there a particular reason we are describing this as final?

api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
```diff
@@ -292,3 +293,18 @@ final case class BoolBoostedQuery(q: String) extends ElasticsearchPartialQuery {
...
+  private final val lshFields = List("inferredData.lshEncodedFeatures")
+  private lazy val documentRef = DocumentRef(index, q)
+
+  lazy val elasticQuery =
+    moreLikeThisQuery(lshFields)
+      .likeDocs(List(documentRef))
+      .copy(
+        minTermFreq = Some(1),
+        maxQueryTerms = Some(1000),
+        minShouldMatch = Some("1")
```
warrd, line 308: Potentially some of these could be default arguments rather than defined inline, I dunno :man_shrugging:

api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
jamieparkinson: No idea, I was copying the convention that already exists in this file. fwiw I would like to revisit how query types are managed as we come back into search improvements work

api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ElasticsearchQuery.scala
jamieparkinson: Yeah, again I was copying the convention which is that the only arguments are required ones - I think this approach makes sense as it means the queries can be used without having to understand the actual ES query specification that's going on

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/CollectionService.scala
```diff
@@ -36,13 +35,16 @@ import uk.ac.wellcome.platform.api.Tracing
...
   implicit ec: ExecutionContext)
     extends Tracing {
 
   def retrieveTree(index: Index,
                    expandedPaths: List[String]): Future[Result[Collection]] =
-    makeEsRequest(index, expandedPaths)
+    elasticsearchService
+      .executeSearchRequest(
+        CollectionRequestBuilder(index, expandedPaths).request
+      )
```
warrd, line 47: Nice bit of refactoring so that we use the elastic search service here instead of redefining stuff :+1:

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesService.scala
```diff
@@ -50,10 +55,28 @@ class ImagesService(searchService: ElasticsearchService)(
...
+      .executeSearchRequest(
+        ImagesRequestBuilder.requestVisuallySimilar(
+          index = index,
+          id = image.id.canonicalId,
+          n = nVisuallySimilarImages
+        )
+      )
+      .map {
+        _.map {
+          _.hits.hits.map(hit => jsonTo(hit.sourceAsString)).toList
```
warrd, line 76: I think with these nested anonymous functions it would be clearer if we defined the input arguments (i.e. `resp => ..` etc) rather than using the `_` syntax, as right now slightly hard to know what they refer to

api/api/src/test/scala/uk/ac/wellcome/platform/api/images/ImagesSimilarityTest.scala
```diff
@@ -0,0 +1,53 @@
...
+           |  "id": "${image.id.canonicalId}",
+           |  "locations": [${digitalLocation(image.location)}],
+           |  "visuallySimilar": [
+           |    ${images.tail.map(imageResponse).mkString(",")}
+           |  ],
+           |  "source": {
+           |    "id": "${image.source.id.canonicalId}",
+           |    "type": "Work"
+           |  }
+           |}""".stripMargin
```
warrd, line 31: bikeshed: funny indentation going on here imo. also i dont think we need the whole `stripMargin` thing, the string is never displayed in raw form as is parsed to a json obj and reformatted before being displayed in any test failures

common/display/src/main/scala/uk/ac/wellcome/display/models/ImagesIncludes.scala
```diff
@@ -0,0 +1,22 @@
...
+}
+
+object SingleImageIncludes {
+  import ImageInclude._
+
+  def apply(visuallySimilar: Boolean = false): SingleImageIncludes =
+    SingleImageIncludes(
+      List(
+        if (visuallySimilar) Some(VisuallySimilar) else None
+      ).flatten
```
warrd, line 20: Can do `option.toList` instead of `List(option).flatten` here

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayImage.scala
```diff
@@ -30,7 +34,12 @@ object DisplayImage {
...
       locations = Seq(DisplayDigitalLocation(image.location)),
-      source = DisplayImageSource(image.source)
+      source = DisplayImageSource(image.source),
+      visuallySimilar = None
     )
 
+  def apply(image: AugmentedImage, similar: Seq[AugmentedImage]): DisplayImage =
+    DisplayImage(image).copy(
+      visuallySimilar = Some(similar.map(DisplayImage.apply))
+    )
```
warrd, line 44: What is the advantage of having a separate `apply` for this? Could we not pass this an as an optional argument to the other `apply`?

Also might be worth having some tests that things are included correctly, like in the `DisplayWork`

api/api/src/test/scala/uk/ac/wellcome/platform/api/images/ImagesSimilarityTest.scala
jamieparkinson: I agree but it's like this across like 10+ files so will leave as is for now

common/display/src/main/scala/uk/ac/wellcome/display/models/ImagesIncludes.scala
jamieparkinson: Yeah, going to keep as is though because I anticipate more includes will be added and so want it to match https://github.com/wellcomecollection/catalogue/blob/master/common/display/src/main/scala/uk/ac/wellcome/display/models/WorksIncludes.scala#L42

api/api/src/test/scala/uk/ac/wellcome/platform/api/images/ImagesSimilarityTest.scala
warrd: Cool might do a whole clean up of the bikeshed PR at some point if its niggling me

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayImage.scala
jamieparkinson: Good point re the apply.

Yeah, I will add some tests, but as a general point I'm not very convinced of the value of the display model tests. They effectively get re-tested in the API tests, and we don't add them consistently so they're not especially useful as documentation either...

jamieparkinson: Actually, the apply doesn't work. Because the case class has default values, and defining an `apply` in the companion object is an overload, it doesn't compile if that `apply` has default arguments: `multiple overloaded alternatives of method apply define default arguments`.

PR #663, opened 2020-07-02
remove OR query

jamesgorrie:
Our recall is way too high.

We would have preferred to have run this as part of a test, but it is also raising some incredibly irrelevant and problematic results, so will be running it as an iteration and marking the date to run analysis.


kenoir:
:shipit:

PR #664, opened 2020-07-02
Don't suppress historical library works & images

jamieparkinson:


jamieparkinson:
Yeah - they were an exception anyway, the first time the merger had ever suppressed anything in this way, so it would be a confusing test to see now that no suppression logic exists

PR #667, opened 2020-07-06
Use the new DNS role throughout the catalogue API

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/4621


alexwlchan:
This is passing in Travis, but it's not updated the status here. Merging anyway.

PR #674, opened 2020-07-15
add title index mapping

jamesgorrie:
Adding the title field that is now analysed differently.

We are adding [asciifolding](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-asciifolding-tokenfilter.html) and [shingles (with asciifolding)](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-shingle-tokenfilter.html).

We're removing the [English filter](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#english-analyzer) as we've seen issues it can raise when searching for specifics, namely the Marie/Mary problem - which when searching specifically they can have very very different meanings. That will be left on things like `description`.

If `title: "açaí palm"` `aqui` and `açaí` will both match on `title` and `title.shingles`.

This is a cutdown version of the mapping:
```JS
{
  "settings": {
    "analysis": {
      "filter": {
        "shingle_filter": {
          "type": "shingle",
          "min_shingle_size": 2,
          "max_shingle_size": 4
        },
        "asciifolding_filter": {
          "type": "asciifolding",
          "preserve_original": true
        },
        "english_stemmer": {
          "type":       "stemmer",
          "language":   "english"
        },
        "english_possessive_stemmer": {
          "type":       "stemmer",
          "language":   "possessive_english"
        }
      },
      "analyzer": {
        "en_analyzer": {
          "tokenizer": "standard",
          "filter": [
            "lowercase",
            "asciifolding_filter",
            "english_possessive_stemmer",
            "english_stemmer"
          ]
        },
        "asciifolding_analyzer": {
          "tokenizer": "standard",
          "filter": [
            "lowercase",
            "asciifolding_filter"
          ]
        },
        "shingle_analyzer": {
          "tokenizer": "standard",
          "filter": [
            "lowercase",
            "asciifolding_filter",
            "shingle_filter"
          ]
        }
      }
    }
  },
  "mappings": {
    "dynamic": false,
    "properties": {
      "title": {
        "type": "text",
        "analyzer": "asciifolding_analyzer",
        "fields": {
          "english": {
            "type": "text",
            "analyzer": "en_analyzer"
          },
          "shingles": {
            "type": "text",
            "analyzer": "shingle_analyzer"
          },
          "keyword": {
            "type": "keyword"
          }
        }
      }
    }
  }
}

```

TODO
- [x] Work on the shape of the field, it's not quite there yet, will base it on queries we'll make
- [ ] Tests

### DIFF
--- a/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
+++ b/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
@@ -10,10 +10,14 @@ import com.sksamuel.elastic4s.ElasticDsl.{
   textField,
   tokenCountField
 }
+import com.sksamuel.elastic4s.json.{XContentBuilder, XContentFactory}
 import com.sksamuel.elastic4s.requests.analysis.{
   Analysis,
   CustomAnalyzer,
-  PathHierarchyTokenizer
+  PathHierarchyTokenizer,
+  ShingleTokenFilter,
+  StemmerTokenFilter,
+  TokenFilter
 }
 import com.sksamuel.elastic4s.requests.mappings.{
   FieldDefinition,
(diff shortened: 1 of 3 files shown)


harrisonpim:
```
        "shingle_analyzer": {
          "tokenizer": "standard",
          "filter": [
            "asciifolding_filter",
            "shingle_filter"
          ]
        }
```

should the `asciifolding_filter` be here?

jamesgorrie:
> should the asciifolding_filter be here?

Yes - assuming we want that so that `açaí` and `aqui` matches on the shingles too?

harrisonpim:
After doing a bit of peer review with @jamesgorrie, i think i have a clearer understanding of what this new title mapping does for us, and how it will play with the query we intend to write. Summarising that understanding here..

- the JSON above is a stub of the full mapping, focusing only on the `title` field and its subfields
- we could query all of the `title` subfields at once, using `title.*`
- however, because of examples we've seen in the search logs and anecdotes, we will instead boost title subfields to different levels
- the keyword field is the most heavily boosted field in the query, because an _exact_ title match is so unlikely but indicates a very clear intention.
- an english analyser is used to build the `en` subfield, which will be boosted more than other fields, but less than other title subfields
- a shingle analyser creates the `shingles` subfield, which is boosted more than the `en` subfield. the `shingle_analyzer` combines ascii folding and shingles (from unigrams up to 4-grams), so a string like "açaí à la carte" is tokenised as:

  > `["açaí", "à", "la", "carte", "açaí à", "à la", "la carte", "açaí à la", "à la carte", "açaí à la carte",`
  >
  >`"acai", "a", "la", "carte", "acai a", "a la", "la carte", "acai a la", "a la carte", "acai a la carte"]`

- the rationale for the difference in analyser structure and boosting between the `en` and `shingle` subfields is illustrated by the following example:
  - given two documents (“Mary who does tailoring", “Marie the tailor”) and a query (“Mary tailoring”)
  - it's desirable to return both documents - both are matched by the analysers described above
  - “Mary who does tailoring" is a more exact match, so the un-stemmed tokens provided in the ascii folded `shingle` field should be more heavily boosted

harrisonpim:
I feel confident about the above, but I'm still wondering why we shouldn't shingle the `en` field as well as the `ascii_folding` one? As I understand it, there's no downside to strongly scoring english-analysed tokens which match the query tokens' ordering

jamesgorrie:
> I'm still wondering why we shouldn't shingle the `en`

We could do that, but would probably created an `title.en_shingled` field. I haven't done a lot of work on how scoring works with arrays.

e.g. `title.en:["acai", "a", "la", "carte"]` vs `title:"acai a la carte"`.

Through a rudimentary test, the shingles scores higher on exact matches. For now I am going to assume we need it, and work out how when we do the query to get this in.

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
```diff
@@ -22,17 +26,87 @@ import com.sksamuel.elastic4s.requests.mappings.{
 }
 import com.sksamuel.elastic4s.requests.mappings.dynamictemplate.DynamicMapping
 
+// TODO: Patch back to elastic4s
```
jamieparkinson, line 29: Looks like they've changed the way analyzers/filters are done too, so it's a patch + upgrade + change existing code. Great! 🙃

jamesgorrie: And they only upgrade with es version bumps...

PR #675, opened 2020-07-15
Decouple METS Adapter deployment from SSM

kenoir:
Move the METS adapter to the new deployment schedule.

### DIFF
--- a/infrastructure/modules/worker/main.tf
+++ b/infrastructure/modules/worker/main.tf
@@ -1,5 +1,9 @@
+locals {
+  deployment_service_name = var.deployment_service_name == "" ? var.name : var.deployment_service_name
+}
+
 module "service" {
-  source = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/service?ref=v2.4.1"
+  source = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/service?ref=v3.0.0"
 
   task_definition_arn            = module.task_definition.arn
   service_name                   = var.name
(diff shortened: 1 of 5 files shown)


infrastructure/modules/worker/main.tf
```diff
@@ -1,5 +1,9 @@
+locals {
+  deployment_service_name = var.deployment_service_name == "" ? var.name : var.deployment_service_name
```
jamesgorrie, line 2: Is there a reason to not just make this compulsory? I sometimes get lost where things get magically created, and have to go digging them out.

infrastructure/modules/worker/main.tf
```diff
@@ -10,10 +14,17 @@ module "service" {
   desired_task_count             = var.desired_task_count
   security_group_ids             = var.security_group_ids
   use_fargate_spot               = var.use_fargate_spot
+
+  propagate_tags = "SERVICE"
+
+  tags = {
+    "deployment:service" : local.deployment_service_name
+    "deployment:env" : var.deployment_service_env
```
alexwlchan, line 22:
```suggestion
    "deployment:service" = local.deployment_service_name
    "deployment:env"       = var.deployment_service_env
```

makefiles/functions.Makefile
```diff
@@ -82,9 +82,9 @@ endef
 define publish_service_ssm
 	$(ROOT)/docker_run.py \
     	    --aws --dind -- \
-                wellcome/weco-deploy:2.0.0  \
+                wellcome/weco-deploy:3.1.0  \
                 --project-id="$(2)" \
-                --role-arn="$(DEV_ROLE_ARN)" \
+                -v \
```
alexwlchan, line 87: What does this `-v` flag do?

jamesgorrie: [This](https://github.com/wellcomecollection/weco-deploy/blob/master/src/deploy/deploy.py#L66)?

infrastructure/modules/worker/main.tf
kenoir: This lets you override the default service name - which is necessary where the service name will not match an expected static pattern (catalogue-DATE services are an example of this).

makefiles/functions.Makefile
kenoir: Verbosity! I think in CI this process should be verbose for easier debugging, at least right now.

infrastructure/modules/worker/main.tf
jamesgorrie: Ah ok, so it's expected the name is what we want, and this is to override.

`deployment_service_name_override` or a comment?  I would feel like I should give this if coming in fresh

PR #676, opened 2020-07-17
Use latest version of weco-deploy

kenoir:
Including:
- YAML config
- Set up for METS adapter deployment


.wellcome_project
```diff
@@ -1,108 +1,69 @@
...
-    "name": "Catalogue API",
-    "role_arn": "arn:aws:iam::760097843905:role/platform-ci",
-    "aws_region_name": "eu-west-1",
-    "github_repository": "wellcometrust/catalogue",
-    "tf_stack_root": "unused"
-  }
-}
+mets_adapter:
+  environments:
+  - id: prod
```
jamesgorrie, line 3: Do we need latest in these lists?

calm_adapter/terraform/locals.tf
```diff
@@ -8,6 +8,8 @@ locals {
   vpc_id          = data.terraform_remote_state.shared_infra.outputs.catalogue_vpc_delta_id
   private_subnets = data.terraform_remote_state.shared_infra.outputs.catalogue_vpc_delta_private_subnets
 
+  calm_adapter_image = "${aws_ecr_repository.ecr_repository_calm_adapter.repository_url}:env.prod"
```
jamesgorrie, line 11: This is neat.

.wellcome_project
kenoir: no

PR #681, opened 2020-07-20
Apply deployment tags to pipeline services in ECS

kenoir:
This adds the tags required for automating ECS deployments using the `weco-deploy` tool.

Further work is required to update image references in services to allow the tool to deploy to ECS (deployments should continue to function as before).


kenoir:
For https://github.com/wellcomecollection/platform/issues/4676

PR #682, opened 2020-07-20
Use new tags in pipeline task defs

kenoir:
Required for ECS deployment using the `weco-deploy` tool.

Depends on https://github.com/wellcomecollection/catalogue/pull/681

### DIFF
--- a/.wellcome_project
+++ b/.wellcome_project
@@ -72,6 +72,63 @@ catalogue_pipeline:
       name: Production
     - id: miro_merging_test
       name: Miro merging test
+  image_repositories:
+    - id: id_minter
+      namespace: uk.ac.wellcome
+      account_id: '760097843905'
+      services:
+        - image-id-minter
+        - work-id-minter
+    - id: inference_manager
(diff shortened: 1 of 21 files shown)


kenoir:
For https://github.com/wellcomecollection/platform/issues/4676

.wellcome_project
```diff
@@ -72,6 +72,63 @@ catalogue_pipeline:
       name: Production
     - id: miro_merging_test
       name: Miro merging test
+  image_repositories:
+    - id: id_minter
+      namespace: uk.ac.wellcome
```
jamieparkinson, line 77: Is it worth using an alias/anchor for the repeated account ID and namespace keys?

kenoir: It's worth doing something to reduce repetition (as previously mentioned by @alexwlchan). I'm not a massive fan of references in YAML. I think the deploy tool should provide a mechanism to specify (and override) defaults.

jamieparkinson: yeah that sounds like a better idea - also not a huge fan of YAML references

PR #686, opened 2020-07-21
remove keyword mapping from notes.content

jamesgorrie:
[Similar to this](https://github.com/wellcomecollection/catalogue/pull/677). Long text fields should not have a keyword mapping. Titles etc should for exact matching, which is why I am leaving them.

Once I get this index working I am going to do a more thorough audit of the mapping.


jamesgorrie:
Apologies, I didn't add the link, and hadn't got my English part of the brain turned on.

`alternativeTitles`, we really do as we want exact matches.

`lettering` - probably not, but maybe. Again pointing to the `Mary/Marie` example.

`physicalDescription` - maybe, I want to do some analysis on it to see if we have repeating terms, if so it could be useful.

PR #688, opened 2020-07-21
New prod API stack to enable long ARNs

kenoir:
This creates a new prod stack (with long long ARNs) it will receive no traffic until the API gateway is updated.

Required for: https://github.com/wellcomecollection/catalogue/pull/683

This change has been applied.


api/terraform/catalogue_api_prod_20200721/main.tf
```diff
@@ -0,0 +1,26 @@
+module "catalogue_api_prod_20200520" {
```
jamesgorrie, line 1: Could we not just call it prod now, like stage?

jamesgorrie: Actually, if it's been applied, no need.

PR #689, opened 2020-07-21
Update reindexer worker, add deployment tags

kenoir:
This terraform has been applied.


reindexer/terraform/reindex_worker/iam_policy_document.tf
```diff
@@ -31,6 +31,19 @@ data "aws_iam_policy_document" "sns_publish_policy" {
...
+    actions = [
+      "sqs:DeleteMessage",
+      "sqs:ReceiveMessage",
+    ]
+
+    resources = [
+      module.reindexer_queue.arn,
+    ]
+  }
+}
```
alexwlchan, line 45: You can also get this from `module.reindexer_queue.read_policy`.

PR #691, opened 2020-07-21
Use new tags in api task defs

kenoir:
Required for ECS deployment using the `weco-deploy` tool.


kenoir:
I believe this will not work as is because the container images are published to a different account from the one expected by the deploy tool. 

The tool will need to be updated to account for this.

kenoir:
Depends on: https://github.com/wellcomecollection/weco-deploy/pull/29

PR #693, opened 2020-07-21
Releasing the title mapping

jamesgorrie:


kenoir:
Not for now, but we should be using aliases no?

PR #694, opened 2020-07-22
Add an include for image stubs on works

jamieparkinson:


common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWorkImageInclude.scala
```diff
@@ -0,0 +1,19 @@
...
+import uk.ac.wellcome.models.work.internal.{Identified, Minted, UnmergedImage}
+
+@Schema(
+  name = "Image",
+  description = "A partial Image included on a work"
+)
+case class DisplayWorkImageInclude(
+  @Schema(description = "The image ID") id: String,
+  @JsonKey("type") @Schema(name = "type") ontologyType: String = "Image"
+)
```
warrd, line 14: We could potentially use the original `DisplayImage` model for this, by making the other fields optional. This would have the advantage of being more DRY, and easier to introduce more data to the include if required.

jamieparkinson: I don't think we can rely on wanting an exact subset of that model though, and those fields aren't optional when we get them from the images endpoint - think my preference is to keep it separate while we don't know what direction it's going to take rather than touching existing code

warrd: Cool. It was mainly the word "partial" it the description which indicated that to me, maybe worth changing the wording in that case?

jamieparkinson: Ah good point, will do

PR #695, opened 2020-07-23
RelatedWorkService

warrd:
## Issue

## Description

First part of the new archive API. Contains a service for querying work relations in elastic, a subsequent PR will use the `RelatedWorkService` for actually rendering the related data in the API.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
@@ -58,7 +58,7 @@ class ElasticsearchService(elasticClient: ElasticClient)(
   def executeSearchRequest(
     request: SearchRequest): Future[Either[ElasticError, SearchResponse]] =
     spanFuture(
-      name = "ElasticSearch#executeQuery",
+      name = "ElasticSearch#executeSearchRequest",
       spanType = "request",
       subType = "elastic",
       action = "query"
@@ -78,7 +78,7 @@ class ElasticsearchService(elasticClient: ElasticClient)(
   def executeMultiSearchRequest(request: MultiSearchRequest)
     : Future[Either[ElasticError, List[SearchResponse]]] =
     spanFuture(
-      name = "ElasticSearch#executeQuery",
+      name = "ElasticSearch#executeMultiSearchRequest",
       spanType = "request",
       subType = "elastic",
       action = "query"
@@ -95,13 +95,10 @@ class ElasticsearchService(elasticClient: ElasticClient)(
               error match {
                 case Some(err) => Left(err)
                 case None =>
-                  Right(
-                    results.collect {
-                      case Right(resp) =>
-                        transaction.addLabel("elasticTook", resp.took)
-                        resp
-                    }.toList
-                  )
+                  val responses = results.collect { case Right(resp) => resp }.toList
+                  val took = responses.map(_.took).sum
+                  transaction.addLabel("elasticTook", took)
+                  Right(responses)
               }
           }
         }
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
@@ -1,16 +1,22 @@
 package uk.ac.wellcome.platform.api.services
 
+import scala.concurrent.{ExecutionContext, Future}
+
 import co.elastic.apm.api.Transaction
 import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.requests.get.GetResponse
-import com.sksamuel.elastic4s.requests.searches.{SearchRequest, SearchResponse}
+import com.sksamuel.elastic4s.requests.searches.{
+  MultiSearchRequest,
+  MultiSearchResponse,
+  SearchRequest,
+  SearchResponse
+}
 import com.sksamuel.elastic4s.{ElasticClient, ElasticError, Index, Response}
 import grizzled.slf4j.Logging
+
 import uk.ac.wellcome.display.models._
 import uk.ac.wellcome.platform.api.Tracing
 import uk.ac.wellcome.platform.api.models._
-
-import scala.concurrent.{ExecutionContext, Future}
 
 case class ElasticsearchQueryOptions(filters: List[DocumentFilter],
                                      limit: Int,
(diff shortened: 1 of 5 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -69,6 +70,35 @@ class ElasticsearchService(elasticClient: ElasticClient)(
...
+        .map(toEither)
+        .map { response =>
+          response.right.flatMap { case MultiSearchResponse(items) =>
+             val results = items.map(_.response)
+             val error = results.collectFirst { case Left(err) => err }
+             error match {
+               case Some(err) => Left(err)
+               case None => Right(
+                 results.collect { case Right(resp) =>
+                   transaction.addLabel("elasticTook", resp.took)
```
warrd, line 93: @jamieparkinson any thoughts on how we want tracing to work with [ES multi search API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html)? An alternative would be to sum these values before adding it to the transaction.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/RelatedWorkRequestBuilder.scala
```diff
@@ -0,0 +1,106 @@
...
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.Index
+import com.sksamuel.elastic4s.requests.searches.{MultiSearchRequest, SearchRequest}
+import com.sksamuel.elastic4s.requests.searches.queries.Query
+
+case class RelatedWorkRequestBuilder(index: Index,
+                                     path: String,
+                                     maxRelatedWorks: Int = 1000) {
+
+  // To reduce response size and improve Elasticsearch performance we only
```
jamesgorrie, line 12: Not sure this improves Elasticsearch performance, but will reduce data across the network, in turn reducing latency.

warrd: Agree that will be there bigger issue, but reading less bytes from the disk / memory will result in a quicker response time.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -69,6 +70,35 @@ class ElasticsearchService(elasticClient: ElasticClient)(
         }
     }
 
+  def executeMultiSearchRequest(
+    request: MultiSearchRequest): Future[Either[ElasticError, List[SearchResponse]]] =
+    spanFuture(
+      name = "ElasticSearch#executeQuery",
+      spanType = "request",
+      subType = "elastic",
+      action = "query"
```
jamesgorrie, line 79: How would we tell the difference between these queries and the other ones? This span seems to be exactly the same as the other one.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/RelatedWorkRequestBuilder.scala
jamesgorrie: Indeed, just thought the comment should point to less bytes, not real ES perf... maybe they're the same though.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
warrd: Good point, how about `spanType = "multiRequest"`?

jamesgorrie: Makes sense, we might need to relook at these anyway as I am not clear what's being recorded in APM.
This will at least let us see if things are super wrong.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
jamesgorrie: Is this `took` per request?
Having them summed would be good, as that analysis in the console will be annoying.

warrd: Yup so currently there would be 3 separate timings (for the children, ancestors and siblings requests),

jamesgorrie: If we had an ID or something to then sum them, so we could average out on how long this whole request takes, that would be useful as that's what people will be experiencing.

warrd:
@jamesgorrie yes exactly: that is currently how things work with `?include=collection`, which this is replacing. Just to note that the longer term plan is to denormalise this data so that `parts`, `partOf`, `suceededBy` and `preceededBy` are stored directly in the index, so this method of making two requests to elasticsearch within an API request is going to be removed in the coming months.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
jamieparkinson: This sort of labelling (eg `multiRequest`) isn't what `spanType` is used for currently - it's a poorly documented field which has some special meanings for APM eg https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-transaction-set-type.

The current pattern, and what I'd prefer, would be to update `name` to `ElasticSearch#executeMultisearchRequest` or something like that. I _think_ it's OK to change `spanType` but just to flag that it's not necessarily as simple as free text, and we already distinguish between the spans using `name`.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -69,6 +75,38 @@ class ElasticsearchService(elasticClient: ElasticClient)(
...
+            case MultiSearchResponse(items) =>
+              val results = items.map(_.response)
+              val error = results.collectFirst { case Left(err) => err }
+              error match {
+                case Some(err) => Left(err)
+                case None =>
+                  Right(
+                    results.collect {
+                      case Right(resp) =>
+                        transaction.addLabel("elasticTook", resp.took)
```
jamieparkinson, line 101: It looks like this might not be doing what you think it is - it's repeatedly updating the value of `elasticTook`, once for each item in `results`. As you say, probably best to either sum them or - maybe better - sum them _and_ have them as separate labels so we can see how long the constituent queries took.

warrd: Ah ok thanks for clarifying what each field is used for.

I think I will just update it to sum the values: as this is something which isn't hanging around for long I don't reckon we need more granularity than that.

warrd:
@jamieparkinson updated now

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
jamesgorrie: > sum them and have them as separate labels so we can see how long the constituent queries took.

This please

PR #696, opened 2020-07-24
Include related works in API

warrd:
## Issue

## Description

Uses the `RelatedWorkService` introduced in #695 to expose related works in the API.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksController.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksController.scala
@@ -13,7 +13,9 @@ import uk.ac.wellcome.platform.api.models.ApiConfig
 import uk.ac.wellcome.platform.api.services.{
   CollectionService,
   ElasticsearchService,
-  WorksService
+  WorksService,
+  RelatedWorkService,
+  RelatedWorks
 }
 import uk.ac.wellcome.platform.api.Tracing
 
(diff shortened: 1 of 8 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksController.scala
```diff
@@ -65,13 +67,17 @@ class WorksController(
...
+              if (includes.anyRelation) {
+                retrieveRelatedWorks(index, work).map { relatedWorks =>
+                  workFound(work, relatedWorks, None, includes)
+                }
+              } else if (includes.collection) {
                 val expandedPaths = params._expandPaths.getOrElse(Nil)
-                retrieveTree(index, work, expandedPaths).map {
-                  workFound(work, _, includes)
+                retrieveTree(index, work, expandedPaths).map { tree =>
+                  workFound(work, None, tree, includes)
```
jamieparkinson, line 77: with this number of arguments it might be worth naming the parameters as we pass them - otherwise we're descending into positional argument hell

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksErrorsTest.scala
```diff
@@ -7,31 +7,31 @@ class WorksErrorsTest extends ApiWorksTestBase {
       assertIsBadRequest(
         "/works?include=foo",
         description =
-          "include: 'foo' is not a valid value. Please choose one of: ['identifiers', 'items', 'subjects', 'genres', 'contributors', 'production', 'notes', 'collection', 'images']"
+          "include: 'foo' is not a valid value. Please choose one of: ['identifiers', 'items', 'subjects', 'genres', 'contributors', 'production', 'notes', 'collection', 'images', 'parts', 'partOf', 'precededBy', 'succeededBy']"
```
jamieparkinson, line 10: bit of a bikeshed but these tests annoy me, I don't think they should break every time we add an include - do you reckon we should change them to (eg) just test whether the message starts with `include: 'foo' is not a valid value`?

warrd: I think its a nice feature of an API documenting what the valid values are. We could change the tests to not check the actual values, using a regex or something, maybe one for a later date.

PR #697, opened 2020-07-24
New multi matcher query 

jamesgorrie:
* Removed all the abstractions from how we create elastic queries as it seemed to start depicting what queries we'd write, which felt wrong
* Removed a whole bunch of tests that will now be covered by [rank eval](https://rank-teal.vercel.app/) (automation needed)


jamieparkinson:
Very much in favour of this - already looking nicer to reason about. Only comment so far is that I would remove the `ImagesQuery` and `WorksQuery` traits as they aren't doing anything - and if we do need something to express that sort of distinction I guess it would look more like a `QueryType[Image]` etc

jamesgorrie:
@jamieparkinson nice catch, I thought I'd done that. Definitely trying not to encode any meaning in that isn't used.

PR #699, opened 2020-07-27
Fix/weco deploy version

jamesgorrie:
Seems this was breaking the build as the defaults weren't set. Bumped the version while we're at it.


jamesgorrie:
Just like to be explicit

PR #700, opened 2020-07-27
Use the new version of scala-libs

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4596

In particular uses all the new-style locations code.

PR #703, opened 2020-07-28
Reduce the verbosity of logging in the tests

alexwlchan:
Spun out of https://github.com/wellcomecollection/catalogue/pull/700; part of https://github.com/wellcomecollection/platform/issues/4596


alexwlchan:
> Are we sure we don't want to be fixing warnings rather than turning them off?

We already run with warnings-as-errors at the Scala level; logged warnings are often innocuous, e.g. I think you get a warning log from the AWS SDK for every authentication method it tries but can't find – which is useful in aggregate (if it can't find *any*), but annoying in the noise.

PR #705, opened 2020-07-28
Make BigMessageSender conform to MessageSender

alexwlchan:
Follows #704; spun out of #700, the continuing steps towards https://github.com/wellcomecollection/platform/issues/4596

One of the big uses of ObjectLocation in the catalogue is in the big-messaging library; elsewhere it's mostly abstracted away behind things like the VersionedHybridStore classes. This patch tries to constrain the sprawl of big-messaging:

* **BigMessageSender now conforms to MessageSender**, our generic trait for sending messages (used for sending directly to SNS); this means we can swap it out for a pure, in-memory, sans-big-messaging version in tests that aren't testing big-messaging directly. Less Docker images to pull in tests.

  In turn, this means the inner workings of our apps are less coupled to big-messaging, if we wanted to move away from it in the future.

* **We only have an S3 implementation of BigMessageSender**; the in-memory implementation is gone. The S3 implementation is the one we actually care about, and makes the Location wrangling easier.

    We could make big-messaging generic against the object store, but I think it's way more likely that we'd spend the development time to do away with it entirely. I thus invoke YAGNI and say we simplify to one implementation, and do the work to make it more generic if we ever need it.

* A lot of the big-messaging tests were only testing other tests (e.g. a fixture that was only ever invoked by that test). All that code is gone now.

* This gets things in an easier state for swapping out ObjectLocation for the new-style S3ObjectLocation, which hopefully I'll be able to do next.

Although the diff is big, a lot of it is just de-denting code a few levels, now we don't have to set up an SNS topic and S3 bucket in every test.


alexwlchan:
This is enormous, so I'm going to break it down into smaller pieces (again).

alexwlchan:
I realise this PR is monstrously big; suggestions for how to break it into smaller parts welcomed.

PR #710, opened 2020-07-29
Concept indexing and contributor boosting

### LINKED ISSUE #709: Contributor / concepts fields
- [x] Add ascii folding to concepts
- [x] Boost contributors to more than title, but title above other concepts

jamesgorrie:
Fixes #709 

- adds ascii folding to labels so `remy` would match `Rémy`
- boost contributors above title so that searching for `Ursula Philip` would match things that have had her contribute to them over having them be in the title 

requires reindex


jamesgorrie:
> Just don't want a repeat of the subjects boosting issues.

What is this?

> It kind of feels like you want to scale down the boost dramatically as document frequency of a phrase increases

The boost in this case is just saying that a signal to that field is more valuable than from other fields. Not too sure this applies here if I have understood what you've said completely.

jamieparkinson:
An example of the issue - before the current title boosting search, this work https://wellcomecollection.org/works/wc4dm8e5 was coming up top for a search for "cats" because of the subject match. So just wanted to know if there were any weird, overly-general, contributors that would cause similarly unexpected results.

Re boosting I'm thinking aloud of ways to stop that kind of problem - so if there's a subject/contributor/whatever that crops up all over the place we don't want the boost. Some kind of way of expressing, "a signal from that field is more valuable than from other fields _unless it's a value that crops up everywhere_" - although I know that's not how it actually works.

jamesgorrie:
@jamieparkinson hopefully that's less of an issue as we aren't boosting `Concept`s, but rather contributors. This could be organisations too, so probably worth a look about (Given an org might generally have more general terms in it's name).

PR #717, opened 2020-07-30
Create Hypotheses

taceybadgerbrook:
This page is to form the first page in a new section called, "Hypotheses"


docs/search/Hypotheses
```diff
@@ -0,0 +1,9 @@
+#Exploring
```
jamesgorrie, line 1:
```suggestion
# Exploring
```

docs/search/Hypotheses
```diff
@@ -0,0 +1,9 @@
+# Exploring
```
jamesgorrie, line 1:
```suggestion
# Exploring
```

jamesgorrie:
In favour of #720

PR #719, opened 2020-07-30
Create Query design

taceybadgerbrook:
This is the first page in a new section called "Query design"


jamesgorrie:
In favour of #720

PR #721, opened 2020-07-31
deploy new pipeline

### LINKED ISSUE #722: Deploy new pipeline with ID search

jamesgorrie:
Fixes #722 

New ID search work.

Also moves search query to 100%. I wouldn't suggest we do this, but it is so much more better than what we had, this is the baseline, and we will move forward from here.

### DIFF
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -1,7 +1,7 @@
-module "catalogue_pipeline_20200717" {
+module "catalogue_pipeline_20200731" {
   source = "./stack"
 
-  namespace = "catalogue-20200717"
+  namespace = "catalogue-20200731"
 
   release_label = "prod"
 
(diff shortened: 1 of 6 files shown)


pipeline/terraform/main.tf
```diff
@@ -62,6 +62,70 @@ module "catalogue_pipeline_20200717" {
...
+  # reindexer topic names.
+
+  sierra_adapter_topic_arns = [
+    local.sierra_reindexer_topic_arn,
+    local.sierra_merged_bibs_topic_arn,
+    local.sierra_merged_items_topic_arn,
+  ]
+
+  miro_adapter_topic_arns = [
+    local.miro_reindexer_topic_arn,
```
warrd, line 92: Are these meant to be uncommented?

jamesgorrie: Yip as they'll be reading from the reindexer

warrd: Just for initial reindex though right? Once its all done we have to remember to comment them again?

jamesgorrie: Indeed, I'll swap that to prod when that's done, just wanted to get this in in case people are working on the pipeline, and don't want them to destroy it.

This deployment process is a little shaky, needs a rethink at some stage.

PR #728, opened 2020-08-03
Run inference tasks on EC2

jamieparkinson:
Required in order to fulfil the latest changes in https://github.com/wellcomecollection/docs/blob/8602f2ed4a09293aa8a0f0922e5c6cd48fe3eb09/rfcs/021-data_science_in_the_pipeline/README.md

### DIFF
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -92,7 +92,7 @@ module "catalogue_pipeline_20200727" {
   ]
 
   miro_adapter_topic_arns = [
-    //    local.miro_reindexer_topic_arn,
+    local.miro_reindexer_topic_arn,
     local.miro_updates_topic_arn,
   ]
 
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -92,7 +92,7 @@ module "catalogue_pipeline_20200727" {
   ]
 
   miro_adapter_topic_arns = [
-    local.miro_reindexer_topic_arn,
+    //    local.miro_reindexer_topic_arn,
     local.miro_updates_topic_arn,
   ]
 
--- a/pipeline/terraform/stack/cluster.tf
+++ b/pipeline/terraform/stack/cluster.tf
@@ -4,7 +4,7 @@ resource "aws_ecs_cluster" "cluster" {
 }
 
 module "inference_capacity_provider" {
-  source = "../../../../terraform-aws-ecs-service/modules/ec2_capacity_provider"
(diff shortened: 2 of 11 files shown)


pipeline/terraform/main.tf
```diff
@@ -92,7 +92,7 @@ module "catalogue_pipeline_20200727" {
   ]
 
   miro_adapter_topic_arns = [
-    //    local.miro_reindexer_topic_arn,
+    local.miro_reindexer_topic_arn,
```
kenoir, line 95: Presumably this is back on because with this change you want to be able to reindex miro things?

pipeline/terraform/stack/cluster.tf
```diff
@@ -1,3 +1,26 @@
...
+  source = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/ec2_capacity_provider?ref=v3.1.0"
+
+  name = "${local.namespace_hyphen}_inference_capacity_provider"
+
+  // Setting this variable from aws_ecs_cluster.cluster.name creates a cycle
+  // The cluster name is required for the instance user data script
+  // This is a known issue https://github.com/terraform-providers/terraform-provider-aws/issues/12739
+  cluster_name = local.namespace_hyphen
+
+  instance_type           = "c5.2xlarge"
```
kenoir, line 16: very premium (@ 34c an hour) - but fine if this spends most of the time scaled to 0.

jamieparkinson: The price should work out slightly less than what Fargate was costing us - this is also a spot instance

pipeline/terraform/main.tf
jamieparkinson: It is mistake

pipeline/terraform/stack/cluster.tf
jamieparkinson: My workings:
- Maxed out Fargate was $0.197 / hr, we ran <= 8 of them
- At spot pricing a `c5.2xlarge` is *maximum* $0.34, but actually has been consistently between $0.15-0.20 (over the last 3 months). We run <= 4 of them.
- I chose a 2xlarge (big enough for 2 tasks) as a compromise: for full reindexes we want maximum horizontal scaling to maximise throughput (so we fit 2 tasks on each host, 8 tasks max), in general use we want maximum speed at low-ish throughput so we give the 1 or 2 tasks we expect to see plenty of POWER. Anecdotally it seemed to run 1.5x-2x as fast as Fargate as I believe Pytorch is able to exploit certain features of the c5 hardware.

jamieparkinson: More important than that logic is that I intend to keep an eye on utilisation and throughput and tweak these values where necessary!

PR #740, opened 2020-08-07
use TransformedBaseWork type param on sendT for proper serialisation

jamesgorrie:
I've gone through the codebase and it's not used anywhere else with anything bar a `TransformedBaseWork`.
Not sure where to write the test, might try get the fix in and do that.

Chaars @warrd and @alexwlchan for the explanations on this.


jamesgorrie:
> I've gone through the codebase and it's not used anywhere else with anything bar a TransformedBaseWork.

From the PR  description

jamesgorrie:
@warrd I've added a test testing we can decode from `TransformedBaseWork` as that's what's needed on the recorder side.

PR #743, opened 2020-08-10
Buildkite

kenoir:
Configuration for CI in buildkite.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,33 @@
+steps:
+  - command: .buildkite/scripts/run_autoformat.py
+  - wait
+  - command: .buildkite/scripts/run_job.py internal_model
+  - command: .buildkite/scripts/run_job.py display
+  - command: .buildkite/scripts/run_job.py elasticsearch
+  - command: .buildkite/scripts/run_job.py big_messaging
+  - command: .buildkite/scripts/run_job.py ingestor_common
+  - command: .buildkite/scripts/run_job.py transformer_common
+  - wait
+  - command: .buildkite/scripts/run_job.py api
+  - command: .buildkite/scripts/run_job.py reindex_worker
+  - command: .buildkite/scripts/run_job.py ingestor_works
+  - command: .buildkite/scripts/run_job.py ingestor_images
+  - command: .buildkite/scripts/run_job.py id_minter
+  - command: .buildkite/scripts/run_job.py matcher
+  - command: .buildkite/scripts/run_job.py merger
+  - command: .buildkite/scripts/run_job.py recorder
+  - command: .buildkite/scripts/run_job.py transformer_miro
+  - command: .buildkite/scripts/run_job.py transformer_sierra
+  - command: .buildkite/scripts/run_job.py transformer_mets
+  - command: .buildkite/scripts/run_job.py transformer_calm
+  - command: .buildkite/scripts/run_job.py sierra_reader
+  - command: .buildkite/scripts/run_job.py sierra_bib_merger
+  - command: .buildkite/scripts/run_job.py sierra_item_merger
+  - command: .buildkite/scripts/run_job.py sierra_items_to_dynamo
+  - command: .buildkite/scripts/run_job.py mets_adapter
+  - command: .buildkite/scripts/run_job.py calm_adapter
+  - command: .buildkite/scripts/run_job.py inference_manager
+  - command: .buildkite/scripts/run_job.py feature_inferrer --changes-in pipeline/inferrer/feature_inferrer
+  - command: .buildkite/scripts/run_job.py snapshot_generator
+  - wait
+  - command: .buildkite/scripts/complete_build.py
--- a/.buildkite/scripts/git_utils.py
+++ b/.buildkite/scripts/git_utils.py
@@ -1,0 +1,59 @@
+# -*- encoding: utf-8
+
+import os
+import subprocess
+import sys
+
+
+def git(*args):
+    """Run a Git command and return its output."""
+    cmd = ["git"] + list(args)
+    try:
+        return subprocess.check_output(cmd).decode("utf8").strip()
+    except subprocess.CalledProcessError as err:
+        sys.exit(err.returncode)
+
+
+def get_changed_paths(*args, globs=None):
+    """
+    Returns a set of changed paths in a given commit range.
+
+    :param args: Arguments to pass to ``git diff``.
+    :param globs: List of file globs to include in changed paths.
+    """
+    if globs:
+        args = list(args) + ["--", *globs]
+    diff_output = git("diff", "--name-only", *args)
+
+    return set([line.strip() for line in diff_output.splitlines()])
+
+
+def remote_default_branch():
+    """Inspect refs to discover default branch @ remote origin."""
+    return git("symbolic-ref", "refs/remotes/origin/HEAD").split("/")[-1]
+
+
+def remote_default_head():
+    """Inspect refs to discover default branch HEAD @ remote origin."""
+    return git("show-ref", f"refs/remotes/origin/{remote_default_branch()}", "-s")
+
+
+def local_current_head():
+    """Use rev-parse to discover hash for current commit AKA HEAD (from .git/HEAD)."""
+    return git("rev-parse", "HEAD")
+
+
+def get_sha1_for_tag(tag):
+    git("fetch")
+    return git("show-ref", "-s", tag)
+
+
+def current_branch():
+    return os.environ["BUILDKITE_BRANCH"]
+
+
+def is_default_branch():
+    current_branch_name = current_branch()
+    default_branch_name = remote_default_branch()
+
+    return current_branch_name == default_branch_name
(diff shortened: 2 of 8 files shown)


kenoir:
See https://buildkite.com/wellcome-collection/catalogue/builds/61 for a manual run of this branch.

kenoir:
Waiting for https://github.com/wellcomecollection/dockerfiles/pull/95, then this PR should be updated to remove the `buildkite` label from the `sbt_wrapper` image.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,33 @@
...
+  - command: .buildkite/scripts/run_job.py sierra_bib_merger
+  - command: .buildkite/scripts/run_job.py sierra_item_merger
+  - command: .buildkite/scripts/run_job.py sierra_items_to_dynamo
+  - command: .buildkite/scripts/run_job.py mets_adapter
+  - command: .buildkite/scripts/run_job.py calm_adapter
+  - command: .buildkite/scripts/run_job.py inference_manager
+  - command: .buildkite/scripts/run_job.py feature_inferrer --changes-in pipeline/inferrer/feature_inferrer
+  - command: .buildkite/scripts/run_job.py snapshot_generator
+  - wait
+  - command: .buildkite/scripts/complete_build.py
```
warrd, line 33: Is there anything in particular I need to do to set up `relation_embedder` beyond adding a `run_job` command here?

.buildkite/scripts/sbt_dependency_tree.py
```diff
@@ -0,0 +1,70 @@
...
+            folders |= d.all_folders()
+        return folders
+
+    def depends_on(self, other):
+        if self == other:
+            return True
+        if other in self.dependencies:
+            return True
+        for d in self.dependencies:
+            return d.depends_on(other)
```
warrd, line 34: Hmm this will return on the first iteration of the loop, is this what you actually want?

warrd: Maybe it should be more like this?

```python
return (
    self == other or
    other in self.dependencies or
    any(d.depends_on(other) for d in self.dependencies))
)
```

.buildkite/scripts/sbt_dependency_tree.py
```diff
@@ -0,0 +1,70 @@
...
+        self.metadata_dir = metadata_dir
+        self.projects = {}
+
+        self._get_all_projects()
+
+    def _get_all_projects(self):
+        for f in os.listdir(self.metadata_dir):
+            self.get_project(name=os.path.splitext(f)[0])
+
+    def get_project(self, name):
```
warrd, line 49: bikeshed: `get_project` indicates to me this returns something, i would prefer this to be called `init_project` or something like that

.buildkite/pipeline.yml
kenoir: The `run_job.py` script does the same thing as the one for `Travis` essentially so you still need to ensure there are `make` tasks available.

.buildkite/scripts/sbt_dependency_tree.py
kenoir: This is copied without modification from the Travis build, so i'm not certain (other than that this will reproduce existing behaviour). Perhaps @alexwlchan can shed some light.

.buildkite/scripts/git_utils.py
```diff
@@ -0,0 +1,59 @@
...
+    Returns a set of changed paths in a given commit range.
+
+    :param args: Arguments to pass to ``git diff``.
+    :param globs: List of file globs to include in changed paths.
+    """
+    if globs:
+        args = list(args) + ["--", *globs]
+    diff_output = git("diff", "--name-only", *args)
+
+    return set([line.strip() for line in diff_output.splitlines()])
```
alexwlchan, line 28: nit: there's syntactic sugar for creating sets from comprehensions:

```suggestion
    return {line.strip() for line in diff_output.splitlines()}
```

.buildkite/scripts/git_utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def local_current_head():
+    """Use rev-parse to discover hash for current commit AKA HEAD (from .git/HEAD)."""
+    return git("rev-parse", "HEAD")
+
+
+def get_sha1_for_tag(tag):
+    git("fetch")
+    return git("show-ref", "-s", tag)
```
alexwlchan, line 48: minor: prefer explicit flag names to short options:

```suggestion
    return git("show-ref", "--hash", tag)
```

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+"""
+
+import os
+import subprocess
+import sys
+
+
+def branch_name():
+    """Return the name of the branch under test."""
+    return os.environ["BUILDKITE_BRANCH"]
```
alexwlchan, line 17: minor: I wonder if it's worth pulling stuff like this into a separate file, say `provider.py`?

Then this file would have `from provider import branch_name`, and all the provider-specific code is in one place (and can be swapped out again later).

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def git(*args):
+    """Run a Git command and return its output."""
+    cmd = ["git"] + list(args)
+    try:
+        return subprocess.check_output(cmd).decode("utf8").strip()
+    except subprocess.CalledProcessError as err:
+        print(err)
+        sys.exit(err.returncode)
```
alexwlchan, line 40:
```
from git_utils import git
```

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+
+def get_changed_paths(*args):
+    """
+    Returns a set of changed paths in a given commit range.
+
+    :param commit_range: Arguments to pass to ``git diff``.
+    """
+    diff_output = git("diff", "--name-only", *args)
+
+    return set([line.strip() for line in diff_output.splitlines()])
```
alexwlchan, line 56:
```
from git_utils import get_changed_paths
```

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+if __name__ == "__main__":
+
+    make("format")
+
+    # If there are any changes, push to GitHub immediately and fail the
+    # build.  This will abort the remaining jobs, and trigger a new build
+    # with the reformatted code.
+    if get_changed_paths():
```
alexwlchan, line 66: minor: maybe replace this with a method that returns a boolean `has_uncommitted_changes()`?

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+        print("*** There were changes from formatting, creating a commit")
+
+        git("config", "user.name", "Buildkite on behalf of Wellcome")
+        git("config", "user.email", "wellcomedigitalplatform@wellcome.ac.uk")
+
+        git(
+            "remote",
+            "add",
+            "ssh-origin",
+            "git@github.com:wellcomecollection/catalogue.git",
```
alexwlchan, line 76: I'd pull this out into a constant somewhere, so it's easy to see what we have to change when we port this to other repos.

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+        # We checkout the branch before we add the commit, so we don't
+        # include the merge commit that Buildkite makes.
+        git("fetch", "ssh-origin")
+        git("checkout", "--track", f"origin/{branch_name()}")
+
+        git("add", "--verbose", "--update")
+        git("commit", "-m", "Apply auto-formatting rules")
+        git("push", "ssh-origin", "HEAD:%s" % branch_name())
```
alexwlchan, line 86: nit: let's be consistent with how we format strings (with three lines above)

```suggestion
        git("push", "ssh-origin", f"HEAD:{branch_name()}")
```

.buildkite/scripts/run_job.py
```diff
@@ -0,0 +1,129 @@
...
+        return True
+    if any(p.startswith("project/") for p in interesting_paths):
+        print("*** Relevant: project/")
+        return True
+
+    for path in interesting_paths:
+        if path.startswith((".buildkite", "docs/")):
+            continue
+
+        if path.endswith(".tf"):
```
alexwlchan, line 51: minor: I've just discovered we trigger a full storage-service rebuild for Python scripts which cannot affect SBT projects. 🤦‍♀️ 

```suggestion
        if path.endswith((".py", ".tf")):
```

.buildkite/scripts/sbt_dependency_tree.py
alexwlchan: Yeah, this is probably a bug.

kenoir:
@alexwlchan i'm going to merge here and address review comments in an upcoming PR.

PR #745, opened 2020-08-10
Add a build badge ... check if Buildkite status appears

kenoir:
Adds a build badge, really a test for triggering builds in Buildkite and viewing status.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,33 @@
+steps:
+  - command: .buildkite/scripts/run_autoformat.py
+  - wait
+  - command: .buildkite/scripts/run_job.py internal_model
+  - command: .buildkite/scripts/run_job.py display
+  - command: .buildkite/scripts/run_job.py elasticsearch
+  - command: .buildkite/scripts/run_job.py big_messaging
+  - command: .buildkite/scripts/run_job.py ingestor_common
+  - command: .buildkite/scripts/run_job.py transformer_common
+  - wait
+  - command: .buildkite/scripts/run_job.py api
+  - command: .buildkite/scripts/run_job.py reindex_worker
+  - command: .buildkite/scripts/run_job.py ingestor_works
+  - command: .buildkite/scripts/run_job.py ingestor_images
+  - command: .buildkite/scripts/run_job.py id_minter
+  - command: .buildkite/scripts/run_job.py matcher
+  - command: .buildkite/scripts/run_job.py merger
+  - command: .buildkite/scripts/run_job.py recorder
+  - command: .buildkite/scripts/run_job.py transformer_miro
+  - command: .buildkite/scripts/run_job.py transformer_sierra
+  - command: .buildkite/scripts/run_job.py transformer_mets
+  - command: .buildkite/scripts/run_job.py transformer_calm
+  - command: .buildkite/scripts/run_job.py sierra_reader
+  - command: .buildkite/scripts/run_job.py sierra_bib_merger
+  - command: .buildkite/scripts/run_job.py sierra_item_merger
+  - command: .buildkite/scripts/run_job.py sierra_items_to_dynamo
+  - command: .buildkite/scripts/run_job.py mets_adapter
+  - command: .buildkite/scripts/run_job.py calm_adapter
+  - command: .buildkite/scripts/run_job.py inference_manager
+  - command: .buildkite/scripts/run_job.py feature_inferrer --changes-in pipeline/inferrer/feature_inferrer
+  - command: .buildkite/scripts/run_job.py snapshot_generator
+  - wait
+  - command: .buildkite/scripts/complete_build.py
--- a/.buildkite/scripts/run_autoformat.py
+++ b/.buildkite/scripts/run_autoformat.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+# -*- encoding: utf-8
+"""
+This script does autoformatting in Buildkite on pull requests.
+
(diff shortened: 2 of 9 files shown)


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,33 @@
...
+  - command: .buildkite/scripts/run_job.py sierra_bib_merger
+  - command: .buildkite/scripts/run_job.py sierra_item_merger
+  - command: .buildkite/scripts/run_job.py sierra_items_to_dynamo
+  - command: .buildkite/scripts/run_job.py mets_adapter
+  - command: .buildkite/scripts/run_job.py calm_adapter
+  - command: .buildkite/scripts/run_job.py inference_manager
+  - command: .buildkite/scripts/run_job.py feature_inferrer --changes-in pipeline/inferrer/feature_inferrer
+  - command: .buildkite/scripts/run_job.py snapshot_generator
+  - wait
+  - command: .buildkite/scripts/complete_build.py
```
jamesgorrie, line 33: Maybe I am being silly, but if we have two branches, branch `A` and `B`. If `B` finishes, then `A` finishes, would `B` not be latest although maybe not latest?

Perhaps I am missing the point as to why it exists though.

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+    make("format")
+
+    # If there are any changes, push to GitHub immediately and fail the
+    # build.  This will abort the remaining jobs, and trigger a new build
+    # with the reformatted code.
+    if get_changed_paths():
+        print("*** There were changes from formatting, creating a commit")
+
+        git("config", "user.name", "Buildkite on behalf of Wellcome")
+        git("config", "user.email", "wellcomedigitalplatform@wellcome.ac.uk")
```
jamesgorrie, line 70: For future reference, is there a key in the buildkite config?

.buildkite/scripts/run_autoformat.py
```diff
@@ -0,0 +1,97 @@
...
+        # We checkout the branch before we add the commit, so we don't
+        # include the merge commit that Travis makes.
+        git("fetch", "ssh-origin")
+        git("checkout", "--track", f"origin/{branch_name()}")
+
+        git("add", "--verbose", "--update")
+        git("commit", "-m", "Apply auto-formatting rules")
+        git("push", "ssh-origin", "HEAD:%s" % branch_name())
+
+        # We exit here to fail the build, so Travis will skip to the next
```
jamesgorrie, line 88:
```suggestion
        # We exit here to fail the build, so Buildkite will skip to the next
```

.buildkite/scripts/run_job.py
```diff
@@ -0,0 +1,129 @@
...
+    """Run a Make command, and check it completes successfully."""
+    check_call(["make"] + list(args))
+
+
+def should_run_sbt_project(repo, project_name, changed_paths):
+    project = repo.get_project(project_name)
+
+    interesting_paths = [p for p in changed_paths if not p.startswith(".sbt_metadata")]
+
+    if ".travis.yml" in interesting_paths:
```
jamesgorrie, line 37: Is this still going to work? I suppose it would, but maybe worth moving over to another config file?

.buildkite/scripts/run_job.py
```diff
@@ -0,0 +1,129 @@
...
+        return True
+    if "build.sbt" in interesting_paths:
+        print("*** Relevant: build.sbt")
+        return True
+    if any(p.startswith("project/") for p in interesting_paths):
+        print("*** Relevant: project/")
+        return True
+
+    for path in interesting_paths:
+        if path.startswith((".travis", "docs/")):
```
jamesgorrie, line 48: Same here, no sure this actually matches anything though.

.buildkite/pipeline.yml
kenoir: Yes you are correct. A race condition might occur producing an inaccurate value which could result in more tests/builds being run than necessary (but not the other way round I think). I'm ok with that outcome for now.

jamesgorrie: Super, just checking I understood.

.buildkite/scripts/run_autoformat.py
kenoir: Yes, ish. Buildkite loads a weco-bot key present in S3 into the ssh-agent - there is no key on disk. I chose to allow it to pull it out of an S3 bucket over the SecretsManager solution which is quite a bit more involved.

.buildkite/scripts/run_job.py
kenoir: No. It's not going to work, it should reference `.buildkite/pipeline.yml`.

README.md
```diff
@@ -2,6 +2,7 @@
 
 [![Build Status](https://travis-ci.org/wellcomecollection/catalogue.svg?branch=master)](https://travis-ci.org/wellcomecollection/catalogue)
 
+[![Build status](https://badge.buildkite.com/0ca819db1215b66ecb17019d8ee5331d8e537094d051141219.svg)](https://buildkite.com/wellcome-collection/catalogue)
```
jamesgorrie, line 5: This isn't working, so I assume needs an updated location?

kenoir:
@jamesgorrie I was expecting reviews on the other PR (this one's a test for something else) i've folded changes from your review into the other PR and will close this one - the feedback is appreciated 👍

jamesgorrie:
@kenoir not sure how I got that confused, sorry for the hassle.

PR #747, opened 2020-08-11
Clean up Buildkite scripts a little

kenoir:


warrd:
What is the plan for the bug in `Project` class in `sbt_dependency_tree.py`?

kenoir:
> What is the plan for the bug in Project class in sbt_dependency_tree.py?

Going to stare at it for a bit.

PR #748, opened 2020-08-11
Download images in inference manager

jamieparkinson:
This partially fulfils the [updated RFC](https://github.com/wellcomecollection/docs/tree/master/rfcs/021-data_science_in_the_pipeline) for data science in the pipeline by downloading images in the manager rather than the inferrer. Next step will be to add another inferrer for colour palettes.


warrd:
Any chance of a description / link to issue number? Not entirely sure what I am reviewing

jamieparkinson:
@warrd good point, I've updated the description and linked to the RFC

pipeline/inferrer/inference_manager/src/main/scala/uk/ac/wellcome/platform/inference_manager/services/InferenceManagerWorkerService.scala
```diff
@@ -34,49 +32,59 @@ class InferenceManagerWorkerService[Destination, Input, Output](
   def run(): Future[Done] =
     msgStream.runStream(
       className,
-      _.via(createRequest)
-        .via(inferrerClientFlow)
+      _.asSourceWithContext {
```
warrd, line 35: Ah nice, this is the functionality I was hoping to use in other places using akka streams and our `MessageStream`, but we were on an older version of akka. Do you think it helps simplify the code rather than passing tuples down between each flow stage?

jamieparkinson: I'm kind of in 2 minds about it - where it works well, it's absolutely an improvement, but where you have to do "interop" with normal flows (like for the request pools) it's a bit of a mess. 

Since I was trying to do flows everywhere in this application it seemed like a good idea so I didn't have `Message` tuples in unexpected places, but in the cases where the flow is only at the top level I might be inclined not to bother...

PR #759, opened 2020-08-14
Terraform 0.13 upgrade

jamieparkinson:
This is what `terraform 0.13upgrade` did, I also had to edit a load of state files manually because of [this](https://github.com/hashicorp/terraform/issues/25752) - seems to be OK now 🤞


kenoir:
I've just discovered this issue myself in https://github.com/wellcomecollection/platform-infrastructure. Editing the state manually to fix this problem is really 😱 .

PR #787, opened 2020-08-18
Perform matcher locking on graph components rather than nodes

jamieparkinson:
_DISCLAIMER:_ This might not actually be correct behaviour, but I _think_ it is


jamesgorrie (approved):
_Seems_ to make sense to me.

PR #788, opened 2020-08-18
Create boosting.md

taceybadgerbrook:
Creating simplified dictionary of search algorithm tools separate from Relevance tests.

### DIFF
--- a/docs/search/query-design/boosting.md
+++ b/docs/search/query-design/boosting.md
@@ -1,0 +1,1 @@
+Boosts items in search result order by attributing higher scores for fields matching words or “tokens” used in query. Results are sorted by scores in descending order.


docs/search/query-design/boosting.md
```diff
@@ -0,0 +1 @@
+Boosts items in search result order by attributing higher scores for fields matching words or “tokens” used in query. Results are sorted by scores in descending order.
```
jamesgorrie, line 1: Could we put the title of the document in here?

```suggestion
# Boosting

Boosts items in search result order by attributing higher scores for fields matching words or “tokens” used in query. Results are sorted by scores in descending order.
```

jamesgorrie: We could also link through to the elastic documentation?
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-boost.html

PR #794, opened 2020-08-18
pipeline_storage module

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4672

## Description

The existing `ingestor_common` module is copied across to a new `pipeline_storage` module, which will be used for imlementation of the [Pipeline Intermediate Storage RFC](https://github.com/wellcomecollection/docs/pull/38). Subsequent PRs will point the ingestors to the new module and remove the `ingestor_common` module.


common/pipeline_storage/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+sqs:
+  image: s12v/elasticmq
+  ports:
+    - "9324:9324"
```
alexwlchan, line 4: I haven't read down yet, but I'm immediately curious about what a messaging service is doing in a storage module.

warrd: Yes I also agree with this I think there is code in here which seems more specific than we would want for `pipeline_storage`. How about pulling out only some of the code, and leaving `ingestor_common` to have the service related stuff?

alexwlchan: I think that’d be much better – have `ingestor_common` depend upon this module, so the ingestors still get all its goodies, but the enricher-or-whatever-it’s-called-now doesn’t need the ingestor services.

warrd: Ok will give it a go, hopefully quite easily untangled

jamesgorrie (commented):
There's loads of classes / objects called `IngestorBlah`.

Do we want to generify those bits of the service as ingestors are going to be dependant on this lib?

warrd:
Yep good point, I think in a future PR might make sense to change some of the naming. Up for suggestions.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/builders/IngestorConfigBuilder.scala
```diff
@@ -0,0 +1,41 @@
...
+import uk.ac.wellcome.pipeline_storage.models.IngestorConfig
+
+import scala.concurrent.duration._
+import scala.util.Try
+
+object IngestorConfigBuilder {
+  def buildIngestorConfig(config: Config): IngestorConfig = {
+
+    // TODO: Work out how to get a Duration from a Typesafe flag.
+    val flushInterval = 1 minute
```
jamesgorrie, line 13: Assuming for now this will be the same for all PS services?

warrd: This will probably need to be modified per service, good spot

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/exceptions/IngestorException.scala
```diff
@@ -0,0 +1,8 @@
+package uk.ac.wellcome.pipeline_storage.exceptions
+
+case class IngestorException(e: Throwable) extends Exception(e.getMessage)
+
+case object IngestorException {
```
jamesgorrie, line 5: What's the benefit of this pattern?

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/IndexerTest.scala
```diff
@@ -0,0 +1,164 @@
...
+
+    val analysis = WorksAnalysis()
+
+    val title = textField("title")
+    val data = objectField("data")
+
+    val mapping = properties(Seq(title, canonicalId, version, data))
+      .dynamic(DynamicMapping.Strict)
+  }
+}
```
jamesgorrie, line 164: Assuming we will be creating an API over the storage layer, and be able to write in memory tests?

jamesgorrie (approved):

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/exceptions/IngestorException.scala
warrd: What part do you mean? If it is that it is a `case object` rather than standard `object` I agree that looks unnecessary, or do you mean something else?

jamesgorrie: Ah, no, just wrapping the exceptions in a case class that doesn't seem to do much else.

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/IndexerTest.scala
warrd: Yep :+1:

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/exceptions/IngestorException.scala
warrd: I guess the difference is that we can specify a specific exception type (rather than generic Exception), and can also pattern match on it, but not entirely sure on the original intention.

On a more general note, error handling in our code is very variable, and I think at some point might make sense to come up with some common patterns that we use everywhere.

jamesgorrie: Yip - definitely think that it would be good to do that as context switching / inventing another way takes up unnecessary time on what is a pretty well trodden domain.

PR #803, opened 2020-08-19
add ability to add search templates on index creation

jamesgorrie:
Allows us to store a [search template](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html#search-template-api-query-params) along with the index being created.

The `rank_eval` app will then look up the template to use against the index, and report back on it.

This is by no means perfect, but gets us into a place to start testing it and iterating on it.

The Elastic4s stuff - I am waiting on hearing back from them as to what approach to take as the search template API has changed significantly, and actually uses 2 APIs, and I want to know if they want to replicate that in the lib.


common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksQuery.scala
```diff
@@ -0,0 +1,57 @@
+package uk.ac.wellcome.elasticsearch
```
jamieparkinson, line 1: Has this been removed in the API project? Might be being stupid but can't see it

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksQuery.scala
```diff
@@ -0,0 +1,57 @@
...
+import com.sksamuel.elastic4s.requests.common.Operator.{AND, OR}
+import com.sksamuel.elastic4s.requests.searches.queries.BoolQuery
+import com.sksamuel.elastic4s.requests.searches.queries.matches.MultiMatchQueryBuilderType.BEST_FIELDS
+import com.sksamuel.elastic4s.requests.searches.queries.matches.{
+  FieldWithOptionalBoost,
+  MultiMatchQuery
+}
+import uk.ac.wellcome.elasticsearch.WorksAnalysis.whitespaceAnalyzer
+
+case object WorksMultiMatcher {
```
jamieparkinson, line 13: Has this been removed in the API project? Might be being stupid but can't see it

jamesgorrie: Indeed, fixed.

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ImagesServiceTest.scala
```diff
@@ -47,7 +47,7 @@ class ImagesServiceTest
   }
 
   describe("retrieveSimilarImages") {
-    it("gets visually similarImages") {
+    ignore("gets visually similarImages") {
```
jamesgorrie, line 50: We've ignored this as @jamieparkinson is working on making these better.

warrd, line 50: Is this test coming back later? Maybe either remove it or add a comment why it's ignored.

jamieparkinson: I'm working on a PR that includes this right now

jamesgorrie: Just left a note, @jamieparkinson is on writing more meaningful features for this, so waiting on that.

It fails given the version we are running in prod. Makes me think running `latest` elastic containers could be a good thing... Or be more conservative when updating the catalogue cluster.

warrd (approved):
Looks good, but for someone like me who hasn't been involved in all the search work I think could be some more documentation about the different classes etc to help give context about what is going on.

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ImagesServiceTest.scala
warrd: :+1:

jamesgorrie:
@warrd good point, it needs that as I think it would probably lead to a little tidy up and refactoring. Coming up.

PR #805, opened 2020-08-20
Inferrer fixes

jamieparkinson:
Notably I forgot to include an important environment variable - also had to change some mysterious numbers to stop akka-http erroring


jamesgorrie (approved):
bloop

PR #807, opened 2020-08-20
Remove 20200819 pipeline

jamieparkinson:
It has done its job and the index is full of bad things


alexwlchan (approved):
One of the bad things it contains is a reminder that we're already into late August. 😭

jamieparkinson:
I propose we change the pipeline/index naming conventions to protect our collective mental health

alexwlchan:
> I propose we change the pipeline/index naming conventions to protect our collective mental health

Particularly when pipelines are created for specific purposes, I wouldn't mind naming that way.

e.g. `catalogue-loris_expts`, `catalogue-inferrer_morememory`

jamesgorrie:
[We used to call the indexes along those lines](https://github.com/wellcomecollection/catalogue/commit/ba78bdcb5f4ea6f33e3d82c5ae95600b3cc322d9#diff-abff2a9b44708c7537e4a5e72b530571R30)

That's probably a better policy for the stack.

jamesgorrie:
You could also just version it? 1,2,3,4,5 etc?

warrd:
I like having dates in there personally, but maybe having additional context too would be useful

alexwlchan:
We used version numbers in the strings because for a while we ran a V1 and a V2 API concurrently. The version numbers were to distinguish which API an index was attached to, because they had different internal models.

Nick’s right – the date is useful, especially when we need to delete old indexes. Being able to see when an index was created (and when it was intended to be used) is helpful.

Personally my preferred approach would be something like

```
catalogue-{date}-{description}
```

PR #809, opened 2020-08-21
Forgotten palette improvements

harrisonpim:
Somehow I left these final updates to the palette inferrer out of my branch in the data-science repo.
Bringing them back in over here.

### DIFF
--- a/pipeline/inferrer/palette_inferrer/app/main.py
+++ b/pipeline/inferrer/palette_inferrer/app/main.py
@@ -1,16 +1,21 @@
+import base64
+
 from fastapi import FastAPI, HTTPException
 
 import src.http as http
 from src.batching import BatchExecutionQueue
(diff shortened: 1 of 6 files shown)


jamieparkinson (approved):
I imagine the CI is failing because of linting errors - `make lint-python` should show you them but not being able to see the errors in buildkite (same for test failures) is a point that @kenoir might know about

kenoir:
Errors should be visible, from both linting and running regular tests. I don't know why they aren't here - what happens when you run the lint task locally?

jamieparkinson:
Interesting, here is an example of an error I couldn't see the other day - https://buildkite.com/wellcomecollection/catalogue/builds/219#dbd18798-95d8-402c-963f-25a29fb17c79

Running the lint task locally gives me a load of tedious whitespace errors

PR #811, opened 2020-08-21
Use weco-datascience package in inferrers

harrisonpim:


pipeline/inferrer/palette_inferrer/requirements.txt
```diff
@@ -5,11 +5,11 @@
 #    pip-compile
 #
 aiodns==2.0.0             # via aiohttp
-aiofile==3.0.0            # via -r requirements.in
-aiohttp[speedups]==3.6.2  # via -r requirements.in
-async-timeout==3.0.1      # via -r requirements.in, aiohttp
+aiofile==3.0.0            # via -r requirements.in, weco-datascience
```
jamieparkinson, line 8: We probably want to remove the dependencies that are already satisfied transitively via weco-datascience (these lovely comments from pip-compile will make that very easy).

jamieparkinson (approved):
Do you want to rebase to squash out all the commits? since the changes are actually pretty minimal

pipeline/inferrer/palette_inferrer/requirements.txt
jamieparkinson: Actually that's not quite right - we want to remove the dependencies which are _only_ used within weco-datascience. If it's explicitly used in both application and library then of course we should keep them and not rely on the transitive dep.

harrisonpim: bah, thought this had been done. must have got lost in the merges

harrisonpim:
> Do you want to rebase to squash out all the janky commits? since the changes are actually pretty minimal

don't care at all about the commit history within this PR - was planning to squash and merge, so no need to rebase afaik?

PR #818, opened 2020-08-25
Output and terminate properly from subprocess

kenoir:
Output the result of `make` so we can see what's going on!

### DIFF
--- a/.buildkite/scripts/commands.py
+++ b/.buildkite/scripts/commands.py
@@ -9,10 +9,7 @@ def _subprocess_run(cmd, exit_on_error=True):
 
     output = []
     pipe = subprocess.Popen(
-        cmd,
-        encoding="utf8",
-        stdout=subprocess.PIPE,
-        stderr=subprocess.STDOUT,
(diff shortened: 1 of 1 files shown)


alexwlchan:
Is it just the output of `make` you want to see in logs? Was the previous behaviour of Git output being captured okay?

.buildkite/scripts/commands.py
```diff
@@ -4,27 +4,36 @@
...
+        output.append(stdout_line)
+
+    # Extract results
+    pipe.communicate()
+    return_code = pipe.returncode
+
+    if return_code != 0 and exit_on_error:
+        sys.exit(return_code)
+
+    return "\n".join(output).strip()
```
alexwlchan, line 27: If you do stick with this, you want `"".join(output)` instead of `"\n".join(output)`; the latter is inserting an extra newline between consecutive lines.

kenoir:
> Is it just the output of `make` you want to see in logs? Was the previous behaviour of Git output being captured okay?

I think it's preferable to output for both - Buildkite builds are private and I don't think we expose any secrets now with our commands 🤞 so it's less pressing to suppress that output.

jamieparkinson (approved):

PR #824, opened 2020-08-27
Search image identifiers in works search

jamieparkinson:


api/api/src/test/scala/uk/ac/wellcome/platform/api/elasticsearch/WorksQueryTest.scala
```diff
@@ -153,6 +155,50 @@ class WorksQueryTest
...
+        )
+
+        val query = "sourceIdentifier123"
+
+        insertIntoElasticsearch(index, work, workNotMatching)
+
+        assertResultsMatchForAllowedQueryTypes(index, query, List(work))
+      }
+    }
+
```
jamieparkinson, line 201: No test for `images.otherIdentifiers` because there is no way in which it can become populated at the moment - I left it in the query for consistency with other fields / in case it does become populated

jamesgorrie (approved):

PR #827, opened 2020-08-27
partOf should contain nested ancestors

warrd:
## Issue

## Description

The `partOf` field currently incorrectly contains an array of all ancestors, but should contain only direct parents, with early ancestors nested within.

Here we allow relation fields to be nested, currently with only `partOf` containing nesting.

NOTE:  For backwards compatibility we include the unnested ancestors ordered from root to parent, with the parent additionally containing the nested data. This will be removed once the front end is switched over.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/RelatedWorkService.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/RelatedWorkService.scala
@@ -13,13 +13,6 @@ import uk.ac.wellcome.models.Implicits._
 import uk.ac.wellcome.json.JsonUtil.fromJson
 import uk.ac.wellcome.platform.api.Tracing
 
-case class RelatedWorks(
-  parts: List[IdentifiedWork],
-  partOf: List[IdentifiedWork],
-  precededBy: List[IdentifiedWork],
-  succeededBy: List[IdentifiedWork],
-)
-
 class RelatedWorkService(elasticsearchService: ElasticsearchService)(
   implicit ec: ExecutionContext)
     extends Tracing {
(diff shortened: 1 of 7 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/RelatedWorkService.scala
```diff
@@ -31,7 +24,7 @@ class RelatedWorkService(elasticsearchService: ElasticsearchService)(
   def retrieveRelatedWorks(index: Index,
                            work: IdentifiedWork): Future[Result[RelatedWorks]] =
     work.data.collectionPath match {
-      case None => Future.successful(Right(RelatedWorks(Nil, Nil, Nil, Nil)))
+      case None => Future.successful(Right(RelatedWorks(Some(Nil), Some(Nil), Some(Nil), Some(Nil))))
```
jamieparkinson: Could be worth adding a `def empty` for this case to the companion object as it's used in a couple of places / looks nice

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
```diff
@@ -187,4 +187,40 @@ case object DisplayWork {
 
   def apply(work: IdentifiedWork): DisplayWork =
     DisplayWork(work = work, includes = WorksIncludes())
+
+  def apply(work: IdentifiedWork, includes: WorksIncludes, relatedWorks: RelatedWorks): DisplayWork =
```
jamieparkinson: :+1: a better place for this to live imo

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/RelatedWorkService.scala
warrd: There is already a `def none` for when its `None`, this is `Some(Nil)` :see_no_evil: .. This means there actually no related works, the other means we don't actually know at that point, maybe could be `unknown` / `none` or something?

jamieparkinson: Yeah the naming is tricky... how about `RelatedWorks.nil`?

warrd: :+1: I like `RelatedWorks.nil` and `RelatedWorks.unknown`

kenoir (approved):

PR #828, opened 2020-08-27
Reduce staging desired task count to 1

jamieparkinson:
afaict we don't want/need 3 instances of the staging API (?)


jamesgorrie (approved):

alexwlchan:
Is the catalogue API backed by API Gateway?

I don't have a reference to hand (this is documented, but finding it is beyond me at this instant) – Amazon recommends you run at least as many tasks as there are available zones within your region.

Sending a request to an API GW that has to cross the AZ boundary suffers from horrific latency. The eu-west-1 region has three AZs, so we run three instances.

jamieparkinson:
It is backed by API gateway but as this isn't for production traffic I am not so bothered about the AZ latency / replication - there are still 3 for the production API

PR #833, opened 2020-08-28
Adding analysis sub pages

taceybadgerbrook:
less than 3 words; 3+ words; subsequent analysis


docs/search/query-design/README.md
```diff
@@ -60,16 +60,16 @@ Searches containing more than one token are matched to fields as phrases represe
...
 
 
-### Stripping
+# Stripping
 
 To control for variation in punctuation in queries (eg “gray's anatomy,” vs “grays anatomy”), all punctuation is removed before matching.
 
 
-### Shingling
+# Shingling
```
jamesgorrie, line 73: We've left these as `h3`s as that's semantically correct. There is the document heading (`h1`), then the section heading (`h2`) and then these as subsections (`h3`) - could we leave them as such?

docs/search/query-development-guide.md
```diff
@@ -0,0 +1,186 @@
+Hypotheses of user expectations and search behaviour are validated by analysis.  Once validated, different query tools are deployed and tested in order to meet these expectations.
```
jamesgorrie, line 1:
```suggestion
# Query development guide

Hypotheses of user expectations and search behaviour are validated by analysis.  Once validated, different query tools are deployed and tested in order to meet these expectations.  
```

docs/search/query-development-guide.md
```diff
@@ -0,0 +1,186 @@
+Hypotheses of user expectations and search behaviour are validated by analysis.  Once validated, different query tools are deployed and tested in order to meet these expectations.  
+
+In cases where links to analysis or relevance testing are missing, this work is either pending or redundant. In cases where links to query tools are missing, this query type is served by the tools used for Concepts, subjects.
```
jamesgorrie, line 3: > In cases where links to query tools are missing, this query type is served by the tools used for Concepts, subjects.

Is this suggesting that we aren't using query tools for concepts?

docs/search/query-development-guide.md
```diff
@@ -0,0 +1,186 @@
...
+
+
+# [Concepts, subjects](https://docs.wellcomecollection.org/catalogue/search/hypotheses/concepts-subjects.md)
+
+### Analysis conducted
+[Less than 3-word searches](https://docs.wellcomecollection.org/catalogue/search/analysis/less-than-3-word-searches.md)
+
+[Searches with 3 words or more](https://docs.wellcomecollection.org/catalogue/search/analysis/summary-3-plus-word-searches.md)
+
+### Query tools used
```
jamesgorrie, line 13: I'm not sure query tools is the write phraseology here. I would consider Elasticsearch to be a tool, or Kibana. These are parts of [Elasticsearche's DSL](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html).

jamesgorrie: I'm also a little confused as to what we mean that we used these tools for the analysis?

jamesgorrie (commented):
I am unclear on what the query development guide document is outlining or informing a reader about?

It feels like we might have some crossed wires around the query design and how we have and plan on developing it.

docs/search/query-design/README.md
taceybadgerbrook: The headings are all supposed to be the same level. Making the bottom ones into h3 makes it look like they are variants of the preceding h1.

docs/search/query-development-guide.md
taceybadgerbrook: The tools used for Concepts, subjects are the default generic tool set. Any hypotheses without query tools listed use the default, ie Concepts, subjects. Maybe it's just easier if I copy and paste the tools for each section.

docs/search/query-design/README.md
jamesgorrie: # Query design (h1) 

## Term (h2)

My bad, shouldn't be `h3`, but definitely `h2`, as should the other terms as `h1` is the title of the document.

harrisonpim: Keeping them all at the same level makes sense to me, but james definitely has a point about them being `h1`s. Could you add an `h1` title for the whole page and make each section into an `h2` or `h3`?

docs/search/query-development-guide.md
jamesgorrie: > default generic tool set

I'm not sure I know what that is?

docs/search/query-development-guide.md
taceybadgerbrook: I agree that between "search," "query" and ways queries are built, the vocabulary space is getting a little congested.  I think referring to Bool boosting as a DSL or AST query is going to make this even more confusing.

The idea was to present what goes on under the hood as a combination of different search strategies depending on user search input. What do you think of changing this from "query tool," to "search tool," or "search strategy?" 

There are 2 different sections: Analysis conducted and Query tools used. Analysis conducted is about how we used data analysis to validate the hypotheses (eg do users search for contributors?)  Query tools used is about the combinations of tools used to serve up the best results.

taceybadgerbrook:
The guide is for non-developers and external users to see how we've created search without going into too much implementation detail. It outlines how we start with hypotheses, validate them with analysis, build queries and then test them.

The guide is intended to be flexible. If we decide to use a new DSL or AST expression, this can be added to the Query design glossary and a link added to the guide.

docs/search/query-design/README.md
taceybadgerbrook: h2 works for me. As long as they are headings of some description and the links in the guide will need to match.

docs/search/query-development-guide.md
taceybadgerbrook: Right now, Contributors and Titles have special query configurations which are more involved than queries that are Concept, subjects.  For this reason, I'm saying the query for Concepts, subjects is the generic/default tool set because it's the one that returns the results if there are no matches for Contributor or Title fields.

Leave it with me. I'll drop the bit about cases where query tools are missing and instead paste in the same combo as for Concepts, subjects.

docs/search/query-development-guide.md
harrisonpim: of those options, "search strategy" is the least confusing to me

harrisonpim: Personally, I don't think i'd find this kind of 'field-first' guide very helpful if I were trying to understand how our queries have developed. 

We're trying to develop new queries by focusing on user intentions first, not fields. Because so many of the individual analyses/tests we've done touch multiple fields, the shift in perspective for this document results in loads of repeated links.

PR #837, opened 2020-09-03
Relation embedder core logic

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4672

## Description

Includes the main logic for the relation embedder:

* Given some work at the input, query for all works in the tree that will require denormalisation
* In turn fetch each work from the store, and query for their relations
* Add the relation data to the works and store them in the output index

NOTE: model changes to `Work` are required for this with have not yet been made, so there is no tests yet for the `RelationEmbedderWorkerService`

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/RelatedWorks.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/RelatedWorks.scala
@@ -1,5 +1,8 @@
 package uk.ac.wellcome.models.work.internal
 
+import scala.util.Try
+import scala.annotation.tailrec
+
 /** Holds relations for a particular work. This is a recursive data structure,
   * so related works can in turn hold their relations. An Option[List[_]] is
   * used to represent cases both where the work has no relations and where the
@@ -15,6 +18,11 @@ case class RelatedWorks(
   partOf: Option[List[RelatedWork]] = None,
   precededBy: Option[List[RelatedWork]] = None,
   succeededBy: Option[List[RelatedWork]] = None,
+)
+
+case class RelatedWork(
+  work: IdentifiedWork,
+  relatedWorks: RelatedWorks
 )
 
 object RelatedWorks {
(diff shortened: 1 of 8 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/RelatedWorks.scala
```diff
@@ -42,12 +50,110 @@ object RelatedWorks {
...
-  work: IdentifiedWork,
-  relatedWorks: RelatedWorks
-)
+  type TokenizedPath = List[PathToken]
+  type PathToken = List[PathTokenPart]
+  type PathTokenPart = Either[Int, String]
+
+  /** Creates RelatedWorks from unnested and unsorted data.
+    *
+    * @param path Children of the work
```
jamesgorrie, line 60: Is `path` `Children of the work`?

warrd: Ooops yep

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/RelatedWorks.scala
```diff
@@ -42,12 +50,110 @@ object RelatedWorks {
...
+          case Some(workPath) =>
+            tokenizedPathOrdering.compare(mainPath, workPath) >= 0
+        }
+    }
+    val parent = ancestors.sortBy(tokenizePath) match {
+      case head :: tail =>
+        Some(
+          tail.foldLeft(RelatedWork(head, RelatedWorks(partOf = Some(Nil)))) {
+            case (relatedWork, work) =>
+              RelatedWork(work, RelatedWorks.partOf(relatedWork))
```
jamesgorrie, line 83: Would this create a `parent` that doesn't have it's `siblings` nor `children`?

warrd: This is what has been requested from the API: `partOf` is nested but nothing else.

warrd: Yep that's right.  This is what has been requested from the API: `partOf` is nested but nothing else.

jamesgorrie: Ah, perfect, so you can traverse up, but then not down another path etc... thanks for clarifying.

pipeline/relation_embedder/src/test/scala/uk/ac/wellcome/relation_embedder/RelatedWorksServiceTest.scala
```diff
@@ -51,110 +51,182 @@ class RelatedWorksServiceTest
   val works =
     List(workA, workB, workC, workD, workE, workF, work4, work3, work2, work1)
```
jamesgorrie, line 52: Maybe it's annoying, but i wonder if it might make the tests easier to read if the works were named
```scala
  val workA = work("a")
  val workA1 = work("a/1")
  val workA1B = work("a/1/b")
  val workA1C = work("a/1/c")
  val workA2 = work("a/2")
  val workA2D = work("a/2/d")
  val workA2E = work("a/2/e")
  val workA2EF = work("a/2/e/f")
  val workA3 = work("a/3")
  val workA4 = work("a/4")
```

Could just be because I'm seeing the changes in the PR out of context of where we set the vals

pipeline/relation_embedder/src/test/scala/uk/ac/wellcome/relation_embedder/RelatedWorksServiceTest.scala
```diff
@@ -51,110 +51,182 @@ class RelatedWorksServiceTest
...
-            parts = List(
-              relatedWork(work1),
-              relatedWork(work2),
-              relatedWork(work3),
-              relatedWork(work4)),
-            partOf = Nil,
-            precededBy = Nil,
-            succeededBy = Nil
-          )
+    it("Returns no affectedd works when work is not part of a collection") {
```
jamesgorrie, line 81:
```suggestion
    it("Returns no affected works when work is not part of a collection") {
```

jamesgorrie (approved):
Lots of list and ordering stuff, great!

PR #838, opened 2020-09-03
make elastic index creator more sensible

jamesgorrie:
Could probably have done some fancy bits and bobs with a `EitherT[Future, Error, Response[T]]` but started to feel overkill where a for comprehension might achieve the same thing.

The logic is also slightly different
* Check if the index exists
* If not => try create it
* If so => try update it
* Then add the search template

This seemed more logical as almost always the index will exist, allowed us to remove the custom error checking, and, I suspect, we might gain some tiny tiny improvements as a `PUT` that returns a fail is probably slower than a `HEAD` request.

This fixes the fact that if the search template changes and the index already exists, it doesn't update.

### DIFF
--- a/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
+++ b/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
@@ -22,18 +22,15 @@ class ElasticsearchIndexCreator(
   val analysis = config.analysis
   val searchTemplate = config.searchTemplate
 
-  private def exists = elasticClient.execute(indexExists(index.name))
+  private def exists =
+    elasticClient.execute(indexExists(index.name)).map(_.result.isExists)
 
   private def createOrUpdate: Future[Unit] = {
     for {
-      existsResp <- exists
-      createResp <- if (existsResp.result.isExists) {
-        update
-      } else {
-        put
-      }
+      doesExist <- exists
+      createResp <- if (doesExist) update else put
       resp <- searchTemplate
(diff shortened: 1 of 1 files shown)


common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
```diff
@@ -23,18 +16,29 @@ class ElasticsearchIndexCreator(
...
+      createResp <- if (existsResp.result.isExists) {
+        update
+      } else {
+        put
+      }
+      resp <- searchTemplate
+        .map(template => putSearchTemplate(template))
+        .getOrElse(Future.successful(createResp))
+    } yield { handleEsError(resp) }
   }
```
jamieparkinson, line 39: Might suggest this as a slightly clearer version

```suggestion
  private def exists =
    elasticClient.execute(indexExists(index.name)).map(_.result.isExists)

  private def createOrUpdate: Future[Unit] = {
    for {
      doesExist <- exists
      createResp <- if (doesExist) update else put
      resp <- searchTemplate
        .map(putSearchTemplate)
        .getOrElse(Future.successful(createResp))
    } yield { handleEsError(resp) }
  }
```

jamesgorrie: I suppose I find it just a little weird to return a `Future[Boolean]` as all the others futures are of `Future[Response[T]]` but that doesn't matter as we're not using `EitherT`.

Thanks!

jamieparkinson (approved):

PR #839, opened 2020-09-03
Use elastic4s built in circe deserialisation

warrd:
There is functionality for this rather than having to pass in `sourceAsString` everywhere. Maybe a slight improvement I dunno :man_shrugging:


warrd:
It still uses any decoders defined

jamesgorrie (approved):
Nice!

PR #842, opened 2020-09-07
Start updating location type CSV to contain LocationType.id

jamesgorrie:
A start at remodelling some of the location stuff to add filters.


jamesgorrie:
Closing as these won't actually be used.

See #857

PR #843, opened 2020-09-07
Attach licenses & access conditions to image locations

jamieparkinson:
Really this should have been there all along

### DIFF
--- a/pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroImage.scala
+++ b/pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroImage.scala
@@ -1,31 +1,15 @@
 package uk.ac.wellcome.platform.transformer.miro.transformers
 
 import uk.ac.wellcome.models.work.internal.{
-  DigitalLocation,
   Identifiable,
   IdentifierType,
-  LocationType,
   SourceIdentifier,
   UnmergedImage,
   Unminted
 }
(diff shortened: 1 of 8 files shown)


pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroImage.scala
```diff
@@ -36,9 +20,6 @@ trait MiroImage {
         value = miroRecord.imageNumber
       ),
       version = version,
-      location = DigitalLocation(
-        url = buildImageApiURL(miroRecord.imageNumber, "info"),
-        locationType = LocationType("iiif-image")
-      )
+      location = getLocations(miroRecord).head
```
alexwlchan, line 23: Is there any case in which `getLocations()` could return multiple locations? And if not, why not have a `getLocation()` method instead?

alexwlchan (approved):
Looks good, modulo `getLocation[s]` comment.

pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroImage.scala
jamieparkinson: Yeah, I didn't want to mess around with this as at the moment it can only ever return 1 location, but it's probably clearer for all involved if I change it to `getLocation` in existing code too

alexwlchan (approved):

PR #845, opened 2020-09-07
Don't supress Calm works with status "Third-party metadata"

warrd:


pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -33,6 +33,15 @@ object CalmTransformer
     ("Arrangement", ArrangementNote(_))
   )
 
+  // As much as it might not look like it, these values mean it should
+  // not be suppressed.
+  val nonSupressedStatus = List(
```
jamieparkinson, line 38: I might prefer to do this as a pattern match with a separate branch for each string (rather than `.contains` or `case "a" | "b" | "c" =>`, but if it's to stay like this
```suggestion
  val nonSuppressedStatuses = List(
```

jamieparkinson (approved):

jamesgorrie (approved):

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
warrd: Ha deja vu 
https://github.com/wellcomecollection/catalogue/pull/619#discussion_r435095448

jamieparkinson: Ha, at least I'm picking the same nits

PR #846, opened 2020-09-07
add search templates route to API

jamesgorrie:
Mucking around trying to get the search templates from the index (which are stored in the index state, which was the first of many fiddly things) I had a spark of why not let the API expose it's search templates.

There's nothing sensitive in it, and then we can just consume it from where ever we want to, in this case a rank_eval lambda.

It will also help with testing against staging


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchTemplates.scala
```diff
@@ -0,0 +1,22 @@
...
+import com.sksamuel.elastic4s.requests.searches.queries.{Query, QueryBuilderFn}
+import io.circe.generic.extras.semiauto.deriveConfiguredEncoder
+import io.circe.Encoder
+import uk.ac.wellcome.display.json.DisplayJsonUtil._
+
+case class SearchTemplates(templates: List[SearchTemplate])
+object SearchTemplates {
+  implicit def encoder: Encoder[SearchTemplates] =
+    deriveConfiguredEncoder
+}
```
alexwlchan, line 12: I’m surprised you need to declare encoders explicitly; we usually handle this by importing `JsonUtil._`.

api/api/src/test/scala/uk/ac/wellcome/platform/api/ApiSearchTemplatesTest.scala
```diff
@@ -0,0 +1,34 @@
...
+class ApiSearchTemplatesTest
+    extends ApiWorksTestBase
+    with Matchers
+    with JsonHelpers {
+
+  it("renders a list of available search templates") {
+    checkJson { json =>
+      json.isObject shouldBe true
+      val templates = getKey(json, "templates")
+      templates.isEmpty shouldBe false
```
alexwlchan, line 17: Slightly more explicit:

```suggestion
      templates should not be empty
```

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchTemplates.scala
```diff
@@ -0,0 +1,22 @@
...
+
+case class SearchTemplate(id: String, index: String, query: String)
+
+object SearchTemplate {
```
alexwlchan, line 16: minor: I’m not a huge fan of variables/classes whose names differ by a single character.

warrd: I agree with this. Also is `SearchTemplates` necessary, why not just use `List[SearchTemplate]` directly?

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchTemplates.scala
warrd: Personally I think it's better to derive explicitly, as otherwise we end up in a situation where the same encoder is automatically derived multiple times and really slow compilation times.

This should be an `implicit val` rather than `implicit def` by the looks of thing though.

jamesgorrie: This also slows down compilation significantly.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchTemplates.scala
jamesgorrie: As I would then need to add it to the top level JSON object:
`{ "templates": [] }`

Most of this is for JSON rendering, so I don't see a massive problem with it.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchTemplates.scala
jamesgorrie: `valèd it up.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchTemplates.scala
warrd: Ah I get it. How about calling it `SearchTemplatesResponse` to indicate it's the top level response that is returned at that endpoint?

jamesgorrie: Yip, that makes sense.

warrd (approved):

PR #848, opened 2020-09-08
Create 20200908 pipeline

jamieparkinson:


pipeline/terraform/main.tf
```diff
@@ -61,3 +61,67 @@ module "catalogue_pipeline_20200827" {
   # Inferrer data
   inferrer_model_data_bucket_name = aws_s3_bucket.inferrer_model_core_data.id
 }
+
+module "catalogue_pipeline_20200908" {
```
jamesgorrie, line 65: Did we say we wanted more meaningful names at some stage?

jamesgorrie (approved):

pipeline/terraform/main.tf
jamieparkinson: It was touched upon but no decision made... I quite like not having to think about naming for smaller, just-rolling-forwards deployments, but for bigger departures (especially long-lived stuff) it seems wise

jamesgorrie: Less thinking, more fun, agreed.

PR #850, opened 2020-09-08
Don't include unnested ancestors in partOf

warrd:
This was previously included for backwards compatibility, but not required since https://github.com/wellcomecollection/wellcomecollection.org/pull/5504


jamieparkinson (approved):

jamieparkinson:
Should digirati know about this too?

warrd:
We've told @tomcrane that this is coming up in the digirati channel

tomcrane:
I see you...

Issue #817, opened 2020-08-25
Location filters


jamesgorrie:
Add location filters to the API.

## Locations
| Type | Label |
|------|-------|
| OpenShelves | Open shelves | 
| ClosedStores | Closed stores | 
| DigitalResource | Online | 

We're removing the current `locationType`(`iiif-presentation`, `iiif-image`) and will add filtering on the above locations via the `items.locations.type` query parameter. e.g. `?items.locations.type=OpenShelves,ClosedStores`.

These will be aggregated so available to the frontend with the above labels and counts of how many are in each aggregation the same way we do with `workType`. e.g.
![Screenshot 2020-09-10 at 10 06 30](https://user-images.githubusercontent.com/31692/92707601-54dca880-f34d-11ea-9990-4744f8e3544b.png)

Note: I've just nabbed the interface with what is live, it will follow whatever filter pattern we have at the time of implementation.

PR #852, opened 2020-09-08
add new locations model

jamesgorrie:
ref #817 

We're moving towards a location model that is focussed on the question "Where/how can I access this?".

Having three different `Location`s seems to cover the bases, and removing the `DigitalLocation` and `PhysicalLocation` as they aren't useful. If we found a use for them they would be a property of the `Location` e.g.
```scala
trait LocationType
object LocationType {
  case object Physical
  case object Digital
}
trait Location
case class OpenShelves() extends Location { type Format: LocationType.Physical }
case class DigitalResource() extends Location { type Format: LocationType.Digital }
```

This will then allow for `?items.locations.type=OpenShelves,DigitalResource` and to be able to facet on that. As we're not storing the label of an item in ES, I might have to have a think about how those are constructed.

The spacing is all a bit strange, thanks `scalafmt`.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location.scala
@@ -1,0 +1,37 @@
+package uk.ac.wellcome.models.work.internal
+
+sealed trait Location {
+  val accessConditions: List[AccessCondition]
+}
+object Location {
+  case class OpenShelves(accessConditions: List[AccessCondition],
+                         shelfmark: String,
+                         sectionName: String)
+      extends Location
+  case class ClosedStores(accessConditions: List[AccessCondition])
+      extends Location
+  case class DigitalResource(accessConditions: List[AccessCondition],
+                             source: DigitalSource)
+      extends Location
+}
+
+sealed trait DigitalSource {
+  val url: String
+  val license: Option[License]
+  val credit: Option[String]
+}
+
+object DigitalSource {
+  case class IIIFPresentation(url: String,
+                              license: Option[License],
+                              credit: Option[String])
+      extends DigitalSource
+  case class IIIFImage(url: String,
+                       license: Option[License],
+                       credit: Option[String])
+      extends DigitalSource
+  case class ExternalSource(url: String,
+                            license: Option[License],
+                            credit: Option[String])
+      extends DigitalSource
+}


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/LocationV2.scala
```diff
@@ -0,0 +1,37 @@
+package uk.ac.wellcome.models.work.internal
+
+trait LocationV2 {
```
warrd, line 3: Should be `sealed trait` here and for `DigitalSource`

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location2.scala
```diff
@@ -0,0 +1,37 @@
...
+sealed trait Location2 {
+  val accessConditions: List[AccessCondition]
+}
+object Location2 {
+  case class OpenShelves(accessConditions: List[AccessCondition],
+                         shelfmark: String,
+                         sectionName: String)
+      extends Location2
+  case class ClosedStores(accessConditions: List[AccessCondition])
+      extends Location2
```
warrd, line 12: Regarding naming, in my opinion the singular makes sense for these: (i.e.`OpenShelf` and `ClosedStore`): the location is an open shelf location, a closed store location or a digital resource location.

jamieparkinson:
Did this refer to https://github.com/wellcomecollection/docs/tree/master/rfcs/020-locations_requesting? From what I can see it's similar but not necessarily identical

jamesgorrie:
@jamieparkinson as the thinking has progressed, this supersedes that.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location2.scala
jamesgorrie: The way we speak about it is definitely in the plural as that is the name of the location.
we would not say give us books that are on an open shelf, but rather give us books that are on the open shelves.

I think if singularise the word we make the location too specific.

jamieparkinson:
Is it worth RFCing this? It's quite a far-reaching change and it would be good to have more context

jamesgorrie:
It felt like a small enough change that the discussions could happen over PRs. Happy to, but personally feel it's overkill.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location2.scala
warrd: Yeh I guess I am thinking it more from a semantic angle than strictly to do with what language is used. In general we don't really use pluralised class names, and find it strange having an ADT defined like that. Regarding the language, these two statements don't really  conflict with each other in my opinion:

* Give us the books that are on the open shelves
* Give us the books that have the open shelf location type

jamieparkinson:
I definitely agree that the current model is lacking. Feel like it's a significant-ish change just because it's to our "core" model, and even though it's small it's a breaking change for the API, the API consumers (including for images), the transformers, the merger, the inference manager and the ingestors.

jamieparkinson:
Possibly stupid question: where do these now go? https://github.com/wellcomecollection/catalogue/blob/master/common/internal_model/src/main/resources/location-types.csv

jamesgorrie:
I don't think setting the standard for every core model change we need an RFC.

This change is not really changing anything that is used (we use the `iiif-image,iiif-presentation` filter, but we know that's wrong) - but rather it's implementing the first phase of something we can use. 

This model was never actually put into practise and now that is has been, we have a much better understanding of where to start.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location2.scala
jamesgorrie: Semantically they are the same, but we never speak about an Open shelf.

I would rather keep language the same - these locations are known and have a way for them to be described and I would prefer to keep that. The name of the location is the Open shelves. An open shelf isn't used in the current vernacular.

I can't think that having a rule to never pluralise is useful or makes sense in helping with understanding or reading code.

As another example if we were talking about the location "The marshes", we wouldn't singularise that to "The Marsh".

jamesgorrie:
> Possibly stupid question: where do these now go? https://github.com/wellcomecollection/catalogue/blob/master/common/internal_model/src/main/resources/location-types.csv

These will be removed as we will read the data from other fields and map that to the new types.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Location2.scala
warrd: OK I get your marshes example, it still seems a bit weird to me but acknowledge that is probably to do with me not really being familiar with the terminology :+1:

jamesgorrie: Sorry to be contrived with the example - but the point I was trying to get across is the location is a know entity and is called the `Open shelves` which is what I am trying to encapsulate in this type class. Maybe because it's not a abstract concept but rather a representation of a known entity.

warrd (commented):
I don't quite understand how this new model is going to be used in practice. Are we going add both locations to the internal model and index both? It would maybe be better just moving to the new one (and thus calling it `Location` rather than `Location2`), and serialising it differently in the API depending on query string or whatever. Is that something which is possible?

jamesgorrie:
@warrd We're deprecating the old `Location` model.

We need to support the frontend from any downtime with this breaking change, to do this we will:
* add a new property to the model called `work.items.locations2`. This will only be returned if we have an `include=location2` 
* we can then build the the frontend for this behind a feature toggle
* once we're happy with the model and the frontend, we will make the frontend read from `locations2` by default
* we will then move this model over to the `work.items.locations` property
* then we update the frontend to read from ☝️ 
* remove the `locations2` from the API  

We will be indexing both for a period of time.

It's slightly long winded process, but it's simple and without down time.

jamesgorrie:
We have to store / index both so that the frontend keeps working.

Yea, I could have a go at renaming the old one to `LegacyLocation` but can't see how there would be less swapping?

jamesgorrie:
Were you thinking of having a single model that could serialise to both display models?

warrd:
* Add new `Location` model and rename old one as `LegacyLocation`
* Rename the original internal model field to `legacyLocation` and add `location` field for the new one
* Add a qurey parameter `legacyLocation` which defaults to true
* Switch over front end to not use legacy location
* Remove legacy location model, field and API functionality

warrd:
> Were you thinking of having a single model that could serialise to both display models?

Yes that was my original idea, failing that my other suggestion is above :point_up:

jamesgorrie:
> > Were you thinking of having a single model that could serialise to both display models?
> 
> Yes that was my original idea, failing that my other suggestion is above ☝️

The transformation rules are all different and we want to ultimately not use the sierra fields that we're using to create these locations.

Which has made me think I need to have a thinking about the `items.locations.locationType` filter too.

warrd:
> but can't see how there would be less swapping?

Your idea involves running a reindex for `location2` and an additional reindex once its renamed. My one you just have a single reindex with `location` as the new model and `legacyLocation` as the old model

jamesgorrie:
I would still run a new reindex to remove the `legacyLocation` property as a tidy up.

jamesgorrie:
I want to avoid the `V2` tidy up extravaganza we had before (and still have bits laying about in the code for).

Saying that is has made me think being able to run a new version of the API here would be handy...

warrd:
Not strictly necessary, you can just wait until someone else needs to reindex for another reason. I guess also don't like the naming `Location2` very much

jamesgorrie:
> you can just wait until someone else needs to reindex 

I'd like to clean it up as a matter of finishing the work.

warrd:
Well it's not really necessary once you have cleaned up the code, there is no harm the data sitting there for a few weeks as will clean up itself

warrd:
But the naming thing is the main issue for me. If you are keen on doing things your way round then go for it, but maybe a better name for `Location2`? `UpcomingLocation`? It seems to me renaming the other one as legacy most matches the state of the models imo though.

warrd:
(I accept we are probably in the bikeshed right now btw, especially as these names are only temporary)

jamesgorrie:
Again, the naming is temporary so no-one is going to have to look at it beside me for a bit.

Saying that I do like the idea of deprecating the old model first, albeit we haven't given anyone anything to use as an alternative, but.. 🤷 🦐

jamesgorrie:
@warrd I've pulled in the deprecated fix. How's this?

warrd (approved):
:ok_hand:

PR #859, opened 2020-09-09
Scoped Id model

warrd:
This is to

* Prevent future clashes with `WorkState` types containing the same names (i.e. `Identified`)
* Help readability of the code (e.g. seeing `Id.WithSourceIdentifier` / `Id.Minted` is more understandable)
* Be consistent with other ADT definitions, such as `Work`, `WorkState` etc


pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/transformers/MiroContributors.scala
```diff
@@ -1,6 +1,6 @@
 package uk.ac.wellcome.platform.transformer.miro.transformers
 
-import uk.ac.wellcome.models.work.internal.{Agent, Contributor, Unminted}
+import uk.ac.wellcome.models.work.internal.{Agent, Contributor, IsState}
```
jamesgorrie, line 3: Assuming `IdState`?

warrd: Yep

warrd: Working on another branch at the moment, expecting a fair few failures on this so probably should have marked as WIP

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -3,117 +3,117 @@ package uk.ac.wellcome.models.work.internal
 import uk.ac.wellcome.models.work.text.TextNormalisation._
 import uk.ac.wellcome.models.parse.parsers.DateParser
 
-sealed trait AbstractRootConcept[+Id] extends HasIdState[Id] {
+sealed trait AbstractRootConcept[+DataId] extends HasId[DataId] {
```
jamesgorrie, line 6: Could you explain why `DataId`?

warrd: Cos doesn't clash with the name `Id` and @jamieparkinson had already used that name elsewhere

warrd: Up for other names if people got different suggestions

warrd: (I think the reason for it being `DataId` is to signify it is the ID type used for `WorkData` rather than images / works)

jamesgorrie: Ah OK - so it's the IdState that the `WorkData` which then drills down to any Identifiable object within that data?

jamesgorrie: But it doesn't have to be inherited from a container type, so what about `IdState`?

warrd: :thinking: I get what you are saying, but also `IdState` isn't quite right in its own way. I decided to name the type simply `Id` so the introduction of word state could be a bit confusing  ("`IdState` which is a of type `Id`"). Maybe I am over thinking things. @jamieparkinson what's your thoughts?

warrd: For context the `DataId` originally comes from here:

https://github.com/wellcomecollection/catalogue/blob/e87516f4c8aad2ba5ab1c8bcdf423dfcda911db8/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala#L99-L104

But you are right as far as the individual models are concerned they shouldn't really care where it comes from.

jamesgorrie: Sorry if it wasn't clear, I was asking if `DataId` here is the state inherited from the containing `WorkData`? e.g. if the `WorkData` is `Minted` then these `Concept`s will be `Minted`. My concern there is that the naming then assumes that these types will always live within `WorkData`, which feels weird especially as Concepts will live separately later.

Na, I get what you're saying, especially as we now have state as an indicator of actual state in the `Work` model. Saying that, Identification is similar/the same to the Finite state machine we are using in the `Work` model.

warrd: > My concern there is that the naming then assumes that these types will always live within WorkData, which feels weird especially as Concepts will live separately later.

Yep that's what I understood what you meant, and completely agree with that.

jamieparkinson: I think I have a marginal preference for `IdState` even though I do understand / agree with the drawbacks you give, because I think `Id` is a bit more confusing - given `Id` is an abbreviation of `Identifier`, what is an `Identified Identifier` / `Identifiable Identifier` / `Unidentifiable Identifier`?

warrd: Ah man, I just spent the whole morning going from `IdState` to `Id` :see_no_evil:.. But yeh, we want to get this right so let's do whatever makes the most sense (hopefully can find the old stuff in my reflog or commit history if decide to go back to that)

Hmmm I have an idea. Why don't we always name the type parameters `State` like we do are planning to do with `Work`:

```scala
sealed traiit Work[State <: WorkState]

...

// Note no constraints on what sort of state here
case class Concept[State](
  state: State,
  label: String
)

...
```

That way it doesn't really specify a certain sort of state / ID and we can happily use different types depending on whether we are in concepts pipeline, work pipeline, images pipeline.

For example we can currently use the `IdState` as we have now, but in future we might want to use `ConceptState` in the concept pipeline etc.

warrd: Yep exactly, this has that use case in mind

jamieparkinson (approved):

PR #860, opened 2020-09-09
Locations display model

### LINKED ISSUE #856: Create display model for new location model

jamesgorrie:
Ref #817 
Fixes #856 

Notes & questions:
* We've ditched `locationType`
* `label` field on each location which will be used for the frontend
* Properties specific to the location such as `shelfmark`, `shelfLocation`, `url` etc are stored on the location directly
* ~I didn't know whether to go with `shelfMark` or `shelfMark`, but if we are to use `shelfLocation` it felt more right to keep it camel cased~ `shelfmark` it is 
* Introduction of the `DigitalResource.format`. I've used format as it's a hint to the frontend on how it might render or handle the resource. `iiif-image` and `iiif-presentation` are examples. I've made it optional for when we're not sure. We could introduce the type `Link` as a catch all, which would just render a link
* ~`items.locations.format` could be an object e.g. `{ "type": "IIIFPresentation", "label": "IIIF presentation" }`. It might useful  for the frontend e.g. "View this resource (PDF)" <- PDF being the label?~ This will be an object
* The serialisation doesn't work, waiting on model changes before continuing

The query will be `?items.location.type=OpenShelves,DigitalResource`. We will use `?items.location.type=DigitalResource` at first to replicate only having digital resources available.

With the intention of producing this JSON:

```JS
{
  "items": [
    {
      "locations": [
        {
          "type": "OpenShelves",
          "label": "Open shelves",
          "accessConditions": [
            {
              "status": {
                "id": "open",
                "label": "Open",
                "type": "AccessStatus"
              },
              "terms": "The papers are available subject to the usual conditions of access to Archives and Manuscripts material.",
              "type": "AccessCondition"
            }
          ],
          "shelfmark": "TP.35.AA8-9",
          "shelfLocation": "History of Medicine"
        },
        {
          "type": "ClosedStores",
          "label": "Closed stores",
          "accessConditions": [
            {
              "status": {
                "id": "closed",
                "label": "Closed",
                "type": "AccessStatus"
              },
              "terms": "Bloop",
              "type": "AccessCondition"
            }
          ]
        },
        {
          "type": "DigitalResource",
          "label": "Online",
          "accessConditions": [
            {
              "status": {
                "id": "open",
                "label": "Open",
                "type": "AccessStatus"
              },
              "type": "AccessCondition"
            }
          ],
          "url": "https://wellcomelibrary.org/iiif/b30598977/manifest",
          "license": {
            "id": "inc",
            "label": "In copyright",
            "url": "http://rightsstatements.org/vocab/InC/1.0/",
            "type": "License"
          },
          "credit": "Credit stuff here",
          "format": { "type": "IIIFPresentation", "label": "IIIFPresentation" }
        },
        {
          "type": "DigitalResource",
          "label": "Online",
          "accessConditions": [
            {
              "status": {
                "id": "open",
                "label": "Open",
                "type": "AccessStatus"
              },
              "type": "AccessCondition"
            }
          ],
          "url": "https://place.org/an/external/link",
          "license": {
            "id": "inc",
            "label": "In copyright",
            "url": "http://rightsstatements.org/vocab/InC/1.0/",
            "type": "License"
          },
          "credit": "Credit stuff here"
        }
      ]
    }
  ]
}
```

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/json/DisplayJsonUtil.scala
@@ -3,6 +3,11 @@ package uk.ac.wellcome.display.json
 import io.circe.generic.extras.{AutoDerivation, Configuration}
 import io.circe.{Encoder, Printer}
 import io.circe.syntax._
+import uk.ac.wellcome.display.models.DisplayLocation.{
+  DisplayClosedStores,
+  DisplayDigitalResource,
+  DisplayOpenShelves
+}
 import uk.ac.wellcome.display.models._
 
 /** Format JSON objects as suitable for display.
@@ -55,11 +60,17 @@ trait DisplayJsonUtil extends AutoDerivation {
     case concept: DisplayAbstractConcept => concept.asJson
   }
 
-  implicit val locationEncoder: Encoder[DisplayLocationDeprecated] = {
+  implicit val locationDeprecatedEncoder: Encoder[DisplayLocationDeprecated] = {
     case digitalLocation: DisplayDigitalLocationDeprecated =>
       digitalLocation.asJson
     case physicalLocation: DisplayPhysicalLocationDeprecated =>
       physicalLocation.asJson
+  }
+
+  implicit val locationEncoder: Encoder[DisplayLocation] = {
+    case closedStores: DisplayClosedStores       => closedStores.asJson
+    case openShelves: DisplayOpenShelves         => openShelves.asJson
+    case digitalResource: DisplayDigitalResource => digitalResource.asJson
   }
 }
 
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayLocation.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayLocation.scala
@@ -1,0 +1,93 @@
+package uk.ac.wellcome.display.models
+
+import io.circe.generic.extras.JsonKey
+import io.swagger.v3.oas.annotations.media.Schema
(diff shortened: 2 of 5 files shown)


jtweed:
I think this is very close and is a fantastic step forward. My only comments:

1. It should be `shelfmark` as it's a single word noun commonly used in the library world
2. Format should be an object and it should be reference data like `workType`, not classes

The rationale for the second point is that I think we should considering format could be any kind digital file, ie anything in PRONOM. I'm thinking forward to born digital support here.

jamesgorrie:
@jtweed great, those were my leanings as well, I'll carry no with that. Comment updated.

jamieparkinson (approved):
Implementation looks good, although the UpperCamelCase queries look a bit strange to me - sort of like leaking implementation details into the interface (in other words, what would these queries look like if we rewrote the API in python? `closed_stores`?)

jamesgorrie:
@jamieparkinson we had a conversation about this, and it's because the query params are meant to be JSON paths, which would point to the `UpperCamelCase` value.

The other type of query is when we point to the `id` prop of an object (e.g. `WorkType`), which it was decided that it is inferred that the JSON path has `.id` i.e. `workType` equates to `workType.id`. 

I personally think we should try choose 1 over the other, and if I had the choice, I would move over to using the `type` as IDs seem as arbitrary as type, but we use that for ontology stuff, so why not just use that.

That would mean that we express `workType` as
```JS
{
  "type": "Books",
  "id": "a",
  "label": "Books"
}
```

With the aim to deprecate `.id`.

Perhaps it might be worth opening the discussion of how to standardise displaying and filtering on enums?

PR #861, opened 2020-09-09
Work state modelling RFC

warrd:
https://github.com/wellcomecollection/docs/pull/39

This modifies the work model so we have a `Work` ADT (`Work.Standard`, `Work.Redirected`, `Work.Invisible`) parameterised by some `WorkState`.

To introduce the changes more gradually here there are only 2 work states, `Unidentified` and `Identified`, which corresponds to what we have with the current modelling. In a future PR these states will be updated.


jamieparkinson (approved):
Pretty impossible to review this meaningfully but looks like all the bases I can think of are covered - I guess your plan is to get it into a new pipeline and see what happens?

warrd:
Yep set up a new pipeline, do a reindex, then check the API is working on staging.

PR #867, opened 2020-09-16
Filter by colour

jamieparkinson:
This allows API users to pass a comma-separated list of colours (well, `color`s) to filter by. As such there is no scoring of colour similarity but it's likely we'll tackle that next. It also fixes a [very silly off-by-one error](https://github.com/wellcomecollection/catalogue/pull/867/files#diff-33b0eaf9b9d3a3a798032a703417da0aL56) that I introduced in https://github.com/wellcomecollection/catalogue/pull/864.

In order to construct the colour queries the API needs to know some of the parameters (bin sizes) that were used in the creation of the colour signature tokens (ie by the palette inferrer). The change in https://github.com/wellcomecollection/catalogue/pull/864 encoded the parameters in the signatures themselves, and this PR includes the functionality to extract these from the index at API startup.

I initially made this a parameter that could be passed as an environment variable, but actually finding a way of sharing that between the pipeline and api stacks _and_ respecting differing co-existing environments and indices proved quite tricky and no less ugly than this solution - but I'm open to discussion on this. I did add a few tests to try to ensure that a failure of this mechanism wouldn't bork the whole API.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
@@ -6,7 +6,7 @@ import akka.http.scaladsl.Http
 import com.typesafe.config.Config
 import uk.ac.wellcome.elasticsearch.ElasticConfig
 import uk.ac.wellcome.elasticsearch.typesafe.ElasticBuilder
-import uk.ac.wellcome.platform.api.models.ApiConfig
+import uk.ac.wellcome.platform.api.models.{ApiConfig, QueryConfig}
 import uk.ac.wellcome.typesafe.config.builders.AkkaBuilder
 import uk.ac.wellcome.typesafe.config.builders.EnrichConfig._
 import uk.ac.wellcome.typesafe.WellcomeTypesafeApp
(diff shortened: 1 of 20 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/Main.scala
```diff
@@ -43,7 +43,11 @@ object Main extends WellcomeTypesafeApp {
           .getOrElse("context.json"),
       )
 
-    val router = new Router(elasticClient, elasticConfig, apiConfig)
+    val queryConfig =
+      QueryConfig.fetchFromIndex(elasticClient, elasticConfig.imagesIndex)
```
warrd, line 47: Would this need to be refreshed from the index every so often? :thinking:

jamieparkinson: Good question - I think in this case no, because changing the parameters of the inferrer requires a complete reindex (as the vectors only make sense with respect to each other) and the hardcoded index means the API task is bound to the index it starts with

api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ColorQuery.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object ColorQuery {
+  type Rgb = (Int, Int, Int)
+
+  def hexToRgb(hex: String): Rgb = {
+    val n = Integer.parseInt(hex, 16)
+    (
+      (n >> 16) & 0xFF,
+      (n >> 8) & 0xFF,
+      n & 0xFF
```
warrd, line 49: l337

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/QueryConfig.scala
```diff
@@ -0,0 +1,65 @@
...
+                }
+            }
+            .left
+            .map(_.asException)
+            .toTry
+            .map {
+              case Some(bins) => bins
+              case None =>
+                throw new RuntimeException(
+                  "Could not extract palette parameters from data in index")
```
warrd, line 61: I think you want to `flatMap` that try and return either success / failure rather than throwing it here

warrd (approved):
Nice work, just comment about throwing the exception

harrisonpim (approved):

PR #878, opened 2020-09-22
API type filter

jamieparkinson:
This follows up from https://github.com/wellcomecollection/catalogue/pull/876, adding a `type` filter which enables us to distinguish between / filter for standard works, collections, series, and sections. The `type` name corresponds to the `type` aka `ontologyType` aka `data.workType` on works, and is not related to the `format` which was previously known as `workType`.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
@@ -332,7 +332,7 @@ trait MultipleWorksSwagger {
         description = "Filter by the type of the searched works",
         required = false,
         schema = new Schema(
-          allowableValues = Array("Work", "Collection", "Series", "Section")
+          allowableValues = Array("Standard", "Collection", "Series", "Section")
         )
       ),
       new Parameter(
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
@@ -332,7 +332,7 @@ trait MultipleWorksSwagger {
         description = "Filter by the type of the searched works",
         required = false,
         schema = new Schema(
-          allowableValues = Array("Standard", "Collection", "Series", "Section")
+          allowableValues = Array("Collection", "Series", "Section")
         )
       ),
       new Parameter(
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
@@ -327,6 +327,15 @@ trait MultipleWorksSwagger {
         required = false
       ),
       new Parameter(
+        name = "type",
+        in = ParameterIn.QUERY,
+        description = "Filter by the type of the searched works",
+        required = false,
+        schema = new Schema(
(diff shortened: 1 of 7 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/swagger/SwaggerDocs.scala
```diff
@@ -326,6 +326,15 @@ trait MultipleWorksSwagger {
         description = "Filter by the format of the searched works",
         required = false
       ),
+      new Parameter(
+        name = "type",
+        in = ParameterIn.QUERY,
+        description = "Filter by the type of the searched works",
+        required = false,
+        schema = new Schema(
+          allowableValues = Array("Standard", "Collection", "Series", "Section")
```
jamieparkinson, line 335: There's a bit of a corner we need to navigate here: we want to return `"type": "Work"` for `WorkType.Standard` works, but that means we either need to 
- make the filter a value (`"Standard"`) that API users never see (as in, `type=Standard` gives you works with `"type": "Work"`), as I've done here
- add a reverse mapping from displayed `type` to internal `workType` so we can filter for `type=Work` but the ES search is performed for `workType: Standard`.
- choose to display "Standard"
- choose to store "Work"

What do we think?

kenoir: Can you fill out the context a bit here in the PR description?

jamieparkinson: I'll try my best but as discussed previously I don't actually have a very clear understanding of the application of this work...

jamesgorrie: I would just have the values `"Collection", "Series", "Section"` and leave the model as `Work` on things that don't have a specialisation.

We have decided not to have an optional label field indicating this specialisation, so we have to document that it is more an inheritance model i.e. These three extend `Work`.

jamieparkinson: What you're describing there is what we do already - this question is about the filter name. I am going to rename `WorkType.Standard` to `WorkType.Work` rather than mapping (external) `"Work"` to (internal) `"Standard"`, unless you think there's an issue with colliding names there?

jamesgorrie: Great.

I still wouldn't have a filter for `Work` though as that should return everything given the inheritance model.

I am working on the location filter and rather than having a `Seq[String]` as the filter property, I have a `Seq[LocationType]` which is an enum of the available values, which could help with this.

jamieparkinson: Ah, I see what you're saying - in this case we have a problem because it isn't an inheritance model, and we _would_ want to be able to filter for standard works, so perhaps we shouldn't be using the `"type"` field for this info?

jamesgorrie: > and we would want to be able to filter for standard works

Where's the use case for this? 

> perhaps we shouldn't be using the "type" field for this info?

Yea, I'm not sure it is the _perfect_ place for this, but after many backwards-and-forwardsing this was the choice. As per my above comment, and as I have expressed previously, I am not 100% sure where we are going to use this filter soit's hard to tell what is the _perfect_ place for it.

jamesgorrie (approved):

PR #879, opened 2020-09-22
tweak contributors query

### LINKED ISSUE #736: score according to best match in contributors list
i think we can still improve the way we're scoring contributors - we should be looking for the best matches within each contributor in the list, not the overall match across the list of contributors

eg. when searching for `william blake`,
```{id: "12345", "contributors": ["william blake", "someone else"]}```
should be scored much higher than 
```{id: "67890", "contributors": ["william someone", "blake else"]}```

i think i'm seeing examples of the second example being given the same score as the first

harrisonpim:
looks like the solution to the contributors issue is to use a nested query on the contributors and changing its default `score_mode` to `max` https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-nested-query.html

syntax is probably all wrong here - might need some help sorting out the actual scala.

resolves #736


jamesgorrie (commented):
Do we need to make the field nested in this case? If we do maybe we can consider having a field used only in search and `copy_to` the value?

As per the conversation with @jamieparkinson having the index as the representation of the model and searchable index is going to start showing it's cracks.

PR #880, opened 2020-09-23
add location type filter

jamesgorrie:
After much tooing-and-frowing - it seems like this is the much easier way to do this.

We will be adding some more data on ClosedStores / OpenShelves later.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
@@ -3,13 +3,13 @@ package uk.ac.wellcome.platform.api.models
 import uk.ac.wellcome.models.work.internal.{AccessStatus, WorkType}
 import java.time.LocalDate
 
-import uk.ac.wellcome.display.models.LocationTypeRequest
+import uk.ac.wellcome.display.models.LocationTypeQuery
 
 sealed trait DocumentFilter
 sealed trait WorkFilter extends DocumentFilter
 sealed trait ImageFilter extends DocumentFilter
 
-case class ItemLocationTypeFilter(locationTypes: Seq[LocationTypeRequest])
+case class ItemLocationTypeFilter(locationTypes: Seq[LocationTypeQuery])
     extends WorkFilter
 
 case class ItemLocationTypeIdFilter(locationTypeIds: Seq[String])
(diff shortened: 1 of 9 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/FiltersAndAggregationsBuilder.scala
```diff
@@ -79,18 +79,19 @@ class FiltersAndAggregationsBuilder(
...
-    case VisibleWorkFilter         => None
-    case _: LanguageFilter         => Some(AggregationRequest.Language)
-    case _: GenreFilter            => Some(AggregationRequest.Genre)
-    case _: SubjectFilter          => Some(AggregationRequest.Subject)
-    case _: LicenseFilter          => Some(AggregationRequest.License)
-    case _: IdentifiersFilter      => None
-    case _: CollectionPathFilter   => None
-    case _: CollectionDepthFilter  => None
-    case _: AccessStatusFilter     => None
+    case _: FormatFilter             => Some(AggregationRequest.Format)
```
jamieparkinson, line 82: Is there a meaningful change in ordering here? Can't see anything but sort of hard to read

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
```diff
@@ -1,14 +1,18 @@
 package uk.ac.wellcome.platform.api.models
 
 import uk.ac.wellcome.models.work.internal.AccessStatus
-
 import java.time.LocalDate
 
+import uk.ac.wellcome.display.models.LocationTypeRequest
```
jamieparkinson, line 6: Not totally sure about the `Request` naming... but can't think of anything better.

jamieparkinson (approved):

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/FiltersAndAggregationsBuilder.scala
jamesgorrie: Put the item bits together, and the tabbing changed

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DocumentFilter.scala
jamesgorrie: We've used it elsewhere, so copying. I think I came up with it and felt weird at the time. Query?

jamieparkinson: Yeah think I prefer query!

PR #883, opened 2020-09-24
Use latest TF modules to add deployment label to services

kenoir:
This updates the ECS service modules to add a deployment:label tag to services - this is populated by the latest version of weco-deploy with the release id so that it is possible to identify which services and tasks originated from which release.

*Note:* This terraform has been applied.


jamesgorrie (approved):
Super good!

PR #884, opened 2020-09-24
add location type aggregation

jamesgorrie:
I had to invent a new piece of model specifically for the aggregations. It does seem a little odd that we:
* decode into the internal model from elastic specific JSON
* convert those to display models
* encode the display models as JSON

It feels like we should be able to just encode elastic JSON => display model as that's the only place it is used. Thoughts?


jamieparkinson:
How come you can't use the same type as for the filter? `LocationTypeQuery`

jamesgorrie:
> How come you can't use the same type as for the filter? `LocationTypeQuery`

Good shout. I suppose they are different things, and I will hopefully do some refactoring on the Location types when removing the new types that we don't need any more and add more descriptive data to the locations.

jamieparkinson (approved):
🎉

PR #887, opened 2020-09-28
CI environment has same auth as local

kenoir:
We should now be able to remove the comment/uncomment auth mechanism as CI will behave the same way as local.


jamesgorrie (approved):
Love it

jamieparkinson (approved):
🎉

PR #889, opened 2020-09-28
Modify work states to be Source -> Merged -> Denormalised -> Identified

warrd:
## RFC

wellcomecollection/docs#39

## Description

#861 laid the groundwork for separating work pipeline state from the sort of work. Here we build on that my introducing 4 distinct states for a work to be in, depending on where the work is in the pipeline:

* `Source` is the initial state of a work generated by the `transformer`
* `Merged` represents a work that has been output from the `matcher` / `merger`, with data potentially merged from multiple works
* `Denormalised` indicates a work which has had relations added by the `relation_embedder`
* `Identified` is the final state of the work after it has been through the `id_minter` 

The pipeline from this perspective is a finite state machine taking the work through the different states.

As well as providing clarity about what stage of the pipeline the work is in, this will enable embedding of related works, by adding a `relations` field to the `Denormalised` and `Identified` states:

```scala
sealed trait WorkState

object WorkState {
  ...

  case class Denormalised(
    relations: Relations[Denormalised],
  ) extends WorkState

  case class Identified(
    canonicalId: String,
    relations: Relations[Identified],
  ) extends WorkState
}

case class Relations[State <: WorkState](
  parts: Option[List[Work[State]]],
  partOf: Option[List[Work[State]]],
  precededBy: Option[List[Work[State]]],
  succeededBy: Option[List[Work[State]]],
)
```


kenoir:
Can you fill out the description to state the reason / outcome of this work?

kenoir (commented):
Renaming the states makes it much clearer what state works are in 👍

warrd:
> Can you fill out the description to state the reason / outcome of this work?

Will do, still WIP for now

warrd:
@kenoir this is ready for review now

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -113,21 +112,55 @@ sealed trait WorkState {
 
 object WorkState {
 
-  // TODO: for now just 2 states, in the end the states will correspond to the
-  // block comment above
-
-  case class Unidentified(
+  case class Source(
```
jamieparkinson, line 115: Are there any other "interesting" files to look at beside this one? ie ones that aren't just renaming etc

warrd: This is the main one, but probably also important to pay attention to what's changed in the merger (it now takes in `Work[Source]` and outputs `Work[Merged]`)

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -113,21 +112,55 @@ sealed trait WorkState {
...
   }
 }
+
+object WorkFsm {
+
+  import WorkState._
+
+  implicit class TransitionSourceWork(work: Work[Source]) {
+
+    def transitionToMerged(isMerged: Boolean): Work[Merged] = {
```
jamieparkinson, line 154: Not sure about the naming here - `transitionToMerged(isMerged = false)` seems sort of self-contradictory. Now I think about it, perhaps the naming problem is really with the `isMerged` field (the contradiction is in `Merged(isMerged = false)`)... Some suggestions I haven't thought about much:

- `pure`
- `hasMultipleSources`
- `passedThrough`
- `singleSource` (negated)

warrd: I think `hasMultipleSources` is really clear. @jamesgorrie any thoughts?

jamesgorrie: Why not just `toMerged` and `hasMultipleSources`? Could that be derived from the `otherIdentifiers` property?

warrd:
I have modified the way you make a FSM transition on a `Work` with my latest commit https://github.com/wellcomecollection/catalogue/pull/889/commits/bb47235503ba583e976d3ad32811b5370b8fe36d.

Before:

```scala
work.transitionToMerged(hasMultipleSources)
```

After:

```scala
work.transition[Merged](hasMultipleSources)
```

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
warrd: The `merged` field is never actually used as far as I can tell, I think it would probably be useful sometimes for filtering things in the elasticsearch index, in which case we need an explicit field rather than a derived one

jamieparkinson (approved):
:tada:

kenoir:
Thanks for filling out the description. Really helps me see what's going on.

PR #890, opened 2020-09-28
Scored colour filter

jamieparkinson:
The current implementation of the images colour filter is applied in the ES filter context and so does not take into account the scoring of the colour query. This means that the results can, confusingly, show very good colour palette matches much further down than poorer matches. This PR moves the filter into the query context (inside a `bool.must`) and so affects the scoring.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
@@ -24,7 +24,9 @@ case class ElasticsearchQueryOptions(filters: List[DocumentFilter],
                                      aggregations: List[AggregationRequest],
                                      sortBy: List[SortRequest],
                                      sortOrder: SortingOrder,
-                                     searchQuery: Option[SearchQuery])
+                                     searchQuery: Option[SearchQuery]) {
+  lazy val (scoredFilters, unscoredFilters) = filters.partition(_.scored)
+}
 
 class ElasticsearchService(elasticClient: ElasticClient)(
   implicit ec: ExecutionContext
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
@@ -50,7 +50,7 @@ class ElasticsearchService(elasticClient: ElasticClient)(
       .request(
         queryOptions,
         index,
-        scored = queryOptions.searchQuery.isDefined
+        scored = queryOptions.searchQuery.isDefined || queryOptions.scoredFilters.nonEmpty
       )
       .trackTotalHits(true)
     Tracing.currentTransaction.addQueryOptionLabels(queryOptions)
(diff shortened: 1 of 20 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -48,7 +50,7 @@ class ElasticsearchService(elasticClient: ElasticClient)(
       .request(
         queryOptions,
         index,
-        scored = queryOptions.searchQuery.isDefined
+        scored = queryOptions.searchQuery.isDefined || queryOptions.scoredFilters.nonEmpty
```
jamesgorrie, line 53: It doesn't feel like it makes that much sense to have this on a filter per filter bases, and might lead to confusion.

Why not have it as an explicit value. on the `ElasticsearchQueryOptions`?

jamesgorrie: The reason it feels odd is it is on a per request basis, and it feels hidden within the filters.

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesRequestBuilder.scala
```diff
@@ -36,9 +37,12 @@ class ImagesRequestBuilder(queryConfig: QueryConfig)
...
           .getOrElse(boolQuery)
-          .filter(queryOptions.filters.collect {
-            case filter: ImageFilter => buildImageFilterQuery(filter)
-          })
+          .must(
+            buildImageFilterQuery(queryOptions.scoredFilters)
+          )
+          .filter(
+            buildImageFilterQuery(queryOptions.unscoredFilters)
+          )
```
jamesgorrie, line 45: Again feels strange as these are - could we not have `DocumentFilters` and `DocumentQueries`?

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
jamieparkinson: I'm not sure I understand the comment about it not making sense on "filter per filter bases" but I agree that having it all live in `filters` / `DocumentFilter` feels odd (it was just the minimal change given how much the query config gets passed around through various case classes etc). That said, this is still a filter because it's a `must`, not a `should`, and so it's hard to know how to name it in such a way that indicates how it should be used...

Are you suggesting that we have `scoredFilters` and `unscoredFilters` as explicit fields that are populated "manually" on `ElasticsearchQueryOptions` and `ImagesSearchOptions`? Or something else?

kenoir:
@jamieparkinson Please could you fill out the description to contextualise the change a bit?

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
jamesgorrie: I am concerned as we have created abstractions on top of elastic DSL, which nearly always feel as though they can be expressed as part of the DSL.

Currently we have starting at the request =>
`QueryParams` => `SearchOptions`
`SearchOptions` => `ElasticsearchQueryOptions`
`ElasticsearchQueryOptions` => `ElasticsearchDSL`

I wonder if we had to start with what we want in Elastic i.e. and work up to the `SearchOptions` we might have something similar, if anything we could cut out `SearchOptions` => `ElasticsearchQueryOptions`.

For this instant we might have:
```
scala case SearchOptions(filters: List[Filter], musts:List[Musts], size: Int, sort: List[Sort])
```

Which can then be converted directly into ES queries.

In short, I would do what you've suggested explicitly saying these are `mustQueries` to avoid having to figure out what that infers further down the line.

jamesgorrie (approved):
Neato!

PR #894, opened 2020-09-29
Add Relations model

warrd:
Adds a new `Relations` model for storing related works on a particular work.

Note this is similar to the existing [`RelatedWorks`](https://github.com/wellcomecollection/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/RelatedWorks.scala) model, although now things are a bit more general and also simpler as the related works are not restricted to being in the `Identified` state, and we can use `Option[List[Work[State]]]` directly than requiring a seperated `RelatedWork` type.


jamieparkinson (approved):
Lovely typeclass in `Transition`

PR #896, opened 2020-09-30
Remove ontologyType fields from internal_model

warrd:
Apart from `SourceIdentifier`, where it is used for generation of the canonical ID, `ontologyType` is an unnecessary field.


common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayGenre.scala
```diff
@@ -18,7 +18,10 @@ case class DisplayGenre(
...
             includesIdentifiers: Boolean): DisplayGenre =
-    DisplayGenre(label = genre.label, concepts = genre.concepts.map {
-      DisplayAbstractConcept(_, includesIdentifiers = includesIdentifiers)
-    }, ontologyType = genre.ontologyType)
+    DisplayGenre(
+      label = genre.label,
+      concepts = genre.concepts.map {
+        DisplayAbstractConcept(_, includesIdentifiers = includesIdentifiers)
+      }
+    )
```
alexwlchan, line 26: ✨

alexwlchan (approved):
So much nicer!

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
```diff
@@ -86,8 +86,8 @@ object Aggregations extends Logging {
 
   implicit val decodeLocationTypeFromLabel: Decoder[LocationTypeQuery] =
     Decoder.decodeString.map {
-      case "DigitalLocation"  => LocationTypeQuery.DigitalLocation
-      case "PhysicalLocation" => LocationTypeQuery.PhysicalLocation
+      case "DigitalLocationDeprecated"  => LocationTypeQuery.DigitalLocation
```
jamieparkinson, line 89: Are we sure we want this? @jamesgorrie?

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
```diff
@@ -179,7 +173,7 @@ case object DisplayWork {
         if (includes.images)
           Some(work.data.images.map(DisplayWorkImageInclude(_)))
         else None,
-      ontologyType = displayWorkType(work.data.workType),
+      ontologyType = DisplayWorkType(work.data.workType),
```
jamieparkinson, line 176: I'm not sure that this should be a whole class of its own as it's just a string and so is a bit confusing to read - that's why I didn't do it like this the first time round

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
jamesgorrie: This should stay the same, I need to do the work to go and get rid of the `Deprecated` stuff.

common/display/src/main/scala/uk/ac/wellcome/display/models/LocationTypeQuery.scala
```diff
@@ -1,8 +1,9 @@
 package uk.ac.wellcome.display.models
 
 sealed trait LocationTypeQuery { this: LocationTypeQuery =>
-  def name = this.getClass.getSimpleName.stripSuffix("$")
+  def name = s"${this.getClass.getSimpleName.stripSuffix("$")}Deprecated"
```
jamesgorrie, line 4: As per the above comment, is there a reason we need to change this?

jamesgorrie (changes requested):
Around locations, they should be stored as `DigitalLocation`.

I need to go and finish the work un-deprecating the locations.

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
warrd: It needs to be used in a few places hence why I split it out

jamieparkinson: Generally don't really like an overloaded `apply` that returns a type other than the class (even when it's a singleton / object), and even more so when that type is a primitive

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
warrd: Not sure I understand what the issue with this is :thinking: 

This currently queries the deprecated locations field, but using the `ontologyType` rather than `type`. Can you explain what the problem is?

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
warrd: Yeh I get you. I am not bothered about it too much personally, I guess I am viewing `DisplayWorkType` as an alias for `String` here, I could be more explicit and add the type alias but scala is annoying in that it doesn't allow type alias at top level.

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayImageSource.scala
```diff
@@ -24,6 +24,8 @@ object DisplayImageSource {
     imageSource: ImageSource[DataState.Identified]): DisplayImageSource =
     new DisplayImageSource(
       id = imageSource.id.canonicalId,
-      ontologyType = imageSource.ontologyType
+      ontologyType = imageSource match {
+        case SourceWorks(work, _) => DisplayWorkType(work.data.workType)
```
jamieparkinson, line 28: What if the canonical and redirected works have different `workType`s? `Work` is never going to be wrong (as the other types are regarded as subclasses of `Work`) so I think it might be more correct to say that an `imageSource` of type `SourceWorks` always has `ontologyType = "Work"`

warrd: The problem is that the `type` field of the work with that ID is not necessarily going to be `Work`, so thought it would make more sense to match that. I am not sure inheritance really applies here personally, but unsure about this whole ontology business in general so can just return `Work` at all times if you think that makes sense.

jamieparkinson: > the type field of the work with that ID is not necessarily going to be Work

That's true, but remember there are 2 works in `SourceWorks` that are not necessarily going to have the same `type` field - so I think we're forced to return `Work` at all times. The inheritance thing was a biiiiig discussion last week that I suspect we'll want to return to at some point, but for now we are claiming (although not enforcing in code) that these types inherit as I described.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
jamesgorrie: The problem is with adding the `Deprecated` part to the string. It should just be `DigitalLocation` and `PhysicalLocation`.

The `Deprecated` classes are going to be un-deprecated as the work done there was made redundant.

warrd: At that stage we can remove the `Deprecated` suffix then. For now it is required as that is how they are serialised

jamesgorrie: Yip - Let's do it like that.

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayImageSource.scala
warrd: Cool makes sense, will change to `Work` :+1: 

Any chance for a very brief summary of the discussion?

jamieparkinson: Some context here: https://github.com/wellcomecollection/catalogue/pull/878#discussion_r492741671

We didn't want to expose `Standard` (as a `type`) in the API and also didn't want to have a `WorkType.Work` given that all `Work`s are of type `Work` (🤯 ). Therefore we say that you can have works of `type` `Collection`, `Series`, or `Section`, all of which are subsets of `Work`. I don't think any of us are totally happy with this (I know I'm not) but it seemed to be the most pragmatic solution for the situation we were/are in.

warrd: :+1: makes sense.

Maybe we could remove `WorkType.Standard` altogether and make `workType` an optional field? (potentially would make sense to rename it in that case, `WorkSpecialisation` or something..)

jamieparkinson: I think this is where the inheritance stuff falls down because it _is_ useful to know that something is a `WorkType.Standard`, so it doesn't feel like a good idea to make it optional - there's tension between the API/Display model and the internal model

jamesgorrie (approved):

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayImageSource.scala
warrd: I don't see the difference between having it as `None` vs having a separate case class for it. To me it is clearer with the optional field as it's existence suggests it is a more specialised thing, rather than a completely different type.

PR #899, opened 2020-09-30
Enforce that "WorksIncludes.includeAll()" includes everything in the compiler

alexwlchan:
This comes from a throwaway comment from Jonathan in preservation planning: we should make sure that the snapshot generator really includes everything.

The previous code for doing this was less than enlightening:

```scala
def includeAll(): WorksIncludes = WorksIncludes(
  true, true, true, true, true, true, true, true, true, true, true, true
)
```

The underlying `apply()` method it's calling has a list of default boolean parameters, all of which default to `false`. It would have be completely possible for us to add a new WorksIncludes parameter, and forget or not notice to update `includeAll()` to match. 

Which we did. This code is wrong.

Or is it?

It‘s not immediately obvious. You have to count `true`s and compare them to the definition of WorksIncludes, which is somewhat annoying.

This PR changes the implementation so that the compiler will check this for us. The WorksIncludes class now holds the booleans directly, and it doesn't default any of the arguments – if we miss one in includeAll(), the code won't compile.

For https://github.com/wellcomecollection/platform/issues/4815

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/WorksIncludes.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/WorksIncludes.scala
@@ -17,7 +17,7 @@ object WorkInclude {
   case object SucceededBy extends WorkInclude
 }
 
-case class BetterWorksIncludes(
+case class WorksIncludes(
   identifiers: Boolean,
   items: Boolean,
   subjects: Boolean,
@@ -34,9 +34,9 @@ case class BetterWorksIncludes(
   def anyRelation: Boolean = parts || partOf || precededBy || succeededBy
 }
 
-case object BetterWorksIncludes {
-  def apply(includes: WorkInclude*): BetterWorksIncludes =
-    BetterWorksIncludes(
+case object WorksIncludes {
+  def apply(includes: WorkInclude*): WorksIncludes =
+    WorksIncludes(
       identifiers = includes.contains(WorkInclude.Identifiers),
       items = includes.contains(WorkInclude.Items),
       subjects = includes.contains(WorkInclude.Subjects),
@@ -51,8 +51,8 @@ case object BetterWorksIncludes {
       succeededBy = includes.contains(WorkInclude.SucceededBy),
   )
 
-  def includeAll(): BetterWorksIncludes =
-    BetterWorksIncludes(
+  def includeAll(): WorksIncludes =
+    WorksIncludes(
       identifiers = true,
       items = true,
       subjects = true,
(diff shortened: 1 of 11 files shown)


jamieparkinson (approved):
Every time I've changed the `true true true true true` bit I've felt bad about it, thank you for actually doing something about it! 🎉

common/display/src/main/scala/uk/ac/wellcome/display/models/WorksIncludes.scala
```diff
@@ -17,57 +17,53 @@ object WorkInclude {
...
-  def contributors = includes.contains(WorkInclude.Contributors)
-  def production = includes.contains(WorkInclude.Production)
-  def notes = includes.contains(WorkInclude.Notes)
-  def images = includes.contains(WorkInclude.Images)
-  def parts = includes.contains(WorkInclude.Parts)
-  def partOf = includes.contains(WorkInclude.PartOf)
-  def precededBy = includes.contains(WorkInclude.PrecededBy)
-  def succeededBy = includes.contains(WorkInclude.SucceededBy)
-  def anyRelation = parts || partOf || precededBy || succeededBy
+case class WorksIncludes(
```
kenoir, line 20: Nicer.

common/display/src/main/scala/uk/ac/wellcome/display/models/WorksIncludes.scala
```diff
@@ -17,57 +17,53 @@ object WorkInclude {
...
-      if (images) Some(Images) else None,
-      if (parts) Some(Parts) else None,
-      if (partOf) Some(PartOf) else None,
-      if (precededBy) Some(PrecededBy) else None,
-      if (succeededBy) Some(SucceededBy) else None,
-    ).flatten
+case object WorksIncludes {
+  def apply(includes: WorkInclude*): WorksIncludes =
+    WorksIncludes(
+      identifiers = includes.contains(WorkInclude.Identifiers),
```
kenoir, line 40: Nicerer.

common/display/src/main/scala/uk/ac/wellcome/display/models/WorksIncludes.scala
```diff
@@ -17,57 +17,53 @@ object WorkInclude {
...
+      precededBy = includes.contains(WorkInclude.PrecededBy),
+      succeededBy = includes.contains(WorkInclude.SucceededBy),
   )
 
-  def includeAll(): WorksIncludes = WorksIncludes(
-    true, true, true, true, true, true, true, true, true, true, true, true
-  )
+  def includeAll(): WorksIncludes =
+    WorksIncludes(
+      identifiers = true,
```
kenoir, line 56: Nicerist

kenoir (approved):

PR #900, opened 2020-09-30
Denormalise relations

warrd:
This updates the `relation_embedder` so that the `Relations` model is added to the state in the transition from `Merged` -> `Denormalised`.

This PR is already getting quite big, so will introduce a further PR for tests of the `RelationEmbedderWorkerService` (currently just tests of the `RelationsService` who's job is to actually fetch the `Relations` for some work rather than perform the denormalisation).


common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
```diff
@@ -195,5 +193,37 @@ case object WorksIndexConfig extends IndexConfig {
...
+
+object MergedWorkIndexConfig extends WorksIndexConfig {
+
+  val state = objectField("state").fields(
+    sourceIdentifier,
+    booleanField("hasMultipleSources"),
+  )
+}
+
+object DenormalisedWorkIndexConfig extends WorksIndexConfig {
```
kenoir, line 212: 👍

kenoir (approved):

PR #903, opened 2020-10-01
Refactor work generators

warrd:
For the `relation_embedder` tests the current work generators are not suitable, as we only have generators for works in `Source` or `Identified` states.

Rather than following the current pattern and adding new methods such as `createDenormalisedWorkWith` (which have verbose implementations and are not composable), I have attempted a redesign of the current generators.

Examples:

```scala
// before
createInvisibleSourceWorkWith(version = 2, items = List(item))
// after
sourceWork(version = 2).invisible().items(List(item))

// before
createIdentifiedWorkWith(canonicalId = "123", workType = WorkType.Collection)
// after
identifiedWork(canonicalId = "123").workType(WorkType.Collection)
```

### DIFF
--- a/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
+++ b/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
@@ -1,0 +1,141 @@
+package uk.ac.wellcome.models.work.generators
+
+import uk.ac.wellcome.models.work.internal._
+import WorkState._
+
+trait WorkGenerators extends IdentifiersGenerators {
+
+  def sourceWork(sourceIdentifier: SourceIdentifier = createSourceIdentifier)
+    : Work.Visible[Source] =
+    Work.Visible[Source](
+      state = Source(sourceIdentifier),
+      data = WorkData[DataState.Unidentified](),
+      version = 1
+    )
+
+  def mergedWork(
+    sourceIdentifier: SourceIdentifier = createSourceIdentifier,
+    hasMultipleSources: Boolean = false
+  ): Work.Visible[Merged] =
+    Work.Visible[Merged](
+      state = Merged(sourceIdentifier, hasMultipleSources),
+      data = WorkData[DataState.Unidentified](),
+      version = 1
+    )
+
+  def denormalisedWork(
+    sourceIdentifier: SourceIdentifier = createSourceIdentifier,
+    hasMultipleSources: Boolean = false,
+    relations: Relations[DataState.Unidentified] = Relations.none
+  ): Work.Visible[Denormalised] =
+    Work.Visible[Denormalised](
+      state = Denormalised(sourceIdentifier, hasMultipleSources, relations),
+      data = WorkData[DataState.Unidentified](),
+      version = 1
+    )
+
+  def identifiedWork(
+    sourceIdentifier: SourceIdentifier = createSourceIdentifier,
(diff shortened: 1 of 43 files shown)


kenoir:
Looks nice to me. How will we ensure that `LegacyWorkGenerators` are ultimately removed?

warrd:
If people are happy with the general form of this, I can start moving some of the tests across, but maybe more of a gradual thing than doing it all at once. In my opinion it's not that bad having both hanging around for a while with this particular bit of code (it's limited to tests and nothing fundamentally different about the two implementations), as long as we only use the newer method from now on.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
```diff
@@ -0,0 +1,136 @@
+package uk.ac.wellcome.models.work.generators
+
+import uk.ac.wellcome.models.work.internal._
+import WorkState._
+
+trait WorkGenerators extends IdentifiersGenerators {
+
+  def sourceWork(
```
jamieparkinson, line 8: What determines which are intrinsic properties (ie the parameters here) vs ones to be set via the fluent interface?

kenoir:
> as long as we only use the newer method from now on.

Can we actually annotate the old ones deprecated?

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
warrd: Anything in `state` can't be in the fluent interface. I could move the `version` outside of the parameters to the fluent interface too if you think that would make things better?

warrd:
> Can we actually annotate the old ones deprecated?

Yes we can, but creates a warning on compilation and we fail compilation on warnings.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
warrd: :+1: will do that in any case so that the intitial function creates a work with some state and the subsequent operations apply to the work independent of its state.

kenoir:
> Yes we can, but creates a warning on compilation and we fail compilation on warnings.

Is it easy to selectively _not_ fail on deprecation warning? Cause if this is going to sit around for a while, i'd like to use annotation.

warrd:
Hmm I don't know if it is possible at least according to [this](https://groups.google.com/g/scala-user/c/rD7roMLF7Kc?pli=1) from 2015. Also would probably drive me nuts if every time I ran a compile I got a bunch of deprecation warnings, though suppose that would probably incentivise me to fix it though :man_shrugging:

alexwlchan:
Stick a warning log inside the body of the methods?

Also I'm happy to crank through these, I have a fairly light workload at the moment, and it would be good for me to work on some low-stakes catalogue code.

warrd:
> Stick a warning log inside the body of the methods?

I really don't think this will help, the log output in tests is so noisy anyway (its actually quite a big problem for me when trying to work out what particular test has failed on CI or locally)

I think the fact that is has "legacy" in the name is good enough, and am happy to help with a drive to get it all switched across, I don't think will be that much work tbh.

jamieparkinson (approved):

kenoir (approved):

PR #904, opened 2020-10-01
Include a schema for the pageSize and page in the Swagger output

alexwlchan:
You get something like:

```json
    "schema" : {
        "default" : 10,
        "format" : "int64",
        "maximum" : 100,
        "minimum" : 1,
        "type" : "integer"
    }
```

In particular I hope this will expose the maximum page size in the docs on developers.wellcomecollection.org.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/Pagination.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/Pagination.scala
@@ -3,6 +3,11 @@ package uk.ac.wellcome.platform.api.rest
 import akka.http.scaladsl.model.Uri
 import uk.ac.wellcome.platform.api.models.{ResultList, SearchOptions}
 
+object PaginationLimits {
+  val minSize = 1
+  val maxSize = 100
+}
+
 trait Paginated { this: QueryParams =>
   val page: Option[Int]
   val pageSize: Option[Int]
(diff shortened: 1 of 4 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/Pagination.scala
```diff
@@ -13,8 +18,12 @@ trait Paginated { this: QueryParams =>
         .filterNot(_ >= 1)
         .map(_ => "page: must be greater than 1"),
       pageSize
-        .filterNot(size => size >= 1 && size <= 100)
-        .map(_ => "pageSize: must be between 1 and 100")
+        .filterNot { size =>
```
jamieparkinson, line 21: Fun fact: I learned the other day about how Scala treats contiguous ranges as special cases and can do certain operations in O(1) upon them, so you could do (not that I'm necessarily recommending it)

```scala
pageSize
  .filterNot { size =>
	import PaginationLimits._
    minSize to maxSize contains size
  }
```

jamieparkinson (approved):

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/Pagination.scala
alexwlchan: Nice! But yes, I think I will not do that today.

PR #905, opened 2020-10-02
Start to strip out the use of LegacyWorkGenerators

alexwlchan:
A first chunk of https://github.com/wellcomecollection/platform/issues/4821

This is all the low-hanging fruit that could be replaced easily. I’m going to keep snipping away at LegacyWorkGenerators for a bit longer, but this is enough for one pull request.

### DIFF
--- a/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
+++ b/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
@@ -65,9 +65,6 @@ trait WorkGenerators extends IdentifiersGenerators {
   def denormalisedWorks(n: Int): List[Work.Visible[Denormalised]] =
     (1 to n).map(_ => denormalisedWork()).toList
 
-  def identifiedWorks(n: Int): List[Work.Visible[Identified]] =
-    (1 to n).map(_ => identifiedWork()).toList
-
   implicit class WorkOps[State <: WorkState](work: Work.Visible[State]) {
 
     def invisible(invisibilityReasons: List[InvisibilityReason] = Nil)
--- a/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
+++ b/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
@@ -3,11 +3,15 @@ package uk.ac.wellcome.models.work.generators
 import uk.ac.wellcome.models.work.internal._
 import WorkState._
 
+import scala.util.Random
+
 trait WorkGenerators extends IdentifiersGenerators {
+  private def createVersion: Int =
+    Random.nextInt(100) + 1
 
   def sourceWork(
     sourceIdentifier: SourceIdentifier = createSourceIdentifier,
-    version: Int = 1
+    version: Int = createVersion
   ): Work.Visible[Source] =
     Work.Visible[Source](
       state = Source(sourceIdentifier),
--- a/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
+++ b/common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
@@ -5,12 +5,14 @@ import WorkState._
 
 trait WorkGenerators extends IdentifiersGenerators {
 
-  def sourceWork(sourceIdentifier: SourceIdentifier = createSourceIdentifier)
-    : Work.Visible[Source] =
+  def sourceWork(
+    sourceIdentifier: SourceIdentifier = createSourceIdentifier,
+    version: Int = 1
+  ): Work.Visible[Source] =
     Work.Visible[Source](
       state = Source(sourceIdentifier),
       data = initData,
-      version = 1
+      version = version
     )
 
   def mergedWork(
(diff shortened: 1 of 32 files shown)


jamieparkinson (approved):

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
```diff
@@ -59,9 +70,6 @@ trait WorkGenerators extends IdentifiersGenerators {
   def denormalisedWorks(n: Int): List[Work.Visible[Denormalised]] =
     (1 to n).map(_ => denormalisedWork()).toList
 
-  def identifiedWorks(n: Int): List[Work.Visible[Identified]] =
-    (1 to n).map(_ => identifiedWork()).toList
```
warrd, line 63: Any reason in particular for removing this?

alexwlchan: It's not used anywhere, AFAICT? I think I thought I was looking at LegacyWorkGenerators when I ran "Find Usages", oops.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
```diff
@@ -3,14 +3,23 @@ package uk.ac.wellcome.models.work.generators
...
+    Random.nextInt(100) + 1
+
+  private def chooseFrom[T](seq: T*): T =
+    seq(Random.nextInt(seq.size))
 
-  def sourceWork(sourceIdentifier: SourceIdentifier = createSourceIdentifier)
-    : Work.Visible[Source] =
+  def sourceWork(
+    sourceIdentifier: SourceIdentifier = createSourceIdentifier,
+    version: Int = createVersion
```
warrd, line 17: As per [this discussion](https://github.com/wellcomecollection/catalogue/pull/903#discussion_r498259452) I removed version from this initial work creator function and moved it to the `version()` operation. I am not too bothered either way about having `version` here, but just in case you missed that.

Also, I am not sure the need for this to be random by default, is there some tests which require this?

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
warrd: I'm not sure it is used yet, might be required to replicate the original functionality, I dunno (`sourceWorks` was required for example).

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
alexwlchan: Ah, I just missed that. Will fix.

> I am not sure the need for this to be random by default, is there some tests which require this?

It's not so much that tests require it; more that it forces us to be explicit when we care about the exact value of the version/not.

In some tests we really care that the version is exactly 1; in others we care that some value is the same as the Work, but we don't care what that value is.

Randomising the number means we can't say `version shouldBe 1` in a test if we're relying on implicit knowledge of a default.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
alexwlchan: Yup, I think it's useful and I'll put it back.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
warrd: Hmm personally I would rather use the implicit knowledge of it being a `1` by default, as in the other case we still sort of use the implicit knowledge of it being a random int between 1 and 100, and have all the indeterminacy that comes with that that (we would for example fail 1 in 100 times in the cases where we are doing `version shouldBe 1`, which imo is much worse than failing every time as less likely to be fixed). Up to you though.

alexwlchan: I prefer the non-deterministic approach – it forces us to be explicit. Rather than failing 1 in 100 times, it fails 99 in 100 times (any time it selects a version that isn't `1`), so it bubbles up very quickly.

e.g. in https://github.com/wellcomecollection/catalogue/pull/905/files#diff-7524ba0dce08c9f4380c755e2a5c8b59

We had `workAv1` but never did the test assert that we actually set the version to 1! So the test is quite fragile; it relies on implicit behaviour of the test helpers. Randomising the version exposes this implicit assumption, and forces the test to be explicit that yes, this really is version 1.

I'm going to look at using the `version` helper for consistency. I've had to tweak the name to `withVersion` (otherwise Scala doesn't know if `work.version` is the attribute or the modifier method).

warrd: Cool, no worries, we do have all sorts of random failures in our catalogue tests so maybe I'm just being over sensitive about that. I guess I would expect a test of `version` to always set it explicitly, but I agree doing it random does force that requirement.

alexwlchan: Yeah, and IME it's usually pretty obvious when your failure is caused by a version issue.

(Although honestly given how often I'll be running the tests during this work, I expect to stamp out pretty much all the non-determinism caused by this change as I go along.)

PR #907, opened 2020-10-02
Use the new format for SnapshotJob

alexwlchan:
Following the spec in https://github.com/wellcomecollection/platform/issues/4817

* We use the structured S3ObjectLocation type that we have everywhere else
* The SnapshotJob includes the request time from the scheduler Lambda, so we can tell if there's a long delay between the scheduler and the snapshot generator running

Plus bug fixes and code tidies.


kenoir (approved):

kenoir:
Worth being aware - there is not stage instance of the snapshot generator at the moment, so it will not get automatically deployed - though we could certainly create one.

alexwlchan:
I’m not going to merge this until Monday, so I can deploy both parts together.

PR #908, opened 2020-10-02
Image versioning and merging

jamieparkinson:
This is quite a fiddly PR / handful of bugs so I will try to give some context and explanation here:

## The symptoms

Some images have a `source.canonicalWork` which is a redirected work, and some works have fewer images in the image search than they claim to in the work data itself. These are 2 sides of the same coin in that they relate to incorrect/inconsistent denormalisation of works and image.

## The root cause

Images have a `version` which comes from their original source work. For works, we have some [additional logic](https://github.com/wellcomecollection/catalogue/blob/7a5747f646fac82855e4813298d77c9640e9dab9/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala#L67) that ensures that merged (and redirected) works overwrite their same-versioned but un-merged counterparts. However, we don't have this for images - so once an image has been ingested, a later merge which affects its sources and the redirects thereof does _not_ make it into the index.

## Pulling the thread

Addressing this issue, I encountered some further bugs that have been present for some time, but haven't noticeably affected us:

- The merged version override assumes merges of _pairs_ of works only (as per the previous merger design). This means that when a work that has one other work merged into it has another work merged in, this change is not reflected.
- The `canonicalWork` on image `source`s refers to the unmerged version of that work - therefore making this denormalisation inconsistent.

I think these haven't affected us because of the order-of-magnitude difference in the number of the Sierra source works vs other sources - therefore meaning that the Sierra (target) works were nearly always coming through the pipeline last. 

## How this PR addresses these issues

- `ImageSource`s gain a `version` to reflect changes to the denormalised source data even when the image itself does not change.
- `hasMultipleSources` in `WorkState` is replaced with an integer, `nMergedSources`, to handle the cases where merge state changes after one merge has already occurred (non-pairwise merging).
- `nMergedSources` is propagated to the `SourceWork` implementation of `ImageSource` in order to be able to handle merge state changes in the denormalised work data on images.
- Image "merging" (attaching of source data) is performed at the top level of the merger rather than within the field-level rule.
- Work redirecting (and therefore merge state) is decoupled from the field-level rules by requiring that the `Merger` itself decides which rules' sources should be redirected. This enables `nMergedSources` to be counted in a clean and deterministic way that includes images (in the same way that `hasMergedSources` also included image merging even where no redirects occurred).

## Interesting changes

- [New image indexing tests](https://github.com/wellcomecollection/catalogue/pull/908/files#diff-8517ccf071721b49c637c7fd4db349afR20)
- [How `nMergedSources` is used](https://github.com/wellcomecollection/catalogue/pull/908/files#diff-1dbafb0799491431076d44fa2df925c7R35)
- [Explicit redirects / source counting / image merging in the merger](https://github.com/wellcomecollection/catalogue/pull/908/files#diff-99f24e63635e3af7d9bb5a9766c38e3aR153)
- [`ImageSource` versioning](https://github.com/wellcomecollection/catalogue/pull/908/files#diff-21d7671c131b3525b3c6549ad38d71f0R14)


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
```diff
@@ -108,32 +132,53 @@ object PlatformMerger extends Merger {
...
+        nSources <- State.inspect[MergeState, Int](_.size)
       } yield
         MergeResult(
-          mergedTarget = work.transition[Merged](true),
-          images = images
+          mergedTarget = work.transition[Merged](nSources),
+          images = unmergedImages.map { image =>
+            image mergeWith (
+              canonicalWork = work.toSourceWork,
+              redirectedWork = sources
```
jamieparkinson, line 172: I made the decision to do this search for the image source because the alternative that I could think of would be to have `ImagesRule` return a kind of intermediary type that might look something like 
```scala
case class ImageMergeResult(image: UnmergedImage, source: Work.Visible[Source])
```
and I decided that the alternative introduced more confusion/complexity/inconsistency compared to the other rules than the search did in this one place.

kenoir:
Thank you for adding the context and breaking out the interesting bits - I can read this and roughly follow what is going on - that would be impossible without the extra information 👍

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -31,8 +30,13 @@ object Indexable {
     new Indexable[AugmentedImage] {
       def id(image: AugmentedImage) =
         image.id.canonicalId
+
       def version(image: AugmentedImage) =
-        image.version
+        10 * (image.version + image.source.version) + (image.source match {
```
kenoir, line 35: Why do we multiply by 10?

jamieparkinson: Bleurgh, knew I'd forgotten something - same reason as the very detailed comment on the works indexer, I will clarify this

common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/test/fixtures/ElasticsearchFixtures.scala
```diff
@@ -127,6 +127,14 @@ trait ElasticsearchFixtures
     assertElasticsearchEventuallyHas(index, works: _*)
   }
 
+  def assertElasticsearchEventuallyHasImage(
+    index: Index,
+    images: AugmentedImage*): Seq[Assertion] = {
+    implicit val id: CanonicalId[AugmentedImage] =
+      (t: AugmentedImage) => t.id.canonicalId
```
alexwlchan, line 134: nit:

```suggestion
      (image: AugmentedImage) => image.id.canonicalId
```

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
```diff
@@ -23,36 +23,36 @@ trait WorkGenerators extends IdentifiersGenerators {
...
+      state = Denormalised(sourceIdentifier, nMergedSources, relations),
       data = initData,
       version = createVersion
     )
 
   def identifiedWork(
     sourceIdentifier: SourceIdentifier = createSourceIdentifier,
     canonicalId: String = createCanonicalId,
-    hasMultipleSources: Boolean = chooseFrom(true, false),
+    nMergedSources: Int = chooseFrom(0, 1, 2, 3),
```
alexwlchan, line 48: nice

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/ImageIndexableTest.scala
```diff
@@ -0,0 +1,155 @@
...
+
+          whenReady(insertFuture) { result =>
+            assertIngestedImageIs(
+              result = result,
+              ingestedImage = updatedSourceImage,
+              index = index
+            )
+          }
+      }
+    }
```
alexwlchan, line 59: It took me a moment to follow this test, because there are two versions going around. (FWIW, the storage-service has explicit types for versions to make this easier.)

Minor notes:

* Does it matter what the version of the original image is? `version = 2` looks significant, but I’m not sure why.
* How do you know the version of `source` is 1? It is because the default on `createIdentifiedWork` happens to be 1, but that's not immediately obvious reading the test.

   We care that it increases in the updated image; the exact value is less important. I would prefer something like:

    ```scala
    val updatedWork = sourceWork.toSourceWork.copy(version = sourceWork.version + 1)
    ```

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/ImageIndexableTest.scala
```diff
@@ -0,0 +1,155 @@
...
+      val source = createIdentifiedWork
+      val originalImage = createAugmentedImageWith(
+        parentWork = source,
+        redirectedWork = None,
+      )
+      val mergedSourceImage = originalImage.copy(
+        source = SourceWorks(
+          canonicalWork = createIdentifiedWork.toSourceWork,
+          redirectedWork = Some(source.toSourceWork)
+        )
```
alexwlchan, line 71: We're just replacing the redirectedWork, right? The canonicalWork is unimportant? If so, how about:

```suggestion
        source = originalImage.source.copy(
          redirectedWork = Some(source.toSourceWork)
        )
```

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/ImageIndexableTest.scala
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    it(
+      "doesn't override an image with a redirected source with a non-redirected one") {
+      val originalImage = createAugmentedImage
+      val nonRedirectedImage = originalImage.copy(
+        source = SourceWorks(
+          canonicalWork = createIdentifiedWork.toSourceWork,
+          redirectedWork = None
+        )
```
alexwlchan, line 120: Ditto above. Emphasise what's changing.

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/WorkIndexableTest.scala
```diff
@@ -132,7 +131,30 @@ class WorkIndexerTest
...
+          )
+          whenReady(insertFuture) { result =>
+            assertIngestedWorkIs(
+              result = result,
+              ingestedWork = mergedWork3,
+              index = index
+            )
+          }
+      }
+    }
```
alexwlchan, line 154: It looks to me like this test will create three completely independent works, all with different canonical IDs. That doesn't seem right.

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/fixtures/ElasticIndexerFixtures.scala
```diff
@@ -52,4 +52,13 @@ trait ElasticIndexerFixtures extends ElasticsearchFixtures with Akka {
...
     (doc: T) => indexable.id(doc)
+
+  def ingestInOrder[T](indexer: ElasticIndexer[T])(documents: T*)(
+    implicit ec: ExecutionContext): Future[Either[Seq[T], Seq[T]]] =
+    documents.tail.foldLeft(indexer.index(List(documents.head))) {
+      (future, doc) =>
+        future.flatMap { _ =>
+          indexer.index(List(doc))
+        }
+    }
```
alexwlchan, line 63: ✨

alexwlchan: Minor: usually our tests just import the global ExecutionContext at the call site rather than taking it implicitly, although it's not a big deal.

pipeline/ingestor/ingestor_common/src/test/scala/uk/ac/wellcome/platform/ingestor/common/fixtures/IngestorFixtures.scala
```diff
@@ -34,7 +34,7 @@ trait IngestorFixtures
         withBigMessageStream[T, R](queue) { messageStream =>
           val ingestorConfig = IngestorConfig(
             batchSize = 100,
-            flushInterval = 5 seconds
+            flushInterval = 1 seconds
```
alexwlchan, line 37: Why change this?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
```diff
@@ -44,20 +50,39 @@ trait Merger extends MergerLogging {
...
 
-          val remaining = (sources.toSet -- mergeResultSources)
-            .map(_.transition[Merged](false))
-          val redirects = mergeResultSources
+          val remaining = (sources.toSet -- redirectedSources)
+            .map(_.transition[Merged](0))
+          val redirects = redirectedSources
             .map(redirectSourceToTarget(target))
-            .map(_.transition[Merged](false))
+            .map(_.transition[Merged](0))
```
alexwlchan, line 85: Might be worth some keyword arguments to remind the reader what this parameter is.

pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/services/PlatformMergerTest.scala
```diff
@@ -164,14 +165,15 @@ class PlatformMergerTest
 
     val expectedRedirectedWork =
       Work.Redirected[Merged](
-        state = Merged(miroWork.sourceIdentifier, false),
+        state = Merged(miroWork.sourceIdentifier, 0),
```
alexwlchan, line 168: Keyword arguments!

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,53 @@ trait Indexable[T] {
...
 }
 
+/**
+  * Versioning:
+  *
+  * When the merger makes the decision to merge some works, it modifies
+  * the content of the affected works. Despite the content of these works
+  * being modified, their version remains the same. However, the merger
+  * attaches the number of merged sources (`nMergedSources`) to the target
+  * work and may choose to redirect some of the source works.
```
alexwlchan, line 34: Not a change for this PR, but I wonder if this speaks to a problem with the word "version"?

The "version" on a Work corresponds to the version of the work in the source system, yes? Maybe we should rename it to "sourceVersion" in a future PR, to emphasise everywhere that it's not counting distinct versions of the work?

(I was going to say "a la SourceIdentifier", but now I wonder – should it be a field *on* SourceIdentifier?)

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,53 @@ trait Indexable[T] {
...
+  * version still running through the pipeline.
+  *
+  * We also need to make sure that, if a work is modified by a source in
+  * such a way that it shouldn't be merged (or redirected) anymore, the
+  * new unmerged version is ingested and never replaced by the previous
+  * merged version still running through the pipeline.
+  *
+  * We can do that by ingesting works into Elasticsearch with a document
+  * version derived from a combination of work version, number of merged
+  * sources and work type.
```
alexwlchan, line 47: Yeah, see "work version" is confusing here. 😂 

How does the work type play into this?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,53 @@ trait Indexable[T] {
...
+  *
+  * More specifically, by multiplying the work version by 10, we make sure
+  * that a new version of a work always wins over previous versions
+  * (merged or unmerged, provided that there are fewer than 10 merged
```
alicefuzier, line 51: Is this assumption still valid?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,53 @@ trait Indexable[T] {
...
+  * sources, which we think is a reasonable assumption).
```
alexwlchan, line 52: Maybe we could drop a warning log here if nMergedSources is greater than 10?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,53 @@ trait Indexable[T] {
...
 object Indexable {
 
   implicit val imageIndexable: Indexable[AugmentedImage] =
     new Indexable[AugmentedImage] {
       def id(image: AugmentedImage) =
         image.id.canonicalId
+
       def version(image: AugmentedImage) =
-        image.version
+        10 * (image.version + image.source.version) + (image.source match {
```
alexwlchan, line 69: Do the image version and image source version move in step, or could they get out of sync?

If they're moving together: why do we need them both?

If they're moving separately: is one more than important than the other? e.g. if I have two images `{version=4, sourceVersion=6}` and `{version=6, sourceVersion=4}`, which one should win?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,53 @@ trait Indexable[T] {
...
+          case SourceWorks(_, Some(_), nMergedSources) => nMergedSources + 1
+          case SourceWorks(_, None, nMergedSources)    => nMergedSources
+          case _                                       => 0
```
alexwlchan, line 72: I wonder if it's worth tweaking this to emphasise that the `+1` is for a RedirectedWork?

And maybe I've missed something – when would we ever hit the fallthrough case?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,53 @@ trait Indexable[T] {
...
+  * (merged or unmerged, provided that there are fewer than 10 merged
+  * sources, which we think is a reasonable assumption).
+  *
+  * We make sure that a merger modified work always wins over other works
+  * with the same version, by adding the number of merged sources to
+  * work.version * 10.
+  *
+  * The exact same logic is applied to image versions, using both the
+  * merge state and the version numbers of the image's SourceWorks.
+  */
```
alicefuzier, line 60: If I've read this correctly, I think there's still a scenario where the correct version wouldn't make it to the index:
-work A version 1 is linked to work B version 1
- work A version 1 is unlinked from work B
-work A version1 is linked to work C <-- this would not make it to the index 
In practice this may be a convoluted example that never happens though..

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: I _think_ so (I added the bit about < 10 merged versions because that's new) - can you think of a counterexample?

alexwlchan (approved):
Really nice stuff, especially the explanation and comments. Good work pulling this through.

I went over it with a fine-toothed comb, so more comments than usual, but I don't think there's anything too substantial. Mostly making sure this code is as clear as possible for the next person who reads it.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: I think you're probably right (and I think there are probably some similar but different problem scenarios too). I'm kind of willing to let these go for now unless we see them happening as they seem to be bugs that've been there all along, what do you think?

alicefuzier: Actually, in the above scenario, how would unlinked workA make it to the index? the number of merged sources goes down in that case

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/WorkIndexableTest.scala
jamieparkinson: Oops! Think I lost this in a merge, will fix

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
alicefuzier: I think the above scenario is probably convoluted and I wouldn't worry. I would put a comment for whoever might need to investigate this in the future though. 
But I think the ability to unlink is a requirement, and I'm not entirely sure I understand how it works

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/fixtures/ElasticIndexerFixtures.scala
jamieparkinson: It got upset when I tried to do that for some reason 🤷‍♂️

pipeline/ingestor/ingestor_common/src/test/scala/uk/ac/wellcome/platform/ingestor/common/fixtures/IngestorFixtures.scala
jamieparkinson: Because these tests were really very sloooooooowwwwwwwwwwwwwwwww and this sped them up a bit (still quite slow though, I spent some time looking into speeding them up and concluded that it was kind of intrinsic to what they're doing)

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
alicefuzier, line 69: What source version id this one? The calculated one for es or the one from the transformer?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
jamieparkinson: Cannae do that unfortunately (at least, as far as I can tell, I did look) because this is a generic transition method where the keyword arg would be `args`

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: That's a very good point - it's appealing to put it on `SourceIdentifier` certainly, worth thinking about

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: Whether it's redirected or invisible - will clarify

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: Good point re fallthrough (also means that the fallthrough removed it won't compile due to being non-exhaustive if we add another type of source).

alexwlchan: Yeah, and then we'll have to come back and decide what the right behaviour is :)

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: You're right, and I feel similarly confused by it 🙃  - we explicitly say (and have always said) that a non-redirected work _doesn't_ override a redirected one, but what if it's become non-redirected because it's been unlinked? Perhaps we are relying on a change to the source data (and hence the version in the source) to cause that, and because changes to merger rules usually (always) require a reindex it doesn't matter. I'll add a comment!

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: They could get out of sync (and as a wider motivation I wanted to add a `version` to the `source` as it will likely be helpful when we end up having non-work image sources).

It's a good point regarding the information lost under addition... I suspect it doesn't matter very much for now, but perhaps what we want is something like (adjust constant factors and bracket expansions to your taste):

```scala
def version(image: AugmentedImage) = 
  10 * (2*image.version + image.source.version) + ...
```

What do you think?

It's frustrating (but necessary and inevitable) that we have to coerce our vectors of versions into a single scalar version for ES using what are effectively arbitrary constants, but I suppose at least it's explicit.

jamieparkinson: From the transformer

jamieparkinson: Note for future: we decided it wasn't worth doing this for now

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
alicefuzier: Not really, I guess a warning log as suggested by @alexwlchan would be enough for us to spot problems

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
alicefuzier: Yeah, I'm quite confused because now I can't understand how this ever worked. I don't think we can always assume that if something gets unlinked it's a new version in the sourcedata. For example, work A might be unlinked from work B because of a change in workB sourcedata. So workB advances in version, but workA doesn't and because the number of linked works is lower, the unlinked one never makes it in the index

alicefuzier: so I think how this used to sort of work is that previously we used the `merged` flag set by the merger to determine the version. Which would be set to `true` even when unlinking because that gets set always if the merger finds a target and potential redirects regardless of whether any of the rules got applied. So the final es version used to be the same for the linked and unlinked version of workA which would mean the second to arrive would make it to the index because the es versioning is greater _or equal_.

Bottom line is, I don't think this case works now. and I _think_ this may be a problem but admittedly the way it used to work is way too obscure

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -25,14 +24,77 @@ trait Indexable[T] {
...
+            case SourceWorks(_, Some(redirected), nMergedSources) =>
+              nMergedSources + 1
+            case SourceWorks(_, None, nMergedSources) => nMergedSources
+          })
+
+      private def warnOnTooManySources(
+        image: AugmentedImage): AugmentedImage = {
+        image.source match {
+          case SourceWorks(_, _, nMergedSources) if nMergedSources >= 10 =>
+            warn(
```
alicefuzier, line 92: Thinking about this again: should this just be a warning log? To go see warnings we need to know that there's something to take a look at, and that would be because points out to us that the data in the API is inconsistent. I'm not sure if that's what we want. I'm tempted to say make the message fail if `nMergedSources >= 10`, cause at this point we can't guarantee data consistentcy.

jamieparkinson: Good point, will change that

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: Ohh I see the difference now, good deduction. I will try keeping it like this (since it's more explicit / less obscure) and see what happens I think

PR #914, opened 2020-10-05
Use intermediate pipeline storage in merger and ID minter

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4828

## Description

This updates the communication between the `merger` and `id_minter` to use an elasticsearch index (in line with the intermediate storage RFC) instead of `big_messaging`.

Do not merge this yet: we need to set up the new infra before we can push this to `master`.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
@@ -102,7 +102,7 @@ object Indexable extends Logging {
   implicit def workIndexable[State <: WorkState]: Indexable[Work[State]] =
     new Indexable[Work[State]] {
 
-      def id(work: Work[State]): String = work.state.id
+      def id(work: Work[State]): String = work.id
 
       def version(work: Work[State]) =
         work match {
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
@@ -4,7 +4,6 @@ import scala.concurrent.Future
 import grizzled.slf4j.Logging
 
 import uk.ac.wellcome.models.work.internal._
-import WorkState.Identified
 
 abstract class Indexer[T: Indexable] {
 
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
@@ -99,18 +99,17 @@ object Indexable extends Logging {
       }
     }
 
-  implicit val workIndexable: Indexable[Work[Identified]] =
-    new Indexable[Work[Identified]] {
+  implicit def workIndexable[State <: WorkState]: Indexable[Work[State]] =
+    new Indexable[Work[State]] {
 
-      def id(work: Work[Identified]) =
-        work.state.canonicalId
+      def id(work: Work[State]): String = work.state.id
 
-      def version(work: Work[Identified]) =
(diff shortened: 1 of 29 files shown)


warrd:
Tests passing locally for this, but getting buildkite errors on `id_minter`

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -89,17 +88,16 @@ object Indexable extends Logging {
...
-        work.state.canonicalId
+      def id(work: Work[State]): String = work.id
 
-      def version(work: Work[Identified]) =
+      def version(work: Work[State]) =
         work match {
           case Work.Visible(_, _, state) if state.nMergedSources >= 10 =>
             throw new RuntimeException(
-              s"Work ${work.state.canonicalId} has ${work.state.nMergedSources} >= 10 merged sources; versioning/ingest may be inconsistent")
+              s"Work ${work.id} has ${work.state.nMergedSources} >= 10 merged sources; versioning/ingest may be inconsistent")
```
alexwlchan, line 100: I guess this (and adding `nMergedSources`) to `Work[Source]` in the previous file is to allow us to index stuff in intermediate Elasticsearch under the Work IDs?

I’m not sure that’s correct – I was expecting us to use transient IDs (probably UUIDs) for the inter-app messages, possibly combined with something like time-based indices to clean up old documents (https://www.elastic.co/blog/ttl-documents-shield-and-found)

Sidebar: time-based indices might solve the "cleaning up indices in Elasticsearch" issue. If the inter-app message is `(index, ID)`, and the sender writes to a time-based issue, we might be free to use Elastic Cloud.

alexwlchan (commented):
We'll discuss this tomorrow no doubt; I'm not entirely sure using the Work IDs is the correct approach for this intermediate storage (but I may have missed a previous discussion).

pipeline/id_minter/id_minter_works/src/main/scala/uk/ac/wellcome/platform/id_minter_works/Main.scala
```diff
@@ -32,10 +40,14 @@ object Main extends WellcomeTypesafeApp {
       )
     )
 
+    val elasticClient = ElasticBuilder.buildElasticClient(config)
+    val index = Index(config.requireString("es.index"))
```
alexwlchan, line 44: Minor: long term, it'd be nice if these flags reflected the type of index, e.g.

```suggestion
    val index = Index(config.requireString("es.denormalisedWorksIndex"))
```

project/Dependencies.scala
```diff
@@ -235,14 +235,15 @@ object CatalogueDependencies {
...
-    ExternalDependencies.mockitoDependencies ++
-      ExternalDependencies.mySqlDependencies ++
+    ExternalDependencies.mySqlDependencies ++
       ExternalDependencies.circeOpticsDependencies
 
   val matcherDependencies: Seq[ModuleID] =
     ExternalDependencies.mockitoDependencies ++
       ExternalDependencies.scalaGraphDependencies
 
+  val mergerDependencies: Seq[ModuleID] = Nil
```
alexwlchan, line 245: Why specify dependencies as nil? This seems odd to me – I would rather omit it entirely in `build.sbt`, so it's obvious it has no extra dependencies.

alexwlchan (approved):
Couple of tiny notes, but nothing that should block merging this and getting it deployed/tested.

pipeline/id_minter/id_minter_works/src/main/scala/uk/ac/wellcome/platform/id_minter_works/Main.scala
warrd: It's not the `denormalisedWorksIndex` that is read from yet, that will be once https://github.com/wellcomecollection/catalogue/pull/950 is merged

PR #922, opened 2020-10-06
Many merged sources

jamieparkinson:
There are 2 things in here:

- It turns out that there can be a _lot_ more than 10 sources - It looks like the maximum we can currently have is about 650, so I've bumped the multiplier to accommodate this.

- After discussing with @warrd, I changed `nMergedSources` to `numberOfSources` (which behaves much like `nMergedSources + 1` as, although it's slightly more confusing when used in the indexing logic, it makes more sense from a modelling perspective and may be more useful in future.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
@@ -85,7 +85,7 @@ object Indexable extends Logging {
             case SourceWorks(_, _, nMergedSources)
                 if nMergedSources >= versionMultiplier =>
               throw new RuntimeException(
-                s"Image ${image.id.canonicalId} has $nMergedSources >= $versionMultiplier merged sources; versioning/ingest may be inconsistent")
+                s"Image ${image.id.sourceIdentifier.toString} has $nMergedSources >= $versionMultiplier merged sources; versioning/ingest may be inconsistent")
             case SourceWorks(_, Some(redirected), nMergedSources) =>
               nMergedSources + 1
             case SourceWorks(_, None, nMergedSources) => nMergedSources
@@ -103,7 +103,7 @@ object Indexable extends Logging {
           case Work.Visible(_, _, state)
               if state.nMergedSources >= versionMultiplier =>
             throw new RuntimeException(
-              s"Work ${work.state.canonicalId} has ${work.state.nMergedSources} >= $versionMultiplier merged sources; versioning/ingest may be inconsistent")
+              s"Work ${work.state.sourceIdentifier.toString} has ${work.state.nMergedSources} >= $versionMultiplier merged sources; versioning/ingest may be inconsistent")
           case Work.Visible(version, _, state) =>
             (version * versionMultiplier) + state.nMergedSources
           case Work.Redirected(version, _, _) =>
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
@@ -79,16 +79,17 @@ object Indexable extends Logging {
       def id(image: AugmentedImage) =
         image.id.canonicalId
 
-      def version(image: AugmentedImage) =
-        versionMultiplier * (image.version + image.source.version) +
-          (image.source match {
-            case SourceWorks(_, _, nSources) if nSources > versionMultiplier =>
-              throw new RuntimeException(
-                s"Image ${image.id.sourceIdentifier.toString} has $nSources > $versionMultiplier sources; versioning/ingest may be inconsistent")
-            case SourceWorks(_, Some(redirected), nSources) => nSources
-            case SourceWorks(_, None, nSources) =>
-              nSources - 1 // The number of additional sources
-          })
+      def version(image: AugmentedImage) = {
+        val versionOffset = image.source match {
+          case SourceWorks(_, _, nSources) if nSources >= versionMultiplier =>
+            throw new RuntimeException(
+              s"Image ${image.id.sourceIdentifier.toString} has $nSources >= $versionMultiplier sources; versioning/ingest may be inconsistent")
+          case SourceWorks(_, Some(redirected), nSources) => nSources
+          case SourceWorks(_, None, nSources) =>
+            nSources - 1 // The number of additional sources
+        }
+        versionMultiplier * (image.version + image.source.version) + versionOffset
+      }
     }
 
   implicit val workIndexable: Indexable[Work[Identified]] =
@@ -100,9 +101,9 @@ object Indexable extends Logging {
       def version(work: Work[Identified]) =
         work match {
           case Work.Visible(_, _, state)
-              if state.nSources > versionMultiplier =>
+              if state.nSources >= versionMultiplier =>
             throw new RuntimeException(
-              s"Work ${work.state.sourceIdentifier.toString} has ${work.state.nSources} > $versionMultiplier sources; versioning/ingest may be inconsistent")
+              s"Work ${work.state.sourceIdentifier.toString} has ${work.state.nSources} >= $versionMultiplier sources; versioning/ingest may be inconsistent")
           case Work.Visible(version, _, state) =>
             (version * versionMultiplier) + (state.nSources - 1) // The number of additional sources
           case Work.Redirected(version, _, _) =>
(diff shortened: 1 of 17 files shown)


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -97,14 +107,20 @@ object Indexable extends Logging {
...
       def version(work: Work[Identified]) =
         work match {
-          case Work.Visible(_, _, state) if state.nMergedSources >= 10 =>
-            throw new RuntimeException(
-              s"Work ${work.state.canonicalId} has ${work.state.nMergedSources} >= 10 merged sources; versioning/ingest may be inconsistent")
+          case Work.Visible(_, _, state)
+              if state.nSources >= versionMultiplier =>
+            warnOnTooManySources(work)
+            // Clamp the increment so as not to override higher versions
+            versionMultiplier - 1
```
alexwlchan, line 114: Do you not want to multiply by the version? This is consistently going to return `version = 999`, no matter what the source version is.

alexwlchan (commented):
I'm not sure the clamping is correct.

I also don't love `nSources` as a name; it reads weird to me. I wasn't super happy with `nMergedSources` either, but I'm struggling for alternatives. `sourceRecordCount`? To reflect that this is a property of the `source`, a la SourceIdentifier.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -68,25 +68,35 @@ trait Indexable[T] {
...
-              nMergedSources + 1
-            case SourceWorks(_, None, nMergedSources) => nMergedSources
+            case SourceWorks(_, _, nSources) if nSources >= versionMultiplier =>
+              warnOnTooManySources(image, nSources)
+              // Clamp the increment so as not to override higher versions
+              versionMultiplier - 1
+            case SourceWorks(_, Some(redirected), 1)        => 1
+            case SourceWorks(_, Some(redirected), nSources) => nSources
+            case SourceWorks(_, None, 1)                    => 0
+            case SourceWorks(_, None, nSources)             => nSources - 1
```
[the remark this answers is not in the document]
jamieparkinson: Yes it is, I changed it from this because I found it confusing (in my mind, the spirit of line 90 is really `(nSources - 1) + 1` but I think I've changed my mind again and will change it back

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
jamieparkinson: oops - yes!

jamieparkinson:
re naming - the number is primarily not a property of the image `source` but rather a property of the work state, and where it appears on the `source` I kept it the same name to clarify that it is the same number

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -68,25 +68,35 @@ trait Indexable[T] {
...
-            case SourceWorks(_, Some(redirected), nMergedSources) =>
-              nMergedSources + 1
-            case SourceWorks(_, None, nMergedSources) => nMergedSources
+            case SourceWorks(_, _, nSources) if nSources >= versionMultiplier =>
+              warnOnTooManySources(image, nSources)
+              // Clamp the increment so as not to override higher versions
+              versionMultiplier - 1
+            case SourceWorks(_, Some(redirected), nSources) => nSources
+            case SourceWorks(_, None, nSources)             => nSources - 1
           })
```
alexwlchan, line 93: I wonder if this might be clearer if you broke out the bit after the second `+`?

As in:

```scala
def version(image: AugmentedImage) = {
  val sourceOffset = (image.source match { … })

  versionMultiplier * (image.version + image.source.version) + sourceOffset
```

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/ImageIndexableTest.scala
```diff
@@ -143,19 +143,34 @@ class ImageIndexableTest
...
           redirectedWork = None,
-          nMergedSources = 10
+          nSources = nSources
         )
       )
+      val slightlyFewerSourcesImage = lotsOfSourcesImage.copy(
+        source = lotsOfSourcesImage.source
+          .asInstanceOf[SourceWorks[DataState.Identified]]
+          .copy(nSources = nSources - 1)
+      )
```
alexwlchan, line 159: This test doesn't feel right to me – you're testing a side-effect of the implementation, rather than the behaviour we want.

What we want:

> If I ingest V1 with lots of sources, V2 will still win out no matter how many sources it has.

I think you want to bump the version and drop `nSources` to the minimum (1?) for a more meaningful test.

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/WorkIndexableTest.scala
```diff
@@ -132,11 +131,11 @@ class WorkIndexableTest
...
 
     it("overrides a merged work with one that has been merged again") {
-      val mergedWork1 = identifiedWork(nMergedSources = 1)
+      val mergedWork1 = identifiedWork(nSources = 2)
       val mergedWork2 =
-        mergedWork1.copy(state = mergedWork1.state.copy(nMergedSources = 2))
+        mergedWork1.copy(state = mergedWork1.state.copy(nSources = 3))
       val mergedWork3 =
-        mergedWork1.copy(state = mergedWork1.state.copy(nMergedSources = 3))
+        mergedWork1.copy(state = mergedWork1.state.copy(nSources = 4))
```
alexwlchan, line 138: nit: I would name these `mergedWork2, mergedWork3, mergedWork4` so that you the variable name matches the number of sources.

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/WorkIndexableTest.scala
```diff
@@ -178,13 +177,25 @@ class WorkIndexableTest
...
+        case (index, indexer) =>
+          val invisibleWorkInsertFuture =
+            ingestInOrder(indexer)(lotsOfSourcesWork, slightlyFewerSourcesWork)
+          whenReady(invisibleWorkInsertFuture) { result =>
+            assertIngestedWorkIs(
+              result = result,
+              ingestedWork = slightlyFewerSourcesWork,
+              index = index)
           }
       }
```
alexwlchan, line 200: Ditto above.

jamesgorrie (commented):
I'm with @alexwlchan on the naming. Without the type, I would mistake the name to mean "a list of n sources". We use `count` elsewhere for `Int` types.

alicefuzier:
Why do we want to clamp it instead of failing? I'm a bit worried that by not failing we won't be notified if by tweaking rules or whatever we increase that number again. If 1000 is not enough maybe we should think of a different versioning algorithm, like prepending the work version digit to `nSources` instead of multiplying. But I'm not convinced that not making it fail is the right strategy

alexwlchan (commented):
Couple of comments on the tests and readability of the versioning logic.

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/ImageIndexableTest.scala
jamieparkinson: You are describing a separate behaviour to what this is testing - as we clamp _at_ the maximum, this is testing that above the maximum number of sources, even a work with fewer (but still too many) sources will override an existing one. I think this is important to document in the test, so I will add another test for the behaviour you're describing.

alicefuzier:
BTW  I'm starting to think that the current algorithm for determining the elasticsearch version is no longer adequate for our use cases and we should maybe take a step back and go over them again and possibly figure out a new approach instead of tweaking the current one until it doesn't break. 
I'm starting to think that the number of sources isn't a very good discriminator for a version number and we should go back to the drawing board with this

jamieparkinson:
I think `nThings` is a common way to name a count variable and also followed `hasMergedSources` quite naturally but I could rename it to `sourceCount`?

> Why do we want to clamp it instead of failing? I'm a bit worried that by not failing we won't be notified if by tweaking rules or whatever we increase that number again. If 1000 is not enough maybe we should think of a different versioning algorithm, like prepending the work version digit to nSources instead of multiplying. But I'm not convinced that not making it fail is the right strategy

The logic behind clamping was just that it meant we could fall back to the previous behaviour when this `nSources` approach didn't work. However, I think prepending digits is a great idea - I think it's exactly equivalent to having an adaptively large power-of-10 `versionMultiplier`, which is perfect!

warrd:
Just to jump into the bikeshed, I also don't like the name `nSources`. I think these sort of names are good when e.g. based off a scientific paper where that sort of naming is common, but not with this sort of domain modelling where we want things to be more descriptive / less jargony. My preferred name is probably `numberOfSources`.

alicefuzier (approved):

PR #923, opened 2020-10-06
Add snapshot reporter lambda

kenoir:
For https://github.com/wellcomecollection/platform/issues/4817

The output in Slack should appear as follows (top successful message if snapshot found, bottom failure message if not found).

Screenshot 2020-10-07 at 10 58 25

Note: The terraform in this PR is applied.


snapshots/snapshot_reporter/src/snapshot_reporter.py
```diff
@@ -0,0 +1,132 @@
...
+def get_snapshots(es_client, days_to_fetch):
+    index = "snapshots"
+
+    date_filter = [
+        {"range": {"snapshotJob.requestedAt": {"gte": f"now-{days_to_fetch}d/d"}}}
+    ]
+    es_query = {"bool": {"filter": date_filter}}
+    sort_by = [{"snapshotJob.requestedAt": {"order": "desc"}}]
+
+    response = es_client.search(index=index, body={"query": es_query, "sort": sort_by})
```
alexwlchan, line 51: It's a small thing, but I generally prefer to construct the Elasticsearch requests in one big dict, i.e.

```python
request_body = {
    "query": {
        "bool": {
            "filter": {"range": {"snapshotJob.requestedAt": {"gte": f"now-{days_to_fetch}d/d"}}}
        }
    },
    "sort": [{"snapshotJob.requestedAt": {"order": "desc"}}],
}
```

cos then (trailing commas aside) you can drop that straight into the Elastic dev tools for experimenting, and it's easier to follow through to the docs.

alexwlchan (commented):
Looking good. ✨

snapshots/snapshot_reporter/src/snapshot_reporter.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def prepare_slack_payload(snapshots, days_to_fetch):
+    def _snapshot_message(snapshot):
+        index_name = snapshot["snapshotResult"]["indexName"]
+        document_count = snapshot["snapshotResult"]["documentCount"]
+        s3_size = snapshot["snapshotResult"]["s3Size"]["bytes"]
+
+        requested_at = parser.parse(snapshot["snapshotJob"]["requestedAt"]).strftime(
+            "%A, %B %e, %I:%M %p"
```
alexwlchan, line 63: nit: if you look at the rendered output, you'll see there's an extra space `Tuesday, October  6`. You don't want the space, you just want the leading zero trimmed. I think this does that.

```suggestion
            "%A, %B %-d, %I:%M %p"
```

snapshots/snapshot_reporter/src/snapshot_reporter.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+        started_at = parser.parse(snapshot["snapshotResult"]["startedAt"])
+
+        finished_at = parser.parse(snapshot["snapshotResult"]["finishedAt"])
+
+        return "\n".join(
+            [
+                f'*"{index_name}" on {requested_at}*',
+                f"{humanize.naturalsize(s3_size)} containing {humanize.intword(document_count)} documents ({document_count} exactly)",
```
alexwlchan, line 73: nit: consider adding commas for readability.

```suggestion
                f"{humanize.naturalsize(s3_size)} containing {humanize.intword(document_count)} documents ({humanize.intcomma(document_count)} exactly)",
```

snapshots/snapshot_reporter/src/snapshot_reporter.py
```diff
@@ -0,0 +1,132 @@
...
+                f"Took {humanize.precisedelta(finished_at - started_at)}",
```
alexwlchan, line 74: nit: is precisedelta really what we want? Telling me it took about 18 minutes is useful. Telling me it took 18 minutes and 8.83 seconds is probably too much precision.

snapshots/snapshot_reporter/src/snapshot_reporter.py
```diff
@@ -0,0 +1,132 @@
...
+            ]
+        )
+
+    heading = f"Snapshots in the last {days_to_fetch} days"
```
alexwlchan, line 78: nit: "1 days"

😢

alexwlchan (commented):
Now I've seen the output, a couple of tiny suggestions. I am very picky, so do ignore if you don't think it's worth the time.

At a higher level: in my mind, this Lambda is about alerting us to issues with the snapshot reporter. So I'd change the output somewhat:

* Unlike the storage service, we're only interested in the last snapshot. The data is there if somebody wants to see how many other snapshots were taken, but something like:

    > The last snapshot was of index *works-20200923* at *5pm yesterday*. It is 457.8MB and contains 1.2 million documents. [link to reporting cluster]

* The exact number isn't important in this message, nor is the time taken.

* What if there aren't any snapshots in the reporting period? That seems something important it  should flag specially, optionally with a pre-built link to the appropriate log group in the logging cluster – see the ingest inspector for how we do this elsewhere. https://github.com/wellcomecollection/ingest-inspector/blob/master/server.py#L92-L140

snapshots/snapshot_reporter/src/snapshot_reporter.py
jamesgorrie, line 73: Why do we have the rounded and exactly amounts? Could we just have the exactly amounts?

snapshots/snapshot_reporter/src/snapshot_reporter.py
kenoir: I find parsing the whole block confusing TBH, but I see the benefit in being able to drop it into the ES console.

kenoir:
> The exact number isn't important in this message, nor is the time taken.

I disagree, the exact number is useful when the granularity is to the nearest hundred thousand and i'm looking at the reports over a few days.

The time taken is also interesting. A very short time would be a warning something is wrong, and the time it takes is useful to know if we start to push up against visibility timeouts.

snapshots/snapshot_reporter/src/snapshot_reporter.py
kenoir: A number in millions without some extra formatting is hard to parse visually (e.g. how much is 11212422?). I think simply using comma notation might fix that though.

snapshots/snapshot_reporter/src/snapshot_reporter.py
kenoir: If we refer to only a single snapshot in the report (as you suggest) this goes away nicely.

snapshots/snapshot_reporter/src/snapshot_reporter.py
kenoir: I am inclined to agree that minute resolution is all that is required here.

kenoir:
> What if there aren't any snapshots in the reporting period? That seems something important it should flag specially.

Yes, that's incoming.

snapshots/snapshot_reporter/src/snapshot_reporter.py
jamesgorrie: Yea, big numbers are hard to process... good point.

snapshots/snapshot_reporter/src/snapshot_reporter.py
```diff
@@ -0,0 +1,135 @@
...
+    if snapshots:
+        snapshot = snapshots[0]
+
+        header_block = _create_header_block(":white_check_mark: Catalogue Snapshot")
+        section_block = _create_section_block(_snapshot_message(snapshot))
+    else:
+        header_block = _create_header_block(
+            ":interrobang: Catalogue Snapshot not found"
+        )
+        section_block = _create_section_block("No snapshot found within the last day.")
```
alexwlchan, line 97: I would include a link to the snapshot generator logs here so it's really easy to go debug, but it's not a big deal.

alexwlchan (approved):
Nice stuff. One suggestion, but it's not blocking.

PR #938, opened 2020-10-08
Add init() method to Indexer trait

warrd:
In #914 we will be using an `Indexer[Work[Merged]]` for intermediate pipeline storage in the `merger`. In order to reduce some boilerplate, here we add an `init()` method to `Indexer`: for the `ElasticIndexer` this calls of too `ElasticsearchIndexCreator` to set up the mappings given some config.

### DIFF
--- a/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
+++ b/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
@@ -13,7 +13,11 @@ import uk.ac.wellcome.elasticsearch.WorksAnalysis._
 trait IndexConfig {
   def mapping: MappingDefinition
   def analysis: Analysis
+}
 
+/** Mixin for common fields used within an IndexConfig in our internal models.
+  */
+trait IndexConfigFields {
   // `textWithKeyword` and `keywordWithText` are slightly different in the semantics and their use case.
   // If the intended field type is keyword, but you would like to search it textually, use `keywordWithText` and
   // visa versa.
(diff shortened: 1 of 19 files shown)


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
```diff
@@ -10,12 +9,23 @@ import com.sksamuel.elastic4s.circe._
...
-class ElasticIndexer[T: Indexable](client: ElasticClient, index: Index)(
+import uk.ac.wellcome.elasticsearch.{
+  ElasticsearchIndexCreator,
+  IndexConfig,
+  NoStrictMapping
+}
+
+class ElasticIndexer[T: Indexable](client: ElasticClient,
+                                   index: Index,
+                                   config: IndexConfig = NoStrictMapping)(
```
alexwlchan, line 20: Do we really want a default here? I'm not sure one config is more preferable than any other.

alexwlchan (approved):

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
jamesgorrie: It would be good to have this with no default, especially as `NoStrictMapping` isn't  a common used config, and is the outlier.

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
```diff
@@ -93,3 +93,8 @@ trait IndexConfig {
   val canonicalId = lowercaseKeyword("canonicalId")
   val version = intField("version")
 }
+
+object NoStrictMapping extends IndexConfig {
```
jamesgorrie, line 97: Would `DynamicMappingConfig` make more sense, as I think this is the use case?

I thought we were going to use `dynamic:false` though? So we could still specify mapping without them being indexed?

PR #939, opened 2020-10-09
Make it easier to debug snapshot recorder failures

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4838

* Print the record received by the snapshot recorder Lambda
* Copy the message to a separate queue

Plus replacing a couple of modules with our published versions, so we get more code reuse.


snapshots/terraform/stack/snapshot_generator/messaging.tf
```diff
@@ -0,0 +1,31 @@
+module "snapshot_generator_input_queue" {
```
kenoir, line 1: 👍 nicer

kenoir (approved):

PR #948, opened 2020-10-13
Stop using Mockito where it's unnecessary

alexwlchan:
Following a conversation with @warrd about how Mockito's runtime shenanigans are frequently a source of pain, I decided to go see if I could clean up some of our uses of Mockito. Spoiler: yes.

* Use a `MemoryMetrics[…]` instance instead of a `mock[MetricsSender]`, which is exactly what the in-memory implementations are meant to be used for.
* Remove `MockitoSugar` from a bunch of test classes that weren't using it.

I've also evicted it from the ID minter in this PR: https://github.com/wellcomecollection/catalogue/pull/914

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -144,8 +144,7 @@ lazy val relation_embedder = setupProject(
 lazy val recorder = setupProject(
   project,
   "pipeline/recorder",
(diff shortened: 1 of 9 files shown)


warrd (approved):
:grinning:

Issue #934, opened 2020-10-08
Ingest versioning is not consistent with respect to merging


jamieparkinson:

alicefuzier:
- [ ] Make METS Adapter store createdDate from the bag manifest https://github.com/wellcomecollection/catalogue/pull/946
- [ ] Update METS store to add createdDate
- [ ] Change transformers to add modifiedDate in the work
- [ ] Change merger to pick version as the highest timestamp in a gruoup of works

PR #957, opened 2020-10-15
Script to update the mets adapter

alicefuzier:
Part of https://github.com/wellcomecollection/catalogue/issues/934


alicefuzier:
Yeh, I'll add a comment. The futures stuff is something I stole from the storage-service that alex wrote. They have a nice blog post explaining it so I'll add a link to that.
As per how long it will take, I don't know really. I would expect it to be nearly finished by now though

PR #958, opened 2020-10-15
Change transformers to add modifiedTime in Source state

alicefuzier:
Part of https://github.com/wellcomecollection/catalogue/issues/934

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
@@ -1,5 +1,7 @@
 package uk.ac.wellcome.models.work.internal
 
+import java.time.Instant
+
 /** Work is the core model in the pipeline / API.
   *
   * It is parameterised by State, meaning the same type of Work can be in a
@@ -130,7 +132,8 @@ sealed trait WorkState {
 object WorkState {
 
   case class Source(
-    sourceIdentifier: SourceIdentifier
+    sourceIdentifier: SourceIdentifier,
+    modifiedTime: Instant
   ) extends WorkState {
 
     type WorkDataState = DataState.Unidentified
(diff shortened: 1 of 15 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -130,7 +132,8 @@ sealed trait WorkState {
 object WorkState {
 
   case class Source(
-    sourceIdentifier: SourceIdentifier
+    sourceIdentifier: SourceIdentifier,
+    modifiedTime: Instant
```
warrd, line 136: Maybe it would be worth adding the `modifiedTime` to `WorkData` rather than the state? I can see this being useful when looking at works in the other indices beside the source.

alicefuzier: @jamieparkinson and I have discussed hit a bit. The main annoying thing about putting it in the `WorkData` is that it cannot have a default, which makes it a bit annoying. 
Also, I think it's more akin to s`sourceIdentifier` so it makes sense to keep it next to it IMO.
 It is going to be copied across to the other `WorkState`s as part of the merger work so it is going to make it into the index anyway

warrd: It could also go inside the work types themselves (i.e. `Visible`, `Invisible`, `Redirected`) but then would that would require some duplication.

Fwiw I reckon `sourceIdentifier` should have probably been at that level rather than in the state, the only reason it's like that is because the `id_minter` needs it there.

alicefuzier: I see. I can have a look at how it looks moving it into the work types. I think duplication is inevitable. Even if it stays in `WorkState` it's going to have to be copied in all of them

kenoir (approved):

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
alicefuzier: Having discussed this with @jamieparkinson , we're going to leave this here for now. It helps in the merger part of this work if it stays in `WorkState` because we can delegate the update to the transition function. It's doable even without delegating to the transition but it looks cleaner this way.

PR #960, opened 2020-10-15
Sorry

warrd:


jamesgorrie (approved):
✨

jamesgorrie (commented):
Any reason not to use `info`?

warrd:
@jamesgorrie yup cos we don't log `info` :see_no_evil:

warrd:
I am trying connecting to the ES cluster locally rather than doing this, but might need to do this at a later point if that is not fruitful.

jamesgorrie:
@warrd I am pretty certain we log info?
https://github.com/wellcomecollection/catalogue/blob/master/pipeline/id_minter/id_minter_works/src/main/resources/logback.xml#L14

warrd:
Nope, none of it makes it through to kibana, I think we only pass through `warn` and above with firelens

PR #964, opened 2020-10-16
Make failures in the ID minter more explicit, earlier

alexwlchan:
Pulling messages from [catalogue-20201016_work_id_minter_dlq](https://eu-west-1.console.aws.amazon.com/sqs/v2/home?region=eu-west-1#/queues/https%3A%2F%2Fsqs.eu-west-1.amazonaws.com%2F760097843905%2Fcatalogue-20201016_work_id_minter_dlq), I see lots of messages like:

```json
{"jsonString": "\"sierra-system-number/b28972995\"", "type": "InlineNotification"}
```

The ID minter then passes that JSON string directly to the ElasticRetriever, which returns nothing. Normally we'd catch that, but because the ID minter reads `Json` instead of a stronger type, that nothing gets coerced to an empty list, and we see all the downstream weirdness we've been experiencing.

This patch:

* Logs the ID we're trying to look up in the ElasticRetriever. If you see a DEBUG log go by with this JSON string, you know something is up.
* Throws an exception from the Retriever if asked to retrieve a non-existent document – this should never happen in the normal operation of the pipeline. Don't let a weird response float downstream.
* Assert immediately in the ID minter if asked to mint an empty set of sourceIdentifiers, which is a sign something weird is happening. DOn't wait to hit it while trying to get the `.tail` of an empty list.
* Updates the merger to emit the work ID as a pure string, rather than an InlineNotification


jamesgorrie (approved):
Very neat

warrd:
Would you mind also reverting this commit? https://github.com/wellcomecollection/catalogue/commit/fb3c3b6bb10c8d34189579e212a5e29df7bc1fa2

PR #970, opened 2020-10-20
Fix ID minter dependencies

jamieparkinson:
I noticed that https://buildkite.com/wellcomecollection/catalogue/builds/940#0e7a0428-4ab4-423f-aa36-237cd018c0da didn't build/test the image minter despite the changes _only_ being in that project.

I'm pretty sure the issue is this leftover project JSON file, which was still picked up by the build logic and - because the original ID minter project folder is the parent of the 3 minter projects now - was overriding the actual project lookup and giving a false negative dependency.

### DIFF
--- a/.sbt_metadata/id_minter.json
+++ b/.sbt_metadata/id_minter.json
@@ -1,8 +1,0 @@
-{
-  "id" : "id_minter",
-  "folder" : "pipeline/id_minter",
-  "dependencyIds" : [
-    "internal_model",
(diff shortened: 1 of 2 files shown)


alexwlchan:
Wonder if we can make sbt do this automatically.

alexwlchan (approved):
Nice catch. I've pushed an extra commit to make sbt clean out these entries automatically in future.

PR #971, opened 2020-10-20
bump weco-deploy deploy timeout to an hour

jamesgorrie:
To help us find the sweet spot without doing too many commits / deploys.
This is part of a bigger piece of work around deployment perf budgets.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -85,4 +85,4 @@ steps:
               "--from-label", "latest",
               "--environment-id", "staging",
               "--description", $BUILDKITE_BUILD_URL,
-              "--confirmation-wait-for", 1200]
+              "--confirmation-wait-for", 3600]


.buildkite/pipeline.yml
```diff
@@ -85,4 +85,4 @@ steps:
               "--from-label", "latest",
               "--environment-id", "staging",
               "--description", $BUILDKITE_BUILD_URL,
-              "--confirmation-wait-for", 1200]
+              "--confirmation-wait-for", ${WECO_DEPLOY_STAGE_DEPLOY_TIMEOUT:-1200}]
```
kenoir, line 88: Where will this env var be set from?

jamesgorrie: CI? Or would you have it read from the tool, and pass it through?

alexwlchan: Is the intention that we'd eventually replace this with a hard-coded value, and ditch the env var?

jamesgorrie: Ditch the env var. It's hard to tell how long it's taking, all I know is "more than 9-10 minutes".

Unless maybe the app takes a never value, which would also flag up deploys that aren't running?

kenoir: Where in CI would you set that value? I think you'd end up putting it in `pipeline.yml`.

Buildkite lacks the "put the var in a box in UI" option that [you might be expecting](https://buildkite.com/docs/pipelines/environment-variables#defining-your-own).

jamesgorrie: Good point.

I've pushed the timeout to an hour to start getting back how long it actually takes. I have put a reminder in to check it on Thursday.

Thoughts?

kenoir (approved):

PR #962, opened 2020-10-15
Use latest modifiedTime in merger results

jamieparkinson:
Due to the autoformatter going a bit weird (probably from my terrible merge conflict resolutions) this also includes some linting fixes to seldom-touched Python files 🤷‍♂️

PR #972, opened 2020-10-20
Fix formatting

warrd:
Autoformat is failing on `master` builds.


jamesgorrie (approved):

jamieparkinson (changes requested):
There are more than these 2 - let's fix them all at once. You can see them by running `make format && make lint` and/or copy them from my commit where I just saw the same issue https://github.com/wellcomecollection/catalogue/pull/962/commits/6b5a119a0d2ec904db5c0d9caf2ea251768ddf4c

warrd:
@jamieparkinson have cherry picked that commit

jamieparkinson (approved):

jamesgorrie (commented):
Does python linting not have `autofix`? Or have we not implemented it into the build?

warrd:
The issue is that changes were pushed straight to `master`, and the autoformat build step does't have permission to push to `master` so fails.

jamieparkinson:
PS I don't believe all of these changes were covered by the autofix capabilities

PR #973, opened 2020-10-20
Fix ElasticRetrieverTest

warrd:


alexwlchan:
Ah, is this to ensure there’s *something* in the index?

warrd:
I'm not sure actually, this is from @jamieparkinson PR

jamieparkinson:
The change that @alex.chan made caused it to throw on the assertion rather than failing how the test expects (and therefore not testing the thing that it claims to), so yeah, have to make sure there's something in the index

warrd:
Can I get an approval for this?

jamieparkinson (approved):

PR #976, opened 2020-10-21
Use modifiedTime to version images and works when indexing

alicefuzier:


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala
```diff
@@ -74,28 +74,13 @@ trait Indexable[T] {
...
-              s"Image ${image.id.sourceIdentifier.toString} has $numberOfSources >= $versionMultiplier sources; versioning/ingest may be inconsistent")
-          case SourceWorks(_, Some(redirected), numberOfSources) =>
-            numberOfSources
-          case SourceWorks(_, None, numberOfSources) =>
-            numberOfSources - 1 // The number of additional sources
-        }
-        versionMultiplier * (image.version + image.source.version) + versionOffset
-      }
+      def version(image: AugmentedImage) =
+        image.modifiedTime.toEpochMilli
```
warrd, line 83: :tada:

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -128,6 +128,7 @@ sealed trait WorkState {
 
   val sourceIdentifier: SourceIdentifier
   val numberOfSources: Int
+  val modifiedTime: Instant
```
jamieparkinson, line 131: Is this being used anywhere?

jamieparkinson (approved):
Great to see all the deletions 🎉

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
alicefuzier: yes, it's used [here](https://github.com/wellcomecollection/catalogue/blob/ca5cbbae8835829db436c2c73b5390c3478c7ea0/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/Indexer.scala#L92) to avoid pattern matching. It's in all the states already anyway

jamieparkinson: Oh yeah, got confused 🙃

PR #979, opened 2020-10-22
Don't attempt to merge empty sets of works

warrd:
We are getting the following error on the `merger`:

```
Cannot index an empty list of documents
```

This is because we still try to apply the merge for out of date documents, and with `big_messaging` previously we would just not send the message. We should catch this earlier rather than allow pipeline storage to store an empty list.

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/MergerWorkerService.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/MergerWorkerService.scala
@@ -15,7 +15,7 @@ import uk.ac.wellcome.messaging.MessageSender
 import uk.ac.wellcome.messaging.sns.NotificationMessage
 import uk.ac.wellcome.messaging.sqs.SQSStream
 import uk.ac.wellcome.pipeline_storage.Indexer
-import WorkState.Merged
+import WorkState.{Merged, Source}
 
 class MergerWorkerService[WorkDestination, ImageDestination](
   sqsStream: SQSStream[NotificationMessage],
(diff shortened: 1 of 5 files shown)


pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/services/MergerWorkerServiceTest.scala
```diff
@@ -330,6 +330,34 @@ class MergerWorkerServiceTest
     }
   }
 
+  it(
+    "processes the message when out of date works, emitting nothing at the output") {
```
alexwlchan, line 334:
```suggestion
  it("doesn't send anything if the works are outdated") {
```

alexwlchan (approved):
Ah, that's what it was! Nice catch.

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/MergerWorkerService.scala
```diff
@@ -36,17 +36,24 @@ class MergerWorkerService[WorkDestination, ImageDestination](
...
         }
       }
-      lastUpdated = workSets.flatMap(_.flatten.map(_.state.modifiedTime)).max
-      _ <- Future.sequence {
-        workSets.map(applyMerge(_, lastUpdated))
+      nonEmptyWorkSets = workSets.filter(_.flatten.nonEmpty)
+      _ <- nonEmptyWorkSets match {
+        case Nil => Future.successful(Nil)
+        case _ =>
+          val lastUpdated = nonEmptyWorkSets
```
jamieparkinson, line 47: Perhaps put this in a separate method now it's embiggened?

jamieparkinson (approved):

PR #983, opened 2020-10-23
Fix paired filters and aggregations behaviour

jamieparkinson:
It turns out that this behaviour has always been broken: [this](https://github.com/wellcomecollection/catalogue/compare/fix-paired-filters-aggregations?expand=1#diff-0096661338e42f3cce180fd6202bfdf75f27fda20a443e2c963d4a282cf7a0a2L45) combined with [this](https://github.com/wellcomecollection/catalogue/compare/fix-paired-filters-aggregations?expand=1#diff-0096661338e42f3cce180fd6202bfdf75f27fda20a443e2c963d4a282cf7a0a2L52) has always meant that the subaggregation did nothing.

In practice this meant that (eg) if you had 2 aggregations - say `workType` and `language` - and a filter that paired with one of them - say a `workType` filter - then the unpaired aggregation (`language`) would _not_ have the filter applied. Fixing this introduced an inconsistency where empty buckets for `language` were displayed when a paired filter/aggregation was present but were not displayed otherwise, so there's a fix for that here too.

### DIFF
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/FiltersAndAggregationsBuilderTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/services/FiltersAndAggregationsBuilderTest.scala
@@ -91,6 +91,28 @@ class FiltersAndAggregationsBuilderTest extends AnyFunSpec with Matchers {
         .subaggs should have length 0
     }
 
+    it("applies paired filters to non-paired aggregations") {
+      val formatFilter = FormatFilter(Seq("bananas"))
+      val sut = new FiltersAndAggregationsBuilder(
+        List(AggregationRequest.Format, AggregationRequest.Language),
+        List(formatFilter),
+        requestToAggregation,
+        filterToQuery
+      )
+
+      sut.filteredAggregations should have length 2
+      val formatAgg =
+        sut.filteredAggregations.head.asInstanceOf[MockAggregation]
+      val languageAgg =
+        sut.filteredAggregations(1).asInstanceOf[MockAggregation]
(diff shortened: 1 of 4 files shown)


api/api/src/test/scala/uk/ac/wellcome/platform/api/services/FiltersAndAggregationsBuilderTest.scala
```diff
@@ -91,6 +91,28 @@ class FiltersAndAggregationsBuilderTest extends AnyFunSpec with Matchers {
         .subaggs should have length 0
     }
 
+    it("applies paired filters to non-paired aggregations") {
+      val formatFilter = FormatFilter(Seq("bananas"))
+      val sut = new FiltersAndAggregationsBuilder(
```
jamesgorrie, line 96: `sut`?

jamesgorrie (approved):
Tests make it nice and clear. Good spot/fix

api/api/src/test/scala/uk/ac/wellcome/platform/api/services/FiltersAndAggregationsBuilderTest.scala
jamieparkinson: It's what I used in the other tests in Jan so just made it match, it's a rather old fashioned term for [System Under Test](https://en.wikipedia.org/wiki/System_under_test)

jamieparkinson: I think I got it from here maybe https://docs.docker.com/docker-hub/builds/automated-testing/

PR #984, opened 2020-10-23
Tests for relation embedder worker

warrd:
https://github.com/wellcomecollection/platform/issues/4829

These were previously missing

### DIFF
--- a/pipeline/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationEmbedderWorkerService.scala
+++ b/pipeline/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationEmbedderWorkerService.scala
@@ -51,6 +51,9 @@ class RelationEmbedderWorkerService[MsgDestination](
       .mapAsync(2)(workIndexer.index)
       .collect { case Left(failedWorks) => failedWorks.toList }
       .mapConcat(identity)
+      .mapAsync(2) { work =>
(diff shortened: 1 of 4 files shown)


pipeline/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationEmbedderWorkerService.scala
```diff
@@ -48,16 +48,19 @@ class RelationEmbedderWorkerService[MsgDestination](
...
       .mapConcat(identity)
-      .toMat(Sink.seq)(Keep.right)
-      .run()
-      .flatMap {
-        case Nil => Future.successful(())
-        case failedWorks =>
-          Future.failed(
-            new Exception(s"Failed indexing works: $failedWorks")
-          )
+      .mapAsync(2) { work =>
```
jamieparkinson, line 61: Does this need `mapAsync() { _ => Future.fromTry(...) }`? `Future.fromTry` returns an already completed future (ie nothing here is async) so I think you should be able to throw the `Failure(...)` directly without redundantly wrapping it

jamieparkinson (approved):

pipeline/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationEmbedderWorkerService.scala
warrd: Good point :+1:

PR #987, opened 2020-10-23
use cross_fields over best_fields for works query

### LINKED ISSUE #981: Contributor / title search not surfacing right result
When searching for "bulloch history of bacteriology" we would expect this result to show up:
https://wellcomecollection.org/works/rkcux48q

jamesgorrie:
Fixes #981 

A few other ideas were discussed, as listed below:

## Use `minimum_should_match`
This _might_ fix the immediate problem, but doesn't really match what we're trying to do. That would make the query match `title` on `bulloch history of bacteriology` by essentially removing `bulloch` from the query. It wouldn't match on the `contributors` field, whereas we want it to match on both.

## Create a new combo-field
We could create a new field and `copy_to` and search on that. This feels useful, but using the `cross_fields` feels like it essentially does the same thing.

## Create a new part of the query which is `cross_fields` only on `title` and `contributors`
Again, could be useful but we went down the over-specific querying before, which we have all but removed.

I am trying to find a pre-stage way of sharing the results from the rank_eval as I have just tested them locally, but haven't quite got a solution yet bar a screen grab:

## before
Screenshot 2020-10-23 at 16 20 15

## after
Screenshot 2020-10-23 at 16 20 56

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/WorksQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/WorksQuery.scala
@@ -3,7 +3,10 @@ package uk.ac.wellcome.platform.api.elasticsearch
 import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.requests.common.Operator.{AND, OR}
 import com.sksamuel.elastic4s.requests.searches.queries.BoolQuery
-import com.sksamuel.elastic4s.requests.searches.queries.matches.MultiMatchQueryBuilderType.BEST_FIELDS
+import com.sksamuel.elastic4s.requests.searches.queries.matches.MultiMatchQueryBuilderType.{
+  BEST_FIELDS,
+  CROSS_FIELDS
+}
 import com.sksamuel.elastic4s.requests.searches.queries.matches.{
   FieldWithOptionalBoost,
   MultiMatchQuery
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):
Didn't know about cross_fields, interesting (and [great docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#type-cross-fields) from Elastic too).

Are there any kinds of queries which might be hurt by this / is there a tradeoff here and if so what is it?

jamesgorrie:
> Are there any kinds of queries which might be hurt by this / is there a tradeoff here and if so what is it?

The main difference at the moment is that scores will be accumulated, whereas with `best_fields` would use the score of the best matched fields. This is offset with the boosting, so the most important fields are scored higher.

Given we use boosting on important fields, I can't see how this would affect things negatively, but this could create a situation where if I search for `The big brown-green cat`, which is a book I know the title of, but we have a book called `The big green cat` by `Susan Brown`, the latter will be returned first as we boost `contributors` over `title`.

However this should be [`prefix_query`](https://github.com/wellcomecollection/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/WorksQuery.scala#L34) we use. However if I searched for `big brown-green cat` that wouldn't match on that query, so _might_ be an issue.

Another consideration if that we search across the same field analysed in multiple ways, i.e. `title`, `title.english`, `title.shingles`, and again with the accumulative scoring, we would have the sum of these scores added together.

We could optimise against this by having those three fields in their own `multi_match` query with `best_fields`, and add that to the new `cross_fields` - but that might have knock on affects too. I would prefer to keep evolving the `rank_eval` test and  write the query against that.

ping @harrisonpim

harrisonpim:
loving the quality of the documentation/discussion on this - feels like we have a really good grasp of the trade-offs!

PR #992, opened 2020-10-26
Single variable for index config

jamieparkinson:
Thought that, since we use this to match index consumers to pipelines, it should be configured in the same way as pipelines - ie via a single variable for the date https://github.com/wellcomecollection/catalogue/blob/938ddb72dd54d573ab7e8bb30fe7b7ad1db96540/pipeline/terraform/main.tf#L4


jamesgorrie (approved):
🚤

PR #993, opened 2020-10-26
Bump scala-libs to 24.1.0

alexwlchan:
This brings a newer version of Akka, which we need to update our version of elastic4s later.


jamesgorrie (approved):
Do we need to leave a RELEASE file?

alexwlchan:
The release file goes on the library, not the services.

PR #994, opened 2020-10-26
Only store json containing necessary fields for relations in ES

warrd:
Relations by the `relation_embedder` are stored in ES  as (unindexed) json. We currently serialise all possible fields for the work data, even though in the case of relations only a minimal number are used. Here we remove all null fields from the serialised data.

Example of a currently stored relation:
```json
{
  "data": {
    "title": "Frederick Belding Power Papers",
    "otherIdentifiers": [],
    "mergeCandidates": [],
    "alternativeTitles": [],
    "format": null,
    "description": null,
    "physicalDescription": null,
    "lettering": null,
    "createdDate": null,
    "subjects": [],
    "genres": [],
    "contributors": [],
    "thumbnail": null,
    "production": [],
    "language": null,
    "edition": null,
    "notes": [],
    "duration": null,
    "items": [],
    "collectionPath": {
      "path": "WF/E/07",
      "level": {
        "type": "Series"
      },
      "label": "WF/E/07"
    },
    "images": [],
    "workType": "Series"
  },
  "id": {
    "sourceIdentifier": {
      "identifierType": {
        "id": "calm-record-id",
        "label": "Calm RecordIdentifier"
      },
      "ontologyType": "SourceIdentifier",
      "value": "cbae884c-142f-4e42-b1a3-5945d1983692"
    },
    "otherIdentifiers": [],
    "identifiedType": "Identified"
  },
  "depth": 1
},
```

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
@@ -5,8 +5,8 @@ import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.requests.bulk.{BulkResponse, BulkResponseItem}
 import com.sksamuel.elastic4s.requests.common.VersionType.ExternalGte
 import com.sksamuel.elastic4s.{ElasticClient, Index, Response}
-import com.sksamuel.elastic4s.circe._
-import io.circe.Encoder
+import com.sksamuel.elastic4s.{Indexable => ElasticIndexable}
+import io.circe.{Encoder, Printer}
 import grizzled.slf4j.Logging
 
 import uk.ac.wellcome.elasticsearch.{ElasticsearchIndexCreator, IndexConfig}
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
Very nice. Elastic treats a few things as empty
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-exists-query.html

alexwlchan:
Why aren't we filling out all the work data in the relation embedder?

Could we fork the WorkData[…] model and use it where these objects would go, so it's explicit that you'd never expect these fields to be non-null?

I'm not opposed to fiddling with Circe, but it's always a bit fragile and relies on Circe not changing. If we can accomplish the same end without changing how we use Circe, I'd prefer that.

warrd:
We could potentially have multiple `WorkData` models, which would def be an improvement in some ways (like what you suggest about being sure those fields exist), but potentially bad in others (maintenance of both, bugs introduced by accidental drift if not implemented well etc). So I am not against that idea, but it's not really what I am working currently, and would probably take some thought to implement correctly.

warrd:
> Why aren't we filling out all the work data in the relation embedder?

As we do not need all the data denormalised, there is only a limited set of stuff the front end actually needs from related works.

alexwlchan:
👍

Makes sense. Maybe put a ticket on the backlog to come back and think about multi-WorkData more carefully at some point?

PR #1001, opened 2020-10-28
Bump up ingestor memory and reduce task count

warrd:


jamieparkinson (approved):

pipeline/terraform/main.tf
```diff
@@ -77,23 +77,23 @@ module "catalogue_pipeline_20201027" {
...
 
   mets_adapter_topic_arns = [
-    local.mets_reindexer_topic_arn,
+    # local.mets_reindexer_topic_arn,
     local.mets_adapter_topic_arn,
   ]
 
   calm_adapter_topic_arns = [
-    local.calm_reindexer_topic_arn,
+    # local.calm_reindexer_topic_arn,
```
jamesgorrie, line 96: Didn't go you mean to do  this? Aren't we still reindexing?

jamesgorrie (approved):
If you meant to unsubscribe from reindexing

pipeline/terraform/main.tf
warrd: The messages are already on the queue, so was just doing this in case someone pulls up another pipeline. I haven't actually applied this part yet though.

PR #1002, opened 2020-10-28
Redirect merged miro

jamieparkinson:
Contrary to what I wrote [here](https://github.com/wellcomecollection/catalogue/pull/632#issue-432548486), we **do** want to redirect/suppress Miro works that can be linked to Sierra works, regardless of whether we can merge in identifiers/items/locations etc. This PR fixes that behaviour [here](https://github.com/wellcomecollection/catalogue/compare/redirect-merged-miro?expand=1#diff-24547383116feb40679e907622ca0adfa5b08487775933fbcb5e073994780735R158).

In addition, given our ongoing struggle to reason about what we want from merging (regardless of the code we write), I've added a feature test in the given-when-then style which aims to act as documentation of our intentions and the actual behaviour, rather than a fully comprehensive test which I think is already covered. This aims to be as readable as possible and gives output that looks like:

> Given a Sierra work and 3 Miro works
When the works are merged
Then the Miro works are redirected to the Sierra work
And images are created from the Miro works
And the merged Sierra work's images contain all of the images


kenoir (approved):
This is a brilliant idea.

PR #1004, opened 2020-10-29
Upgrade sbt to 1.4.1

jamieparkinson:
It looks like _maybe_ some of the unnecessary recompiles could have been caused by this issue https://github.com/sbt/sbt/issues/4168 so this PR upgrades SBT to 1.4.1, which resolves that issue. This also requires using Coursier for dependency resolution rather than Ivy

It also enables [cached resolution](https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html) which has improved build times for me locally, at least.

### DIFF
--- a/docker_run.py
+++ b/docker_run.py
@@ -137,6 +137,22 @@ if __name__ == "__main__":
         cmd += ["--volume", "%s/.sbt:/root/.sbt" % os.environ["HOME"]]
         cmd += ["--volume", "%s/.ivy2:/root/.ivy2" % os.environ["HOME"]]
 
+        # Coursier cache location is platform-dependent
+        # https://get-coursier.io/docs/cache.html#default-location
+        linux_coursier_cache = ".cache/coursier/v1"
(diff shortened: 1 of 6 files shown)


docker_run.py
```diff
@@ -137,6 +137,22 @@ def parse_args():
         cmd += ["--volume", "%s/.sbt:/root/.sbt" % os.environ["HOME"]]
         cmd += ["--volume", "%s/.ivy2:/root/.ivy2" % os.environ["HOME"]]
 
+        # Coursier cache location is platform-dependent
+        # https://get-coursier.io/docs/cache.html#default-location
+        linux_coursier_cache = ".cache/coursier/v1"
+        if sys.platform == "darwin":
```
kenoir, line 143: Yuck.

kenoir (approved):

docker_run.py
jamieparkinson: I know!

PR #1005, opened 2020-10-30
Add ISSN and digcode identifiers

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4867 and https://github.com/wellcomecollection/platform/issues/4865

This patch updates the Sierra transformer to get the ISSN from MARC 022 and the Wellcome digcode from MARC 759. I copied the comments about digcodes directly from https://github.com/wellcomecollection/platform/issues/4865.

PR #1006, opened 2020-10-30
internal_model: Be explicit about the ontologyType of a SourceIdentifier

alexwlchan:
Spotted while writing https://github.com/wellcomecollection/catalogue/pull/1005

The ontologyType of a SourceIdentifier is meaningful – it's used to distinguish identifiers in the ID minter. e.g. `sierra/1234567` will be minted two different canonical IDs if it's a Work or an Item.

Forcing callers to be explicit about what ontologyType they want will ensure we get meaningful identifiers – as this shows, we have an issue where the Calm transformer *hasn't* set this field previously, so the identifiers will have the wrong ontologyType. e.g. if a Calm-transformed record links to a Sierra bib, the canonical ID won't match the ID on the Sierra work.

Fixing that is beyond the scope of this patch, but at least prevents it getting any worse.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
@@ -3,7 +3,7 @@ package uk.ac.wellcome.models.work.internal
 /** An identifier received from one of the original sources */
 case class SourceIdentifier(
   identifierType: IdentifierType,
-  ontologyType: String = "SourceIdentifier",
+  ontologyType: String,
   value: String
 ) {
   override def toString = s"${identifierType.id}/$value"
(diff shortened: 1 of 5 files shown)


jamesgorrie (approved):

kenoir:
@alexwlchan so my understanding of this is there is an issue with a swathe of calm sourced things not being matchable?

alexwlchan:
> so my understanding of this is there is an issue with a swathe of calm sourced things not being matchable?

So it depends on what you mean by "matchable".

The pipeline matcher will probably be fine, because it uses the mergeCandidates field which we don't surface publicly. We can change that as much as we like.

It's more matching stuff in terms of our canonical IDs. e.g. you could do

```
GET /calm-work-1
{
  "identifiers: [
    {
      "type": "sierra-system-number",
      "sourceIdentifier": "b12345678",
      "canonicalId": "axiuqw12"
    }
    ...
}
```

```
GET /sierra-work-b12345678
{
  "identifiers: [
    {
      "type": "sierra-system-number",
      "sourceIdentifier": "b12345678",
      "canonicalId": "dfkhfduh"
    }
    ...
}
```

So the same record from Sierra somehow has two, different canonical IDs from the pipeline,

PR #1007, opened 2020-10-30
Try to make the relation embedder more efficient

alexwlchan:
This was my first thought when looking at the relation_embedder code, and somewhat speculative.

Why do we retrieve works one at a time? Would it be more efficient to fetch them in batches? We've already seen an improvement from indexing in bulk, so let's do it for retrieval too.

My naive approach would be to load all the works into memory in one go, but I don't think that works – thinking about it some more, I'm guessing some relation graphs are way too big to load into memory? No doubt @warrd can shed more light on this.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
@@ -39,4 +39,8 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
           }
       }
   }
+
+  override def lookupMultipleIds(ids: Seq[String]): Future[Map[String, T]] =
+    Future.sequence(ids.map { lookupSingleId })
+      .map { documents => ids.zip(documents).toMap }
 }
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
@@ -21,7 +21,6 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
     with Logging {
 
   final def lookupSingleId(id: String): Future[T] = {
-    debug(s"Index $index: looking up ID $id")
     client
       .execute {
         get(index, id)
@@ -41,8 +40,6 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
   }
 
   override def lookupMultipleIds(ids: Seq[String]): Future[Map[String, T]] = {
-    debug(s"Index $index: looking up multiple IDs $ids")
-
     client
       .execute {
         multiget(ids.map { get(index, _) })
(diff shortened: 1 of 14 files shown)


warrd:
The largest trees consist of about 1000 works I seem to remember, so we probably shouldn't be fetching them all into memory at once.

I was going to suggest batching them, it looks like you have already done this (into batches of 25)? Also worth remembering that each of these streams is run within another stream (the `SQSStream`), which already has a concurrency of 10. So this will result in 250 works in memory at a time I think..

alexwlchan:
> The largest trees consist of about 1000 works I seem to remember, so we probably shouldn't be fetching them all into memory at once.

Okay, good to know I was on the right track.

What happens if we load them all into memory? Have we tried it? I ran a quick script that gets the size (as JSON) of the biggest works in the catalogue snapshot:

```
[(324787, 1),
 (146644, 1),
 (133423, 1),
 (114830, 1),
 (107223, 1),
 (98219, 1),
 (98033, 1),
 (97964, 1),
 (97040, 1),
 (86168, 1)]
```

If every work was as big as that biggest work, and we had 1000 of them, we'd need ~325MB of memory. Triple that – we have the reduced version in RelationsService, the merged versions, and the denormalised versions. Call that 1GB.

That doesn't feel prohibitive, especially as:

* Work sizes drop off pretty dramatically after that big one
* Not every tree has 1000 works
* Scala's internal data structures may well achieve better efficiency than JSON

warrd:
Another way to make the relation embedder more efficient is indexing order: at the moment reindexing is done in a completely random order, which leads to lots of unnecessary work.

Given the following tree:

```
A (root of archive)
|
|-------
|      |
B      C
|
|
D
```

If we index level 0 (`A`) first, then level 1 (`B`, `C`), then level 2  (`D`), that will result in the least amount of work, as the relation embedder needs to reindex the input work and all its subtree.

We could achieve efficient calm reindexing relatively simply by adding an additional `reindex_order` column to the calm-adapter VHS table, which is set to the depth of the tree. However the whole thing is complicated by the fact that sierra works which are matched with those calm works will be received in random orders: we could always run the calm reindex post sierra reindex to alievaite this, but then the issue still exists during the overnight sierra reharvest.

warrd:
:+1: Sounds good

warrd:
One other potential method for can improving the efficiency is due to the fact that related works only have a limited number of fields that are stored: if when receiving an update we somehow checked whether any of of those fields have actually changed, deciding to swallow the message and do nothing if none of them have (with it presumably being one of the other fields, or maybe nothing at all changed), then we could avoid all the subsequent work.

Not sure whether this is something we want to do either, but something else to bear in mind.

warrd:
> If every work was as big as that biggest work, and we had 1000 of them, we'd need ~325MB of memory

Just occurred to me that the `relation_embedder` only returns a few core fields when querying from elasticsearch, so we will result in much smaller memory usage than these calculations:

https://github.com/wellcomecollection/catalogue/blob/2b74be1c52d68414a065dacfe779f0b66b4c99f8/pipeline/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationsRequestBuilder.scala#L17-L30

warrd:
:point_up: ignore the above that is wrong: we still fetch the full original works, it is the related works that are added into the state of these works which only have a few keys fields.

warrd (approved):

PR #1009, opened 2020-11-02
Allow calm_adapter to make other Calm queries beyond Modified field 

warrd:
There is currently an issue with the Calm adopter in that it just queries the `Modified` field. This can sometimes be blank, so we should also query the `Created` field. Here we also include functionality to query the `RefNo`, and data when there is neither a `Created` or `Modified` field (currently 450 records).

A further PR will update the Calm window generator lambda / cli with the functionality to provide these other messages to the adapter.

### DIFF
--- a/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmQuery.scala
+++ b/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmQuery.scala
@@ -9,7 +9,9 @@ sealed trait CalmQuery {
   def logicalOperator = "OR"
   def relationalOperator = "="
   def queryExpression =
-    keys.map(key => s"($key${relationalOperator}$value)").mkString(logicalOperator)
+    keys
+      .map(key => s"($key${relationalOperator}$value)")
+      .mkString(logicalOperator)
 }
 
 object CalmQuery {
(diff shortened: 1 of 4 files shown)


calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmQuery.scala
```diff
@@ -4,15 +4,40 @@ import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 
 sealed trait CalmQuery {
-  def key: String
+  def keys: List[String]
   def value: String
-  def queryExpression = s"$key=$value"
+  def logicalOperator = "OR"
+  def relationalOperator = "="
+  def queryExpression =
```
jamesgorrie, line 11: Is this clearer than just having a list of `queryExpressions` on the `CalmQuery` being something like `List(s"Created=${formatDate(date)}")` or similar?

This also means you can only share the `logicalOperator` and value across fields. 

Saying this, ☝️ is exactly what we we're doing.

jamesgorrie (approved):

calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmQuery.scala
warrd: I think this is fine, the options are either having the abstraction like we have now or just every case class defining the `queryExpression` directly. I think either methods work but want to get this up and running so going to merge as is.

PR #1010, opened 2020-11-03
Update calm_window_generator lambda / cli for other query types

warrd:
The other half of #1009. The lambda is already deployed and working.

### DIFF
--- a/calm_adapter/calm_window_generator/src/window_generator.py
+++ b/calm_adapter/calm_window_generator/src/window_generator.py
@@ -3,23 +3,48 @@ import json
 
 
 class WindowGenerator:
-    def __init__(self, sns_client, topic_arn, start, end=None):
+    def __init__(self, sns_client, topic_arn, queries):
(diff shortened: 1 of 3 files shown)


calm_adapter/calm_window_generator/src/window_generator.py
```diff
@@ -3,23 +3,48 @@
...
+        return CalmQuery("EmptyCreatedAndModifiedDate")
+
+    @staticmethod
+    def ref_no(ref_no):
+        return CalmQuery("RefNo", refNo=ref_no)
+
+
+def created_or_modified_date_range(start, end=None):
+    end = end or start
+    assert start <= end, "Start window is after end window"
```
jamieparkinson, line 47: `<`?

warrd: Having them the same is valid, it means the range will just be a single day.

jamieparkinson (approved):

PR #1012, opened 2020-11-04
[WIP] Use p tags for Sierra description

warrd:


warrd:
https://github.com/wellcomecollection/catalogue/pull/1124#pullrequestreview-540514183 here instead

PR #1017, opened 2020-11-06
Use scroll API and return works directly in the getAffectedWorks query

alicefuzier:


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
```diff
@@ -20,7 +20,8 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
     extends Retriever[T]
     with Logging {
 
-  final def lookupSingleId(id: String): Future[T] = {
+  final def apply(id: String): Future[T] = {
+    debug(s"Looking up ID $id in index $index")
```
warrd, line 24: `debug` messages don't get logged to ES and just make the tests more noisy so maybe worth removing?

pipeline/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationsService.scala
```diff
@@ -37,41 +37,33 @@ trait RelationsService {
...
-                }
-              works match {
-                case Right(works) =>
-                  info(
-                    s"work ${work.id} ($path) has ${works.size} affected works")
-                  Future.successful(works.map(_.state.sourceIdentifier))
-                case Left(err) => Future.failed(err)
+            Source
+              .fromPublisher(
+                elasticClient.publisher(
```
warrd, line 58: Nice :+1:

warrd (approved):

pipeline/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationsService.scala
alicefuzier: stolen from the snapshot generator

PR #1018, opened 2020-11-06
Cache all relations for an archive up front of denormalising

warrd:


alicefuzier (approved):
Nice work!

pipeline/terraform/stack/service_work_relation_embedder.tf
```diff
@@ -44,7 +44,7 @@ module "relation_embedder" {
   }
 
   subnets             = var.subnets
-  max_capacity        = 10
+  max_capacity        = 0
```
alicefuzier, line 47: It could be good to also reconnect merger and id minter in this PR?

PR #1021, opened 2020-11-09
Parse as many calm dates as possible

jamieparkinson:
This builds on @warrd's work in https://github.com/wellcomecollection/catalogue/pull/70 and allows the calm transformer to parse dates as specified [here](http://www.dswebhosting.info/documents/Manuals/ALM/V10/MANUAL/main_menu/basics/period_field_format.htm).

It also tries to deal with common deviations from that specification as encountered in the real data (from the calm index in the reporting cluster). Over that data, it is able to parse **99.3%** of non-empty date fields.

I tried really hard to get around the enumeration of the permutations of date ranges, unqualified/qualified dates etc etc, but while I suspect it's possible somehow I think the code is probably simpler (if longer) when listing all known variations as I've done here. If you have questions around the fastparse syntax etc you may wish to refer to the discussion in https://github.com/wellcomecollection/catalogue/pull/70.

A lot (>50%) of the changes here are either boilerplate date range conversions in `ToInstantRange.scala` or the huge table of test cases I reproduced from [here](http://www.dswebhosting.info/documents/Manuals/ALM/V10/MANUAL/main_menu/basics/period_field_format.htm) - so don't be scared off by the line count!


common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FuzzyDate.scala
```diff
@@ -25,84 +24,13 @@ case class CenturyAndDecade(century: Int, decade: Int) extends FuzzyDate
...
-    convert[FuzzyDateRange[Day, CalendarDate]](
-      value =>
-        InstantRange(
-          localDate(value.from.day, value.to.month, value.to.year),
-          localDate(value.to),
-          value.label))
+object FuzzyDateRange {
+  def combine[
+    FF <: FuzzyDate,
+    TT <: FuzzyDate
```
warrd, line 30: Can we get nicer / more descriptive type parameter names here?

jamieparkinson: Actually since I'm using an existential type for the unused params (rather than `FT`, `TF` as I think I'd originally intended) would `F` and `T` to match existing/other code suffice?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/periods/Lex.scala
```diff
@@ -0,0 +1,65 @@
...
+    P(int ~ "'".? ~ "s" filter (_ % 10 == 0))
+
+  def season[_: P]: P[String] =
+    P(StringIn("spring", "summer", "autumn", "fall", "winter").! map {
+      case "fall" => "autumn"
+      case season => season
+    })
+
+  def lawTerm[_: P]: P[String] =
+    P(StringIn("michaelmas", "hilary", "easter", "trinity").!)
```
kenoir, line 37: Just popping to the buttery to stick some gin on the battles.

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/periods/PeriodParser.scala
```diff
@@ -0,0 +1,199 @@
...
+  def yearDivision[_: P]: P[FuzzyDateRange[MonthAndYear, MonthAndYear]] =
+    seasonYear | lawTermYear
+
+  def seasonYear[_: P]: P[FuzzyDateRange[MonthAndYear, MonthAndYear]] =
+    P((Lex.season map {
+      case "spring" => (3, 5)
+      case "summer" => (6, 8)
+      case "autumn" => (9, 11)
+      // Winter YEAR refers to the year in which the winter starts
+      // https://www.metoffice.gov.uk/weather/learn-about/weather/seasons/winter/when-does-winter-start
```
kenoir, line 175: nice comment

kenoir (approved):

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/periods/PeriodParser.scala
jamieparkinson: Realised that while I thought this was the case had absolutely no evidence for it and briefly convinced myself that since there are more winter months in the following year it was the other way round.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FuzzyDate.scala
warrd: I think descriptive type parameters make code much more readable: having a `T` is fine when it's really clear from the context (e.g. `Vec[T]`), but when it's not, and especially when there is more than one, imo descriptive is much more preferable. I still have no idea what `T` and `F` mean.

I have just looked at the code though and see that it was me who originally wrote `T` and `F`, so take the blame for that :see_no_evil:

jamieparkinson: `T`o and `F`rom - they are used so extensively (and consistently) that I think I'll keep them as is. If you're in that area of the codebase you quickly get used to them.

warrd: Yep I guessed that meaning, OK sounds good :+1:

PR #1022, opened 2020-11-10
Create images from digitised ephemera

jamieparkinson:
Does what it says on the tin


jamesgorrie (commented):
What happens with these rules if there is no associated work?

jamieparkinson:
Not sure that I understand the question - what do you mean by an associated work?

This change creates searchable images when we have a linked METS file and Sierra ephemera or picture, whereas previously it would only do this for pictures. If METS and Sierra are linked but the Sierra is not one of these material types we merge items, identifiers etc but do not create images from the METS work.

warrd (approved):

PR #1023, opened 2020-11-11
Add a "languages" field to the internal and ES model

alexwlchan:
A first step towards https://github.com/wellcomecollection/platform/issues/4864

I’ll add the other pieces (populating the field in the transformer, reading the field in the API) in separate patches.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
@@ -90,7 +90,7 @@ case class WorkData[State <: DataState](
   thumbnail: Option[LocationDeprecated] = None,
   production: List[ProductionEvent[State#MaybeId]] = Nil,
   language: Option[Language] = None,
-  languages: List[Language] = List.empty,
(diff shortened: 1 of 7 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -90,6 +90,7 @@ case class WorkData[State <: DataState](
   thumbnail: Option[LocationDeprecated] = None,
   production: List[ProductionEvent[State#MaybeId]] = Nil,
   language: Option[Language] = None,
+  languages: List[Language] = List.empty,
```
jamieparkinson, line 93: All the other empty lists here are `Nil` which would be my preference - either way it should be consistent

jamieparkinson (approved):

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
alexwlchan: I'd missed that, will fix.

PR #1029, opened 2020-11-11
Expose filters/aggregations/includes for languages in the API

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4864

I struggled to get the aggregations working. This is the JSON that comes back from a languages aggregation as currently implemented:

<details><summary>Elasticsearch response</summary>

```json
{
  "doc_count_error_upper_bound": 0,
  "sum_other_doc_count": 0,
  "buckets": [
    {
      "key": "en",
      "doc_count": 3,
      "sample_doc": {
        "hits": {
          "total": {
            "value": 3,
            "relation": "eq"
          },
          "max_score": 0.0,
          "hits": [
            {
              "_score": 0.0,
              "_type": "_doc",
              "_source": {
                "data": {
                  "languages": [
                    {
                      "label": "English",
                      "id": "en"
                    }
                  ]
                }
              },
              "_id": "ylhsqawzwo",
              "_index": "index-tn9gqnctm"
            }
          ]
        }
      }
    },
    {
      "key": "sv",
      "doc_count": 2,
      "sample_doc": {
        "hits": {
          "total": {
            "value": 2,
            "relation": "eq"
          },
          "max_score": 0.0,
          "hits": [
            {
              "_score": 0.0,
              "_type": "_doc",
              "_source": {
                "data": {
                  "languages": [
                    {
                      "label": "English",
                      "id": "en"
                    },
                    {
                      "label": "Swedish",
                      "id": "sv"
                    },
                    {
                      "label": "Turkish",
                      "id": "tr"
                    }
                  ]
                }
              },
              "_id": "tnmjufcamy",
              "_index": "index-tn9gqnctm"
            }
          ]
        }
      }
    },
    {
      "key": "tr",
      "doc_count": 1,
      "sample_doc": {
        "hits": {
          "total": {
            "value": 1,
            "relation": "eq"
          },
          "max_score": 0.0,
          "hits": [
            {
              "_score": 0.0,
              "_type": "_doc",
              "_source": {
                "data": {
                  "languages": [
                    {
                      "label": "English",
                      "id": "en"
                    },
                    {
                      "label": "Swedish",
                      "id": "sv"
                    },
                    {
                      "label": "Turkish",
                      "id": "tr"
                    }
                  ]
                }
              },
              "_id": "tnmjufcamy",
              "_index": "index-tn9gqnctm"
            }
          ]
        }
      }
    }
  ]
}
```

</details>

A language is currently implemented as `case class Language(label: String, id: Option[String])`. I can't see a way to get the ID and the label out of an Aggregation result.

If we're going to lean more heavily on the ISO/LOC language codes, does it make sense to lean more heavily on the pre-defined language codes, and use those as the canonical identifier for a language through the pipeline?

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
@@ -108,7 +108,9 @@ object Aggregations extends Logging {
 
   implicit class EnhancedEsAggregations(aggregations: Elastic4sAggregations) {
 
-    def decodeAgg[T: Decoder](name: String, documentPath: Option[String] = None): Option[Aggregation[T]] = {
+    def decodeAgg[T: Decoder](
+      name: String,
+      documentPath: Option[String] = None): Option[Aggregation[T]] = {
       aggregations
         .getAgg(name)
         .flatMap(
@@ -166,7 +168,8 @@ object AggregationMapping extends Logging {
   private val bucketSampleDoc =
     root.sample_doc.hits.hits.index(0)._source.json
 
-  def aggregationParser[T: Decoder](json: String, path: Option[String]): Try[Aggregation[T]] =
+  def aggregationParser[T: Decoder](json: String,
+                                    path: Option[String]): Try[Aggregation[T]] =
     parse(json).toTry
       .map { parsedJson =>
         for {
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
@@ -10,7 +10,6 @@ import com.sksamuel.elastic4s.requests.searches.aggs.responses.{
 import com.sksamuel.elastic4s.requests.searches.SearchResponse
 import grizzled.slf4j.Logging
 import uk.ac.wellcome.display.models.LocationTypeQuery
-import uk.ac.wellcome.json.JsonUtil._
 import uk.ac.wellcome.models.work.internal._
 
 case class Aggregations(
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
@@ -10,6 +10,7 @@ import com.sksamuel.elastic4s.requests.searches.aggs.responses.{
 import com.sksamuel.elastic4s.requests.searches.SearchResponse
 import grizzled.slf4j.Logging
 import uk.ac.wellcome.display.models.LocationTypeQuery
+import uk.ac.wellcome.json.JsonUtil._
 import uk.ac.wellcome.models.work.internal._
 
 case class Aggregations(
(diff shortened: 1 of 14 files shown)


jamesgorrie:
@alexwlchan we've done this with `Format`s where [we use the `fromCode`](https://github.com/wellcomecollection/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala#L69-L75). This is if we have an `enum` with all know values to look up from.

You can also use the [top_hits method](https://github.com/wellcomecollection/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala#L129-L137), which I don't think currently supports working through the source document having a list that we then need to traverse to find the relevant value (matching the `key` of the aggregation on the `id` of the list item).

alexwlchan:
I've done some more digging, and learnt a couple of things:

* We have a lot of weird values in the language field, and we need to support passing a Language without an ID. Both for data quality purposes, and because as James says, being able to search by language name is important.
* Unidentified languages are excluded from aggregations right now, so excluding them in the new world is no worse than the status quo.

I've pushed a change that makes this more explicit, and I'll gradually ramp up the strictness around this stuff in future patches.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
```diff
@@ -92,15 +104,17 @@ object Aggregations extends Logging {
...
+      implicit decoder: Decoder[T]): Option[Aggregation[T]] = {
       aggregations
         .getAgg(name)
         .flatMap(
-          _.safeTo[Aggregation[T]]((json: String) =>
-            AggregationMapping
-              .aggregationParser[T](json, documentPath)).recoverWith {
+          _.safeTo[Aggregation[T]](
+            (json: String) => {
+              println(s"@@AWLC json = $json")
```
jamieparkinson, line 114: left in by accident?

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
```diff
@@ -34,8 +35,12 @@ object Aggregations extends Logging {
...
           productionDates = e4sAggregations
             .decodeAgg[Period[IdState.Minted]]("productionDates"),
-          language = e4sAggregations
-            .decodeAgg[Language]("language", Some("data.language")),
+          language = e4sAggregations.decodeAgg[Language](
+            "language",
+            documentPath = Some("data.language")
+          ),
+          languages = e4sAggregations.decodeAgg[Language]("languages")(
+            decodeLanguageList),
```
jamieparkinson, line 43: Is this a temporary (until the old `language` is removed) explicit override of the implicit decoder here? If so, would it make more sense to do this the other way round ie to override the decoder for `language` pending its removal / unambiguous implicit?

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
alexwlchan: Yup, will remove.

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
alexwlchan: Yes, although looking at it some more I don't think it's as necessary as I thought it was.

I hadn't looked at how we populate languages previously – turns out we're pretty strict about it, and always use the code, so maybe I can tidy this up.

alexwlchan:
@jamieparkinson Thanks for reviewing; I've tidied those up.

jamieparkinson (approved):

alexwlchan:
Adding "DO NOT MERGE" – although we use the codes in `language-codes.csv` for the Calm data, we don't do that for the Sierra data. If we deploy this as-is, language aggregations will break in the API.

PR #1030, opened 2020-11-11
Make the API tests go faster ⚡️

alexwlchan:
The API tests are noticeably slower than other services – often almost twice as long as anything else. This costs money in CI instances, and frustrates developers who want a fast edit-test-debug cycle. I did some investigation into why they're so slow:

* The `SwaggerDocs` class is *incredibly* expensive – as in, multiple seconds to create a new instance. We were creating a new instance every time we created a new `Router` instance, which happens in any test which runs against the public API.

   This is the big win.

* We create a whole stack of indexes we don't need – works indexes in an images API test, images indexes in a works API test, both indexes in tests that don't touch Elasticsearch at all.

   This is a smaller win – each new index costs ~0.3s to create (on my machine), but multiplied over several hundred tests that adds up.

* When we index documents before a tests start, we check they're all indexed by counting the number of documents returned. There's a dedicated API for counting the size of an index, which involves less JSON encoding/decoding – use that instead.

    I don't know how much of a difference this makes, but it surely can't make things worse.

Plus some other minor refactorings and clean ups.


alexwlchan:
18m ~> 13m.

warrd (approved):

jamesgorrie (commented):
Really nice!

PR #1031, opened 2020-11-11
Do not subscribe the relation embedder queue to any topics

jamieparkinson:


alicefuzier (approved):

warrd:
Not sure if you have encountered this but you might have to add `"merged_read"` to here and re-run this script: https://github.com/wellcomecollection/catalogue/blob/master/pipeline/create_pipeline_storage_es_credentials.py#L29

warrd:
There is a chance I left it with permission for both, so might be unnecessary

jamieparkinson:
I think it would already have broken if it was going to (?) as the relation embedder isn't doing anything, this PR is just stopping that queue filling up. Anyway good to know about that script!

warrd:
Yep if it's up and working then no need to do anything else.

PR #1033, opened 2020-11-11
Relation embedder reads paths

alicefuzier:


pipeline/relation_embedder/router/src/main/scala/uk/ac/wellcome/platform/router/RouterWorkerService.scala
```diff
@@ -25,9 +25,10 @@ class RouterWorkerService[MsgDestination](
 
   private def processMessage(message: NotificationMessage): Future[Unit] = {
     workRetriever.apply(message.body).flatMap { work =>
-      work.data.collectionPath.fold(ifEmpty = Future.fromTry(worksMsgSender.sendT(work.id))) { path =>
-        Future.fromTry(pathsMsgSender.sendT(path))
-      }
+      work.data.collectionPath
+        .fold(ifEmpty = Future.fromTry(worksMsgSender.sendT(work.id))) { path =>
+          Future.fromTry(pathsMsgSender.sendT(path))
```
warrd, line 30: These should be `send(work.id)` and `send(path)`: using `sendT` will mean it uses an `Encoder[String]` which wraps these in extra unnecessary quotes due to encoding the content of the string as json.

warrd (approved):

PR #1036, opened 2020-11-12
Add an object for looking up languages according to the MARC Language Codes List

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4864, we're going to be switching to using the three-letter MARC language codes in both CALM and Sierra.

This patch adds a copy of the list, and a Scala object that can look up languages by code or by name. Changing how the language parsing works will come in a separate patch – this can be spun out of that patch as a standalone piece.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/marc/MarcLanguageCodeList.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/marc/MarcLanguageCodeList.scala
@@ -34,9 +34,10 @@ object MarcLanguageCodeList extends Logging {
 
   // Create a lookup from name -> code
   //
-  // Note: languages.xml includes variant names.  We don't currently
-  // consider those in this lookup, but it may be worth adding as a
-  // future enhancement.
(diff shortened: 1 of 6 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/marc/MarcLanguageCodeList.scala
```diff
@@ -0,0 +1,98 @@
...
+// Utilities for parsing the MARC Code List for Languages to looking up
+// language labels.
+//
+// This parses the XML file downloaded from:
+// https://www.loc.gov/standards/codelists/languages.xml
+//
+object MarcLanguageCodeList extends Logging {
+
+  // Create a lookup from code -> name
+  private val codeLookup: Map[String, String] = {
```
warrd, line 16: Might be worth making these `lazy`? Otherwise ever project that uses internal model has to load that huge XML file.

alexwlchan: Done.

warrd (approved):

PR #1044, opened 2020-11-12
Include source workdata otherIdentifiers in image search

jamieparkinson:
Noticed this when testing digcodes - we weren't correctly searching the source work `otherIdentifiers` in the images query


jamieparkinson:
Re why there are 2 `otherIdentifiers`: https://wellcome.slack.com/archives/C3TQSF63C/p1605196018280200

Something that may be worth addressing/simplifying etc in future

warrd (approved):

warrd:
As the `otherIdentifiers` in the `IdState` are never initialised here, how come we need to include them?

warrd:
Here is where we create the `SourceWork` class, it looks like they are always blank: https://github.com/wellcomecollection/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/ImageSource.scala#L25-L56

jamieparkinson:
Yeah, we don't, I just thought I'd leave them there - but probably less confusing to remove them, will do that

PR #1046, opened 2020-11-12
Add a script for adding an expiry policy to recorder buckets

alexwlchan:
We have 20 unused recorder buckets. This is silly.

You can add these expiry policies manually, but who has the time? This script automates the process, and I've already run it against all the existing buckets.


jamieparkinson (approved):
✨ v good, expiry policy is a much better way than emptying the buckets

PR #1047, opened 2020-11-12
A Bold And Exciting New Era of Pipeline Creation

alexwlchan:
I was going to add a new pipeline for https://github.com/wellcomecollection/platform/issues/4864, and it felt like a process with a few rough edges. This patch tries to make the process of creating pipelines a bit smoother:

* I wrote a script to expire objects in the old recorder buckets, so the "terraform plan" output isn't so noisy. See https://github.com/wellcomecollection/catalogue/pull/1046

* Remembering to comment/uncomment individual reindexer topics is clunky and non-obvious. Now, the module gets a structured collection of topics that never changes, and has five new boolean flags to control the behaviour:

   - `enable_reindexing`
   - `enable_miro_reindexing`
   - `enable_sierra_reindexing`
   - `enable_mets_reindexing`
   - `enable_calm_reindexing`

   The module uses these variables to decide whether to include the reindexer topic. This is more explicit at the top level, so we can see what's going on.

   Setting `true` on an individual source overrides the global setting, e.g. you can enable Miro-only reindexing with

    ```hcl
    enable_reindexing      = false
    enable_miro_reindexing = true
    ```

* Copying large blocks of Terraform around to create/destroy stacks is clunky. I can't see two pipelines on my screen at once. Eww!

    I've switched us to using the support for [multiple instances of a module](https://www.terraform.io/docs/configuration/modules.html#multiple-instances-of-a-module) that was added in Terraform 0.13. This substantially reduces the amount of copy/paste we'll need to do when creating new pipelines, and gives you a much better "at a glance" view of the pipelines we have.

### DIFF
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -1,10 +1,26 @@
-module "catalogue_pipeline_20201111" {
+locals {
+  stacks = {
+    "20201111" = {
+      release_label     = "prod"
+      enable_reindexing = false
+    }
+  }
+}
+
+module "stack" {
   source = "./stack"
 
-  pipeline_date = "20201111"
-  release_label = "prod"
+  for_each = local.stacks
 
-  enable_reindexing = false
+  pipeline_date = each.key
+  release_label = each.value.release_label
+
+  enable_reindexing = each.value.enable_reindexing
+
+  enable_sierra_reindexing = lookup(each.value, "enable_sierra_reindexing", false)
+  enable_miro_reindexing   = lookup(each.value, "enable_miro_reindexing", false)
+  enable_calm_reindexing   = lookup(each.value, "enable_calm_reindexing", false)
+  enable_mets_reindexing   = lookup(each.value, "enable_mets_reindexing", false)
 
   account_id      = data.aws_caller_identity.current.account_id
   aws_region      = local.aws_region
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -4,6 +4,13 @@ locals {
       release_label     = "prod"
       enable_reindexing = false
     }
+
+    # A new pipeline with transformers that include support for multiple languages.
+    # See https://github.com/wellcomecollection/platform/issues/4864
+    "2020-11-12" = {
+      release_label     = "stage"
+      enable_reindexing = true
+    }
   }
 }
 
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -4,6 +4,8 @@ module "catalogue_pipeline_20201111" {
   pipeline_date = "20201111"
   release_label = "prod"
 
+  enable_reindexing = false
+
   account_id      = data.aws_caller_identity.current.account_id
   aws_region      = local.aws_region
   vpc_id          = local.vpc_id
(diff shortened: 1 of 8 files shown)


pipeline/terraform/main.tf
```diff
@@ -1,8 +1,33 @@
...
+locals {
+  stacks = {
+    "20201111" = {
+      release_label     = "prod"
+      enable_reindexing = false
+    }
+
+    # A new pipeline with transformers that include support for multiple languages.
+    # See https://github.com/wellcomecollection/platform/issues/4864
+    "2020-11-12" = {
```
alexwlchan, line 10: Minor: I find the eight-digit dates quite difficult to read, and spent a not insignificant amount of time not noticing the difference between `20201111` and `20200111`. This is an attempt to make these a little more readable.

jamieparkinson: I'm not 100% certain that we don't have any code that does a `index_name.split('-')` or equivalent - can't think of or find any examples but something to be aware of

alexwlchan: I ran this yesterday and it populated the index fine, so 🤞 it's an issue.

(That was before I realised I'd not put in the key change. 🤦‍♀️ )

pipeline/terraform/main.tf
```diff
@@ -1,8 +1,33 @@
...
-  pipeline_date = "20201111"
-  release_label = "prod"
+  for_each = local.stacks
+
+  pipeline_date = each.key
+  release_label = each.value.release_label
+
+  enable_reindexing = each.value.enable_reindexing
+
+  enable_sierra_reindexing = lookup(each.value, "enable_sierra_reindexing", false)
```
jamieparkinson, line 27: hmmm, I'm kind of on the fence about this - I know that at least @alicefuzier and I sometimes have the reindexer topics enabled and the update topics disabled (as eg it can be useful to test on a pipeline that's not going to have unexpected stuff coming through), and also it's a lot of code to be able to use some booleans... Perhaps we say that it's worth it for the majority of use cases but I'm not sure.

pipeline/terraform/main.tf
```diff
@@ -1,8 +1,33 @@
-module "catalogue_pipeline_20201111" {
+locals {
+  stacks = {
+    "20201111" = {
```
jamieparkinson, line 3: Is it possible/easy to `-target` a dynamic module?

alexwlchan: Seemed to be when I was doing it yesterday – you just get the resource by square brackets instead of name

e.g. `module.stack["20201111"]` instead of `module.stack.catalogue_20201111`

pipeline/terraform/main.tf
alexwlchan: [nod]

I haven't done much pipeline work, so maybe the old approach is fine – but the first time I spun up the pipeline I got the "comment/uncomment these topics" thing wrong.

A much less destructive approach I could try would be to move the topic definitions to the top of the file. Right now it's

```
module "stack" {
  // names you need to change
  // stuff that never changes
  // topics you have to change (sometimes)
  // more stuff that never changes
}
```

and so it was easy for me to miss the topics.

alicefuzier: 👍  yup definitely for me it's super useful to be able to disable updates and only have reindexes enabled. It's especially useful when I'm testing something that spans multiple services like merging and I can send specific records from the reidexer

alexwlchan: 👍

I'll defer to your pipeline experience. :)

alicefuzier: agree that it's diffult to rmember to uncomment/comment the topics. I also made that mistake multiple times

PR #1048, opened 2020-11-12
Store a fast lookup of MARC tag -> varFields in the Sierra transformer (aka go faster)

alexwlchan:
I was glancing at my reindex stats, and realised the Sierra transformer is a bit of a bottleneck. Partly because it has an order of magnitude more messages to chew through than anything else, but then I also realised that it's quite inefficient.

**The problem:** The Sierra bib data has a list of varfields, and different parts of the transformer loop through them to find varfields with a given MARC tag. This is a slow *O(N)* operation. By my count, we loop through all the varfields **62 times** to transform each Sierra record – and every transformation adds another iteration of this slow loop.

I don't know how many varfields a typical Sierra bib has, but anything more than a few and this loop is going to slow us down.

**The solution:** We store a map from (MARC tag) -> (matching varfields). This means that we run the loop once to build the map, and subsequent lookups of all the varfields with a given MARC tag is now an *O(1)* operation.

I don't have any profiling to back this up, but this seems like pretty low-hanging fruit that can only help.

As a bonus, if you aggregate over multiple MARC tags with SierraQueryOps, you now get them in the original order.

```
Bib: VF(3), VF(1), VF(2)

Before: varfieldsWithTags(1, 2, 3) => VF(1), VF(2), VF(3)
After:  varfieldsWithTags(1, 2, 3) => VF(3), VF(1), VF(2)
```

I don't know that it makes a huge difference, but it's nice to have and reduces the amount of transformation implementation that leaks into the pipeline.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAlternativeTitlesTest.scala
+++ b/pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAlternativeTitlesTest.scala
@@ -42,8 +42,8 @@ class SierraAlternativeTitlesTest
   it("should extract all alternative titles when multiple fields defined") {
     val varFields = List(field130, field240, field246)
     getAlternativeTitles(varFields) shouldBe List(
-      "Apples",
       "Bananas",
+      "Apples",
       "Cherries")
   }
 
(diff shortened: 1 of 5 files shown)


pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAlternativeTitlesTest.scala
```diff
@@ -42,8 +42,8 @@ class SierraAlternativeTitlesTest
   it("should extract all alternative titles when multiple fields defined") {
     val varFields = List(field130, field240, field246)
     getAlternativeTitles(varFields) shouldBe List(
-      "Apples",
       "Bananas",
+      "Apples",
```
alexwlchan, line 46: This is an example of the improved ordering logic – the titles in this test are

```
240 -> Apples
130 -> Bananas
246 -> Cherries
```

Now we retrieve them in MARC order, not transformer implementation order!

jamieparkinson (approved):
fwiw, I vaguely remember that I profiled the Sierra transformer when I did this work https://github.com/wellcomecollection/catalogue/pull/452 and didn't see anything in the flame chart that seemed to dominate so unless I'm remembering wrong I wouldn't expect to see especially big speedups here

PR #1050, opened 2020-11-13
A slightly more pragmatic approach to spinning up new pipelines

alexwlchan:
This takes the dynamism of #1047, and tamps it down a bit.

In particular, you can still add/remove individual topics rather than navigate a stack of boolean variables – but all the HCL you copy to spin up/down a new stack is “interesting” and something you might want to edit.

The idea is to give more prominence to the topics, so you'll notice they're waiting for you to edit them.

Fun fact: apparently there's a bug somewhere in #1047, because when I ran this it told me it wanted to add some reindexer topic subscriptions… which should have been enabled by the previous patch. 🤦‍♀️ 

The most conservative form of this is https://github.com/wellcomecollection/catalogue/commit/b8db0174b4e2523a29ebee8aaaa65151a7365fcd, which just consolidates the boilerplate at the bottom of the module definition.

### DIFF
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -30,6 +30,8 @@ locals {
       ]
     }
 
+    # A new pipeline with transformers that include support for multiple languages.
+    # See https://github.com/wellcomecollection/platform/issues/4864
     "2020-11-12" = {
       release_label = "stage"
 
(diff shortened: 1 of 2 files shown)


alicefuzier (approved):
This is really nice. I like how it hides the boilerplates and it keeps visible only things that change in between pipelines

Issue #1052, opened 2020-11-13
Expose availability online as a top-level work property


jamieparkinson:
Required for https://github.com/wellcomecollection/wellcomecollection.org/issues/5741

jamieparkinson:
New linked PR: https://github.com/wellcomecollection/catalogue/pull/1062

PR #1056, opened 2020-11-13
Top level availableOnline on works

jamieparkinson:
Resolves https://github.com/wellcomecollection/catalogue/issues/1052, and is a prerequisite for https://github.com/wellcomecollection/wellcomecollection.org/issues/5741

`availableOnline` is a boolean indicating whether any of the work's items have a digital location, therefore matching the availability filtering.

One question is - would we want this behind an `include` rather than being always present? Both for neatness and/or because we might not want to search the locations for every result.

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
@@ -84,7 +84,7 @@ case class DisplayWork(
     description = "List of items related to this work."
   ) items: Option[List[DisplayItem]] = None,
   @Schema(
-    description = "Whether the work contains an item with a digital location",
+    description = "Whether the work contains an item which is available online",
     `type` = "Boolean"
   ) availableOnline: Boolean = false,
   @Schema(
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
@@ -3,6 +3,8 @@ package uk.ac.wellcome.display.models
 import io.circe.generic.extras.JsonKey
 import io.swagger.v3.oas.annotations.media.Schema
 import uk.ac.wellcome.models.work.internal.{
+  DigitalLocationDeprecated,
+  Item,
   RelatedWork,
   RelatedWorks,
   Work,
@@ -26,7 +28,7 @@ case class DisplayWork(
   ) title: Option[String],
   @Schema(
     `type` = "List[String]",
-    description = "Alternative titles of  the work."
+    description = "Alternative titles of the work."
   ) alternativeTitles: List[String],
   @Schema(
     `type` = "String",
@@ -81,6 +83,10 @@ case class DisplayWork(
     `type` = "List[uk.ac.wellcome.Display.models.DisplayItem]",
     description = "List of items related to this work."
   ) items: Option[List[DisplayItem]] = None,
+  @Schema(
+    description = "Whether the work contains an item with a digital location",
+    `type` = "Boolean"
+  ) availableOnline: Boolean = false,
   @Schema(
     description = "Relates a work to its production events."
   ) production: Option[List[DisplayProductionEvent]] = None,
@@ -167,6 +173,7 @@ case object DisplayWork {
             DisplayItem(_, includesIdentifiers = includes.identifiers)
           })
         else None,
+      availableOnline = containsDigitalLocation(work.data.items),
       production =
         if (includes.production) Some(work.data.production.map {
           DisplayProductionEvent(_, includesIdentifiers = includes.identifiers)
@@ -230,6 +237,14 @@ case object DisplayWork {
             }
           } else None,
     )
+
+  def containsDigitalLocation(items: List[Item[_]]): Boolean =
+    items.exists { item =>
+      item.locations.exists {
+        case _: DigitalLocationDeprecated => true
+        case _                            => false
+      }
+    }
 
   def displayWorkType(workType: WorkType): String = workType match {
     case WorkType.Standard   => "Work"
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/ApiWorksTestBase.scala
@@ -34,6 +34,7 @@ trait ApiWorksTestBase
       |   "type": "${formatOntologyType(work.data.workType)}",
       |   "id": "${work.state.canonicalId}",
       |   "title": "${work.data.title.get}",
+      |   "availableOnline": ${hasDigitalLocations(work)},
       |   "alternativeTitles": []
       |   ${optionalObject("workType", format, work.data.format)}
       |   ${optionalObject("language", language, work.data.language)}
@@ -58,4 +59,9 @@ trait ApiWorksTestBase
       case WorkType.Series     => "Series"
       case WorkType.Section    => "Section"
     }
+
+  def hasDigitalLocations(work: Work.Visible[Identified]): String =
+    work.data.items
+      .exists(_.locations.exists(_.isInstanceOf[DigitalLocationDeprecated]))
+      .toString
 }
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksIncludesTest.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksIncludesTest.scala
@@ -40,6 +40,7 @@ class WorksIncludesTest
                    "id": "${work0.state.canonicalId}",
                    "title": "${work0.data.title.get}",
                    "alternativeTitles": [],
+                   "availableOnline": false,
                    "identifiers": [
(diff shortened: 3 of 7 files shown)


common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
```diff
@@ -81,6 +83,10 @@ case class DisplayWork(
     `type` = "List[uk.ac.wellcome.Display.models.DisplayItem]",
     description = "List of items related to this work."
   ) items: Option[List[DisplayItem]] = None,
+  @Schema(
+    description = "Whether the work contains an item with a digital location",
```
alexwlchan, line 87: Minor: isn't the point of this field to hide this detail from the user? Would a better description be "Whether the work contains an item which can be viewed online"?

alexwlchan (commented):
> Would we want this behind an include rather than being always present?

That would be my preference; possibly behind the "items" include? Given that this is a piece of information about the items.

jamieparkinson:
It shouldn't/can't go on the items include as the reason for having this field is so the client can tell whether a work is available online without fetching the items (ie in the search results)...

If there were to be an include I think it would just be `availableOnline`

jamesgorrie:
> That would be my preference; possibly behind the "items" include? Given that this is a piece of information about the items.

It is potentially about the items, as you've said, this implementation detail should be hidden from users.

I don't see the needs to make it an include, the reason we do that is to reduce the response size, and we have generally tried to do that by including arrays.

@alexwlchan What is the reason for making it an include?

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
```diff
@@ -231,6 +238,14 @@ case object DisplayWork {
           } else None,
     )
 
+  def containsDigitalLocation(items: List[Item[_]]): Boolean =
+    items.exists { item =>
+      item.locations.exists {
+        case _: DigitalLocationDeprecated => true
+        case _                            => false
+      }
+    }
```
warrd, line 247: Might be beyond the scope of this currently, but maybe this field should be calculated upfront during denormalisation? I wanted to make an optimisation to the API code, where we dynamically suppress fields returned from the ES request in relation to the includes, similarily to what we do here:

https://github.com/wellcomecollection/catalogue/blob/b4e4492ef625425b4c4516ed0eba27176504260e/pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/RelationsRequestBuilder.scala#L12-L26

This assumes the items exists in the `Work` model, which would not be the case if we did that and had not asked for items

warrd: Maybe not during "denormalisation", as not really the domain of the `relation_embedder`, but could be generated by transformers and then the merger works out what to do when we merge multiple items.

At the moment we are returning the whole work from ES on every request, including for the works search page which does not need most of that data.

jamieparkinson: I think this is a really good point (although I'm not sure it is in the current scope) and is a great example of the difficulties we have where the internal model is not really what we want the API to be consuming and converting to the display model.

Since this is derived data, from a modelling point of view it probably shouldn't live on the WorkData... If we stopped being tied to storing the internal model in the index, it could just be calculated at index time by the ingestors. I think having to have the merger work out what to do is unnecessary - as I say it's strictly derived data.

warrd: Not sure I understand why it would be beneficial for this purpose to have a separate model that is stored in the index. Won't we need that field somewhere on the internal model for conversion to the display model?

To me it sounds like the most appropriate place would for it to be in the state of denormalised works, which is the current method of storing derived data.

jamieparkinson: Separating the model that's in the index which gets searched, and the internal model of the pipeline has been something that's come up again and again (and something I know @jamesgorrie is keen on working towards). It would be beneficial here because, whilst it would be good to precalculate this field, that isn't information that's useful as part of the internal model (but there's no other model upon which we can store it).

warrd: I get why it might be beneficial to have a separate derived model for search, but in this case this information is useful as part of the internal model, precisely as we need this information to be serialised to the display model in the API. The search model as I understand it is data that we do not actually need returned from ES and we just use for indexing purposes.

jamieparkinson: I'm making a larger point that I think that this:
> precisely as we need this information to be serialised to the display model in the API

is not a good thing.

It makes small changes require reindexes because the API is so coupled to the internal model, and it makes storing derived data like this hard.

Unfortunately having thought about it I think you're right - this is too dumb a thing to do on every single work of every single request. I'll add the functionality to the merger but I regard this whole problem as a big to-do that will need refinement at some point.

jamesgorrie: @jamieparkinson I can't say I follow here completely. Are we saying that this property will be derived at the merger stage and then stored in ES?

> It makes small changes require reindexes because the API is so coupled to the internal model, and it makes storing derived data like this hard.

👆 this is a very valid point, and if what I've said above is true, does make this harder.

jamieparkinson: Yes - I think that if we are to calculate it in the pipeline it needs to be done by the merger or the ingestor, and since the ingestor currently is unable to do anything other than index a known model I think it needs to be the merger

warrd: Cool, yeah I think I understand the problem you are getting at, but not sure I agree that it is an issue with the internal model and API being coupled. One way to solve this using the current modelling would be to have another `WorkState` named `Derived`, where we calculate data like this right at the end of the pipeline. I think when everything is using pipeline storage we can make tools to run a reindex from a specific point in the pipeline (e.g. just recalculate all the derived data rather than having to do the transforms / merging / denormalisation etc before it), which would avoid the whole reindex scenario.

jamesgorrie: > just recalculate all the derived data rather than having to do the transforms / merging / denormalisation etc before it

That's really nice.

>  right at the end of the pipeline

In this case, I think this is what @jamieparkinson is saying, would be the merger as the in_minter / ingestor are intentionally stupid

jamieparkinson: I think `Derived` is a good solution as I don't want to do it in the merger (it just makes the most sense with the current model)

kenoir: FWIW I agree that indexing the internal model is not ideal now and that converting to a more appropriate structure before indexing is a longer term good idea.

warrd: > converting to a more appropriate structure before indexing is a longer term good idea.

Can someone explain what that structure might look like? I am not sure I understand the issue, as to me it would look exactly the same as the current internal model, but maybe with a few extra fields to help with search and derived fields (in which case a `Work[Derived]` would fit)

jamieparkinson: Separating the pipeline output (internal model) from the API input, and mapping between those 2 data stores separately would make it possible for the API ownership to extend outside of our team and increase the ability of the people working on the website to iterate in the way that they need, rather than being blocked by reindexes, Circe, etc

warrd: I might be missing something, but still really don't understand how this solves these issues. The suggestion is:

```
pipeline internal model    -->    API internal model         -->    API display model
(pipeline)                        (API backend)                     (API frontend)
```

1) Assuming we are storing a new API model, what is doing this conversion from the pipeline model and storing it, surely its pretty much just an extra pipeline stage? If so, how does that circumvent the need for reindexing?

2) As mentioned before, I do not see in what ways this new API model and the pipeline model (e.g. `Work[Derived]`) would actually differ. Can you elaborate on what the differences might be?

3) The display model is the current interface for people outside the team. There is also no fundamental reason non Scala services cannot use the current internal model if required, or present that as an interface. What purpose does having an extra model serve?

jamieparkinson: I think we're proposing pretty much the same thing if and only if pipeline intermediate storage is used and can be reindexed from any point - that is the same as the 2 data stores I'm talking about. I think any differences are mostly just naming, although it would still be great if that "derived" mapping could be done without having to know/care about the rest of the pipeline (even better if it could be in a language that handles JSON better than Java).

I don't really want this to become another confusing GitHub megathread and it would be good to discuss this in person (well, Teams) with other people (as I say I know @jamesgorrie has thoughts) - maybe later this week?

PR #1059, opened 2020-11-16
Router fixes

alicefuzier:
- [x] Make the router index the denormalized work if it doesn't belong to an archive
- [x] Send a `String` instead of `CollectionPath` to the batcher

### DIFF
--- a/pipeline/relation_embedder/router/src/main/scala/uk/ac/wellcome/platform/router/RouterWorkerService.scala
+++ b/pipeline/relation_embedder/router/src/main/scala/uk/ac/wellcome/platform/router/RouterWorkerService.scala
@@ -29,12 +29,14 @@ class RouterWorkerService[MsgDestination](
       work.data.collectionPath
         .fold(ifEmpty = {
           for {
-           worksEither <- workIndexer.index(work.transition[Denormalised] (Relations.none))
-           _ <- worksEither match {
-             case Left(failedWorks) => Future.failed(new Exception(s"Failed indexing $failedWorks"))
-             case Right(_) => Future.successful(())
-           }
-            _ <-Future.fromTry(worksMsgSender.send(work.id))
+            worksEither <- workIndexer.index(
+              work.transition[Denormalised](Relations.none))
+            _ <- worksEither match {
+              case Left(failedWorks) =>
+                Future.failed(new Exception(s"Failed indexing $failedWorks"))
+              case Right(_) => Future.successful(())
+            }
+            _ <- Future.fromTry(worksMsgSender.send(work.id))
           } yield ()
         }) { path =>
           Future.fromTry(pathsMsgSender.send(path.path))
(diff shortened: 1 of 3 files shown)


pipeline/relation_embedder/router/src/main/scala/uk/ac/wellcome/platform/router/RouterWorkerService.scala
```diff
@@ -26,8 +27,19 @@ class RouterWorkerService[MsgDestination](
...
+              work.transition[Denormalised](Relations.none))
+            _ <- worksEither match {
+              case Left(failedWorks) =>
+                Future.failed(new Exception(s"Failed indexing $failedWorks"))
+              case Right(_) => Future.successful(())
+            }
+            _ <- Future.fromTry(worksMsgSender.sendT(work.id))
+          } yield ()
+        }) { path =>
+          Future.fromTry(pathsMsgSender.sendT(path.path))
```
warrd, line 42: See comment here: https://github.com/wellcomecollection/catalogue/pull/1033#discussion_r521924141

pipeline/relation_embedder/router/src/main/scala/uk/ac/wellcome/platform/router/RouterWorkerService.scala
```diff
@@ -26,8 +27,17 @@ class RouterWorkerService[MsgDestination](
   private def processMessage(message: NotificationMessage): Future[Unit] = {
     workRetriever.apply(message.body).flatMap { work =>
       work.data.collectionPath
-        .fold(ifEmpty = Future.fromTry(worksMsgSender.sendT(work.id))) { path =>
-          Future.fromTry(pathsMsgSender.sendT(path))
+        .fold(ifEmpty = {
+          for {
+           worksEither <- workIndexer.index(work.transition[Denormalised] (Relations.none))
```
warrd, line 32: :thinking: it might actually be worth using `groupedWithin` like we do elsewhere to batch up the works before indexing, as we have encountered issues previously with ES performance.

warrd: We don't actually do that everywhere on second thoughts, but e.g. in the merger we do send the whole merged works in one go rather than individually.

alicefuzier: yeah, I realised this too. This isn't batching and it will probably cause problems. I feel that if we're going to use the pipeline storage on other pipeline services it should be somehow provided by that though. Most services in the pipeline work in a "one in one out" way like the router. The merger and the relation embedder are exceptions in this sense and it might be worth to do it in the merger too anyway.

warrd (approved):

PR #1062, opened 2020-11-17
Add an Indexed WorkState and use it to hold availableOnline

jamieparkinson:
As per @warrd's suggestion in https://github.com/wellcomecollection/catalogue/pull/1056#discussion_r524237356, this adds a new final state to the `WorkState` FSM, `Derived`. This is used to hold derived data which is calculated immediately before indexing, in the ingestor. 

As was discussed in the last PR, this combined with full adoption of `pipeline_storage` will enable us to make (some) data changes without full reindexes being required. In this first instance, we use the `derivedData` to hold an `availableOnline` property.

A **lot** (the vast majority) of this PR is boilerplate changing of `Identified` -> `Derived` and `identifiedWork(s)` -> `derivedWork(s)` so a rough guide to the interesting bits follows:

- [the new `Derived` state and transition](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-392d4159297494dd464cb120a082e26fc55aa3d855242e33cd47cc27112b6119R172)
- [the `DerivedData` object](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-048a9aa8b5a2ed3e3b186f96eb5a271bb51c769ab5fa4305494fcdae1de31b09R5) which holds and calculates the derived data.
- [changes to the ingestor](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-8756ce3eeda67b331911d652c2077fa617569e9953c82d9a76b80f537f58f1b6R13) for transforming documents immediately before ingesting them.
- [a typeclass for recalculating derived state with our fluent-style WorkGenerators](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-df0213c8bdec7dcfdb1e7fcf9cd30837a679f2bcc8e6ff231907618afd0fd917R213). This only affects tests, and is the best/only way that I could find to get the generators to automagically recalculate the state when building up the work data.
- [the index config for the `Derived` state](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-116ca4a08da8f56353739dad1fb1fb78921f321bdfbd5398009cabb08fef4686R264)
- [the new property on `DisplayWork`s](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-543816e295c0647a438e198c83cbeae948df282137703f655b84ce2cf826c3a1R87)
- [a minor point](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-42df172f7d382a01da5fc05d65a36c8b4337b4428ea1ea85b833967e1157442dR168) about the expected behaviour of `availableOnline` for related works
- [moved `mapData` off of `Work`s](https://github.com/wellcomecollection/catalogue/pull/1062/files#diff-24547383116feb40679e907622ca0adfa5b08487775933fbcb5e073994780735R125) into the merger, to clarify that we don't expect/want `WorkData` to ever be changed.


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -20,17 +20,6 @@ sealed trait Work[State <: WorkState] {
...
-  def mapData(f: WorkData[State#WorkDataState] => WorkData[State#WorkDataState])
-    : Work[State] =
-    this match {
-      case Work.Visible(version, data, state) =>
-        Work.Visible(version, f(data), state)
-      case Work.Invisible(version, data, state, reasons) =>
-        Work.Invisible(version, f(data), state, reasons)
-      case Work.Redirected(version, redirect, state) =>
-        Work.Redirected(version, redirect, state)
-    }
```
warrd, line 32: :wave:

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -175,6 +168,20 @@ object WorkState {
...
+  case class Derived(
+    sourceIdentifier: SourceIdentifier,
+    canonicalId: String,
+    modifiedTime: Instant,
+    derivedData: DerivedData,
+    relations: Relations[DataState.Identified] = Relations.none
+  ) extends WorkState {
+
+    type WorkDataState = DataState.Identified
+    type TransitionArgs = WorkData[WorkDataState]
```
warrd, line 181: Hmm not really ideal that we have to pass this in as an arg. Maybe the `state` function in `Transition` should have access to the `WorkData` always instead?

warrd (approved):
Nice.

At the moment looks very clean the transformation happening in the ingestors. However I am thinking if we move everything in the pipeline across from `big_messaging` / VHS to `pipeline_storage`, I am not sure the ingestors will actually still exist (the final stage of the pipeline, currently the ID minter, will just do the indexing itself).

So something to bear in mind for the future, we might either need to do the deriving in the `id_minter` or somewhere, or have a seperate service just for it.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
jamieparkinson: Yeah, I went through the same thought process - seemed like changing too much at the time but I'll give it a go now to see how it looks

warrd:
Another consideration.. Having `derivedWork()` / `Work[Derived]` littered throughout the code base is maybe a bit more confusing than what we had before with `identifiedWork()` / `Work[Identified]`.

I know we would end up getting used to it, but it might make sense to think of other names for this state, which make more sense where the work is used. Some initial ideas which may not be very good:
```
Work[Final]
Work[API]
Work[Queryable]
```

@jamesgorrie might have some thoughts on this?

jamieparkinson:
+1 on thinking about naming, definitely not 100% sure about `Derived`

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -20,10 +20,10 @@ sealed trait Work[State <: WorkState] {
   def identifiers: List[SourceIdentifier] =
     sourceIdentifier :: data.otherIdentifiers
 
-  def transition[OutState <: WorkState](args: OutState#TransitionArgs)(
+  def transition[OutState <: WorkState](args: OutState#TransitionArgs = ())(
```
warrd, line 23: Really baffled how this compiles :thinking: `TranstionArgs` isn't necessarilly of type `Unit` so confused how using that as a default works here

warrd: It catches this when you apply it to a type which isn't `Unit`:

```
[E]      type mismatch;
[E]       found   : Unit
[E]       required: uk.ac.wellcome.models.work.internal.WorkState.Denormalised#TransitionArgs
[E]          (which expands to)  uk.ac.wellcome.models.work.internal.Relations[uk.ac.wellcome.models.work.internal.DataState.Unidentified]
[E]      Error occurred in an application involving default arguments.
[E]      L251:     work.transition[Denormalised]()
[E]                               ^
```

jamieparkinson: I was going to comment that on it haha - part of me wonders if it will compile in CI and it's just that my IDE hasn't noticed it...

warrd: No I checked it out locally and it works!

jamesgorrie:
@warrd could you remind me again of the what steps come before deriving this data? Got lost between this and the denormalisation work, and could be useful to think of what it's doing in context of the whole pipeline.

jamieparkinson:
@jago you might find the diagram here helpful https://github.com/wellcomecollection/catalogue/blob/derived-work-state/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala#L92

jamieparkinson:
Votes from me for `Indexed` and `Catalogued`

kenoir:
>  I am not sure the ingestors will actually still exist (the final stage of the pipeline, currently the ID minter, will just do the indexing itself).

An aside: I think we should keep the ingestor as a separate service whatever - ingesting into the prod index should be scalable separately from any service before (especially a service like  the id_minter coupled to a SQL database.

kenoir (commented):
Nice way to do this, +1 `Indexed`

warrd:
> An aside: I think we should keep the ingestor as a separate service whatever - ingesting into the prod index should be scalable separately from any service before (especially a service like the id_minter coupled to a SQL database).

Yep I agree with this upon some thought, it originally seemed weird to me as if the `id_minter` was going to use `pipeline_storage` it would essentially be indexing to ES too, but the final stage is a bit different due to all the index mappings and the fact its the API index rather than one in the pipeline storage cluster.

PR #1073, opened 2020-11-19
Batching in Router and PipelineStorageStream

alicefuzier:
The router is currently sending documents to index in elastic search using the batching API, but it sends then one at a time, so it's effectively not batching. Because messages can only be deleted from the queue when indexing and sending on the id has succeeded, batching affects the whole stream, as we need to hold on to the original message to be able to delete it.

This PR is adding logic to group the messages into batches in a `PipelineStorageStream` which is a decorator for `SQSStream`. In this PR `PipelineStorageStream` is used in the router, but it should be easy to swap usages of SQSStream for `PipelineStorageStream` without having to change the code except for configuration and infrastructure

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
@@ -2,7 +2,7 @@ package uk.ac.wellcome.pipeline_storage
 
 import akka.stream.FlowShape
 import akka.{Done, NotUsed}
-import akka.stream.scaladsl.{Broadcast, Flow, GraphDSL, Merge, Source}
+import akka.stream.scaladsl.{Broadcast, Flow, GraphDSL, Merge}
 import grizzled.slf4j.Logging
 import io.circe.Decoder
 import software.amazon.awssdk.services.sqs.model.Message
@@ -21,8 +21,7 @@ class PipelineStorageStream[T, D, MsgDestination](messageStream: SQSStream[T], d
     implicit decoderT: Decoder[T], indexable: Indexable[D]): Future[Done] =
     run(
       streamName = streamName,
-      source =>
-        source
+      Flow[(Message,T)]
           .mapAsyncUnordered(parallelism = config.parallelism) {
             case (message, t) =>
               debug(s"Processing message ${message.messageId()}")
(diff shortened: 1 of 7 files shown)


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -0,0 +1,96 @@
...
+  private def batchAndSendFlow(implicit indexable: Indexable[D]) = {
+    Flow[(Message, Option[D])]
+      .collect { case (message, Some(document)) => Bundle(message, document) }
+      .groupedWithin(
+        config.batchSize,
+        config.flushInterval
+      )
+      .mapAsyncUnordered(10) { msgs =>
+        storeDocuments(msgs.toList)
+      }
```
alexwlchan, line 65: If I've read this correctly, this groups up to 10 messages at once? Is it worth exposing this as a top level config item, so we can tweak it?

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStreamTest.scala
```diff
@@ -0,0 +1,248 @@
...
+class PipelineStorageStreamTest
+    extends AnyFunSpec
+    with ElasticIndexerFixtures
+    with IdentifiersGenerators
+    with RandomGenerators
+    with ElasticsearchFixtures
+    with SQS
+    with Akka {
+
+  val pipelineStorageConfig = PipelineStorageConfig(1, 1 seconds, 10)
```
alexwlchan, line 37: keyword arguments plz

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStreamTest.scala
```diff
@@ -0,0 +1,248 @@
...
+                            pipelineStorageConfig: PipelineStorageConfig =
+                              pipelineStorageConfig)(
+    testWith: TestWith[
+      (PipelineStorageStream[NotificationMessage, SampleDocument, String],
+       QueuePair,
+       MemoryMessageSender),
+      R]): R =
+    withActorSystem { implicit ac =>
+      withLocalSqsQueuePair(visibilityTimeout) {
+        case q @ QueuePair(queue, _) =>
```
alexwlchan, line 51: Why are you creating a queue pair if you never use the DLQ?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
alicefuzier: ah yes, good spot thanks

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStreamTest.scala
alicefuzier: we do, just not here in the `with` method. We pass the whole `QueuePair` in `testWith` so each test can use it

kenoir (approved):

PR #1076, opened 2020-11-19
Make test logging output more user friendly

warrd:
Currently the test logging is really really noisy making it much harder than it needs to be for parsing the test output both locally and on CI.

IMO 95% of the time or more you are only interested in the test output: whether something has passed, or failed and with what failure / error. So here we make the standard in tests to suppress stdout from inside the tests.

In the certain cases where you do need to see the standard out (such as debugging) it is possible to uncomment a line in `logback-test.xml` to get this displaying. I have improved the output here in a few ways:
1) Make the output coloured
2) Simplify time display so seconds is smallest resolution
3) Display a simplified class name rather than the fully namespaced one
4) Don't display what thread which as this useful is rarely helpful (e.g. `default-akka.actor.default-dispatcher-7`)
5) Don't display output from external libs by default (can be enabled by the user if needed).
6) Don't display the verbose logback status upon startup

## Before:
2020-11-19-174235_959x808_scrot

## Standard:
2020-11-19-172022_958x382_scrot

## With STDOUT enabled:
2020-11-19-175904_959x679_scrot

### DIFF
--- a/api/api/src/test/resources/logback-test.xml
+++ b/api/api/src/test/resources/logback-test.xml
@@ -12,7 +12,7 @@
 
   <root level="DEBUG">
     <!-- Use the CATALOGUE_TEST_LOGGING env var to set which logger to use -->
-    <appender-ref ref="${CATALOGUE_TEST_LOGGING:-standard}"/>
+    <appender-ref ref="${CATALOGUE_TEST_LOGGING:-silent}"/>
   </root>
 
   <!-- Edit these to change the log levels for external libraries -->
--- a/api/api/src/test/resources/logback-test.xml
+++ b/api/api/src/test/resources/logback-test.xml
@@ -1,21 +1,21 @@
 <configuration>
+
+  <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
+
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <encoder>
-      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+      <pattern>%blue(%d{HH:mm:ss}) %magenta([%logger{0}]) %yellow(%-5level) %msg%n</pattern>
     </encoder>
   </appender>
 
   <root level="DEBUG">
-    <appender-ref ref="STDOUT" />
+    <!-- Uncomment the following line to display stdout of tests -->
+    <!-- <appender-ref ref="STDOUT"/> -->
   </root>
 
-  <logger name="com.sksamuel.elastic4s" level="INFO" />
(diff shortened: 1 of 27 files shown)


kenoir (approved):
👏 _much_ better.

jamieparkinson (commented):
I think simpler as default is good but I'm not sure about the stdout suppression - often stuff that doesn't go through the logging adapter is the really useful/interesting stuff. Can't see in your example where the difference with/without stdout is - could you point me to what that looks like?

Another thought is that I might keep the old logging encoder in a comment in case we want it back (eg, every now and then I find the thread logging very useful, even though usually it isn't)

warrd:
> Can't see in your example where the difference with/without stdout is - could you point me to what that looks like?

The difference is between the second and third images.

> often stuff that doesn't go through the logging adapter is the really useful/interesting stuff.

What doesn't go through the logging adapter, you mean like `println`? Should have mentioned that goes through at all times, even when you haven't enabled the logging output.

jamieparkinson:
> The difference is between the second and third images.

Sorry, somehow totally missed that. I much prefer the version with STDOUT, I frequently use the logging output when debugging tests locally. Is there a way to do it conditionally eg disable stdout on presence of the `CI` env var?

warrd:
Yeah maybe a env var or something would be good, as personally I much prefer the one without (note that in other test cases the log output is much longer than in these tests). I will investigate..

alexwlchan:
I like the more compact output, but agree with Jamie that the stdout is useful sometimes.

When I'm using py.test in Python, it does the compact printing unless the test fails, and then it shows you all of stdout and stderr. That feels like a good compromise. Can we get something like that with scalatest?

warrd:
How about like this?

2020-11-20-111016_958x945_scrot

jamieparkinson:
Sadly looks like @alexwlchan's request isn't really possible in Scala :( https://github.com/scalatest/scalatest/issues/405. Yeah, I think that env var is a good idea (although it'll need to be set in CI manually, could you derive it from `CI` or another existing var?

alexwlchan:
> although it'll need to be set in CI manually, could you derive it from CI or another existing var?

Unfortunately the tests in CI run in docker, so the env vars set by CI aren't visible to scalatest.

You can _kinda_ approximate it by asking "am I running as root"?

warrd:
Might be better to just make whatever we want for CI the default option, and then anyone who does't want that can manually export the env var in there `bashrc` or whatever?

I am more bothered about setting it locallly than CI (although I would still prefer it to be quieter), so up for either choice being the default really.

jamieparkinson:
Let's have default quiet (I think I can set the environment variable in IntelliJ and forget about it)

PR #1078, opened 2020-11-19
Remove some unused sbt packager settings

alexwlchan:
Currently this is dropping warnings in the build:

    [warn] there are 35 keys that are not used by any other settings/tasks:[0J
    [warn]
    [warn] * api / composeNoBuild
    [warn]   +- /var/lib/buildkite-agent/builds/buildkite-elasticstack-i-05cfcac96e54357a61/wellcomecollection/catalogue/project/Packager.scala:9


kenoir (approved):

jamieparkinson:
iirc I looked into this and couldn't find a way of removing these settings in a way that didn't break the docker compose plugin - because it's command based rather than task based (or maybe the other way round?) it's not very composable

alexwlchan:
It seems to be fine in the storage service, but I don't want to spend a lot of time on this.

PR #1085, opened 2020-11-20
Declare ecr_repos data blocks outside of stack module

alicefuzier:
This is an attempt at solving one of the issues we have with targeting individual pipelines. Currently, if we add services to on e pipeline, it causes the following error when planning (with target): ```Call to function "zipmap" failed: number of keys (18) does not match number of
values (0).```. This solves this problem by taking the data block out


alexwlchan (approved):
✨

kenoir (approved):

kenoir:
@alicefuzier this looks good - you might want to add a little bit more in the description about _why_ this.

PR #1094, opened 2020-11-24
Add source's languages, title and first contributor to images

jamieparkinson:
This adds the `languages`, the `title` and the first of the `contributors` from a source work to the `source` field on images. The `contributor` and `languages` are behind an `include` to match the behaviour of works. 

We'll also need additional locations on images, but that will require a bit more thought / a separate PR.


jamesgorrie (approved):
Nice and tidy!

PR #1101, opened 2020-11-24
Show all contributors on image sources

jamieparkinson:
It was confusing to have slightly different fields on works and images - this puts _all_ contributors on `image.source.contributors`


jamesgorrie (approved):
What was the reason for choosing 1 in the first place?

alexwlchan (approved):
🧽

jamieparkinson:
Confusion

jtweed:
> What was the reason for choosing 1 in the first place?

Thinking we'd put it directly on the image, but have since decided to keep all descriptive source metadata on the source, so it should match the Work model.

PR #1102, opened 2020-11-24
Read METS in the transformer without assuming an IAM role

alexwlchan:
This is mostly untested – if I plan locally, it wants to tear down the current prod pipeline (https://github.com/wellcomecollection/catalogue/pull/1100) – but hopefully illustrates the concept.

Part of https://github.com/wellcomecollection/platform/issues/4895


alicefuzier (approved):
Thank you!

PR #1104, opened 2020-11-24
Calculate numChildren and numDescendents in relation embedder

warrd:
https://github.com/wellcomecollection/platform/issues/4861

### DIFF
--- a/pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
+++ b/pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
@@ -37,8 +37,6 @@ class ArchiveRelationsCache(
 
   def size = relations.size
 
-  def numParents = parentMapping.size
-
   private def getChildren(
     path: String): List[Relation[DataState.Unidentified]] =
     childMapping
--- a/pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
+++ b/pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
@@ -85,7 +85,7 @@ class ArchiveRelationsCache(works: Map[String, Work[Merged]]) extends Logging {
       }
     numDescendents(childMapping.getOrElse(path, Nil))
   }
-  
+
   private lazy val relations =
     works.map {
       case (path, work) =>
--- a/pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
+++ b/pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
@@ -6,9 +6,7 @@ import grizzled.slf4j.Logging
 import uk.ac.wellcome.models.work.internal._
 import WorkState.Merged
 
-class ArchiveRelationsCache(
-  relations: Map[String, Relation[DataState.Unidentified]])
-    extends Logging {
+class ArchiveRelationsCache(works: Map[String, Work[Merged]]) extends Logging {
 
   def apply(work: Work[Merged]): Relations[DataState.Unidentified] =
     work.data.collectionPath
@@ -36,6 +34,8 @@ class ArchiveRelationsCache(
       }
 
   def size = relations.size
+
+  def numParents = parentMapping.size
 
   private def getChildren(
     path: String): List[Relation[DataState.Unidentified]] =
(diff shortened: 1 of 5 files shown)


pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
```diff
@@ -74,18 +72,43 @@ class ArchiveRelationsCache(
         getAncestors(parentPath, relations(parentPath) :: accum)
     }
 
+  private def getNumChildren(path: String): Int =
+    childMapping.get(path).map(_.length).getOrElse(0)
+
+  private def getNumDescendents(path: String): Int = {
+    @tailrec
+    def numDescendents(stack: List[String], accum: Int = 0): Int =
```
jamieparkinson, line 80: I don't believe this needs to be a closure, it doesn't close over `path`

jamieparkinson (approved):

pipeline/relation_embedder/relation_embedder/src/test/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCacheTest.scala
```diff
@@ -27,25 +27,35 @@ class ArchiveRelationsCacheTest
...
+  val relA = Relation(workA, depth = 0, numChildren = 4, numDescendents = 9)
+  val rel1 = Relation(work1, depth = 1, numChildren = 2, numDescendents = 2)
+  val relB = Relation(workB, depth = 2, numChildren = 0, numDescendents = 0)
+  val relC = Relation(workC, depth = 2, numChildren = 0, numDescendents = 0)
+  val rel2 = Relation(work2, depth = 1, numChildren = 2, numDescendents = 3)
+  val relD = Relation(workD, depth = 2, numChildren = 0, numDescendents = 0)
+  val relE = Relation(workE, depth = 2, numChildren = 1, numDescendents = 1)
+  val relF = Relation(workF, depth = 3, numChildren = 0, numDescendents = 0)
+  val rel3 = Relation(work3, depth = 1, numChildren = 0, numDescendents = 0)
+  val rel4 = Relation(work4, depth = 1, numChildren = 0, numDescendents = 0)
```
alexwlchan, line 39: nit: sorting these would make it easier to follow

alexwlchan (approved):

pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
warrd: It's not a closure, just a function defined within another to hide it from the outside.

pipeline/relation_embedder/relation_embedder/src/test/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCacheTest.scala
warrd: Yes it is weird, I just followed the sorting which existed already for the works which I'm sure I originally sorted according to some specific but indecipherable rules :see_no_evil:

pipeline/relation_embedder/relation_embedder/src/main/scala/uk/ac/wellcome/relation_embedder/ArchiveRelationsCache.scala
jamieparkinson: Exactly, it's not capturing anything - so you're redefining the inner function every time `getNumDescendents` is called

warrd: Hm not sure whether the scala compiler actually does that under the hood, and if so I am more bothered with keeping in organised and hidden than premature optimisation for this case

warrd: I don't know about Scala compile but for Rust for example a closure function will be compiled to a single function which takes an additional struct containing the captured environment.

jamieparkinson: I'm talking more about readability/intent than performance - I spent a while looking for `path` in the inner function because we usually only use inner functions to close over something. It's a nit regardless...

warrd: Ah ok. Personally I don't think expect a closure whenever I see an inner function, the official scala docs has a very similar example for defining a recursive function: https://docs.scala-lang.org/tour/nested-functions.html

PR #1105, opened 2020-11-24
Refactoring and more testing in TransformerWorker

alexwlchan:
I was looking through TransformerWorker, to get an idea of how much work it would be to use pipeline_storage instead.

This patch is not that – it's some code tidies and extra test cases I spotted while reading the code.

This is laying the groundwork to use the same TransformerWorker pattern in all four transformers, and in turn adding pipeline_storage in one go.


jamesgorrie (approved):
Having not looked at the code for a while, these tests have made it way clearer as to what it does. Nice.

PR #1106, opened 2020-11-25
Remove visible filter from relation embedder query

warrd:
This is a mistake, it means non visible works won't make it through the pipeline.


jamesgorrie (approved):
No test to check this?

PR #1118, opened 2020-11-26
Merge thumbnails onto Calm works

jamieparkinson:
This also consolidates our branching on access statuses into one place, therefore enforcing that we make these decisions consistently across the pipeline (rather than separately in the METS transformer and the merger).

Resolves https://github.com/wellcomecollection/platform/issues/4892


warrd (approved):

jamieparkinson:
Waiting on confirmation that "unavailable" is not a restricted access status

PR #1127, opened 2020-11-30
Make the bulk_size parameter configurable in the snapshot generator

alexwlchan:
Tentative fix for https://github.com/wellcomecollection/platform/issues/4901

Previously we'd fetch 1000 works in one go, and now it seems that's a bit too big. Let's try lowering it to 500 and see what happens.


alicefuzier (approved):

snapshots/snapshot_generator/src/main/scala/uk/ac/wellcome/platform/snapshot_generator/models/SnapshotGeneratorConfig.scala
```diff
@@ -0,0 +1,19 @@
+package uk.ac.wellcome.platform.snapshot_generator.models
+
+import com.sksamuel.elastic4s.Index
+
+case class SnapshotGeneratorConfig(
+  index: Index,
+  // How many documents should be fetched in a single request?
```
kenoir, line 7: 👍

kenoir (approved):

PR #1132, opened 2020-11-30
Add a Deleted Work type

alicefuzier:
For https://github.com/wellcomecollection/platform/issues/4893

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksController.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksController.scala
@@ -64,6 +64,8 @@ class WorksController(elasticsearchService: ElasticsearchService,
               workRedirect(work)
             case Right(Some(work: Work.Invisible[Indexed])) =>
               gone("This work has been deleted")
+            case Right(Some(work: Work.Deleted[Indexed])) =>
+              gone("This work has been deleted")
             case Right(None) =>
               notFound(s"Work not found for identifier $id")
             case Left(err) => elasticError(err)
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
@@ -1,31 +1,31 @@
 package uk.ac.wellcome.platform.api.works
 
-import uk.ac.wellcome.models.work.internal._
 import uk.ac.wellcome.models.Implicits._
-import WorkState.Indexed
+import uk.ac.wellcome.models.work.internal.WorkState.Indexed
+import uk.ac.wellcome.models.work.internal._
 
-class WorksTestInvisible extends ApiWorksTestBase {
-  val invisibleWork: Work.Invisible[Indexed] = indexedWork().invisible()
+class WorksTestDeleted extends ApiWorksTestBase {
+  val deletedWork: Work.Deleted[Indexed] = indexedWork().deleted()
 
-  it("returns an HTTP 410 Gone if looking up a work with visible = false") {
+  it("returns an HTTP 410 Gone if looking up a deleted work") {
     withWorksApi {
       case (worksIndex, routes) =>
-        insertIntoElasticsearch(worksIndex, invisibleWork)
-        val path = s"/$apiPrefix/works/${invisibleWork.state.canonicalId}"
+        insertIntoElasticsearch(worksIndex, deletedWork)
+        val path = s"/$apiPrefix/works/${deletedWork.state.canonicalId}"
         assertJsonResponse(routes, path) {
           Status.Gone -> deleted(apiPrefix)
         }
     }
   }
 
-  it("excludes works with visible=false from list results") {
+  it("excludes deleted works from list results") {
     withWorksApi {
       case (worksIndex, routes) =>
         val works = indexedWorks(count = 2).sortBy {
           _.state.canonicalId
         }
 
-        val worksToIndex = Seq[Work[Indexed]](invisibleWork) ++ works
+        val worksToIndex = Seq[Work[Indexed]](deletedWork) ++ works
         insertIntoElasticsearch(worksIndex, worksToIndex: _*)
 
         assertJsonResponse(routes, s"/$apiPrefix/works") {
@@ -34,11 +34,11 @@ class WorksTestInvisible extends ApiWorksTestBase {
     }
   }
 
-  it("excludes works with visible=false from search results") {
+  it("excludes deleted works from search results") {
     withWorksApi {
       case (worksIndex, routes) =>
         val work = indexedWork().title("This shouldn't be deleted!")
-        insertIntoElasticsearch(worksIndex, work, invisibleWork)
+        insertIntoElasticsearch(worksIndex, work, deletedWork)
 
         assertJsonResponse(routes, s"/$apiPrefix/works?query=deleted") {
           Status.OK -> worksListResponse(apiPrefix, works = Seq(work))
--- a/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
+++ b/api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
@@ -32,17 +32,5 @@ class WorksTestDeleted extends ApiWorksTestBase {
           Status.OK -> worksListResponse(apiPrefix, works = works)
         }
     }
-  }
+  }}
 
-  it("excludes deleted works from search results") {
-    withWorksApi {
-      case (worksIndex, routes) =>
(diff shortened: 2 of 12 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksController.scala
```diff
@@ -64,6 +64,8 @@ class WorksController(elasticsearchService: ElasticsearchService,
               workRedirect(work)
             case Right(Some(work: Work.Invisible[Indexed])) =>
               gone("This work has been deleted")
+            case Right(Some(work: Work.Deleted[Indexed])) =>
+              gone("This work has been deleted")
```
alexwlchan, line 68: Now we have this message, is it worth tweaking the one for Invisible?

"This work is not available"?

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
```diff
@@ -0,0 +1,48 @@
...
+    withWorksApi {
+      case (worksIndex, routes) =>
+        val work = indexedWork().title("This shouldn't be deleted!")
+        insertIntoElasticsearch(worksIndex, work, deletedWork)
+
+        assertJsonResponse(routes, s"/$apiPrefix/works?query=deleted") {
+          Status.OK -> worksListResponse(apiPrefix, works = Seq(work))
+        }
+    }
+  }
```
alexwlchan, line 47: Is there anything about `deletedWork` that means it would show up in this search if our deletion filter wasn't working?

Don't you need to attach the word "deleted" somewhere to it?

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  it("excludes deleted works from list results") {
+    withWorksApi {
+      case (worksIndex, routes) =>
+        val works = indexedWorks(count = 2).sortBy {
+          _.state.canonicalId
+        }
+
+        val worksToIndex = Seq[Work[Indexed]](deletedWork) ++ works
+        insertIntoElasticsearch(worksIndex, worksToIndex: _*)
```
alexwlchan, line 29: Maybe simpler:

```suggestion
        insertIntoElasticsearch(worksIndex, deletedWork, works: _*)
```

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestInvisible.scala
```diff
@@ -38,7 +38,7 @@ class WorksTestInvisible extends ApiWorksTestBase {
     withWorksApi {
       case (worksIndex, routes) =>
         val work = indexedWork().title("This shouldn't be deleted!")
-        insertIntoElasticsearch(worksIndex, work, deletedWork)
+        insertIntoElasticsearch(worksIndex, work, invisibleWork)
```
alexwlchan, line 41: I am also wondering if this test is dubious, for similar reasons.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DeletedReasons.scala
```diff
@@ -0,0 +1,8 @@
+package uk.ac.wellcome.models.work.internal
+
+sealed trait DeletedReason
+
+object DeletedReason {
+  case class DeletedFromSource(info: String) extends DeletedReason
+  case class SuppressedFromSource(info: String) extends DeletedReason
+}
```
alexwlchan, line 8: Really nice to have these in the index.

Do we need to do something to ensure they're not searchable, invisible fields?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
```diff
@@ -37,12 +38,10 @@ object CalmTransformer
   def apply(record: CalmRecord, version: Int): Result[Work[Source]] =
     if (shouldSuppress(record)) {
       Right(
-        Work.Invisible[Source](
+        Work.Deleted[Source](
```
alexwlchan, line 41: I'm not sure this is quite right – a work can be suppressed but still present in the source system, right? Possibly I'm splitting hairs, but I'd prefer to keep suppressed-but-extant works as Invisible.

alexwlchan (approved):
Bunch of small comments, but nothing substantial.

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
alicefuzier: ha, I think this test actually doesn't make any sense cause `Deleted` works don't carry information ie the `WorkData` block is always empty. I copied the tests for `Invisible` works where this makes sense and didn't read properly. I'm going to just delete this test. Thanks

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
alicefuzier: Unfortunately the compiler doesn't like that cause it can't figure out the common type between `deletedWork` and `works`. I think this is why it was done like this in `WorkTestInvisible`

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestInvisible.scala
alicefuzier: It is, will fix

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DeletedReasons.scala
alicefuzier: they are invisible cause the works that carry them are not visible. We want them to be searchable cause they are searched by people in #collections-info

api/api/src/test/scala/uk/ac/wellcome/platform/api/works/WorksTestDeleted.scala
alexwlchan: I figured there'd be a sensible reason, thanks for explaining.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DeletedReasons.scala
alexwlchan: Yeah, I’ve realised what I asked doesn't make sense. 😅

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
alicefuzier: Maybe it's a naming issue. The main drive for splitting `Deleted` out of `Invisible` is that `Invisible` is used in the METS transformer (and Sierra I think) to send through works that are incomplete, so shouldn't be shown on their own. METS works are always considered incomplete and sent as `Invisible`. However, the merger can use the information in `Invisible` works to merge into `Visible` works. We don't want that for `Deleted` works. So the discrimination in my mind has been: is what we have in this work usable in the pipeline to enrich other works? yes => make it `Invisible`, no => make it "Deleted". I think `Suppressed` falls into the second case?

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksController.scala
alicefuzier: I find that a bit weird cause the response it's still a gone 410 instead of a 404 not available. I don't think the difference `Invisible` and `Deleted` should be of interest from the API perspective, it's more a difference in treatment in the pipeline. If we had an API only internal model, I don't think we would need these two types

alicefuzier:
@alexwlchan Do the answers to your comments make sense? Can I merge or do you want to talk about those a bit more?

pipeline/transformer/transformer_calm/src/main/scala/uk/ac/wellcome/platform/transformer/calm/CalmTransformer.scala
alexwlchan: If I was really splitting hairs, I'd introduce _another_ work state, so we'd have Invisible/Suppressed/Deleted, but I don't think it's worth holding up this PR.

alexwlchan:
> Do the answers to your comments make sense?

Yep.

> Can I merge or do you want to talk about those a bit more?

Merge away!

PR #1134, opened 2020-11-30
Deploy pipeline 2020-12-02

warrd:
Yes I do realise this date doesn't exist :see_no_evil:

### DIFF
--- a/pipeline/terraform/stack/service_work_batcher.tf
+++ b/pipeline/terraform/stack/service_work_batcher.tf
@@ -2,7 +2,7 @@ module "batcher_queue" {
   source                     = "git::github.com/wellcomecollection/terraform-aws-sqs//queue?ref=v1.1.2"
   queue_name                 = "${local.namespace_hyphen}_batcher"
   topic_arns                 = [module.router_path_output_topic.arn]
-  visibility_timeout_seconds = 1000
+  visibility_timeout_seconds = 1900
   aws_region                 = var.aws_region
   alarm_topic_arn            = var.dlq_alarm_arn
 }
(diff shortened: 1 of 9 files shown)


pipeline/terraform/stack/service_work_batcher.tf
```diff
@@ -26,9 +26,9 @@ module "batcher" {
     queue_url        = module.batcher_queue.url
     output_topic_arn = module.batcher_output_topic.arn
 
-    flush_interval_minutes = 15     // NOTE: this needs to be less than visibility timeout
-    max_processed_paths    = 100000 // NOTE: SQS in flight limit is 120k
-    max_batch_size         = 40
+    flush_interval_minutes = 10    // NOTE: this needs to be less than visibility timeout
+    max_processed_paths    = 40000 // NOTE: SQS in flight limit is 120k
+    max_batch_size         = 25
```
jamesgorrie, line 31: Out of interest, why have we chose these numbers?

warrd: It's a trade off between a few different things, if we lower the first 2 numbers we mean less latency in the pipeline but put more load on the relation embedder. The relation embedder was processing messages faster than they were coming before so decided to lower these to see if we could push it a bit harder / lower the latency. The bottom parameter has a more complicated effect, it will result in more but less complicated search queries in the relation embedder, I am just trying this to see if I can reduce search response time from the cluster.

warrd: Taking these numbers back to before as pushed it too hard.

alicefuzier (approved):

PR #1138, opened 2020-12-01
Look up works from pipeline storage in the matcher

alexwlchan:
Another step towards wellcomecollection/platform#4897; follows https://github.com/wellcomecollection/catalogue/pull/1136

This patch modifies the matcher to read IDs from the transformer, and look those IDs up in the pipeline storage Elasticsearch index. The recorder is still hanging around, but its output doesn't go anywhere.

Screenshot 2020-12-01 at 13 03 18

The next step is to update the merger to read works from pipeline storage, which I'll do in a separate patch.


alicefuzier:
I don't really understand from this PR if the matcher output has changed? If I remember correctly it sends identifiers with their versions, but versions no longer make sense, or do they?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/MemoryRetriever.scala
```diff
@@ -1,8 +1,10 @@
 package uk.ac.wellcome.pipeline_storage
 
+import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
 
-class MemoryRetriever[T](index: Map[String, T] = Map.empty)(
+class MemoryRetriever[T](val index: mutable.Map[String, T] =
+                           mutable.Map[String, T]())(
```
warrd, line 7: nit: I think it makes more sense for this data to be immutable, we always know what is going inside it on creation in the tests. Not a big issue of course, just would confuse be a bit (a bit unexpected, unlike `MemoryIndexer` where the data actually changes through it's behaviour)

warrd (approved):

alexwlchan:
> I don't really understand from this PR if the matcher output has changed? If I remember correctly it sends identifiers with their versions, but versions no longer make sense, or do they?

Good question, I should have clarified: the matcher output hasn't changed.

We do care about versions in the _merger_, so we want to preserve this output. e.g. if the matcher says _"merge Av1 and Bv2"_, and the merger is only able to retrieve Av2 and Bv3, it doesn't do any merging – it waits to hear from the matcher about the latest version of everything.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/MemoryRetriever.scala
alexwlchan: Yeah, I’d definitely like to come back and fix this at some point. It was easier to keep this PR manageable – previously we'd write stuff into a VHS as we were going along.

PR #1140, opened 2020-12-01
Remove the recorder from the pipeline and the repo

alexwlchan:
Another step towards wellcomecollection/platform#4897; follows #1139

This PR expunges the vast majority of recorder code from the repo, and in particular turns off the transformers actually writing to it:

Screenshot 2020-12-01 at 14 43 38

The recorder isn't completely gone; two remnants of it still linger:

* I'm not deleting the ECR repo yet
* I'm not deleting the script for adding an expiry policy to the recorder buckets

The latter will be useful for mopping up the last few buckets; the former we want to keep until the recorder is gone from all our deployments.


kenoir:
Do we need to route messages from the transformer to the matcher, or is that in another PR?

alexwlchan:
That was done a couple of PRs ago.

jamieparkinson:
This is what I like to see! 
image

warrd (approved):

PR #1146, opened 2020-12-02
Fix an issue with ElasticRetriever which was breaking the matcher

alexwlchan:
I discussed this with @warrd in Slack; a brief recap:

* The matcher tells the merger about all the works that _could_ be merged together, and the merger tries to retrieve all of them. This is done by the new SourceWorkLookup class (it uses Retriever[Source[Work]], and replaces the old RecorderPlaybackService. It returns a `Seq[Option[Work]]`, and whether the Option is filled in depends on whether:

     - The work is found
     - The work matches the version sent by the matcher

* The Retriever would previously throw an exception if _any_ of the IDs it was asked to retrieve didn't exist. This is wrong – the merger can use a partially filled in set of works, but only if it knows about them! So I changed the Retriever interface to include a list of:

     - Documents which were found
     - Documents which weren't found, and the throwable that meant they weren't there

* The ElasticRetriever had an additional bug where it would misinterpret documents that weren't in the index, and try to decode them rather than reading `found = false`. This was being masked by the `Retriever.apply()` method, which Nick noticed was swallowing the original exception.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
@@ -40,8 +40,15 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
           // original IDs.
           // See https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-multi-get.html
           val documents = result.docs
-            .map { _.safeTo[T] }
             .zip(ids)
+            .map { case (getResponse, id) =>
+              if (getResponse.found) {
+                id -> getResponse.safeTo[T]
+              } else {
+                id -> Failure(new RetrieverNotFoundException(id))
+              }
+            }
+            .toMap
 
           RetrieverMultiResult(
             found = documents.collect { case (id, Success(t)) => id -> t },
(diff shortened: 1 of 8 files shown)


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
```diff
@@ -40,17 +40,19 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
           // original IDs.
           // See https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-multi-get.html
           val documents = result.docs
-            .map { _.safeTo[T] }
             .zip(ids)
```
warrd, line 43: Are we sure these always return in the order requested?

warrd (approved):

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
alexwlchan: GitHub has clipped the comment; yes:

```
// Documents are guaranteed to be returned in the same order as the
// original IDs.
// See https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-multi-get.html
```

warrd (approved):

PR #1147, opened 2020-12-02
More efficient use of the Retriever in SourceWorkLookup

alexwlchan:
Follows #1146 

The matcher gives the merger a collection of identifiers which could be merged together. These identifiers may have a version, but it's not required. (I'm not entirely sure why this happens, but it's how the models are set up. I think possibly if a source record says “match this other source identifier”, and the other identifier isn’t in the pipeline yet?)

Screenshot 2020-12-02 at 11 34 42

If an identifier doesn't have a version, we can't use it in the merger – the merger will only combine the exact set of identifiers/versions it was given by the matcher.

Screenshot 2020-12-02 at 11 33 44

If we know we're not going to use it, then it's pointless looking up the work associated with this identifier – we'll discard it as soon as it's retrieved. We used to do this in the RecorderPlaybackService, but it got dropped when I added bulk retrievals. This puts it back.

Value: unsure. It will definitely reduce the amount of data we shuttle between us and Elasticsearch, but I don't know by how much.

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/SourceWorkLookup.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/SourceWorkLookup.scala
@@ -21,11 +21,27 @@ class SourceWorkLookup(retriever: Retriever[Work[Source]])(
     )
 
     retriever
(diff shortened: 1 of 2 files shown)


warrd (approved):

warrd:
I think my changes to #1146 does this anyway, so closing this

PR #1151, opened 2020-12-03
Denormalise works after they have been identified

warrd:
https://github.com/wellcomecollection/platform/issues/4907

The terraform is in this PR: https://github.com/wellcomecollection/catalogue/pull/1158

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
@@ -112,13 +112,13 @@ case class WorkData[State <: DataState](
   *      ▼
   *   Merged
   *      |
-  *      | (relation embedder)
-  *      ▼
-  * Denormalised
-  *      |
   *      | (id minter)
   *      ▼
   *  Identified
+  *      |
+  *      | (relation embedder)
+  *      ▼
+  * Denormalised
   *      |
   *      | (ingestor)
   *      ▼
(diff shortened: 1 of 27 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -228,21 +234,27 @@ object WorkFsm {
...
-    def state(state: Merged,
-              data: WorkData[DataState.Unidentified],
-              relations: Relations[DataState.Unidentified]): Denormalised =
-      Denormalised(state.sourceIdentifier, state.modifiedTime, relations)
-
-    def data(data: WorkData[DataState.Unidentified]) = data
-
-    def redirect(redirect: IdState.Identifiable) = redirect
-  }
+  implicit val identifiedToDenormalised =
```
alicefuzier, line 237: why there's no `mergedToIdentified`?

alicefuzier: ah ignore me. Because the idminter is work model agnostic

warrd: Because the `id_minter` does that transition, and its magical / untyped so doesnt need a `Transition[_, _]`

warrd: Yup, snap

alicefuzier (approved):

PR #1152, opened 2020-12-03
Use pipeline_storage in works id minter / ingestor

warrd:
https://github.com/wellcomecollection/platform/issues/4907


pipeline/id_minter/id_minter_works/src/main/scala/uk/ac/wellcome/platform/id_minter_works/Main.scala
```diff
@@ -37,10 +44,17 @@ object Main extends WellcomeTypesafeApp {
       )
     )
 
+    val workIndexer = ElasticIndexerBuilder[Work[Identified]](
+      config,
+      namespace = "denormalised-works",
```
alexwlchan, line 49: This namespace is wrong, and I don't see any application config.

warrd: :+1: will fix that. I was going to wait until #1151 to modify `application.conf` and terraform, as we are not going to deploy this as is, and things will change a fair bit more in that PR.

warrd: https://github.com/wellcomecollection/catalogue/pull/1152/commits/f136592563926473a54b78a12c453fecef047da3

alicefuzier (approved):

PR #1154, opened 2020-12-04
Transformer sends deleted Mets works as deleted works

alicefuzier:
For https://github.com/wellcomecollection/platform/issues/4893


pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsData.scala
```diff
@@ -15,28 +15,40 @@ case class MetsData(
...
-        )
-      )
+             modifiedTime: Instant): Either[Throwable, Work[Source]] = {
+    deleted match {
+      case true =>
+        Right(
+          Work.Deleted[Source](
+            version,
+            state = Source(sourceIdentifier, modifiedTime),
+            None))
```
alexwlchan, line 30: Is this the deletion reason? Let's fill it in with some useful info, and put a keyword argument here so the code is easier to read.

alexwlchan (approved):

PR #1159, opened 2020-12-07
Use a state FSM in the image model

jamieparkinson:
https://github.com/wellcomecollection/catalogue/pull/861 and https://github.com/wellcomecollection/catalogue/pull/889 introduced an ADT representation (`WorkState`) of a finite state machine for representing the changes in a work as it progresses through the pipeline. 

This introduces the same concept for images, and as part of doing that no longer attaches them directly to works - instead works now contain `ImageData`, from which an `Image[Initial]` is created in the merger. 

There is very little here that departs from the work state model - although I did keep the 2 state representations completely distinct, and resisted the desire to add some sort of new shared state type. The state diagram is:

```
      |
      | (merger)
      ▼
    Initial
      |
      | (image id minter)
      ▼
  Identified
      |
      | (inferrer)
      ▼
  Augmented
      |
      | (ingestor)
      ▼
   Indexed
```
Like with works, `Indexed` holds derived data - in this case, the thumbnail (aka the iiif-image location in `locations`).

The majority of changes in this PR are related to updating tests, and the most interesting parts are just:
- [the new image class, state, and transition typeclass.](https://github.com/wellcomecollection/catalogue/compare/image-state?expand=1#diff-4319def0a3de29602e5d5e2483e41b510b521c7838613a09fb9cf4ba6f6377b7R11)
- [the change to work data.](https://github.com/wellcomecollection/catalogue/compare/image-state?expand=1#diff-392d4159297494dd464cb120a082e26fc55aa3d855242e33cd47cc27112b6119R97)


api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesService.scala
```diff
@@ -18,21 +18,21 @@ class ImagesService(searchService: ElasticsearchService,
   private val nVisuallySimilarImages = 5
   private val imagesRequestBuilder = new ImagesRequestBuilder(queryConfig)
 
-  def findImageById(id: String)(
-    index: Index): Future[Either[ElasticError, Option[AugmentedImage]]] =
+  def findImageById(id: String)(index: Index)
+    : Future[Either[ElasticError, Option[Image[ImageState.Indexed]]]] =
```
warrd, line 22: In the `WorkState` stuff we tend to import the ADT types so we can write `Work[Denormalised]` etc, which I find reads a bit nicer and is easier to parse in code with lots of type parameters around. Do you think that would get confusing doing that for `ImageState` when possible (where there are no name clashes), or do you think it would get confusing due to having some of the same names?

jamieparkinson: I actually started off importing the types and then changed to this, because as you say it got confusing due to name clashes.

warrd: :+1: that's annoying, ah well can't have it all

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ImagesIndexConfig.scala
```diff
@@ -7,18 +7,23 @@ import com.sksamuel.elastic4s.requests.mappings.dynamictemplate.DynamicMapping
 
 object ImagesIndexConfig extends IndexConfig with WorksIndexConfigFields {
```
warrd, line 8: Do we need separate index config per state :thinking:? Suppose not required until we move images part of the pipeline to pipeline storage

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -2,70 +2,140 @@ package uk.ac.wellcome.models.work.internal
...
+) {
+  def id: String = state.id
+  def sourceIdentifier: SourceIdentifier = state.sourceIdentifier
+
+  def transition[OutState <: ImageState](args: OutState#TransitionArgs = ())(
+    implicit transition: ImageFsm.Transition[State, OutState],
+    // The transition helper method does not allow transitions across source
+    // DataState boundaries
+    sourceEqualityWitness: ImageSource[State#SourceDataState] =:= ImageSource[
+      OutState#SourceDataState]
```
warrd, line 26: Eh what's this syntax, guess I have to read up on it

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ImagesService.scala
jamesgorrie: Is this true? [Searching through the code](https://github.com/wellcomecollection/catalogue/search?p=2&q=WorkState.) we seem to use a mixture of both?

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedImageData.scala
```diff
@@ -0,0 +1,18 @@
...
+  thumbnail: DigitalLocationDeprecated
+)
+
+object DerivedImageData {
+  def apply(image: Image[_]): DerivedImageData =
+    DerivedImageData(
+      thumbnail = image.locations
+        .find(_.locationType.id == "iiif-image")
+        .getOrElse(
+          // This should never happen
```
jamesgorrie, line 13: If this is the case why do we account for it?

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ImagesIndexConfig.scala
jamieparkinson: Yeah that was my logic

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamesgorrie: Is this saying that the `transition` method is only available iif `ImageSource[State#SourceDataState]` is equal to `ImageSource[OutState#SourceDataState]`.

So this is checking that source work is in the same state... could you give an example of where this might not be the case?

jamesgorrie (commented):
This is looking good real good to see this pattern evolving and being propagated.

Conflicts though

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamieparkinson: This is very cool imo, as it's technically not new syntax - it's infix syntax for a typeclass that looks like `Eq[A, B]` and for which the language provides an implicit for any `Eq[A, A]`. I liked [this](https://stackoverflow.com/a/22717040) explanation, or [here](https://www.scala-lang.org/api/current/scala/$eq$colon$eq.html) are the docs, or additionally I found the cats documentation on [their more powerful alternative](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/evidence/Is.scala) quite interesting too

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedImageData.scala
jamieparkinson: Yeah, I can't think of a reason but if it happens then we really want to know and this custom error message is more helpful that what a plain `get` would throw. We do this elsewhere (sometimes with pattern matching) and it's been useful.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
jamesgorrie: That explanation is great, I found [this one](https://twitter.github.io/scala_school/advanced-types.html#otherbounds) less verbose and quite useful too.

Just checking my understanding, but it's constraining the source work's `DataType` to ensure that they are the same.

Is this so that you can transition `ImageState.Identified[WorkState.Identified] => ImageState.Augmented[WorkState.Identified]` but not `ImageState.Identified[WorkState.Identified] => ImageState.Augmented[WorkState.Unidentified]`?

jamieparkinson: Exactly - it removes the need for an identity method for `source` on the `Transition` instances like we do [here for works](https://github.com/wellcomecollection/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala#L217).

I didn't change it on works because this PR was big enough already, but I believe that this approach provides the exact same conversion via the same mechanism (ie implicit typeclass resolution), just (a) using a language builtin and (b) using implicit conversion of the `source` so it's invisible.

jamieparkinson: Sorry @jamesgorrie missed your question in your first comment - the only time where the `DataState` would change would be from `Unidentified -> Identified` inside the `ImageState` transition `Initial -> Identified`, and we don't and won't provide a `Transition` for this because it happens automagically inside the ID minter (it's opaque to the services either side).

jamesgorrie (approved):

PR #1177, opened 2020-12-09
Add a SourcePayload model for the adapters to send

alexwlchan:
Part of wellcomecollection/platform#4918

These models reflect the rows in our DynamoDB tables.

My thinking on what’s next:

* Have the reindexer parse and send records as these typed models (using the new source info provided in #1176)
* Have the adapters start to send these models, where appropriate

This means we'll know we're getting a consistent model in the transformers.

The transformers won't notice the difference yet – these models all have the same fields as `Version[String, Int]`, which is what the transformers expect right now.

### DIFF
--- a/.sbt_metadata/reindex_worker.json
+++ b/.sbt_metadata/reindex_worker.json
@@ -2,6 +2,7 @@
   "id" : "reindex_worker",
   "folder" : "reindexer/reindex_worker",
   "dependencyIds" : [
+    "internal_model",
     "big_messaging_typesafe"
   ]
 }
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/pipeline/SourcePayload.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/pipeline/SourcePayload.scala
@@ -1,0 +1,37 @@
+package uk.ac.wellcome.models.pipeline
+
+import uk.ac.wellcome.storage.s3.S3ObjectLocation
+
+sealed trait SourcePayload {
+  val id: String
+  val version: Int
+}
+
+case class CalmSourcePayload(
+  id: String,
+  version: Int,
+  location: S3ObjectLocation
+) extends SourcePayload
+
+case class MetsSourcePayload(
+  id: String,
+  version: Int,
+  payload: MetsSourceData
+) extends SourcePayload {
+  val sourceData: MetsSourceData = payload
+}
+
+case class MiroSourcePayload(
+  id: String,
+  version: Int,
+  location: S3ObjectLocation,
+  isClearedForCatalogueAPI: Boolean
+) extends SourcePayload
+
+case class SierraSourcePayload(
+  id: String,
+  version: Int,
+  payload: S3ObjectLocation
+) extends SourcePayload {
+  val location: S3ObjectLocation = payload
+}
--- a/build.sbt
+++ b/build.sbt
@@ -168,7 +168,7 @@ lazy val batcher = setupProject(
 lazy val reindex_worker = setupProject(
   project,
   "reindexer/reindex_worker",
-  localDependencies = Seq(big_messaging_typesafe),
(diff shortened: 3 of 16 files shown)


.sbt_metadata/reindex_worker.json
```diff
@@ -2,6 +2,7 @@
   "id" : "reindex_worker",
   "folder" : "reindexer/reindex_worker",
   "dependencyIds" : [
+    "internal_model",
```
alicefuzier, line 5: could we create a new module/library instead of shoving the adpater/transformer models into the `internal model`?
I don't think the adapters/reindexer need to know about the `Work` model and the other stages of the pipeline don't need to know about the source data. I think it's best to keep them separate

common/internal_model/src/main/scala/uk/ac/wellcome/models/pipeline/SourcePayload.scala
```diff
@@ -0,0 +1,37 @@
+package uk.ac.wellcome.models.pipeline
+
+import uk.ac.wellcome.storage.s3.S3ObjectLocation
+
+sealed trait SourcePayload {
+  val id: String
+  val version: Int
+}
```
warrd, line 8: nit: I am a bit unsure why this is defined as a common interface / ADT rather than parameterically:

```scala
case class SourcePayload[T] {
  id: String,
  version: Int,
  payload: T
```

Both ways will work of couse, but the way it is defined currently indicates to me scenarios where we receive a generic `SourcePayload` and do something on it at runtime (like pattern match on it, or pull out the `id` or `version` defined on the base interface), where our transformers always know at compile time what sort of messages it receives.

alicefuzier: This makes sense to me, although looking at the classes it looks like the Miro data wouldn't fit this model and would need a migration. Probably worth doing anyway and use just a single case class with type parameter

alexwlchan: The problem is that the DynamoDB rows don’t all fit that pattern. My plan is to validate, but pass along the JSON unmodified. The reindexer output won’t change at all, but we’ll know it can be deserialised as the appropriate type.

I should ditch the sealed trait; it’s not adding anything.

warrd: Oh you mean it doesn't fit if we want this to exactly match the columns defined in Dynamo? It would fit if we transformed each row first, but that would require the reindexer to know about these types (which I don't know if that is the plan?)

alexwlchan: I think it’ll make sense when I put it alongside the reindexer patch, but I didn’t finish it in time.

warrd: Yep that makes more sense to me now :+1: . Sealed trait might still be useful if we want the `TransformerWorker` to receive some `Payload <: SourcePayload`.

alexwlchan: It could fit if we transformed each row first, yeah; I’m not sure if that adds much value IMO. You’d add a bunch of extra code to rename a couple of fields.

.sbt_metadata/reindex_worker.json
kenoir: Does that mean the adapter/reindexer services could also be in a separate repo?

alexwlchan: I don’t think so. The point of this work is to unify the transformers, adapters, and reindexer. If anything, it makes it harder to split two of them out.

alexwlchan:
Closing in favour of other, more up-to-date PRs.

PR #1181, opened 2020-12-10
Id minter after transformers

alicefuzier:
This is the current pipeline layout:
Current_pipeline
The change in this PR brings it to this:
new_pipeline
We avoid two instances of the id minter and we avoid sending merged/redirected works to the id minter which are all redundant work


kenoir:
Can we get a sketch of what the pipeline looks like after this change?

alicefuzier:
This is the current pipeline layout:
Current_pipeline
The change in this PR brings it to this:
new_pipeline
We avoid two instances of the id minter and we avoid sending merged/redirected works which are all redundant work

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
```diff
@@ -68,13 +68,6 @@ object ImageState {
...
-  ) extends ImageState {
-    type SourceDataState = DataState.Unidentified
-    type TransitionArgs = Unit
-  }
-
-  case class Identified(
-    sourceIdentifier: SourceIdentifier,
     canonicalId: String
   ) extends ImageState {
     type SourceDataState = DataState.Identified
```
jamieparkinson, line 73: Can we now remove `SourceDataState` entirely (as it's always `Identified`)? Maybe best done in a subsequent PR, I'd be happy to do that if you want

alicefuzier:
Another PR will follow up on this one to get rid of the images id minter

jamieparkinson (approved):
This was remarkably straightforward 🎉

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Image.scala
alicefuzier: Ah yeah makes sense, I didn't spot that. Best in a different PR though, this one is big enough. I can do that if you want

jamieparkinson: Yeah go for it - I think it will get rid of the most unpleasant aspects of the image state stuff

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
```diff
@@ -211,18 +211,18 @@ object SourceWorkIndexConfig extends WorksIndexConfig {
...
-
-object IdentifiedWorkIndexConfig extends WorksIndexConfig {
-
   val fields = Seq(
     keywordField("type"),
     objectField("data").fields(
       collectionPath(analyzedPath)
     )
   )
+  val dynamicMapping = DynamicMapping.False
```
jamesgorrie, line 220: Might have missed the conversation - but why were we changing this?

jamesgorrie (commented):
From my POV - this definitely makes the pipeline easier to understand!

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
warrd: It has been `false` for the pipeline storage indices for a while, although I think we should change it for the other index too (was about to set up a discussion for this).

jamesgorrie: The main reason is that we then cannot query on certain fields, and we have more than just the API that uses the index (tools / reporting etc etc).

PR #1182, opened 2020-12-11
Reduce size of relation data

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4929

## Description

Currently for each relation we store a blob of json, which due to using `WorkData` always contains lots of empty data:

```json
{
  "data": {
    "title": "Murchison, Charles (1830-1879)",
    "otherIdentifiers": [],
    "mergeCandidates": [],
    "alternativeTitles": [],
    "subjects": [],
    "genres": [],
    "contributors": [],
    "production": [],
    "languages": [],
    "notes": [],
    "items": [],
    "collectionPath": {
      "path": "MS3652",
      "level": {
        "type":"Collection"
      },
      "label": "MSS.3652, 3653"
    },
    "imageData": [],
    "workType": "Collection"
  },
  "id": {
    "canonicalId": "daeah2n8",
    "sourceIdentifier": {
      "identifierType": {
        "id": "calm-record-id",
        "label": "Calm RecordIdentifier"
      },
      "ontologyType": "SourceIdentifier",
      "value": "bca287b5-6647-435b-b4a4-be7a4dcbc996"
    },
    "otherIdentifiers": []
  },
  "depth": 0,
  "numChildren": 2,
  "numDescendents": 2
}
```

In this PR we store relation data directly on the `Relation` model, rather than using `WorkData`, leading to much smaller json for the equivilent data:

```json
{
  "title": "Murchison, Charles (1830-1879)",
  "collectionPath": {
    "path": "MS3652",
    "label": "MSS.3652, 3653"
  },
  "workType": "Collection",
  "id": "daeah2n8",
  "depth": 0,
  "numChildren": 2,
  "numDescendents": 2
}
```

The original json (without any whitespace) is 617 bytes, whereas it is now 187 bytes, so a saving of approximately 70%. Currently a full index of works with relations takes up ~200GB of space (with 195GB of that being due to stored relations), so I estimate after this PR hopefully that will be more like 65GB. 

Note that the `collectionPath.level` field has also been removed, which is covered by `workType` already, and we do not store the source identifier on relations as these are not needed by the front end currently.


alicefuzier:
Could be good to change the shards for the catalogue and denormalise indices as well in this PR maybe?In #1185 I set those indices to have 5 shards but sounds like now they would need only 2?

warrd:
@alicefuzier done

alicefuzier (approved):

common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayRelation.scala
```diff
@@ -0,0 +1,48 @@
...
+    description =
+      "The title or other short label of a work, including labels not present in the actual work or item but applied by the cataloguer for the purposes of search or description."
+  ) title: Option[String],
+  @Schema(
+    `type` = "String",
+    description =
+      "The identifier used by researchers to cite or refer to a work."
+  ) referenceNumber: Option[String] = None,
+  @Schema(
+    `type` = "List[Work]",
```
jamieparkinson, line 23: Should this be `List[DisplayRelation]`?

warrd: Well it wouldn't be `DisplayRelation`, as this is a public thing and `Display*`is internal naming.

At the moment as far as the front end is concerned this is just a work but with less fields serialised, as it was before. Maybe up to @jtweed if we want to introduce a separate "relation" concept into the API (which is currently just a pipeline / internal model thing).

jamieparkinson: OK, got it. Bit worried that this makes our exposed API types unsound - if we're saying that these are works, but effectively making some non-optional fields into optional fields, how are we communicating that to consumers?

warrd: The only fields which are missing are optional, but I agree we could accidentally get in a situation where we break this assumption.

jamieparkinson: > The only fields which are missing are optional
Is this true? I think it's missing the only 2 non-optional fields: `alternativeTitles` and `availableOnline`

warrd: Ah yep you are right. Before this PR these two pieces of data would be included but were actually incorrect data I think (always empty / false).

jamieparkinson: Is there not a way that we can make those correct? It seems like a bad idea to maintain 2 different models which differ only in whether a couple of fields are optional; if we were able to make them correct then we could just use a `DisplayWork` full of `None`s for the optional values. 

I'm a bit averse to making `availableOnline` an `Option[Boolean]` because of course there is a notion of whether it is available online, regardless of whether we've worked it out or not, but making it optional would be another solution

warrd: So we could have a method to create a `DisplayWork` from a `Relation`, but there is no way to generate `availableOnline` and `alternativeTitles` unless we store these on each relation (assuming they are non optional).

The whole point of this PR and lots of the troubles of getting a pipeline running nicely is due to the size of relations, and as these fields on relations aren't used by the front end it is kind of necessary to avoid indexing them. To put things in context, a single reindex and reharvest (which didn't even complete) resulted in 2.5 TB of data flowing across the NAT gateway, and the adding of relations increased the index from 5 GB (`works-merged`) to 200 GB (``works-denormalised`).

jamieparkinson: I understand the reason but I think that this separate model introduces a reasonably large inconsistency/complexity into our display model (ie the API, this is a breaking API change) - making those 2 fields optional would've been no worse than this, imo. We now have to maintain 2 different display models - I think we should either be explicit about that or try to avoid doing it.

warrd: :+1: happy to change the display work model to make those fields optional, agree maybe that would be the best solution for now.

kenoir: Actually think we might need some input from @jtweed on this as it involves the structure of the user-facing API. I think I prefer making `availableOnline` and `alternativeTitles` optional and creating a `DisplayWork` from a `Relation` over exposing (or hiding) the `Relation` type.

jamieparkinson: I think it's important that we maintain the semantics of denormalised works, or be sure about whether we want to change them. I'm partly influenced by just having done the types for the archive stuff in https://github.com/wellcomecollection/wellcomecollection.org/pull/5884 - the case of what fields are present is already confusing enough from a client perspective!

warrd: I don't actually think the choice between these options affect the user facing API, the end result is the same: root level works have `alternativeTitles` and `availableOnline` always, and related works have these fields optionally.

I think we are all on the same page that internally having a single `DisplayWork` mode is preferablel to having both `DisplayWork` and `DisplayRelation` (which is actually a work as far as front end is concerned).

I will put in a new PR.

PR #1184, opened 2020-12-14
Use types inside the reindexer

alexwlchan:
A first step towards https://github.com/wellcomecollection/platform/issues/4918

This patch introduces a new `ReindexPayload` type, which corresponds to the structure of the rows in DynamoDB. Rather than passively passing through whatever's in DynamoDB, the reindexer will read values of this type from DynamoDB (`CalmReindexPayload`, `MetsReindexPayload`, …), and send *those* to the transformers.

This patch doesn't change the output of the reindexer. In particular, the reindexer continues to send the `{"id": "…", "version": …}` format required by the current transformers. There's still a 1:1 mapping between what's in DynamoDB and what's being sent by the reindexer – but this opens the door to start transforming these types into a common model used by the adapters and reindexers. That will happen in a future patch.

The `ReindexPayload` type is private to the reindexer. The long term plan is to introduce another type  `SourcePayload`.

* The reindexer would transform instances of ReindexPayload to SourcePayload before sending
* The adapter would send instances of SourcePayload
* The transformers would know how to read SourcePayload

Screenshot 2020-12-14 at 11 03 04


build.sbt
```diff
@@ -168,7 +168,7 @@ lazy val batcher = setupProject(
 lazy val reindex_worker = setupProject(
   project,
   "reindexer/reindex_worker",
-  localDependencies = Seq(big_messaging_typesafe),
+  localDependencies = Seq(mets_adapter),
```
alicefuzier, line 171: Why does the reindex worker depend on the mets adapter?

alexwlchan: For the MetsSourceData model.

I can move it to a separate library if you'd like? I was thinking maybe `pipeline_model`?

alicefuzier: ah yes, I see. I think we need a new library for your other PR too but whether it's done in this PR or the other, I don't mind. `pipeline_model` makes it sound like it should be shared within all pipeline services and instead it's something used only by services that deal with source data. Maybe `source_model`?`source_data_model`?

alexwlchan: Yes, `source_model` is nicer. It will also need MetsSourceData, so I'll add it in the next PR.

reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/models/ReindexSource.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import scala.util.{Failure, Success}
+
+sealed trait ReindexSource
+
+object ReindexSource {
+  case object Calm extends ReindexSource
+  case object Mets extends ReindexSource
+  case object Miro extends ReindexSource
+  case object MiroInventory extends ReindexSource
```
alicefuzier, line 13: what's the difference between miro and miro inventory?

alexwlchan: Miro = the metadata about Miro images, i.e. what gets fed to the transformer

Miro inventory = the metadata about where Miro images are stored. This only goes to the reporting cluster. I only discovered it when writing a test that we could parse the existing reindexer config.

alicefuzier (approved):
Looks good to me, nice!

PR #1185, opened 2020-12-14
Shard indices with relations

alicefuzier:
Resolves https://github.com/wellcomecollection/platform/issues/4930

### DIFF
--- a/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
+++ b/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
@@ -14,6 +14,7 @@ import uk.ac.wellcome.elasticsearch.WorksAnalysis._
 trait IndexConfig {
   def mapping: MappingDefinition
   def analysis: Analysis
+  def shards: Int = 1
 }
 
 /** Mixin for common fields used within an IndexConfig in our internal models.
--- a/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
+++ b/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
@@ -242,6 +242,14 @@ object DenormalisedWorkIndexConfig extends WorksIndexConfig {
   )
 
   val dynamicMapping = DynamicMapping.False
+
+  /**
+    * Denormalised relations make index sizes explode from about 5gb without relations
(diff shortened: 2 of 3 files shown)


common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
```diff
@@ -14,6 +14,7 @@ import uk.ac.wellcome.elasticsearch.WorksAnalysis._
 trait IndexConfig {
   def mapping: MappingDefinition
   def analysis: Analysis
+  def shards: Int = 1
```
alexwlchan, line 17: Is there a reason this is `def` rather than `val`?

alexwlchan (approved):
Looks good. Nice comments in WorksIndexConfig. 👍

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
alicefuzier: Not really. I never remember variables initialization order when they're defined in a trait and overridden in extending classes. Using `def` just feels a little bit safer cause I don't need to know that order since a def is always re-evaluated when called. Where it's overridden in `WorksIndexConfig` it's defined as a val

warrd: Also in general making things `val` in the base interface over constrains things (can't make dynamic in the implementer if required, always has to be fixed at creation time)

warrd:
This appears to be overridden just below :thinking: 

https://github.com/wellcomecollection/catalogue/blob/02d003419978dc7cc5d463bc31bb5a82248113ff/common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala#L39-L50

common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndexConfig.scala
```diff
@@ -272,4 +280,12 @@ object IndexedWorkIndexConfig extends WorksIndexConfig {
         keywordField("info")
       )
     )
+
+  /**
+    * Denormalised relations make index sizes explode from about 5gb without relations
```
warrd, line 285: It is worth mentioning that ~5GB on the denormalised index becomes ~25GB in the catalogue index due the mapping of all the extra fields.

PR #1186, opened 2020-12-14
Add a SourcePayload model

alexwlchan:
The next step towards wellcomecollection/platform#4918; follows https://github.com/wellcomecollection/catalogue/pull/1184

This patch introduces a new SourcePayload model, which encapsulates the data which should be carried between the adapter/reindexer and the transformer. I've added it to the reindexer and the METS adapter; getting it into the Calm and Sierra adapters is going to be trickier (because I need to do some introspection in the index stores).

They'll come in a separate PR.


common/source_model/src/main/scala/weco/catalogue/source_model/mets/MetsSourceData.scala
```diff
@@ -1,9 +1,9 @@
-package uk.ac.wellcome.mets_adapter.models
-
-import java.time.Instant
+package weco.catalogue.source_model.mets
```
warrd, line 1: :+1:

warrd (approved):

PR #1191, opened 2020-12-15
Allow enabling of gzip compression for ES requests

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4929

## Description

This allows enabling compression of requests sent to Elasticsearch, on a per client basis.

This will potentially be useful for reducing NAT gateway costs for indexing works in the pipeline, although it will increase CPU load both on our containers and the ES cluster.

Before this can be merged:
- [x] Elastic4s needs a release >= `7.10.1`.
- [x] `EnrichConfig` in scala-libs should be updated to have a `getBooleanOption` method.
- [ ] Should check ES clusters that we are sending compressed data to are configured to have compression enabled.

### DIFF
--- a/common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/test/fixtures/ElasticsearchFixtures.scala
+++ b/common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/test/fixtures/ElasticsearchFixtures.scala
@@ -14,7 +14,7 @@ import com.sksamuel.elastic4s.requests.get.GetResponse
 import com.sksamuel.elastic4s.requests.indexes.IndexResponse
 import com.sksamuel.elastic4s.requests.indexes.admin.IndexExistsResponse
 import com.sksamuel.elastic4s.requests.searches.SearchResponse
-import com.sksamuel.elastic4s.{ElasticClient, Index, Response}
+import com.sksamuel.elastic4s.{Index, Response}
 import grizzled.slf4j.Logging
 import io.circe.{Decoder, Encoder, Json}
 import io.circe.parser.parse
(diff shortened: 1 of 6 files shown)


alicefuzier:
Is this the update we are waiting for https://github.com/sksamuel/elastic4s/pull/2269?

warrd:
Yep but doesn't appear like they have release that yet: https://github.com/sksamuel/elastic4s/tags

alexwlchan:
elastic4s now has a release 7.10.1: https://search.maven.org/artifact/com.sksamuel.elastic4s/elastic4s-core_2.12/7.10.1/jar

common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/test/fixtures/ElasticsearchFixtures.scala
```diff
@@ -241,7 +251,22 @@ trait ElasticsearchFixtures
...
+    elasticClientWithCompression
+      .execute {
+        indexInto(index.name).doc(doc)
+      }
+      .map { r =>
+        if (r.isError) {
+          error(s"Error from Elasticsearch: $r")
+        }
+        r
+      }
```
alexwlchan, line 269: I would inline this method and the definition of elasticClientWithCompression in the test case, so it's easier to see what that test doing. It seems unlikely we'd reuse this client. Otherwise, looks good.

alexwlchan (approved):

common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/test/fixtures/ElasticsearchFixtures.scala
warrd: I think fine as the `elasticClientWithCompression` is lazy, and we don't want to recreate the client lots of times if for example the call to this method was wrapped in an `eventually`

PR #1192, opened 2020-12-15
Remove image id_minter

alicefuzier:


pipeline/terraform/ecr.tf
```diff
@@ -18,12 +18,8 @@ resource "aws_ecr_repository" "ecr_repository_transformer_calm" {
   name = "uk.ac.wellcome/transformer_calm"
 }
 
-resource "aws_ecr_repository" "ecr_repository_id_minter_works" {
-  name = "uk.ac.wellcome/id_minter_works"
-}
-
-resource "aws_ecr_repository" "ecr_repository_id_minter_images" {
-  name = "uk.ac.wellcome/id_minter_images"
```
alexwlchan, line 26: Is this going to break the existing prod pipeline when you apply it? Services won't be able to pull images.

alicefuzier: argh, yes. Thank you good spot

kenoir (approved):
Looks good, bar @alexwlchan's comment about the ECR repo.

pipeline/terraform/ecr.tf
alexwlchan: Yeah, I’m also wondering if changing the order of services/messaging will break the pipeline?

alicefuzier: I hope not. The same thing was done to move the id minter from after the relation embedder to before. We have to target the apply but other than that it shouldn't have an impact on prod? (Also, the changing of order is already merged in master. this PR is just deleting the image id minter)

PR #1194, opened 2020-12-15
Upgrade to terraform 0.14 in the Calm adapter; allow scaling to zero

alexwlchan:
I don't think there's a good reason the Calm adapter can't scale to zero when it's not in use; if there is, we should document it in the Terraform.


kenoir (approved):
Looks fine. It is suspicious that this isn't scalable to 0 now - really makes me feel there is a reason lost to time. I guess we'll find out.

PR #1195, opened 2020-12-15
Lazily send works from the transformers

alexwlchan:
This patch introduces a new object `WorkComparison` which decides if a Work is "interesting" compared to a previously stored version. (Admittedly, I don't love the name.)

* If the transformed Work is newer than the Work already in the pipeline and has different data, then it is "interesting"
* If the transformed Work is older than the Work already in the pipeline, then it is "not interesting"
* If the transformed Work has the same ~date~ data as the Work already in the pipeline, then it is "not interesting"

If Work is not interesting, then we don't need to send it to the rest of the pipeline.

This should reduce the amount of processing we need to do for the overnight to Sierra harvest. Although it's quite difficult to determine if two arbitrary blobs of JSON from the Sierra API are "meaningfully different", it's pretty easy with our structured Work model.

Part of https://github.com/wellcomecollection/platform/issues/4934


alexwlchan:
Currently this is transformer-specific. We might want to push it into PipelineStorage if we think this would be useful behaviour to have elsewhere in the pipeline.

jamieparkinson:
Possibly stupid question: does ES document versioning (like we do for the catalogue index) not do this for us?

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
```diff
@@ -89,11 +97,38 @@ trait TransformerWorker[Payload <: SourcePayload, SourceData, SenderDest]
         StoreReadError(err, p)
       }
 
+  private def compareToStored(workResult: Result[(Work[Source], StoreKey)]): Future[Result[(Option[Work[Source]], StoreKey)]] =
+    workResult match {
+
+      // Once we've transformed the Work, we query forward -- is this a work we've
```
kenoir, line 104: This is really neat way of doing the comparison that we can do now we store the work at this point.

alexwlchan:
Looking at this in a PR, I think the abstraction is wrong. This should go in PipelineStorage behind a flag – that documents get indexed isn’t a detail that the transformer worker should care about.

alexwlchan:
> does ES document versioning (like we do for the catalogue index) not do this for us?

Not in the way we want.

Elasticsearch will stop you replacing V3 with V2, but it will let you replace any V2 with any V3.

with this comparison system, we would only replace V2 with V3 if V3 contains interesting new data. If not, we wait for a V4 that has interesting changes.

kenoir:
Worth noting there will be a cost associated with retrieving the works from ES even if we don't send them down the pipeline. I expect that's less than the cost incurred by sending them upstream though.

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
alexwlchan: The other neat thing about it is that the "interesting" data is pushed into WorkData, so the comparison is easy.

_"Are these two case classes the same"_ vs _"Are these two case classes the same (except this one field)"_

jamieparkinson:
> Not in the way we want.

I'm wondering what the tradeoff between redundant updates using ES versioning and the extra work of doing the check ourself is...

alexwlchan:
> Worth noting there will be a cost associated with retrieving the works from ES for comparison even if we don't send them down the pipeline. I expect that's less than the cost incurred by sending them upstream though.

Rough back of a napkin calculation:

* After a Work goes from the transformer, it gets retrieved at least four times (matcher, merger, ID minter, ingestor), and indexed five times (transformer + those four). This doesn't account for amplification from the relation embedder.
* We get ~250k new records from Sierra each night. How much might we save?

    Let's be pessimistic and say only 10% of them are "not interesting". That means we'd save ~75k get requests to Elasticsearch and ~125k index requests.

    If it's 50%, you save ~375k get requests and ~625k index requests.

    It it's 90%, you save ~675k get requests and ~1.1M index requests.

* This process will add ~2.5M get requests to every reindex.

    So if 50% of updates from Sierra each night can be dropped, we'd expect this to be strictly better if we reindex less than once a week. That's probably conservative, and it likely pays off faster.

alexwlchan:
> I'm wondering what the tradeoff between redundant updates using ES versioning and the extra work of doing the check ourself is... Any easy way to estimate?

I'm not sure I understand what you're suggesting; if we do redundant updates with ES versioning, I don't think we can tell it was redundant?

alicefuzier:
> If the transformed Work has the same date as the Work already in the pipeline, then it is "not interesting"

Could this be a problem potentially for sources like calm where the date changes only once a day? Or does it change more often, I can't quite remember

alexwlchan:
> Could this be a problem potentially for sources like calm where the date changes only once a day? Or does it change more often, I can't quite remember

possibly, except this is a typo for “data”

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/WorkComparison.scala
```diff
@@ -0,0 +1,30 @@
...
+  def isNewer(storedWork: Work[Source], work: Work[Source]): Boolean = {
+    assert(
+      storedWork.id == work.id,
+      s"Cannot compare two works with different IDs: ${storedWork.id} != ${work.id}"
+    )
+
+    // If the stored work has a higher version, then it automatically wins.
+    if (storedWork.version > work.version) {
+      false
+    }
```
warrd, line 23: :thinking: I am a little hazy on remembering where this version actually comes from, but doesn't it get bumped every time it is ingested into the adapter, and thus will be greater even when the data is the same (such as during a reharvest)?

alexwlchan: Yeah, so I don’t think with this branch very often in practice

jamieparkinson:
To clarify my question:

>(1) If the transformed Work is newer than the Work already in the pipeline and has different data, then it is "interesting"
(2) If the transformed Work is older than the Work already in the pipeline, then it is "not interesting"
(3) If the transformed Work has the same date as the Work already in the pipeline, then it is "not interesting"

Of these 3 points, aren't we already doing (2) with ES versioning using `modifiedTime`, and couldn't we do (3) now by changing the versioning from `external_gte` to `external`?

So I'm wondering, how much of a gain do we get from implementing (1) given that we would have to do all those extra GETs? What percentage of the redundant work lies in (1) compared to (2) and (3)?

alexwlchan:
> Of these 3 points, aren't we already doing (2) with ES versioning using modifiedTime, and couldn't we do (3) now by changing the versioning from external_gte to external?

I don’t think so. I also don’t think I’ve explained it badly, because it seems to be causing confusion.

This logic goes above Elasticsearch by dropping updates in the following case: *the stored Work has the same data and the same or lower version as the transformed Work*.

The version may be bumped by updates in the source system that we don’t care about. For example, the nightly Sierra reharvest. In such a case, the transformer will drop the update – even if the version is newer than what is stored, because the data hasn’t changed.

jamieparkinson:
Does the nightly reharvest bump the `modifiedDate` in the bib record (which becomes the `modifiedTime`, which is used as the ES external version)? I didn't think it did but could be wrong - if it doesn't change, I think we should have a look at changing `external_gte` to `external` as an alternative to this.

alexwlchan:
> Does the nightly reharvest bump the modifiedDate in the bib record?

I’m about 98% sure it does.

Sierra sees it as an updated database record, with new data and so a new modification date.

Among other things, I think there’s a varfield in the Sierra data something like “last synced with Calm”. This counts as a record update in Sierra, so it gets a new modifiedDate in our pipeline.

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/WorkComparison.scala
warrd: Ah I misread the original code, I get it now

alexwlchan: I’m going to put some more comments in this code tomorrow, because it’s clearly not obvious right now.

alexwlchan:
I've pushed a new commit which expands the comment on this code, so hopefully it's easier to see what I'm trying to do here. I don't think we can achieve the same thing with Elasticsearch versioning; if it's not obvious from the current comments, let me know and I'll try to clarify further.

I've also tweaked the API to make the intent clearer. Now it goes:

```scala
if (newWork.shouldReplace(storedWork)) {
   ...
}
```

so it's more explicit that the transformer discards works which shouldn't replace their already-stored counterparts.

alicefuzier:
I haven't read the PR properly, I had exhausted my brain capacity yesterday. But conceptually I think @alexwlchan is right on this. Versioning alone won't help us in this case. `modifiedDate` and the VHS `version` both change so we have to look into the actual data to see if those changes are meaningful

warrd:
This relies on our transformers being completely deterministic in ordering of things etc, not sure of that is actually the case so something to hear in mind.

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
```diff
@@ -104,10 +148,16 @@ trait TransformerWorker[Payload <: SourcePayload, SourceData, SenderDest]
               case TransformerError(_, sourceData, key) =>
                 error(s"$name: TransformerError on $sourceData with $key")
             }
-            Future.failed(err)
-          case Right((work, key)) =>
+
+            throw err
+
+          case Right((None, key)) =>
+            info(s"$name: from $key transformed work; already in pipeline")
```
alicefuzier, line 155: nitpicking, but why do we return the `key` all the way to here only to log and discard it? can we not log in the `compareToStored` method and avoid returning? the return type of `compareToStored` is a bit confusing to me

alexwlchan:
> This relies on our transformers being completely deterministic in ordering of things etc, not sure of that is actually the case so something to bear in mind.

I’ve always understood the transformer output to be deterministic. If that’s not true, it’s a surprise and might scupper this – but I can’t think of any examples where it’s not?

alicefuzier (approved):
👏

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
alexwlchan: Good shout.

alicefuzier:
> This relies on our transformers being completely deterministic in ordering of things etc, not sure of that is actually the case so something to bear in mind

I think they should be? Only case that I can think of where that might be the case is where the source itself doesn't order stuff consistently which might be for sierra items? Although I think we order them by item id number

warrd:
It probably depends on the source systems. E.g. when there are multiple of the same varfields does Sierra guarantee to return them always in the same order? Maybe other things like that worth checking.

alexwlchan:
> Only case that I can think of where that might be the case is where the source itself doesn't order stuff consistently which might be for sierra items? Although I think we order them by item id number

We do.

> E.g. when there are multiple of the same varfields does Sierra guarantee to return them always in the same order?

I would be surprised if this wasn't the case. I'm pretty sure MARC field order is significant, at least in some organisations.

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
alexwlchan: Fixed.

alicefuzier:
> It probably depends on the source systems. E.g. when there are multiple of the same varfields does Sierra guarantee to return them always in the same order? Maybe other things like that worth checking.

I'm 80% sure that Sierra returns varFields in a consistent order whatever that may be (ie, the same item/bib unchanged will come back from the API with the fields in the same order, even though it's not immediately apparent what is the underlying ordering criteria). But also, a lot of things that are lists in our data model are extracted following a certain order. Like the first subject comes from a specific field and the other ones come from other fields. 
So _I think_ we're covered for Sierra (and METS and MIRO). A bit less sure about Calm

kenoir:
>I’ve always understood the transformer output to be deterministic. If that’s not true, it’s a surprise and might scupper this – but I can’t think of any examples where it’s not?

Yeah, i'd go further and say that transformer output *must* be deterministic, if it's not it's broken.

warrd:
> I'm 80% sure that Sierra returns varFields in a consistent order whatever that may be

:+1: hopefully that is the case, not worried about Calm too much as Sierra is the only thing that really matters for this.

> Yeah, i'd go further and say that transformer output must be deterministic, if it's not it's broken.

I am pretty sure all our transformer code is deterministic, should have phrased it originally more regarding the source systems.

PR #1205, opened 2020-12-18
Bump the CPU used by the METS transformer

alexwlchan:


kenoir (approved):

warrd:
Are we going to change container parallelism too?

alexwlchan:
Do we need to? Seems to do okay with just this change.

warrd:
👍

PR #1206, opened 2020-12-18
The Identified -> Merged argument can be an Instant, not an Option[Instant]

alexwlchan:
Spotted on code read.

The only place we call `transition[Merged](None)` is in tests. Be explicit that this transition will always provide a modifiedTime.


kenoir (approved):

jamieparkinson:
fwiw, this was in fact intentional -https://github.com/wellcomecollection/catalogue/pull/962#discussion_r509176496

PR #1209, opened 2020-12-18
Expose a batchRetrieveFlow on PipelineStorageStream

warrd:
Exposes a `batchRetrieveFlow` on `PipelineStorageStream`.

As this is not able to be used by all services that use the pipeline storage stream (such as in the case of transformers, which do not retieve any data from pipeline storage), this is exposed as a flow for now. In a future PR I will have a go at abstracting it so services do not need to use the flow directly, but may revert to that in services where it is complex to do in a universal manner.


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -95,3 +77,72 @@ class PipelineStorageStream[In, Out, MsgDestination](
       }
     )
 }
+
+object PipelineStorageStream extends Logging {
+
+  def batchRetrieveFlow[T](config: PipelineStorageConfig,
```
alicefuzier, line 83: Where is this being used?

warrd: Nowhere yet, I am going to use it in my next PR

alicefuzier (approved):

PR #1210, opened 2020-12-18
Trim trailing spaces from mergeCandidates bnumber

alicefuzier:
Some record with a sierra digital merge candidate have a trailing space. This causes the "impossible" id minter bug where the id minter is able to retrieve the canonical id for that sourceIdentifier, but not match back to that sourceIdentifier and add it to the json

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
@@ -6,5 +6,7 @@ case class SourceIdentifier(
   ontologyType: String,
   value: String
 ) {
+  assert(!value.endsWith(" "))
+
   override def toString = s"${identifierType.id}/$value"
 }
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
@@ -6,7 +6,7 @@ case class SourceIdentifier(
   ontologyType: String,
   value: String
 ) {
-  assert(!value.endsWith(" "))
+  require(value == value.trim, s"SourceIdentifier value contains trailing whitespace: <$value>")
 
   override def toString = s"${identifierType.id}/$value"
 }
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
@@ -6,7 +6,9 @@ case class SourceIdentifier(
   ontologyType: String,
   value: String
 ) {
-  require(value == value.trim, s"SourceIdentifier value contains whitespaces: <$value>")
+  require(
+    value == value.trim,
+    s"SourceIdentifier value contains whitespaces: <$value>")
 
   override def toString = s"${identifierType.id}/$value"
 }
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
@@ -6,7 +6,7 @@ case class SourceIdentifier(
   ontologyType: String,
   value: String
 ) {
-  require(value == value.trim, s"SourceIdentifier value contains trailing whitespace: <$value>")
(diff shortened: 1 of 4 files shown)


alexwlchan (commented):
Excellent debugging!

Is there ever a case of when we want trailing whitespace on an identifier value?

If not, is it worth throwing an assertion/require in SourceIdentifier that will tell us when an identifier value has whitespace? (Not actually trimming it, but forcing us to investigate if it arises again.)

alicefuzier:
That's a good point. The idminter gets unhappy if there are spaces so it's good to flag that up before it reaches the idminter

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
```diff
@@ -6,5 +6,7 @@ case class SourceIdentifier(
   ontologyType: String,
   value: String
 ) {
+  assert(!value.endsWith(" "))
```
alexwlchan, line 9: Couple of notes:

* Right now this will throw a generic Assertion failed message; I'd add a second argument that logs the 

    ```suggestion
      assert(..., s"Value contains trailing whitespace: <$value>")
    ```

* I would make the condition `value == value.trim`, so we handle all trailing whitespace (not just space, but tab/newline) and also leading whitespace.

* Minor: I prefer `require` to `assert` for this sort of thing. From the [Prede(f docs](https://www.scala-lang.org/api/current/scala/Predef$.html#require(requirement:Boolean,message:=>Any):Unit)

    > `require` is distinct from `assert` in that if the condition fails, then the caller of the function is to blame rather than a logical error having been made within [the function] itself.

    But that's just a preference, and I'm not going to fuss if you leave it as an assert.

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidatesTest.scala
```diff
@@ -41,6 +40,17 @@ class SierraMergeCandidatesTest
...
+    it("trims spaces at the end of the bnumber of the mergeCandidate") {
+      val sierraData = createSierraBibDataWith(
+        varFields = create776subfieldsWith(
+          ids = List(s"(UkLW)$mergeCandidateBibNumber ")
+        )
+      )
+
+      SierraMergeCandidates(sierraData) shouldBe
+        physicalAndDigitalSierraMergeCandidate(mergeCandidateBibNumber)
+    }
```
alexwlchan, line 52: Might be worth a comment that this is based on real Sierra data, possibly including an example.

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidatesTest.scala
```diff
@@ -262,38 +272,43 @@ class SierraMergeCandidatesTest
...
+    it("dedupes Calm IDs and adds as mergeCandidates") {
+      val calmIds = (1 to 5).map(_ => randomUUID.toString)
+
+      val bibData = bibDataWith035(calmIds ++ calmIds)
+
+      SierraMergeCandidates(bibData) shouldBe calmIds.map(
+        createCalmMergeCandidate)
+    }
+    it(
+      "creates clam merge candidates if it has a mix of calm and non calm identifiers") {
```
alexwlchan, line 299:
```suggestion
      "creates calm merge candidates if it has a mix of calm and non calm identifiers") {
```

alexwlchan (approved):
Really nice now I can read it on a screen bigger than a postage stamp. 😅 

A couple of suggestions around the assertion, otherwise LGTM.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/SourceIdentifier.scala
```diff
@@ -6,5 +6,9 @@ case class SourceIdentifier(
   ontologyType: String,
   value: String
 ) {
+  require(
+    value == value.trim,
+    s"SourceIdentifier value contains trailing whitespace: <$value>")
```
alexwlchan, line 11: nit:

```suggestion
    s"SourceIdentifier value contains extra whitespace: <$value>")
```

This assertion could fail if there's leading whitespace, e.g. `" b1234"`

alexwlchan (approved):
One tiny nit, but otherwise this looks really good. 👍

PR #1211, opened 2020-12-27
Parse a handful more dates for the date filter

alexwlchan:
I was using the site, and spotted a slightly odd looking date on a works page: **Feb.y 11th 1782** (https://wellcomecollection.org/works/kntad7tz)

Currently this work won't show up using date filters, because the parser doesn't recognise this month. This patch updates the date parser to recognise months like **Feb.y**, and other spellings that can be parsed unambiguously. It's a pretty small change given that current design of the parser, yay.

As part of this, I also did a little data wrangling to get a list of all the period labels currently in the catalogue API – to check there's nothing the parser crashes on, and to get an idea of what it currently *can't* parse.

Two discoveries of note:

* It already parses ~80% of labels. There's some more low-ish fruit (e.g., but there aren't substantial gains to be made here. See [unparseable_labels.txt](https://github.com/wellcomecollection/catalogue/files/5743917/unparseable_labels.txt)
* This change doesn't have a substantial impact on performance. On my very slow not-work laptop, it parses ~1M period labels in less than three minutes (about the same as before the patch).


jamesgorrie (commented):
A small fear here is that we are fixing faulty source data with the mispelloings rather than flagging it up and fixing at the source.

Are things like `mai` expected for some wonderful reason, or is it something that collections info should fix?

warrd:
`mai` is french isnt it?

alexwlchan:
> Are things like mai expected for some wonderful reason, or is it something that collections info should fix?

I think what we're seeing isn't misspellings, but month names in other languages, e.g. "mai" is French for May. AIUI, Collections Info will often transcribe the date as written on the document, rather than converting it directly into the Gregorian calendar.

PR #1215, opened 2021-01-05
Use fixed thread pool executor in METS transfomer

alicefuzier:
Next thing to try if #1214 doesn't fix issues with the mETS trasfomer

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/resources/application.conf
+++ b/pipeline/transformer/transformer_mets/src/main/resources/application.conf
@@ -1,3 +1,13 @@
+akka {
+  actor {
+    default-dispatcher {
+      executor = "thread-pool-executor"
+      thread-pool-executor {
+        fixed-pool-size = 50
+      }
+    }
+  }
+}
 aws.sqs.queue.url=${?transformer_queue_id}
 aws.metrics.namespace=${?metrics_namespace}
 


warrd (approved):

pipeline/transformer/transformer_mets/src/main/resources/application.conf
```diff
@@ -1,3 +1,13 @@
+akka {
+  actor {
+    default-dispatcher {
+      executor = "thread-pool-executor"
+      thread-pool-executor {
+        fixed-pool-size = 50
```
kenoir, line 6: What is this by default?

kenoir: Also be useful to describe how changing its value helps.

alicefuzier: It's the [fork join](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html) executor: https://doc.akka.io/docs/akka/current/general/configuration-reference.html

kenoir: Here is what _i_ think is going on: the thread pool size by default is _very large_ (`Integer.MAX_VALUE`), and the scheduler is creating threads to do work faster than it can actually do that work until it runs out of resources and can no longer schedule work.

kenoir: So for a `fork-join-executor` the max parallelism is `parallelism-max = 64` (from that link).

kenoir: So you are reducing the maximum number of threads from 64 to 50?

warrd: @alicefuzier any idea how big the `physicalFileIdsMapping` can get? [This line](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/transformer/transformer_mets/src/main/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsXml.scala#L140) could have potentially been creating **a lot** of new threads (one for each item in the mapping)

kenoir: @warrd What's is the maximum parallelism for `par` in that case? AFAIK it does not share the default-executor with AKKA.

jamieparkinson: fwiw I wrote the offending parallel map and also spent a lot of time reading about and configuring a similar thing (fixed thread pool) in the ID minter - `physicalFileIdsMapping` can get huge but the JVM won't keep creating threads without bound, as @kenoir says. The problem is almost certainly that `par` is using its own threadpool which Akka has not accounted for - I think removing it will work.

jamieparkinson: You could probably persuade it to use the same pool but I imagine it's not worth the effort https://docs.scala-lang.org/overviews/parallel-collections/configuration.html

jamieparkinson: As I say I would guess that this PR won't end up being necessary but as I remember the fixed size thread pool offers better throughput for blocking tasks than even an equivalently sized fork joining pool, because it is not "greedy" or work-stealing and will not choke itself in the same way

warrd: :+1: @jamieparkinson  ah I was confused as it seemed to be me in the git blame that made that line, but had no recollection of it. I have merged #1214 so hopefully that should help things.

kenoir (approved):

warrd:
@alicefuzier do we want this merged?

alicefuzier:
I think I'm going to close it without merging. Doesn't seem to be needed

warrd:
:thinking: are we sure its not needed? I thought the reason we got all the METS messages through was as we kept retrying it and eventually managed to get them all through?

alicefuzier:
We haven't tried your change removing the `par`. I think that's probably the cause for it dying as @jamieparkinson was saying, so definitely I would like to try that in isolation first

warrd:
Ah ok. Yep might be worth trying that on its own so we know its the actual cause.

PR #1218, opened 2021-01-05
Disable merging for Sierra audiovisual works

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4876; follows https://github.com/wellcomecollection/catalogue/pull/1217

I've created a new `isAudiovisual` Work predicate, which then gets used in two places:

* Inside Sources.findFirstLinkedDigitisedSierraWorkFor, when we work out what the digitised work that should be linked to a physical Sierra work is
* Inside the PartialRule in OtherIdentifiers

I don't think we can get away with just one – the findFirstLinked method is called both to copy across fields *and* to work out the list of "merged sources", and we need to skip audiovisual merging in both places.

There's a separate piece of this work to go through and rerun all the Sierra AV works through the merger once this change is deployed.


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/models/Sources.scala
```diff
@@ -2,17 +2,34 @@ package uk.ac.wellcome.platform.merger.models
...
+        // format and another for the digitised version, where available.
+        //
+        // These bibs routinely contain different data, and we can't consider one of
+        // them canonical, unlike other physical/digitised bib pairs.
+        //
+        // Longer term, this may change based if Collections Information change how
+        // they catalogue AV works.
+        //
+        // See https://github.com/wellcomecollection/platform/issues/4876
+        && !isAudiovisual(target)) {
```
jamieparkinson, line 25: It's annoying that this has to be in both this method and in the partial rules but it's a bit of a non-trivial refactor to resolve that so agreed that it's probably simpler to keep it like this

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Format.scala
```diff
@@ -98,12 +100,12 @@ object Format extends Enum[Format] {
     override val label: String = "Mixed materials"
   }
 
-  case object Audio extends Unlinked {
+  case object Audio extends Unlinked with Audiovisual {
```
jamieparkinson, line 103: How come the sealed trait in the model > maintaining a list of AV formats in the merger?

jamieparkinson (approved):

jamieparkinson:
Ooh actually one more comment - could you add a test case to `MergerFeatureTest` for this?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/models/Sources.scala
alexwlchan: That said, it's pretty easy to determine you need to do this when making changes – I wrote the obvious test case in PlatformMergerTest, and it failed until I updated both locations. It's not a subtle failure.

kenoir (approved):

PR #1222, opened 2021-01-06
Support one to many operation in pipeline storage stream

alicefuzier:
Resolves https://github.com/wellcomecollection/platform/issues/4936

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
@@ -16,7 +16,7 @@ case class PipelineStorageConfig(batchSize: Int,
                                  flushInterval: FiniteDuration,
                                  parallelism: Int)
 
-case class Bundle[T](message: Message, item: T)
+case class Bundle[T](message: Message, item: T, numberOfItems: Int)
 
 class PipelineStorageStream[In, Out, MsgDestination](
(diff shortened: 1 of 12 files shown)


alexwlchan (approved):
This looks reasonable to me, but I'm not super au fait with the Akka flows. You might want a second opinion.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -133,12 +134,37 @@ object PipelineStorageStream extends Logging {
     indexer: Indexer[T])(implicit
                          ec: ExecutionContext,
                          indexable: Indexable[T]) =
-    Flow[(Message, Option[T])]
-      .collect { case (msg, Some(document)) => Bundle(msg, document) }
+    Flow[(Message, List[T])]
+      .collect {
+        case (msg, items @ _ :: _) =>
```
warrd, line 139: This is for filtering out empty items right :thinking: ?

warrd (approved):

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
alicefuzier: Yes, cause empty lists are dealt with in the other branch of the broad cast, the identityFlow

PR #1225, opened 2021-01-06
[WIP] Migrate Calm works to have the correct ontology type

warrd:
- [x] Add a migration for updating the Calm works to have the `Work` ontology type.
- [ ] Test this against a dump of the data (will this query catch anything which shouldn't be a work?)
- [ ] Work out how migrations are run, and do anything else required to run it


warrd:
@alexwlchan applied this to the db manually so no need for migration

PR #1230, opened 2021-01-07
First pass at speeding up ECS deployment

jamieparkinson:
This decreases some timeouts/healthcheck intervals to match (or close to match, as TCP target groups only have a limited number of valid settings) those we use in wc.org.

It also appears that akka-http ignores SIGTERMs and the ECS agent waits an additional 30 seconds to issue a SIGKILL, so I've upgraded akka-http in order to add [coordinated shutdown](https://doc.akka.io/docs/akka-http/current/server-side/graceful-termination.html) which I think should resolve that (and can't hurt if it doesn't).

A weird side-effect of this is that I had to update the IntelliJ Scala plugin in order for the syntax highlighting to agree that the changes to the parameter directives were valid...


alexwlchan (approved):

jamieparkinson:
This is a massive pain to do while keeping continuous uptime so I'm going to close it as it's marginal gains

PR #1236, opened 2021-01-07
Store WorkData on deleted works

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4952

## Description

This is required so deleted works get routed to the batcher, and thus related works correctly remove it from their relations.


alicefuzier (approved):
Thanks!

PR #1239, opened 2021-01-08
Changes for a reindex on 2021-01-08

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4948

### DIFF
--- a/pipeline/terraform/stack/service_id_minter.tf
+++ b/pipeline/terraform/stack/service_id_minter.tf
@@ -77,7 +77,7 @@ module "id_minter" {
 module "id_minter_topic" {
   source = "../modules/topic"
 
-  name       = "${local.namespace_hyphen}_work_id_minter"
+  name       = "${local.namespace_hyphen}_id_minter_output"
   role_names = [module.id_minter.task_role_name]
 
   messages_bucket_arn = aws_s3_bucket.messages.arn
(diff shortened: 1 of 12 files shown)


pipeline/terraform/stack/service_id_minter.tf
```diff
@@ -77,7 +77,7 @@ module "id_minter" {
 module "id_minter_topic" {
   source = "../modules/topic"
 
-  name       = "${local.namespace_hyphen}_work_id_minter"
+  name       = "${local.namespace_hyphen}_id_minter_output"
```
jamesgorrie, line 80: Why this change?

pipeline/terraform/stack/service_merger.tf
```diff
@@ -7,7 +7,7 @@ module "merger_queue" {
   alarm_topic_arn = var.dlq_alarm_arn
 
   # This has to be longer than the flush interval in the merger
```
jamesgorrie, line 9:
```suggestion
  # This has to be longer than the `flush_interval_seconds` in the merger
```
❓

jamesgorrie (approved):
Really nice clear comments on why the changes are being made!

pipeline/terraform/stack/service_id_minter.tf
alexwlchan: The old name comes from a time when we had separate ID minters for images and works; that’s no longer the case.

And although “output” is implicit from this being SNS, we do have topics that aren’t the output of any app and being explicit about that is useful.

jamesgorrie (approved):

PR #1240, opened 2021-01-08
partOf work filter in API

warrd:
https://github.com/wellcomecollection/platform/issues/4956

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/MustQuery.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/MustQuery.scala
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.platform.api.models
 
-sealed trait MustQuery
-sealed trait ImageMustQuery extends MustQuery
+sealed trait WorkMustQuery
+
+sealed trait ImageMustQuery
 
 case class ColorMustQuery(hexColors: Seq[String]) extends ImageMustQuery
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala
@@ -181,8 +181,6 @@ object WorksRequestBuilder
           field = "data.items.locations.locationType.id",
           values = itemLocationTypeIds)
       case PartOfFilter(id) =>
-        termQuery(
-          field = "state.relations.ancestors.id",
-          value = id)
+        termQuery(field = "state.relations.ancestors.id", value = id)
     }
 }
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala
@@ -180,5 +180,9 @@ object WorksRequestBuilder
         termsQuery(
           field = "data.items.locations.locationType.id",
           values = itemLocationTypeIds)
+      case PartOfFilter(id) =>
+        termQuery(
+          field = "state.relations.ancestors.id",
+          value = id)
     }
 }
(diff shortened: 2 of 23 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/MustQuery.scala
```diff
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.platform.api.models
 
-sealed trait MustQuery
-sealed trait ImageMustQuery extends MustQuery
+sealed trait WorkMustQuery
```
alexwlchan, line 3: Where is this trait used? I see it in a bunch of function signatures, but I don’t see an example of something that actually extends it (and surely that would have to be in this file, since it's a sealed trait).

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala
```diff
@@ -176,5 +180,7 @@ object WorksRequestBuilder extends ElasticsearchRequestBuilder {
         termsQuery(
           field = "data.items.locations.locationType.id",
           values = itemLocationTypeIds)
+      case PartOfFilter(id) =>
+        termQuery(field = "state.relations.ancestors.id", value = id)
```
alexwlchan, line 184: Just to double check: the partOf display field is constructed from the `ancestors` field at query time, right? There's not an internal field called partOf we should do a terms query on instead?

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
```diff
@@ -222,6 +222,27 @@ trait WorkGenerators extends IdentifiersGenerators with InstantGenerators {
...
+    def ancestors(works: Work.Visible[Indexed]*): Work.Visible[Indexed] =
+      Work.Visible[Indexed](
+        work.version,
+        work.data,
+        updateState(
+          work.state.copy(
+            relations = work.state.relations.copy(
+              ancestors = works.toList.zipWithIndex.map {
+                case (work, idx) =>
+                  Relation(work, idx + 1, 1, works.length - idx)
```
alexwlchan, line 237: Keyword parameters would make it easier to see what this code is doing.

alexwlchan (approved):
Looks fine to me.

I think this is fine to deploy without a reindex, it's just that these filters will always be empty. Is that correct, or is there a breaking model change I'm not seeing?

api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksRequestBuilder.scala
warrd: Yes that's right: the reason `ancestors` isn't just called `partOf` is that `ancestors` is a flattened list, whilst `partOf` is nested work data structure

api/api/src/main/scala/uk/ac/wellcome/platform/api/models/MustQuery.scala
warrd: We need to provide two type parameters for this:

```scala
case class SearchOptions[DocumentFilter, MustQuery](
  ...
)
```

The `MustQuery` type is used for images and not for works. but we still need to provide something. I added a new type for this, as imo that is much clearer to read than passing in `Unit` or something everywhere, and in future we may actually extend from it (there are a number of other options here such as a type alias, or using an empty `object` type, but both also have downsides too)

warrd:
> I think this is fine to deploy without a reindex, it's just that these filters will always be empty. Is that correct, or is there a breaking model change I'm not seeing?

Yes deploying it will not break anything, however the functionality won't work until we reindex and thus the ancestors IDs are actually in the index mapping.

common/internal_model/src/test/scala/uk/ac/wellcome/models/work/generators/WorkGenerators.scala
warrd: Added keyword params

PR #1243, opened 2021-01-08
Lowercase b numbers in the METS transformer

alexwlchan:
The Sierra source exclusively returns lowercase identifiers (e.g. b1234). The METS source currently returns identifiers as mixed case (e.g. B1234).

This causes an issue for matching and ID minting, because pipeline identifiers are case sensitive:

*   These two works won't be recognised as the same and matched/merged
*   When the ID minter tries to find existing ID for "B1234", it does a case insensitive database search.  It finds the existing ID for "b1234", and gets confused because that's not what it asked for.

Because we know METS identifiers are always b numbers, let's lowercase them to match Sierra.

This is affecting three METS works. All three of them are invisible in the current prod API. The canonical ID of these works will change, but because the old ID was never publicly visible I don't think that's an issue.

* B20442178 / xsfkn7tk
* B20442506 / q9u4bnvx
* B20442233 / r5uw7g74

Part of https://github.com/wellcomecollection/platform/issues/4948

### DIFF
--- a/pipeline/transformer/transformer_mets/src/test/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsDataTest.scala
+++ b/pipeline/transformer/transformer_mets/src/test/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsDataTest.scala
@@ -1,7 +1,7 @@
 package uk.ac.wellcome.platform.transformer.mets.transformer
 
 import java.time.Instant
-import org.scalatest.Inside
+import org.scalatest.{EitherValues, Inside}
 import org.scalatest.funspec.AnyFunSpec
 import org.scalatest.matchers.should.Matchers
 import uk.ac.wellcome.models.work.internal._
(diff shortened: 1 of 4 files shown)


warrd (approved):

pipeline/transformer/transformer_mets/src/test/scala/uk/ac/wellcome/platform/transformer/mets/transformer/MetsDataTest.scala
```diff
@@ -458,4 +461,28 @@ class MetsDataTest
...
+
+    val work = data.toWork(version = 1, modifiedTime = Instant.now()).value
+
+    work.sourceIdentifier shouldBe SourceIdentifier(
+      identifierType = IdentifierType("mets"),
+      ontologyType = "Work",
+      value = "b1234"
+    )
+
+    val mergeCandidates = work.asInstanceOf[Work[Source]].data.mergeCandidates
```
warrd, line 478: Why is `asInstanceOf[..]` required here?

alexwlchan: Good spot; it isn't. I'll remove that.

PR #1245, opened 2021-01-11
Add a script for getting stats about a reindex

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4959

This shows you how far along a reindex is in terms of Elasticsearch documents. It shows us if the pipeline is leaky:
 
Screenshot 2021-01-11 at 08 19 49


jamesgorrie (approved):
Super nice. 

Just checking my understanding here, we expect there to always be as many source docs as indexed docs (but with different states)?

alexwlchan:
> Just checking my understanding here, we expect there to always be as many source docs as indexed docs (but with different states)?

Yup.

PR #1248, opened 2021-01-11
Use pipeline storage for images

warrd:
## Issue

https://github.com/wellcomecollection/platform/issues/4965

## Description

Updates the image section of the pipeline to use `pipeline_storage` rather than `big_messaging`.

**NOTE: due to the size of this PR here I just update the code itself, a future PR will be needed for terraform and configuration of environment variables.**

- [x] Merger
- [x] Inference manager
- [x] Image ingestor

### DIFF
--- a/.sbt_metadata/inference_manager.json
+++ b/.sbt_metadata/inference_manager.json
@@ -3,6 +3,6 @@
   "folder" : "pipeline/inferrer/inference_manager",
   "dependencyIds" : [
     "internal_model",
-    "big_messaging_typesafe"
+    "pipeline_storage_typesafe"
   ]
 }
--- a/.sbt_metadata/ingestor_common.json
+++ b/.sbt_metadata/ingestor_common.json
@@ -3,7 +3,6 @@
   "folder" : "pipeline/ingestor/ingestor_common",
   "dependencyIds" : [
     "elasticsearch_typesafe",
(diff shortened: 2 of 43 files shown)


kenoir (commented):
LGTM. It occurs to me that this has the potential to significantly increase NAT gateway costs when a reindex is run with it - perhaps we should hold off merging until we are happy that has been mitigated by the PrivateLink changes?

warrd:
@kenoir yes definitely would be expected to increase NAT costs, though would hopefully reduce some costs elsewhere. The only issue I have about not merging is that conflicts are likely to pile up as this touches quite a lot of different parts of the code. I am only around for 2 more weeks so might be pretty frustrating for whoever has to fix them.

warrd:
Also worth mentioning that the pipeline storage cluster is already beyond its limit of what it can index, so we probably would need to bump it up or slow things down (e.g. less tasks per service) for this too.

kenoir:
> I have about not merging is that conflicts are likely to pile up as this touches quite a lot of different parts of the code. 

I agree that's an issue. I guess the question to ask is whether master is "deployable" after this change is merged. If we do merge it will put us in a position where a full reindex will likely not work, but that doesn't mean this shouldn't be in master so that we can test it more effectively.

kenoir (approved):

PR #1249, opened 2021-01-12
Merger batching fix

alicefuzier:
This is largely what was in #1222 . The interesting addition is closing the substreams once they've finished as explained [here](https://github.com/wellcomecollection/catalogue/pull/1249/files#diff-33efef025bb293b86b89b608f4f0619e1e031dafb16e6bbdd66d949d388070a6R176)

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
@@ -143,14 +143,14 @@ object PipelineStorageStream extends Logging {
       }
       .mapConcat[Bundle[T]](identity)
       .via(batchIndexFlow(config, indexer))
-      .via(takeSuccessfulListOfBundles(maxSubStreams, 5 minutes))
+      .via(takeCompleteListOfBundles(maxSubStreams, 5 minutes))
       .mapConcat(identity)
       .mapAsyncUnordered(config.parallelism) { bundle =>
         for {
           _ <- Future.fromTry(msgSender.send(indexable.id(bundle.item)))
         } yield bundle
       }
-      .via(takeSuccessfulListOfBundles[T](maxSubStreams, 5 minutes)
+      .via(takeCompleteListOfBundles[T](maxSubStreams, 5 minutes)
         .collect {
           case head :: _ => head.message
         }
@@ -161,7 +161,7 @@ object PipelineStorageStream extends Logging {
   // Each substream emits one message with the complete list of bundles for the same messageId
   // or no message if it didn't receive the correct number of bundles
   // The result is a flow of list of _complete_ bundles
-  def takeSuccessfulListOfBundles[T](maxSubStreams: Int, t: FiniteDuration): Flow[Bundle[T], List[Bundle[T]], NotUsed] = {
+  def takeCompleteListOfBundles[T](maxSubStreams: Int, t: FiniteDuration): Flow[Bundle[T], List[Bundle[T]], NotUsed] = {
     val groupByMessage = Flow[Bundle[T]]
       .groupBy(maxSubstreams = maxSubStreams, f = _.message.messageId(), allowClosedSubstreamRecreation = true)
       .scan(Nil: List[Bundle[T]]) {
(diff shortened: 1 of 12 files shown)


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -1,22 +1,22 @@
...
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future, TimeoutException}
 
 case class PipelineStorageConfig(batchSize: Int,
                                  flushInterval: FiniteDuration,
                                  parallelism: Int)
 
-case class Bundle[T](message: Message, item: T)
+case class Bundle[T](message: Message, item: T, numberOfItems: Int)
```
kenoir, line 19: If T were type constrained to an Iterable, would you be able to drop `numberOfItems`?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -132,14 +133,53 @@ object PipelineStorageStream extends Logging {
...
+        case (bundleList, b) => b :: bundleList
+      }
+      .filter { list =>
+        list.nonEmpty && list
+          .map(_.item)
+          .distinct
+          .size == list.head.numberOfItems
       }
+    // Close the substream with take if it succeeds
+    // or after timeout if it fails
```
kenoir, line 176: How does this resolve the problem that we saw before?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
alicefuzier: I don't think so...
the problem is a single message can result in many documents to ingest and send. We want to batch those independently to how many were related to the same original message, but we want to delete the message only when we know the ingest (and sending) was successful.

To ingest them independently of the original message we squash them together with documents reulting from other messages, but to delete the message when all of them have ingested successfully, we need to count how many we get at the end. That's what `numberOfItems` is used for. Before we squash them, we take note of how many documents per eacch message we have to ingest. After we've ingested, we count how many were successful. If the number match it's fine

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
alicefuzier: The problem was that there's a maximum number of substreams that you can have. I was setting it to Integer.MAX_VALUE, but because I was never closing them, eventually it would be unable to open substreams

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -132,13 +133,54 @@ object PipelineStorageStream extends Logging {
...
+      }
+      .via(groupByMessage[T](maxSubStreams, 5 minutes).collect {
+        case head :: _ => head.message
+      }.mergeSubstreams)
+  }
+
+  // Splits the flow into a substream for each messageId.
+  // Each substream emits one message with the complete list of bundles for the same messageId
+  // or no message if it didn't receive the correct number of bundles
+  def groupByMessage[T](maxSubStreams: Int, t: FiniteDuration) =
```
warrd, line 161: I think it would be easier to read this help having the full function signature including output type here

alicefuzier: I can add it if you like, but the result signature is a bit weird cause it's a subflow. I think it's more confusing than anything but if you think it clarifies then sure

warrd: It was mainly so I could see the output type, leave it out if you think really bad

alicefuzier: I've added it, let me know what you think. I don't find a way to simplify it

alicefuzier: Scratch everything I said...this can output a flow and the signature becomes much better

alicefuzier:
Are there any more comments on this @warrd and @kenoir ? Is it ok to be merged?

warrd (approved):

kenoir (approved):

PR #1253, opened 2021-01-12
Improve the error messages in the merger if you send in an invalid message

alexwlchan:
I was sending some hand-rolled messages to the merger this evening, patching up a reindex. The merger takes a collection of `WorkIdentifier` instances, which is defined as:

```scala
case class WorkIdentifier(identifier: String, version: Option[Int])
```

In my naïveté, I sent the following WorkIdentifier to the merger: `WorkIdentifier(id = "qwkttzg2", version = None)`. I promptly saw a null pointer exception coming from the ElasticRetriever in the merger logs, which was both confusing and wrong. We pretty much never want to be throwing a null pointer error – it indicates a higher-level bug somewhere else, and we should catch it and explain what's actually gone wrong.

Digging in, I discovered that the ElasticRetriever throws a null pointer error if you ask it to retrieve an empty list of IDs. I've replaced that error with an assertion that throws a more explicable error message (*"You should never look up an empty list of IDs!"*). If it turns out there is some scenario when this is useful, it'll be easier to debug.

But why was the merger retrieving an empty list of IDs? Because it only asks the ElasticRetriever to find Works whose WorkIdentifier has a non-empty `version`.

I had to stare at the matcher code for quite a while to convince myself that yes, sometimes this field will be empty and that's okay, and also that the matcher should always include at least one non-empty version in the results it sends. I've added some comments to the matcher graph code so it's easier to follow, and an assertion/comment in the merger to catch this as well.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
@@ -20,7 +20,8 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
     extends Retriever[T]
     with Logging {
 
-  override final def apply(ids: Seq[String]): Future[RetrieverMultiResult[T]] = {
+  override final def apply(
+    ids: Seq[String]): Future[RetrieverMultiResult[T]] = {
     assert(
       ids.nonEmpty,
       "You should never look up an empty list of IDs!"
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
@@ -20,7 +20,12 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
     extends Retriever[T]
     with Logging {
 
-  override final def apply(ids: Seq[String]): Future[RetrieverMultiResult[T]] =
+  override final def apply(ids: Seq[String]): Future[RetrieverMultiResult[T]] = {
+    assert(
+      ids.nonEmpty,
+      "You should never look up an empty list of IDs!"
+    )
+
     client
       .execute {
         multiget(
@@ -59,4 +64,5 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
             notFound = documents.collect { case (id, Failure(e)) => id -> e }
           )
       }
+  }
 }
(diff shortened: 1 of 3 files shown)


alexwlchan:
This is failing because of a flaky test that's fixed in https://github.com/wellcomecollection/catalogue/pull/1254

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
```diff
@@ -20,7 +20,13 @@ class ElasticRetriever[T](client: ElasticClient, index: Index)(
     extends Retriever[T]
     with Logging {
 
-  override final def apply(ids: Seq[String]): Future[RetrieverMultiResult[T]] =
+  override final def apply(
+    ids: Seq[String]): Future[RetrieverMultiResult[T]] = {
```
jamesgorrie, line 24: Just as a thought experiments, as I know that it would have larger knock-on effects, but could this be a [`NonEmptyList`](https://typelevel.org/cats/datatypes/nel.html). We use them [in the merger for lists of work](https://github.com/wellcomecollection/catalogue/blob/48be8c35d1e0299c1ab43e225fa6b164b328c9df/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/FieldMergeRule.scala#L48) and would encapsulate this requirement in the type system.

kenoir: Don't let stray `cats` in.

alicefuzier: I would say `NonEmptyList` makes sense. We already have cats every classpath

alexwlchan: Maybe if I was embarking on a more full-blown refactor of the matcher, but I'm not sure I want to commit to that work right now?

alexwlchan:
I'm going to merge this as-is to get the matcher changes in; if somebody wants to come back and make this a NonEmptyList, let's do that separately.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticRetriever.scala
alicefuzier: yeah. to be honest , on second thought, you still have to create a `NonEmptyList` from a `List` which means checking the `List` elsewhere and having that duplicated code in all usages of `ElasticRetriever` which is probably annoying

jamesgorrie: > Maybe if I was embarking on a more full-blown refactor of the matcher, but I'm not sure I want to commit to that work right now?

Indeedy, hence "thought experiment" - glad we dabbled.

PR #1255, opened 2021-01-13
Build inference manager apps in docker-compose

alexwlchan:
Nick was having issues starting the docker-compose in the inference manager tests, because he doesn't run tests through Make, and the docker-compose relies on images that are built locally.

You can tell docker-compose "build this image from a local Dockerfile", so this patch does that – which slightly reduces our dependency on Make.

I'd like some eyes on this from @jamieparkinson when he's feeling better, as he wrote pretty much all this code and will know if I'm missing something.


warrd (approved):

alexwlchan:
Closing as doesn't work – as Jamie says in Slack, this doesn't play well with sbt.

PR #1259, opened 2021-01-13
Provide more helpful logging messages when running a snapshot

alexwlchan:
Currently, the snapshot generator is repeatedly logging:

```
Received 10000 works from Index(works-2021-01-12)
```

This isn't especially helpful, because I can't easily work out how through a snapshot I am. This patch updates the message to tell you how many works have been retrieved so far, e.g.:

```
Received another 10,000 works (350,000 so far) from Index(index-zdhdqo)
```

We could also fetch the total number of works by getting the `_count` of the index from Elasticsearch, but that would be a more involved patch with futures and stuff. This is just a small patch to make the logging a bit more useful.


jamesgorrie (approved):
Neato

PR #1261, opened 2021-01-13
Add Catalogue pipeline storage Elastic Cloud cluster

kenoir:
This PR adds configuration for an Elastic Cloud cluster in terraform and AWS PrivateLink connection to the infrastructure critical stack.  

The intention is to reduce NAT Gateway costs and better manage infrastructure as code.

These changes add a new cluster, but do not attach it to a new pipeline, which should occur in an upcoming PR.

Part of: https://github.com/wellcomecollection/platform/issues/4949

### DIFF
--- a/infrastructure/critical/locals.tf
+++ b/infrastructure/critical/locals.tf
@@ -7,6 +7,11 @@ locals {
   public_subnets_new  = local.catalogue_vpcs["catalogue_vpc_delta_public_subnets"]
   private_subnets_new = local.catalogue_vpcs["catalogue_vpc_delta_private_subnets"]
 
+  catalogue_pipeline_ec_vpce_domain = "vpce.eu-west-1.aws.elastic-cloud.com"
+  catalogue_pipeline_ec_privatelink_host = "${ec_deployment.pipeline_storage.id}.${local.catalogue_pipeline_ec_vpce_domain}"
+
+  ec_eu_west_1_service_name = "com.amazonaws.vpce.eu-west-1.vpce-svc-01f2afe87944eb12b"
+
   admin_cidr_ingress = data.aws_ssm_parameter.admin_cidr_ingress.value
 
   read_principles = [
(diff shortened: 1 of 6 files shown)


infrastructure/critical/locals.tf
```diff
@@ -7,6 +7,13 @@ locals {
   public_subnets_new  = local.catalogue_vpcs["catalogue_vpc_delta_public_subnets"]
   private_subnets_new = local.catalogue_vpcs["catalogue_vpc_delta_private_subnets"]
 
+  catalogue_pipeline_ec_vpce_domain      = "vpce.eu-west-1.aws.elastic-cloud.com"
+  catalogue_pipeline_ec_privatelink_host = "${ec_deployment.pipeline_storage.id}.${local.catalogue_pipeline_ec_vpce_domain}"
```
jamesgorrie, line 11: I can't see where this is being used?

kenoir: It's an [output](https://github.com/wellcomecollection/catalogue/pull/1261/files#diff-81d64fabe28d970892cf46ff6bc61d5b576d4e7cd2dfb6ceb16e7c5e985d29abR40).

jamesgorrie: That it is - silly me.

kenoir: Services will use that address as the URI for their Elastic Cloud deployment, as it routes requests via the VPC endpoint (and the deployment ID tells Elastic Cloud the other side which deployment to reference).

jamesgorrie (approved):
What do we need locally / in CI / wherever to have this work? I thought you said there was some env vars? Is it worth documenting them if so?

kenoir:
> What do we need locally / in CI / wherever to have this work? I thought you said there was some env vars? Is it worth documenting them if so?

Good point. I haven't added that in this PR. I need to do that.

kenoir:
This is now applied.

PR #1262, opened 2021-01-13
Use the BatchWriteItem API to reduce DynamoDB traffic in the matcher

alexwlchan:
Right now, we write works to the graph database with an individual PutItem call per node. Whenever a single node in a component gets updated, we re-write *every node in the component*.

DynamoDB [pricing for on-demand capacity](https://aws.amazon.com/dynamodb/pricing/on-demand/) is based on the notion of "write request units" – every request is rounded up to the nearest kilobyte, and you pay per kilobyte of data written. Given the matcher entries are small, this means we're probably overpaying for writes.

Using the BatchWriteItem API means we waste way less capacity – we round up once for all the items in a component, rather than once per item. I would hope this reduces the amount of money we spend on DynamoDB in the matcher.

Part of https://github.com/wellcomecollection/platform/issues/4969

### DIFF
--- a/pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/Main.scala
+++ b/pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/Main.scala
@@ -5,6 +5,7 @@ import java.time.Duration
 import scala.concurrent.ExecutionContext
 import akka.actor.ActorSystem
 import com.typesafe.config.Config
+import org.scanamo.auto._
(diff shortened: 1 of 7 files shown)


kenoir (approved):
It'll be interesting to see if the pricing model applies like you've described or not - potentially AWS could still round up cost on each individual request within a batch. In my reading the wording of the linked docs does make it sound like the rounding is per API request though there is enough of ambiguity to make me unsure.

PR #1264, opened 2021-01-14
Only retrieve specific fields from Elasticsearch in the matcher

alexwlchan:
Follows #1263; this is the second part of and closes https://github.com/wellcomecollection/platform/issues/4971

This follows a similar pattern we use in the relation embedder – rather than fetching and decoding the whole Work, we fetch specific fields. This reduces traffic over the NAT Gateway – not a huge amount, but it's low-handing fruit we can easily pick.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticRetriever.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticRetriever.scala
@@ -25,7 +25,8 @@ trait ElasticRetriever[T] extends Retriever[T] with Logging {
   def createGetRequest(id: String): GetRequest
   def parseGetResponse(response: GetResponse): Try[T]
 
-  override final def apply(ids: Seq[String]): Future[RetrieverMultiResult[T]] = {
+  override final def apply(
+    ids: Seq[String]): Future[RetrieverMultiResult[T]] = {
     assert(
       ids.nonEmpty,
       "You should never look up an empty list of IDs!"
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticRetriever.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticRetriever.scala
@@ -1,8 +1,9 @@
-package uk.ac.wellcome.pipeline_storage
+package uk.ac.wellcome.pipeline_storage.elastic
 
-import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success}
+import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
 
+import com.sksamuel.elastic4s.requests.get.{GetRequest, GetResponse}
 import com.sksamuel.elastic4s.{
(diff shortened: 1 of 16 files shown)


warrd (approved):
Nice work

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticRetriever.scala
```diff
@@ -1,24 +1,29 @@
...
+  val client: ElasticClient
+  val index: Index
 
-class ElasticRetriever[T](client: ElasticClient, index: Index)(
-  implicit val ec: ExecutionContext,
-  decoder: Decoder[T])
-    extends Retriever[T]
-    with Logging {
+  def createGetRequest(id: String): GetRequest
+  def parseGetResponse(response: GetResponse): Try[T]
```
warrd, line 26: Couldn't we provide default implementations for these so this remains a `class` instead of a `trait`? I am not a fan of lots of this sort of inheritance generally, but I think it would be appropriate here and avoid the boiler plate of having to create an `ElasticSourceRetriever` for cases which don't need custom behaviour

pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/storage/elastic/ElasticWorkLinksRetriever.scala
```diff
@@ -1,29 +1,55 @@
...
+import scala.util.Try
 
-class ElasticWorkLinksRetriever(client: ElasticClient, index: Index)(
+/** The matcher only needs to know about linking data on works; most of the fields
+  * it can completely ignore.
+  *
+  * This custom retriever fetches only the specific fields that the matcher needs, to
+  * reduce the amount of data we have to get out of Elasticsearch (and pay to send through
+  * the AWS NAT Gateway).
+  */
```
jamesgorrie, line 21: I like it  either way as it's efficient, but would the [`PrivateLink`](https://github.com/wellcomecollection/catalogue/pull/1261) render the NAT gateway reason redundant?

alexwlchan: Yup.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticRetriever.scala
warrd: @alexwlchan did you miss this comment?

alexwlchan: Ah sorry. I considered it, but I'm not a super fan of classes extending classes tbh. We use the `trait … class` pattern everywhere else, and the difference between SourceRetriever and Retriever is very slight.

warrd: I have taken a look at the code again and I think I misread it the first time around, this is fine by me :+1:

PR #1269, opened 2021-01-14
Make pipeline storage es host configurable

kenoir:
In order to use the new pipeline storage ES cluster accessible via a VPC endpoint we need to make the ES host used by the pipeline stack configurable. Using this mechanism we should be able to spin up a new pipeline that uses the VPC endpoint ES host exported from the infrastructure/critical stack, while keeping the existing pipeline working.

Part of https://github.com/wellcomecollection/platform/issues/4949
Depends on https://github.com/wellcomecollection/catalogue/pull/1261

### DIFF
--- a/pipeline/terraform/stack/service_id_minter.tf
+++ b/pipeline/terraform/stack/service_id_minter.tf
@@ -44,7 +44,7 @@ module "id_minter" {
     db_username          = "catalogue/id_minter/rds_user"
     db_password          = "catalogue/id_minter/rds_password"
 
-    es_host     = "catalogue/pipeline_storage/es_host"
+    es_host     = var.pipeline_storage_es_host_secret_id
     es_port     = "catalogue/pipeline_storage/es_port"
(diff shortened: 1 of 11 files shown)


jamesgorrie (approved):
Given you've set the default I'm not sure it is dependant on the other PR?

Assuming from the description/implementation, the other stack updates the value of the host? When would it write to another key? Still agree that that stack should still expose the key as it's writing it, but I can't see in [that PR](https://github.com/wellcomecollection/catalogue/pull/1261/files) where that happens?

kenoir:
> Given you've set the default I'm not sure it is dependant on the other PR?

Only in the sense it's not useful without it.

> Assuming from the description/implementation, the other stack updates the value of the host? 

Not quite. We require a new pipeline to pull the value from the critical stack and set it for that pipeline. This PR enables that but does not add the new pipeline. 

Another PR will appear when i'm ready to run a reindex into that notional pipeline.

PR #1279, opened 2021-01-18
Reduce the number of objects fetched in a single Elasticsearch snapshot request

alexwlchan:
We're still seeing the same issues as https://github.com/wellcomecollection/platform/issues/4901


kenoir (approved):

alexwlchan:
Confirmed that 150 works gets a successful snapshot. 👍

PR #1288, opened 2021-01-19
Setup SNS topic for image ingestor

warrd:


jamesgorrie (approved):

pipeline/terraform/stack/service_image_ingestor.tf
```diff
@@ -61,6 +62,15 @@ module "ingestor_images" {
   shared_logging_secrets = var.shared_logging_secrets
 }
 
+module "image_ingestor_topic" {
+  source = "../modules/topic"
+
+  name       = "${local.namespace_hyphen}_image_ingestor"
```
alexwlchan, line 68:
```suggestion
  name       = "${local.namespace_hyphen}_image_ingestor_output"
```

I know it being an output is implied by it being an SNS topic, but I find it useful to distinguish between topics that are the output of a service, and the result of something like a script.

PR #1289, opened 2021-01-19
Use HSV colour space in palette inferrer

jamieparkinson:
This changes the palette inferrer to cluster and bin colours in HSV space rather than RGB - at the cost of some additional complexity, this should make colour filtering behave more as we expect. We might need to come back and tweak some parameters once it's done.

The colour bins look a bit like this:

image


harrisonpim (approved):

alicefuzier (approved):
I don't really understand much, but the scala bit looks all right to me

PR #1291, opened 2021-01-19
Tweak the rules around the lazy sending of Works in the transformer

alexwlchan:
After discussion with Robert and Alice in Slack, I realised that there is an issue with the transformer sending works.

## The problem

In the 2021-01-13 reindex, the ID minter input queue received ~28k less messages than there are works in the works-source index. That means the transformer successfully indexed those works into pipeline_storage, but it didn't send them to SNS and they didn't land on a DLQ.

This seems to be the rough timeline of events:

* Transformer picks up a message from the queue
* It transforms the message into a Work
* It looks in pipeline_storage – ES doesn't know about this Work, so the transformer indexes it
* The transformer tries to send the Work ID to the ID minter… 💥 something goes wrong
* …
* Transformer picks up the message from the queue again
* It transforms the message into a Work
* It looks in pipeline_storage – ES already has the Work, so the transformer skips sending it
* The message is deleted from the queue

So it's possible for a Work to get dropped at the transformer and never reach the ID minter.

> **Sidebar:** I’m not sure how the transformer picks up the message from the queue. We have an [explicit test for this scenario](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/transformer/transformer_common/src/test/scala/weco/catalogue/transformer/TransformerWorkerTestCases.scala#L243-L264) that says failing to send an SNS notification for a transformed Work should drop a message on the DLQ after a single failure. These messages shouldn't be being retried at all, and yet the transformer logs suggest this is happening. Wassup?

## The solution

**We could remove the lazy sending entirely.** But I really don’t want to do that – it would re-expose the pipeline to the full force of the nightly Sierra reharvests. Back of the napkin, that was costing at least $100/day before the most recent cost optimisations, and achieving almost nothing of value. It would be super expensive and I don't think it's necessary.

**Instead, I've tweaked the lazy sending logic.** Here are the new rules:

1. If the transformed Work is older than the stored Work, we _never_ send it onward
2. If the transformed Work is the same version as the stored Work, we _always_ send it onward
3. If the transformed Work is newer than the stored Work, we only send it onward if it has different WorkData

Case 1 is trivial.

Case 2 applies for retried messages during reindexes. If a pipeline ends up resending a Work several times (e.g. because it failed to send to SNS), we'll keep sending it onwards. By doing this, we can assume the Work will eventually be indexed.

(Previously, we would only send on a transformed Work with the same version as the stored Work if it had different WorkData.)

Case 3 applies for updates after the initial reindex. If the WorkData is different, this is new and interesting and send it onward. If not, we can assume the initial version of the Work with this WorkData was successfully stored (by case 2), so we don't need to send it on again.

For https://github.com/wellcomecollection/platform/issues/4966


warrd (commented):
That logic seems to make sense to me.

I remember having a slightly different approach for a similar problem in the Calm adapter: we set the record ["published"](https://github.com/wellcomecollection/catalogue/blob/master/calm_adapter/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/CalmAdapterWorkerService.scala#L81-L84) after sending it, so we know we don't need to send it again if it is re-encountered. However that would require a change to the work model here, which I think is pretty undesirable for this use case.

alicefuzier:
> Sidebar: I’m not sure how the transformer picks up the message from the queue. We have an explicit test for this scenario that says failing to send an SNS notification for a transformed Work should drop a message on the DLQ after a single failure. These messages shouldn't be being retried at all, and yet the transformer logs suggest this is happening. Wassup?

That's not how I read the meaning of that test. It means the message ends up in the dlq _eventually_. SQSStream doesn't retry anything, so wether something ends up in a dlq or how many times is retried (if any) is down to the sqs config of the queue. The `withLocalSqsQueuePair` creates a queue and a dlq with a redrive policy of 3 redrives so that's why you see 3 redrives in the logs.

alexwlchan:
> That's not how I read the meaning of that test. It means the message ends up in the dlq eventually.

Ah, so I missed a detail – I added some logging to the failing MessageSender to print every time it was called, so I could see every retry. It only prints once, then the message goes to the DLQ.

I think that means it's not retrying?

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
```diff
@@ -119,22 +117,58 @@ trait TransformerWorker[Payload <: SourcePayload, SourceData, SenderDest]
...
+              // same data, it's possible this is SQS retrying a message that failed
+              // last time.  e.g. we indexed the work but didn't send an ID to SNS.
+              //
+              // Send the work, just to be sure it got through the pipeline.
+              else if (storedWork.version == transformedWork.version) {
+                debug(
+                  s"${transformedWork.id}: transformed Work and stored Work have the same version")
+                true
+              } else if (storedWork.data != transformedWork.data) {
+                println(
```
alicefuzier, line 141: println?

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
```diff
@@ -119,22 +117,58 @@ trait TransformerWorker[Payload <: SourcePayload, SourceData, SenderDest]
...
+                true
+              }
+
+              // If we get here, it means the new work and the stored work should have
+              // the same data, but the stored work has a strictly lower version.
+              //
+              // Nothing in the pipeline will change if we send this work, and we
+              // can assume the previous version of the work was sent successfully.
+              else {
+                println(
```
alicefuzier, line 152: ditto

kenoir:
> If the transformed Work is the same version as the stored Work, we always send it onward

This makes sense as we cannot guarantee that storing a work always results in a sent message.

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
```diff
@@ -119,22 +117,58 @@ trait TransformerWorker[Payload <: SourcePayload, SourceData, SenderDest]
...
       // we care about for the pipeline is changed.
-      case Right((newWork, key)) =>
+      case Right((transformedWork, key)) =>
         retriever
-          .apply(workIndexable.id(newWork))
+          .apply(workIndexable.id(transformedWork))
           .map { storedWork =>
-            if (newWork.shouldReplace(storedWork)) {
-              Right(Some((newWork, key)))
+            val shouldSend = {
```
alicefuzier, line 124: nitpicking but I would prefer this in a separate `shouldSend` method so that `compareToStored` is easier to read

alicefuzier (approved):
A couple of comments, but looks good. Thanks for fixing so quickly

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
alexwlchan: Arrgh. I had this in to turn stuff on/off during tests, didn't mean to check it in. Should be debug logs like the others.

pipeline/transformer/transformer_common/src/main/scala/weco/catalogue/transformer/TransformerWorker.scala
alexwlchan: Done.

PR #1308, opened 2021-01-22
Fix/image search query

jamesgorrie:
* Adds boosting to concept fields similar to works query
* removes `minimum_should_match` which was hangover from past query
* boosts ID fields inline with works query

Ran through rank_eval, all passes, once deployed to stage we can check the actual tool.

☝️ we need to find a way to test this in feature branches.


api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ImagesQuery.scala
```diff
@@ -3,33 +3,44 @@ package uk.ac.wellcome.platform.api.elasticsearch
...
+      ("data.genres.concepts.label", Some(10)),
+      ("data.production.*.label", Some(10)),
+      ("data.description", None),
+      ("data.physicalDescription", None),
+      ("data.language.label", None),
+      ("data.edition", None),
+      ("data.notes.content", None),
+      ("data.lettering", None),
+      ("data.collectionPath.path", None),
+      ("data.collectionPath.label", None),
```
alexwlchan, line 36: Tiny nit: the horizontal alignment of these makes it quite tricky to see which fields have the same boost. Possibly flip them round to make it easier to see the relative boost on the different fields?

e.g.

```scala
val boostedFields = Map(
      Some(1000) -> Seq(
        "data.contributors.agent.label"
      ),
      Some(100) -> Seq(
        "data.title",
        "data.title.english",
        "data.title.shingles",
        "data.alternativeTitles",
      ),
      Some(10) -> Seq(
        "data.subjects.concepts.label",
        "data.genres.concepts.label",
        "data.production.*.label",
      ),
      None -> Seq(
        "data.description",
        "data.physicalDescription",
        "data.language.label",
        "data.edition",
        "data.notes.content",
        "data.lettering",
        "data.collectionPath.path",
        "data.collectionPath.label"
      )
    )

    val fields: Map[String, Option[Int]] =
      boostedFields
        .flatMap { case (boost, fieldNames) =>
          fieldNames.map { _ -> boost }
        }
        .flatMap { case (fieldName, boost) =>
          toWorkField(fieldName).map { workField => (workField, boost) }
        }
        .map { case (field, boost) =>
          FieldWithOptionalBoost(field, boost.map(_.toDouble))
        }
```

alexwlchan (approved):

api/api/src/main/scala/uk/ac/wellcome/platform/api/elasticsearch/ImagesQuery.scala
jamesgorrie: The point it to not couple things in this regard, but it does seem good to understand the order. Thanks!

harrisonpim (approved):

PR #1311, opened 2021-01-25
Separate critical and pipeline infra; talk to the catalogue cluster over PrivateLink

alexwlchan:
This is a fairly large PR that addresses two things we learnt in the 2021-01-22 reindex.

* **It's good to flex the size of the ID minter database and the pipeline_storage cluster.** We can't afford to run them at full capacity all the time, and we don't need to – 95% of the time, the pipeline is idling, and paying for those resources is a waste. If we only turn them up during a reindex, we get good performance and reasonable costs.

   But both of these were managed in the `infrastructure/critical` stack, and we don't want to be changing that every time we do a reindex. This PR addresses that in two ways:

   - The pipeline storage Elasticsearch cluster now lives in its own Terraform stack. All the data it contains can be recreated, and an outage in this cluster isn't user-facing. This makes it safer to scale the cluster up/down for a reindex.
   - The `pipeline/stack` module has an option to spin up extra ID minter database instances. The core ID minter database is still defined in the `infra/critical` stack, but now you can add capacity without touching the core resources. This makes it safer to temporarily add ID minter database capacity for a reindex.

* **Using PrivateLink/VPC endpoints makes a big difference in cost.** The NAT Gateway cost of the last reindex was noticeably less spiky, but still a significant portion of the cost.

    Some of this is probably the “last mile” cost of getting Works/Images from the ingestor to the API cluster. Additionally, NAT Gateway is a substantial fraction of the cost in the catalogue account (~$470/mo).

    This PR addresses that by creating the appropriate VPC endpoints and configuration for us to talk to the catalogue cluster using PrivateLink rather than the NAT Gateways. I've applied and tested this in the catalogue pipeline, API and snapshot generator.

    For https://github.com/wellcomecollection/platform/issues/4949


infrastructure/pipeline_storage/README.md
```diff
@@ -0,0 +1,33 @@
...
+    This is moderately inconvenient -- you can't use it by running vanilla Terraform.
+    For now I don't want to make the pipeline stack harder to use.
+
+## Scaling up the cluster during reindexes
+
+We can't afford to run a high-capacity cluster all the time, but we also don't need to.
+The cluster is under heavy load when we're reindexing the catalogue pipeline, but that's it.
+
+*   When the pipeline is idle, we leave it at a single 8GB node.
+*   When we're reindexing, it's helpful to turn it up to 29GB and add an extra zone.
```
jamieparkinson, line 23: Could you add some more specific documentation re how/where to do this? Maybe extracting these features to some `local`s.

infrastructure/pipeline_storage/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Remember to turn it down again afterwards!
```
kenoir, line 25: What about a "reindexing" variable on the stack that takes care of sizeing for you?

kenoir (approved):

pipeline/terraform/main.tf
```diff
@@ -71,26 +74,28 @@ module "catalogue_pipeline_2021-01-22" {
...
     local.mets_adapter_topic_arn,
   ]
 
   calm_adapter_topic_arns = [
-    local.calm_reindexer_topic_arn,
+    /*local.calm_reindexer_topic_arn,*/
     local.calm_adapter_topic_arn,
   ]
 
+  extra_rds_instances = 0
```
jamieparkinson, line 97: Why not just `rds_instances`? You could even add a custom validation rule to make sure it's always >=1, `extra` to me is a bit confusing as it isn't clear what I'm augmenting

infrastructure/pipeline_storage/README.md
kenoir: I think you might be able to hide the scaling functionality entirely in the stack and set a "reindexing" variable to true of false.

infrastructure/pipeline_storage/README.md
alexwlchan: I think we could do that if we moved the pipeline_storage module into the pipeline stack.

How do we feel about that?

pipeline/terraform/main.tf
alexwlchan: Mmm, let me think about this some more.

The "extra" is meant to be "in addition to the base instance, which is always running". Maybe calling it `rds_instances` would make more sense, but I'm also tempted by Robert's idea of a "reindexing" module.

infrastructure/pipeline_storage/README.md
kenoir: So implicitly that means one ES cluster per pipeline, which would make the base cost n*minimum_cost_of_es_cluster. I think maybe. Let's discuss elsewhere.

alexwlchan:
There's definitely more to be done to improve the ergonomics here, but I'm going to merge "as-is" because this is working right now, and changing it further requires more discussion.

infrastructure/pipeline_storage/README.md
alexwlchan: When I wrote that I actually meant defining the cluster at the top of the pipeline stack (still a single instance, but living alongside it), but that doesn't really work with targeted plan/apply.

It also requires some thinking about authentication. I think it's possible, but agree we should discuss further.

PR #1315, opened 2021-01-26
Add contributorAgents to DerivedWorkData

warrd:
We cannot currently aggregate on contributors as we need to aggregate on both the agent label and its type. Here we derive a new string field which combines both of these fields (`"$type:$label"`). In the aggregations we will be able to decode this interpolated field to reconstruct the original `Contributor` object.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
@@ -1,16 +1,31 @@
 package uk.ac.wellcome.models.work.internal
 
 case class DerivedWorkData(
-  availableOnline: Boolean
+  availableOnline: Boolean,
+  contributorAgents: List[String],
 )
 
 object DerivedWorkData {
-  def none: DerivedWorkData = DerivedWorkData(availableOnline = false)
 
-  def apply(workData: WorkData[_]): DerivedWorkData =
+  def none: DerivedWorkData =
     DerivedWorkData(
-      availableOnline = containsDigitalLocation(workData.items)
+      availableOnline = false,
+      contributorAgents = Nil
     )
+
+  def apply(data: WorkData[_]): DerivedWorkData =
+    DerivedWorkData(
+      availableOnline = containsDigitalLocation(data.items),
+      contributorAgents = contributorAgents(data.contributors)
+    )
+
+  private def contributorAgents(contributors: List[Contributor[_]]): List[String] =
+    contributors.map {
+      case Contributor(_, agent, _) =>
+        val ontologyType = agent.getClass.getSimpleName
+        val label = agent.label
+        s"$ontologyType:$label"
+    }
 
   private def containsDigitalLocation(items: List[Item[_]]): Boolean =
     items.exists { item =>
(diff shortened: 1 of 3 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
```diff
@@ -1,17 +1,33 @@
...
+      availableOnline = containsDigitalLocation(data.items),
+      contributorAgents = contributorAgents(data.contributors)
+    )
+
+  private def contributorAgents(
+    contributors: List[Contributor[_]]): List[String] =
+    contributors.map {
+      case Contributor(_, agent, _) =>
+        val ontologyType = agent.getClass.getSimpleName
+        val label = agent.label
```
jamieparkinson, line 27: I think it's probably worth asserting that this doesn't contain a `:` as @alexwlchan was saying yesterday

jamieparkinson (approved):

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
warrd: What is the issue if there is? Some labels are likely to contain that, and if we allow it through I don't see the problem, as in practice we will be splitting at the first occurrence of `:` and checking the prefix is one of `Person`, `Agent`, `Meeting`, `Organisation`..

jamesgorrie (commented):
Assuming then that we are happy for filters to still just be on the label and return results that are of any type "Orgs or people".

Do we have examples where they are the same label, but different type?

Ping @jtweed to make sure that we know this is the case, and will probably resolve it when creating filters on not-the-label.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
jamieparkinson: Hmmm, yeah, makes sense - perhaps a comment about splitting on the first occurrence then? It wasn't obvious to me that it would work like that rather than splitting into a list which we might incorrectly assume always had 2 elements

jamesgorrie:
As a note - I do think we should relook at this and try to remove the idea that the aggregations have to come back as the same types as in the internal model.

It will make the client / interface need to know how to handle all of the types, when all it is actually doing with aggregations is linking to filtered results.

warrd:
> Do we have examples where they are the same label, but different type?

I have no idea bout the actual data, but worth mentioning It's not just that we want to aggregate diffrerent types into different buckets: it's more that we have no way of knowing what type it is otherwise.

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
alexwlchan: Ah, so my suggested assertion was that the ontology type doesn't contain a `:`, not the agent. Which admittedly seems unlikely, but is the sort of thing that could creep in – then you'd get out a different thing than you started with.

warrd: I'm pretty sure that is an illegal character for a class name :man_shrugging:

PR #1318, opened 2021-01-26
Fix and update remove_work script

jamieparkinson:
This updates the script to handle the current model and infra, adds a dry run flag, finds images associated with the work and provides an option to delete them


alexwlchan (approved):

jamesgorrie (approved):
Looks good - took me a little while to work my head around it because we've overloaded the `image` word to be a think in the catalogue (work) and a thing in an image index.

PR #1322, opened 2021-01-27
Refactoring in preparation to add license aggregations on images

alexwlchan:
For wellcomecollection/platform#4994

This PR doesn't change the behaviour of the API at all, but it moves some classes around and adds the "Work" prefix in preparation for creating an "Image" variant later.


api/api/src/main/scala/uk/ac/wellcome/platform/api/models/SearchOptions.scala
```diff
@@ -1,18 +1,41 @@
...
+sealed trait SearchOptions[DocumentFilter, AggregationRequest, MustQuery] {
+  val searchQuery: Option[SearchQuery]
+  val filters: List[DocumentFilter]
+  val aggregations: List[AggregationRequest]
+  val mustQueries: List[MustQuery]
+  val sortBy: List[SortRequest]
+  val sortOrder: SortingOrder
+  val pageSize: Int
+  val pageNumber: Int
+}
```
warrd, line 19: A bit unsure why having this extra trait is useful rather than just being a case class? It doesn't define any extra behaviour, and `WorkSearchOptions` / `ImageSearchOptions` are basically just the original case class with different type parameters.

alexwlchan: I thought this was neater:

* The type parameters always appear together. You'd never have, say, a SearchOptions with WorkAggregations but ImageQuery, and the distinct case classes enforce that.
* You end up propagating type parameters to a bunch of places that don't really care about them, because they're in so many places. If a class is only passing through a SearchOptions that it doesn't inspect directly, it's a faff to update all the signatures when you change it (and those signatures were starting to get unwieldy).
* It makes it clear that these are in some sense distinct; one issue with doing this patch is the number of things that look generic but are really only usable with Works

alexwlchan: It occurs to me now that type aliases might have done the same trick, hmm.

jamieparkinson: I haven't given this suggestion all that much thought but since 3 is quite a lot of type parameters, and as you say they all appear together, could you do this with (abstract) type members rather than parameters? They are strictly equivalent (although not necessarily wrt to the compiler's ability to infer them)

alexwlchan: I'm not super familiar with abstract type members – what's the difference between that and what I have here?

warrd: I agree abstract type members / associated types would be more appropriate here probably as would be even neater as it hides the type away from the main type signature (and thus we can just write `S <: SearchOptions` instead `S <: SearchOptions[_, _, _]`).

I understand your points about it being neater without having to pass 3 type parameters at a time. I generally consider traits to have two purposes: defining interfaces for behaviour or defining ADTs (I kind of think it a bit weird that Scala overloads traits for this, as opposed to Rust which has `trait` vs `enum`, where `trait` you wouldn't be allowed to do something like this). It's not the worst thing however, so if its the neatest option then probably worth it.

Other alternatives would be:
* Type alias
* A trait which just defines a bunch of bundled types like we do [here](https://github.com/wellcomecollection/catalogue/blob/master/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DataState.scala#L10-L13), so we only have 1 rather than 3 params (this is also a slightly weird pattern I admit)

warrd: Worth mentioning the original implementation of `Work` used this pattern of a base trait with lots of `val` members, and that proved to be really inflexible and un-ergonomic, hence moving to `Work[State]` which made abstracting over it much nicer. However `SearchOptions` is a much more isolated type and I don't think the way we model it will have as far reaching issues.

warrd: > I'm not super familiar with abstract type members – what's the difference between that and what I have here?

The [RFC for associated types in Rust](https://github.com/rust-lang/rfcs/blob/master/text/0195-associated-items.md#motivation) has some good motivation / benefits of associated types.

warrd (approved):
Approving this as not a blocker, up to @alexwlchan whether / how they want to address the comment above

alexwlchan:
Thanks Nick, I'm going to merge as-is for now and maybe circle back to it later. 👍

PR #1325, opened 2021-01-28
Batch es reads other services

alicefuzier:


pipeline/terraform/stack/service_matcher.tf
```diff
@@ -52,7 +52,9 @@ module "matcher" {
 
     dynamo_lock_timeout = local.lock_timeout
 
-    es_index = local.es_works_identified_index
+    es_index                           = local.es_works_identified_index
+    read_ingest_batch_size             = 100
```
kenoir, line 56: Any context for why these numbers as a comment would be good.

alexwlchan: Do we need these config parameters on other services?

alicefuzier: They already have them. The matcher was the only one not already using pipeline storage in some way

alicefuzier: > Any context for why these numbers as a comment would be good.

To be honest there isn't a good reason for why these numbers other than this is the batch size/flush interval we've used elsewhere. As we've never batched many reads together these might need tweaking but we can't know before a reindex

alexwlchan (approved):
I looked at this earlier today; looks good!

PR #1329, opened 2021-01-29
Provision an Elasticsearch cluster per pipeline

alexwlchan:
Because hey, we can do this now!

This patch updates the pipeline module to:

* Create a dedicated Elasticsearch cluster
* Provisions the roles and users that the pipeline needs
* Puts those users in the appropriate Secrets Manager secrets

So each pipeline gets ringfenced resources, and all those resources get cleaned up when we're done with a pipeline.

Perhaps the most interesting dev-facing aspect of this is the `is_reindexing` boolean on the `stack` module. If you set this to `true`, we automagically provision extra database resources to ensure the pipeline is reliable, but without having to run them all the time.

This opens the door to using provisioned DynamoDB capacity, and temporarily increasing it for reindexes, per https://github.com/wellcomecollection/platform/issues/4969


pipeline/terraform/stack/elastic_pipeline_storage.tf
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  depends_on = [
+    module.pipeline_storage_secrets,
+    aws_secretsmanager_secret.es_username,
+    aws_secretsmanager_secret.es_password,
+  ]
+
+  provisioner "local-exec" {
+    command = "python3 scripts/create_pipeline_storage_users.py ${var.pipeline_date}"
```
kenoir, line 68: Fancy!

kenoir: What happens if this fails?

kenoir (approved):

pipeline/terraform/stack/elastic_pipeline_storage.tf
alexwlchan: If a local-exec provisioner fails, Terraform marks the resource as tainted and tries to recreate it on the next plan/apply cycle.

alicefuzier:
This is nice!

PR #1333, opened 2021-02-01
Lazy indexing in the works ingestor

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4999

Because the relation embedder runs so much faster than the ingestor, the final step of a reindex often ends with 2.5M+ messages on the ingestor queue while the relation embedder is finished.

Once the relation embedder is done, sending the same ID to the ingestor more than once won't change anything, but it will cause unnecessary index operations in Elasticsearch. This patch modifies the works ingestor to ask ES first _"hey, do you have any of the Works I'm about to index?"_, and to skip indexing anything which is **identical** to what's already in ES.

I've only enabled this in the works ingestor, because that's the only service where we expect to see a non-trivial saving here.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
@@ -34,48 +34,54 @@ class ElasticIndexer[T: Indexable](
     new ElasticsearchIndexCreator(client, index, config).create
 
   final def apply(documents: Seq[T]): Future[Either[Seq[T], Seq[T]]] =
-    Future
-      .fromTry(Try(
-        require(documents.nonEmpty, "Cannot index an empty list of documents")))
-      .flatMap { _ =>
-        debug(
-          s"Indexing ${documents.map(doc => indexable.id(doc)).mkString(", ")}")
-        val inserts = documents.map { document =>
-          indexInto(index.name)
-            .version(indexable.version(document))
-            .versionType(ExternalGte)
-            .id(indexable.id(document))
-            .doc(document)
-        }
+    for {
+      _ <- Future.fromTry(Try(
+        require(documents.nonEmpty, "Cannot index an empty list of documents"))
+      )
 
-        client
-          .execute {
-            bulk(inserts)
-          }
-          .map { response: Response[BulkResponse] =>
-            if (response.isError) {
-              error(s"Error from Elasticsearch: $response")
-              Left(documents)
-            } else {
-              debug(s"Bulk response = $response")
-              val bulkResponse = response.result
-              val actualFailures = bulkResponse.failures.filterNot {
-                isVersionConflictException
-              }
+      ids = documents.map(doc => indexable.id(doc))
+      _ = debug(s"Indexing ${ids.mkString(", ")}")
 
-              if (actualFailures.nonEmpty) {
-                val failedIds = actualFailures.map { failure =>
-                  error(s"Failed ingesting ${failure.id}: ${failure.error}")
-                  failure.id
-                }
+      result <- indexDocuments(documents)
+    } yield result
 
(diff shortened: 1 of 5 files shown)


jamieparkinson:
Is a high throughput of index reads on the catalogue cluster safe?

kenoir:
> Is a high throughput of index reads on the catalogue cluster safe? 

This is a good question.

kenoir:
It feels a little like the expense of reindexing identical things should be in some way mitigated by ES, not by us reading and checking. I'm wondering about whether there is some affordance of ES that would allow us to achieve the same thing with less responsibility on our services?

jamieparkinson:
I was wondering about the impact of changing the versioning behaviour from `gte` -> `gt`? I feel like it's probably `gte` for a reason, but not sure

alexwlchan:
> Is a high throughput of index reads on the catalogue cluster safe?

I'm going on what feels like common sense intuition that index writes are more expensive than index reads. The API cluster is able to handle this volume of writes when a reindex is in its final stages and we're writing millions of mostly unchanged documents.

> It feels a little like the expense of reindexing identical things should be in some way mitigated by ES, not by us reading and checking. I'm wondering about whether there is some affordance of ES that would allow us to achieve the same thing with less responsibility on our services?

I did look for this, but I couldn't find anything. I would suspect not?

For Elasticsearch to know if a document is identical to the one in the index, it would have to completely re-run the indexing process, at which point you've lost 90% of the benefit of de-duplication/

> I was wondering about the impact of changing the versioning behaviour from external_gte -> external? I feel like it's probably gte for a reason, but not sure

Do we reset the version when we merge two Works together? Or does the version come from the original, unmerged Work?

That might work, but I'm pretty sure we don't want it universally.

jamieparkinson:
> Do we reset the version when we merge two Works together? Or does the version come from the original, unmerged Work?

We use the `modifiedTime` as the version, when we merge works together we take the most recent `modifiedTime`. I don't know how/if the relation embedder affects this though...

alexwlchan:
> We use the modifiedTime as the version, when we merge works together we take the most recent modifiedTime. I don't know how/if the relation embedder affects this though...

Ah, so I don't think `gt` will work for us?

Suppose we have `Work(id=A, modifiedTime=2)` and `Work(id=B, modifiedTime=1, mergeCandidates=[A])`.

If `A` gets ingested first, it won't be replaced when `B` comes through and they get merged into `A+B`.

kenoir:
> For Elasticsearch to know if a document is identical to the one in the index, it would have to completely re-run the indexing process, at which point you've lost 90% of the benefit of de-duplication/

There is on-by-default in-built detection if you make an update and that results in no change (https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_detecting_noop_updates).

alexwlchan:
> There is on-by-default in-built detection if you make an update and that results in no change 

That's using the `update` API, in which Elasticsearch starts by calling up the existing version of the document.

We're using the `index` API, [which doesn't do this](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#index-noop):

> When updating a document using the index API a new version of the document is always created even if the document hasn’t changed. If this isn’t acceptable use the _update API with detect_noop set to true. This option isn’t available on the index API because the index API doesn’t fetch the old source and isn’t able to compare it against the new source.
>
> There isn’t a hard and fast rule about when noop updates aren’t acceptable. It’s a combination of lots of factors like how frequently your data source sends updates that are actually noops and how many queries per second Elasticsearch runs on the shard receiving the updates.

kenoir:
> We're using the index API, which doesn't do this:

This begs the question: Why don't we use the update API?

alexwlchan:
> Why don't we use the update API?

It used to be the case that the update API didn't support external versioning, which we rely on.

I can't see any mention of that limitation in the 7.x docs so it might be okay now?

alexwlchan:
Closing for now, I'm not going to work on this in the immediate future.

PR #1335, opened 2021-02-01
Make sure deleted Works don't get dropped by the merger

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5001

I believe this is why five Works keep disappearing from reindexes. If we can identify a target Work, we'd silently drop the deleted Works and they'd never get out of the merger. Now we pull them out separately, and add them to the MergerOutcome once all the merging logic has been applied.

I've also added some extra assertions that catch this issue, so it'll be easier to spot if something similar happens in future.


alicefuzier (approved):
Thank you!

PR #1339, opened 2021-02-03
Default the contributorAgents on DerivedData to Nil

alexwlchan:
For some reason, there are works in the index that don't have a value for this field (e.g. jpubbjnr).  Requesting them through the API returns a 500 error.

There's no harm in setting a reasonable default here, and hopefully it means these works will stop returning 500 errors.

This is hard to test without writing something to write deliberately bad data into a test index.


jamieparkinson (approved):
These works are universal constants - they were not contributed or created; they merely _are_

PR #1342, opened 2021-02-03
Call the staging API environment "stage", the same label we use in the weco-deploy defaults and everywhere else

alexwlchan:


jamieparkinson (approved):
This is fully my fault, and I no longer know why I did it - given that we don't mind the staging API going down, do you want to rename the terraform module/directory as well?

alexwlchan:
> I no longer know why I did it

it’s not an unreasonable choice of name, it’s a faff to change (you have to tear down the load balancer and recreate it), and honestly I’m not sure anybody else hit it

Only ditz here was ever deploying the staging Catalogue API by hand 😂

Issue #1320, opened 2021-01-27
Multiple genres / subjects / contributors search


jamesgorrie:
both the `genres.labels`, `concepts.label` and `contributors.agent.label` need to allow for multiple values.

After a few discussion's we have decided like the other filters, comma separated values would work for this. Values should be wrapped in `"`, and if the contain `"`, have those escaped.

e.g. 
* `"Watson, Jacqueline."`
* `"Watson, Jacqueline.","Wellcome Trust Centre for the History of Medicine at UCL.","University College, London."`
* `"Watson, \"Danger\", Jacqueline.","University College, London."`

jamesgorrie:
Ping @jtweed

jtweed:
Yep, agree with that ^

taceybadgerbrook:
@jamesgorrie what do these multiple selections look like in conversion*? Do queries get lumped into page.query.query, are there new fields, do multiple selections get separated by commas, etc. 

The reason I ask is because I would like to be able to see how users are getting on with this new facility.

PR #1343, opened 2021-02-03
make genre, subject and contributor filters keyword filters

jamesgorrie:
#1320 

After conversations with @jtweed we have decided to make these filters strict filters on the label of the genres, subjects and contributors (concepts).

This will be more inline with what we have for any other filters, i.e. `language`, `workType`, `...`, and how they interact with aggregations, which are used by the client to then link to these filters <sup>*</sup>.

We will leave more freestyle text search to the `query` parameter.

When we move into actual concept search, and how scoring works within that, we will probably have another endpoint, e.g. `/concepts?query=boom` and use the learning from other relevancy search we have applied elsewhere.

---
# Example (taken from tests)

* `contributors.agent.label="Bath, Patricia",Karl Marx,"Darwin \"Jones\", Charles"` searches for any works with any of the below contributors
  * `Bath, Patricia`
  * `Karl Marx`
  * `Darwin "Jones", Charles`
---

<sup>*</sup> this goes back to a conversation as well where I have suggested we create the aggregations with this use case rather than creating a pseudo version of the internal model, RFC to come)

aggs_filters

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
@@ -1,9 +1,9 @@
 package uk.ac.wellcome.platform.api.rest
 
 import java.time.LocalDate
-
 import akka.http.scaladsl.server.{Directive, Directives, ValidationRejection}
 import akka.http.scaladsl.unmarshalling.Unmarshaller
+import com.github.tototoshi.csv.CSVParser
 import io.circe.{Decoder, Json}
 import uk.ac.wellcome.platform.api.models.LicenseFilter
 import uk.ac.wellcome.platform.api.rest.MultipleWorksParams.decodeCommaSeparated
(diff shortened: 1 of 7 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/QueryParams.scala
```diff
@@ -35,7 +35,17 @@ trait QueryParamsUtils extends Directives {
     Decoder.decodeInt.withErrorMessage("must be a valid Integer")
 
   def decodeCommaSeparated: Decoder[List[String]] =
-    Decoder.decodeString.emap(str => Right(str.split(",").toList))
+    Decoder.decodeString.emap(
+      str =>
+        Right(
+          CSVParser
```
jamieparkinson, line 41: Good

jamieparkinson (approved):
I like the sound of your suggestion about how we construct aggregations, looking forward to the RFC

Issue #1167, opened 2020-12-08
Implement a consistent definition of work availability


jamieparkinson:
Currently, we flag works as _"available online"_/`availableOnline` if they contain an item with a digital location. 
Additionally, on wellcomecollection.org we provide filters/aggregations labelled for availabilities _"Online"_ and _"In the library"_: these actually use filters for the presence of a digital location or a physical location, respectively.

In https://github.com/wellcomecollection/catalogue/issues/1162 we want to change the definition of the `availableOnline` flag to be true if and only if:
- the work contains an item with a digital location 
- the location has either an `Open` or `OpenWithAdvisory` access status<sup>*</sup>

<sup>*could do with another confirmation.</sup>

However, this introduces an incompatibility with the filters that weco.org is using!
Since the outcome that we want for users is to see availability (rather that one of its proxies, presence of a digital location), _I think we should make the "Online" filter/aggregation use the flag directly_, thus making the definitions consistent.

This leaves us with the "in the library" availability - is this defined by the presence of a physical location alone, or do access conditions also play a role here? If so, we may want to replace `availableOnline: Boolean` with a union like `availability: online | in-library | on-request | closed` (credit to @jamesgorrie for this idea).

pollecuttn:
For physical items their item status tells you if they're actually available or not: https://github.com/wellcomecollection/platform/blob/master/ontologies/Reference%20data/item-status.csv

Generally it's clear but with something like Permission required it may become available to the requestor in future.

jamieparkinson:
Conclusion:

- We will have an `availability` property that will be a **list** of strings. Initially these will be limited to either `"Online"` and `"In the library"`.
- Archive records will inherit availability from their children.

jamieparkinson:
- [ ] Add availabilities property
- [ ] Add filtering/aggregation on the property
- [ ] Inherit availability from children

PR #1347, opened 2021-02-04
Use the access status to determine availableOnline

alexwlchan:
Based on Alexandra's comment in #wc-platform-feedback. I had a quick look at this, and drummed up a very conservative rule:

If a digital location has:

*   at least one access condition, and
*   every access condition has a non-empty AccessStatus, and
*   every access status is Closed or Unavailable

then it feels reasonable to say this location doesn't count as "available", and we should set `availableOnline = false`.

/cc @jamieparkinson as the original author of this code, and the person who knows most about this

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/DerivedWorkData.scala
@@ -44,8 +44,7 @@ object DerivedWorkData {
   //
   private def isAvailableOnline(items: List[Item[_]]): Boolean =
     items.exists { item =>
-      item
-        .locations
+      item.locations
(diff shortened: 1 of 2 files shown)


jamieparkinson:
Sadly I don't think this is what we want - this requires more product work and @jtweed deprioritised it last time it came up. See this ticket for more details  https://github.com/wellcomecollection/catalogue/issues/1167

alexwlchan:
Ah, alas.

jamieparkinson:
Yeah, it's a frustrating one! Hopefully we'll be able to work out a complete solution soonish, I might try to move it forward if I can

alexwlchan:
Ah, I know why I couldn't find the ticket – I was look in the platform repo, but it's in the catalogue repo.

jamieparkinson:
I think search-related stuff tends to live in the catalogue repo these days

PR #1348, opened 2021-02-04
Extract a general-purpose API client from the calm retriever

jamieparkinson:
As I said in https://github.com/wellcomecollection/catalogue/pull/1334, a bit more work was required to make a general-purpose API client out of the pieces that the adapter was using.

There is actually very little new/changed code here; the changes are mostly extracting pieces of the existing `CalmRetriever`/`CalmRetrieverTest` into `CalmApiClient`/`CalmApiClientTest`.

There might be some more tweaks when I come to use it in the deletion checker but this should get some diff out of the way for now.


calm_adapter/calm_api_client/src/test/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClientTest.scala
```diff
@@ -0,0 +1,155 @@
...
+      case (apiClient, _) =>
+        whenReady(apiClient.summary(1)) { response =>
+          response.id shouldBe "1"
+          response.data.keys should not contain suppressedField._1
+        }
+    }
+  }
+
+  it("fails if there is no cookie in a search response") {
+    val responses = List(searchResponse(1, None))
```
alexwlchan, line 115:
```suggestion
    val responses = List(searchResponse(1, cookiePair = None))
```

calm_adapter/calm_api_client/src/test/scala/uk/ac/wellcome/platform/calm_api_client/fixtures/CalmResponseGenerators.scala
```diff
@@ -0,0 +1,62 @@
...
+            <SearchResult>{n}</SearchResult>
+          </SearchResponse>
+        </soap:Body>
+      </soap:Envelope>.toString,
+      protocol
+    )
+
+  def summaryResponse(
+    data: List[(String, String)],
+    timestamp: Option[Instant] = Some(retrievedAt)): HttpResponse =
```
alexwlchan, line 39: I can't find any instances of this argument being set to a non-default value, maybe ditch it?

You could then do something like:

```scala
def summaryResponse(data: (String, String)*)
```

which IMO makes it slightly nicer when it's called.

calm_adapter/calm_api_client/src/test/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClientTest.scala
```diff
@@ -0,0 +1,155 @@
...
+            "1",
+            Map("RecordID" -> List("1")),
+            retrievedAt)
+        }
+    }
+  }
+
+  it("uses basic auth credentials for requests") {
+    val responses =
+      List(searchResponse(1), summaryResponse(List("RecordID" -> "1")))
```
alexwlchan, line 63: I'd prefer to see keyword arguments throughout here, so it's clear what this number means – it's not obvious that it's number of results rather than, say, the ID.

calm_adapter/calm_api_client/src/test/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClientTest.scala
```diff
@@ -0,0 +1,155 @@
...
+        whenReady(for {
+          _ <- apiClient.search(query)
+          _ <- apiClient.summary(1)
+        } yield ()) { _ =>
+          testHttpClient.requests.map(_.headers.collect {
+            case auth: Authorization => auth
+          }) should contain only List(
+            Authorization(BasicHttpCredentials(username, password))
+          )
+        }
```
alexwlchan, line 75: This is all a bit smushed together; a couple of intermediate variables would help readability.

```suggestion
        val future = for {
          _ <- apiClient.search(query)
          _ <- apiClient.summary(1)
        } yield ()
        
        whenReady(future) { _ =>
          val authHeaders =
            testHttpClient.requests
              .flatMap { _.headers }
              .collect { case auth: Authorization => auth }

          authHeaders.distinct shouldBe List(
            Authorization(BasicHttpCredentials(username, password))
          )
        }
```

calm_adapter/calm_api_client/src/test/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClientTest.scala
```diff
@@ -0,0 +1,155 @@
...
+            case auth: RawHeader => auth
+          }) shouldBe List(
+            List(
+              RawHeader("SOAPAction", "http://ds.co.uk/cs/webservices/Search")),
+            List(
+              RawHeader(
+                "SOAPAction",
+                "http://ds.co.uk/cs/webservices/SummaryHeader"))
+          )
+        }
```
alexwlchan, line 98: Ditto above.

alexwlchan (approved):
Happy for you to merge as-is; since you're changing this code I've suggested a couple of clean ups, but it's nothing major that should be a blocker.

PR #1352, opened 2021-02-08
Add a script for cleaning up old indexes in the API cluster

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5009; related to https://github.com/wellcomecollection/platform/issues/4526

The script fetches some cluster credentials from Elasticsearch, then uses the [cat indices API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-indices.html) to get a list of indexes.

If an index starts with `works-` or `images-` and looks like it's older than the indexes used by the API, it will offer ask for confirmation that it's okay to delete. It only deletes an index with explicit confirmation.

Here's a screenshot of a run where I cleaned up all the current old indexes:

Screenshot 2021-02-08 at 11 39 36

### DIFF
--- a/api/cleanup_unused_api_indexes.py
+++ b/api/cleanup_unused_api_indexes.py
@@ -56,7 +56,7 @@ def list_indexes(es_client):
     """
     Returns a list of indexes in the Elasticsearch cluster, sorted by index name.
     """
-    resp = es_client.get("/_cat/indices", params={"format": "json"})
+    resp = es_client.get("/_cat/indices/works-*,images-*", params={"format": "json"})
     resp.raise_for_status()
 
     return sorted(
--- a/api/cleanup_unused_api_indexes.py
+++ b/api/cleanup_unused_api_indexes.py
@@ -1,8 +1,11 @@
 #!/usr/bin/env python
 
+import click
+import datetime
 import json
 
 import boto3
+import httpx
 
 
 def get_session_with_role(role_arn):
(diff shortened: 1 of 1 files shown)


api/cleanup_unused_api_indexes.py
```diff
@@ -0,0 +1,189 @@
...
+        f"\nConsidering %s (%s docs, %s)"
+        % (
+            click.style(idx["name"], "blue"),
+            humanize.intcomma(idx["doc_count"]),
+            idx["size"],
+        )
+    )
+
+    # We should never delete the prod or staging indexes -- this would cause an instant
+    # outage in the API, which is very bad.
```
jamesgorrie, line 108: Agreed.

api/cleanup_unused_api_indexes.py
```diff
@@ -0,0 +1,189 @@
...
+        base_url=credentials["endpoint"],
+        auth=(credentials["username"], credentials["password"]),
+    )
+
+
+def list_indexes(es_client):
+    """
+    Returns a list of indexes in the Elasticsearch cluster, sorted by index name.
+    """
+    resp = es_client.get("/_cat/indices", params={"format": "json"})
```
jamesgorrie, line 59: To avoid some of the index name checking you can use
```suggestion
    resp = es_client.get("/_cat/indices/works-*,images-*", params={"format": "json"})
```

[See here for more details](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-indices.html#cat-indices-api-path-params)

alexwlchan: Nice!

I'm going to leave the name checking as-is because it's another safety check, and in this sort of script a bit of extra checking never goes amiss.

api/cleanup_unused_api_indexes.py
```diff
@@ -0,0 +1,189 @@
...
+    # prod and staging APIs.  e.g. works-2001-01-01 < works-2002-02-02
+    #
+    # We skip indexes that sort higher, because this might be an index that we're currently
+    # reindexing into, but we haven't pointed an API at yet.  In general, indexes go forward,
+    # not backward.
+    #
+    # We still ask the user to confirm they really want to delete this index, just in case.
+    # We should never offer to delete an index that would cause an outage, but they might
+    # skip an index if, say, they've only just promoted a new index to prod and they
+    # want to be able to roll back to this index.
```
jamesgorrie, line 131: I almost wonder if this is a requirement, keeping a rollbackable index? I'm just not sure how we would work that out...

Saying that, maybe worth a post-using and testing this script.

api/cleanup_unused_api_indexes.py
```diff
@@ -0,0 +1,189 @@
...
+
+    for idx in list_indexes(es_client):
+        maybe_cleanup_index(
+            es_client,
+            idx=idx,
+            prod_index_name=prod_index_name,
+            stage_index_name=stage_index_name,
+        )
+
+    #
```
jamesgorrie, line 186: Needed? Needs refinement?

jamesgorrie (approved):

api/cleanup_unused_api_indexes.py
alexwlchan: Oops, removed, thanks.

PR #1354, opened 2021-02-08
Add new AccessStatus types to reflect the Collection Access Policy

alexwlchan:
Follows #1353. Closes https://github.com/wellcomecollection/platform/issues/5011

This is something I spotted while doing the Location modelling – these types should reflect our collections access framework, but they weren't quite right. This brings them into sync.


jamesgorrie (approved):
Nice

PR #1356, opened 2021-02-08
Automatically invalidate CloudFront for /works when removing a Miro image

alexwlchan:
Use case: I'm a clumsy developer who breaks things when they fiddle around in web consoles. By teaching robots to do things for me, I don't have to risk busting something in the CloudFront console.

There may be a reason this isn't done automatically, in which case let's comment it, otherwise it's one less thing for the next person to do.

I've done a dry run and checked this finds the right CloudFront distro, so I think it's working.

### DIFF
--- a/scripts/remove_work/README.md
+++ b/scripts/remove_work/README.md
@@ -9,6 +9,9 @@ This script:
 - creates CloudFront invalidations for Loris and wellcomecollection.org
 - updates the Miro VHS inventory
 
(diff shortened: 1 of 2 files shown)


jamieparkinson (approved):
Yeah, I've never been sure why this wasn't built in - thank you for adding it!n 

You might want to add a comment for the Loris invalidation saying that it might/probably won't appear to work because of the internal Loris cache (adding whatever's necessary for busting that cache seems like a huge pain especially considering Loris is on its way out)

PR #1358, opened 2021-02-09
Add the license/shelfmark/link_text fields to our Location models

alexwlchan:
Follows https://github.com/wellcomecollection/catalogue/pull/1357; closes https://github.com/wellcomecollection/platform/issues/5012

* The link_text field on DigitalLocation is for text to display instead of the URL in links (e.g. "View this journal"). It comes from Sierra holdings records, and it's useful when the URL itself is an ugly mess.
* The license field on PhysicalLocation is mostly hypothetical (I don't know if we'll populate it immediately), but license is core enough to the model that we do want it
* The shelfmark field on PhysicalLocation will be for shelf data on Sierra items

This patch just adds the field to our models and enough to get our existing tests compiling; actually populating these fields will come in another patch.

### DIFF
--- a/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayLocation.scala
+++ b/common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayLocation.scala
@@ -45,6 +45,10 @@ case class DisplayDigitalLocation(
     description = "Who to credit the image to"
   ) credit: Option[String] = None,
   @Schema(
+    `type` = "String",
+    description = "Text that can be used when linking to the item - for example, 'View this journal' rather than the raw URL"
+  ) linkText: Option[String] = None,
+  @Schema(
     description =
       "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made."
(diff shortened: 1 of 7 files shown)


common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayLocation.scala
```diff
@@ -44,6 +44,11 @@ case class DisplayDigitalLocation(
     `type` = "String",
     description = "Who to credit the image to"
   ) credit: Option[String] = None,
+  @Schema(
+    `type` = "String",
+    description =
+      "Text that can be used when linking to the item - for example, 'View this journal' rather than the raw URL"
+  ) link_text: Option[String] = None,
```
jamesgorrie, line 51: Any reason we're using `link_text` over `linkText`? I saw it in the RFC, should have asked there.

alexwlchan: The reason is that I was doing all the experiments in Python, and forgot we used camelCase in the catalogue. Fixing…

jamesgorrie (approved):

PR #1362, opened 2021-02-09
Fix a flaky test in the matcher

alexwlchan:
Logging from CI shows that this test is flaky, and sometimes it doesn't try to log the component ID 'ABC'.  That would make sense if it doesn't get that component from the store -- if this future hasn't completed yet.

I'm hoping that waiting for the future to complete will make this test stop being flaky in CI.

Tentatively fixes https://github.com/wellcomecollection/platform/issues/5016 – the logging of the exact locks helped me find out what was really behind this failure.


jamieparkinson (approved):
✨

PR #1363, opened 2021-02-09
Deletion checker: do the actual checking

jamieparkinson:
This contains 3 things:
- A [refactor of `CalmQuery`](https://github.com/wellcomecollection/catalogue/pull/1363/files#diff-687e5165a6d8e39e55ac1371e18ff2fc94c2d30f41aaaeec5ad045095f798e41) that enables representation of queries by simple syntax trees, thus enabling boolean composition of individual queries.
- A [refactor of test fixtures](https://github.com/wellcomecollection/catalogue/pull/1363/files#diff-a5af0a850299521fdf020132727c8b65bda97a96b732ddcd7894ddc2abcb2d8d) related to the Calm API client and the HTTP client class that is a dependency of the former.
- **Most importantly**, [the addition of `DeletionChecker`](https://github.com/wellcomecollection/catalogue/pull/1363/files#diff-746e986dbc1e27fa35bf00d38e01fa2c64d0ed5a46cd01b7199e80cc2aaaa0aa) which implements Algorithm 1 of [Wang et al (2017)](https://arxiv.org/abs/1407.2283) in order to efficiently find deletions in a given set of records (see [the RFC](https://github.com/wellcomecollection/docs/tree/master/rfcs/032-calm-deletions#minimising-calm-api-queries) for more context). The tests for this class use property-based testing to confirm adherence to the bounds proved in the paper.

This does not contain e2e-style tests for the `DeletionCheckerWorkerService`, the diff is large enough already and that will come in my next PR.


alexwlchan:
👀

calm_adapter/calm_deletion_checker/src/main/scala/uk/ac/wellcome/platform/calm_deletion_checker/DeletionChecker.scala
```diff
@@ -1,14 +1,81 @@
...
-class DeletionChecker(calmApiClient: CalmApiClient) {
+class ApiDeletionChecker(calmApiClient: CalmApiClient)(
+  implicit ec: ExecutionContext)
+    extends DeletionChecker {
 
-  def deletedRecords(
-    batch: Seq[CalmSourcePayload]): Future[Set[CalmSourcePayload]] =
-    ???
+  def nDeleted(records: Records): Future[Int] =
+    calmApiClient
```
kenoir, line 68: Will how Calm deals with this query affect the performance of this implementation? If for example the time taken to return a result scales linearly with the number of ids to search for? I have not yet had any coffee, so please excuse me if this is a stupid question.

alexwlchan (commented):
I've had a look at the linked paper (👍 for link) and I'm happy this is an accurate implementation of the algorithm it describes. I've put in some comments about where the code could be made clearer, and I think Robert's question about Calm API performance is worth considering, but overall really nice patch.

calm_adapter/calm_deletion_checker/src/main/scala/uk/ac/wellcome/platform/calm_deletion_checker/DeletionChecker.scala
```diff
@@ -1,14 +1,81 @@
...
+  // While there are more optimal algorithms that exist, the implementation of this one is very simple
+  def deletedRecords(batch: Records)(
+    implicit ec: ExecutionContext): Future[Records] = {
+    def findDeleted(records: Records, d: Int): Future[Records] = d match {
+      case 0                      => Future.successful(Set.empty)
+      case n if n == records.size => Future.successful(records)
+      case _ =>
+        val testSet = records.take(M(records.size, d))
+        for {
+          d1 <- nDeleted(testSet)
```
kenoir, line 30: So this works because something like: deletions are rare, so pick bigger sets of records to dismiss + do something clever to work out the optimal size based on how many deletions you see.

kenoir (approved):

calm_adapter/calm_deletion_checker/src/main/scala/uk/ac/wellcome/platform/calm_deletion_checker/DeletionChecker.scala
jamieparkinson: It is not a stupid question - yes, it will affect the performance, but while I didn't benchmark anything properly when I was testing out the big queries they seemed to scale sub-linearly (or at least the things it did that might have scaled linearly were dwarfed by other constant factors). There is probably an optimal batch size to balance API perf and number of queries but for now I think we'll just see if this is good enough...

calm_adapter/calm_api_client/src/test/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClientTest.scala
```diff
@@ -131,17 +128,17 @@ class CalmApiClientTest
 
   it("fails on error responses from the API") {
     val responses = List(HttpResponse(500, Nil, "Oops", protocol))
-    withApiClient(responses) {
+    withCalmClients(responses) {
       case (apiClient, _) =>
         whenReady(apiClient.search(query).failed) { error =>
-          error.getMessage shouldBe "Unexpected status from CALM API: 500 Internal Server Error"
+          error.getMessage shouldBe "Max retries attempted when calling Calm API"
```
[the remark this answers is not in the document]
jamieparkinson: I changed the test fixtures to all use a client with exponential backoff (as is used IRL) rather than using a one-shot client fixture. I may come back round later and see if I can clean things up having removed that one-shot case.

calm_adapter/calm_deletion_checker/src/main/scala/uk/ac/wellcome/platform/calm_deletion_checker/DeletionChecker.scala
```diff
@@ -1,14 +1,81 @@
...
+  CalmSession
+}
 import weco.catalogue.source_model.CalmSourcePayload
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
+
+trait DeletionChecker {
+  type Record = CalmSourcePayload
+  type Records = Set[Record]
```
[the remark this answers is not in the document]
jamieparkinson: Good, point yes on all of these. Totally agree about matching published convention (even if it's weird) and had done that with variables if not types!

calm_adapter/calm_deletion_checker/src/main/scala/uk/ac/wellcome/platform/calm_deletion_checker/DeletionChecker.scala
jamieparkinson: Yep, that's a good summary. This algorithm being "nested" is a constraint that the authors (or rather, a previous paper) made up to make the problem more tractable, it's something along the lines of only recursing through strict subsets.

[This paper](http://www.learningtheory.org/colt2009/papers/004.pdf) is truly optimal and is very clever indeed, I would have liked to use it and did actually implement a good chunk of it but when I got to writing a fast Fourier transform in Scala I realised what I was doing and went back to the simpler one.

calm_adapter/calm_adapter/src/test/scala/uk/ac/wellcome/calm_adapter/CalmRetrieverTest.scala
```diff
@@ -27,11 +25,9 @@ class CalmRetrieverTest
     with Matchers
     with ScalaFutures
     with Akka
+    with CalmApiTestClient
```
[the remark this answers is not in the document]
jamieparkinson: Would be cool if IntelliJ could pick up redundant trait-ing as we have lots of it, maybe one day

PR #1365, opened 2021-02-10
We shouldn't expose the name of our closed stores

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5022

We don't want to expose the internal layout of our closed stores to users, so in these cases we should set the label to "Closed stores".


jamesgorrie (approved):
Great!

PR #1367, opened 2021-02-10
 If the location is "in above", try to guess the locationType

alexwlchan:
Some Sierra locations have names like "bound in above" or "closed in above", which is completely unhelpful for mapping to our fixed vocabulary of location types.

We assume that "in above" refers to a location on one of the other items on this bib, so if we can guess that unambiguously, use that as the location for this item.

This gets the proportion of locations we can map into the new model from 95% to 99%.

Follows #1365; closes https://github.com/wellcomecollection/platform/issues/5019


jamesgorrie (approved):
Really nice tests, made the implementation way easier to understand.

PR #1374, opened 2021-02-11
More Calm API client refactoring

jamieparkinson:
The diff does show a net increase in code... but I did add some tests!

In splitting out a client for the Calm API I had ended up with a `CalmHttpClient` and a `CalmApiClient`, which shared concerns and were generally confusing. This removes the former and tries to keep logic grouped by concern.

These changes make it much easier to test things which use an API client (by providing a `TestCalmApiClient`) without coupling the tests to the underlying HTTP requests/responses. For testing the client itself, there is a `TestHttpCalmApiClient` which works, as before, at the request/response level.

Resolves https://github.com/wellcomecollection/platform/issues/5027


calm_adapter/calm_api_client/src/main/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClient.scala
```diff
@@ -63,5 +99,25 @@ class CalmApiClient(
...
+    extends HttpCalmApiClient(
+      url,
+      username,
+      password,
+      minBackoff,
+      maxBackoff,
+      randomFactor,
+      maxRestarts)
+    with AkkaHttpClient {
+  implicit val actorSystem: ActorSystem = mat.system
```
jamesgorrie, line 122: Just checking my understanding as to why this is necessary - it's so we can test the `CalmApiClient` without having to worry about the AKKA stuff?

jamieparkinson: Not so much akka stuff, it's so we can test it without making real HTTP requests

jamesgorrie: I'll read through this again, but then why is it called `HttpCalmApiClient`?

Would it make sense to have the same pattern of `MemoryCalmApiClient` for testing and `HttpCalmApiClient` for when we interact with HTTP?

jamieparkinson: Hmm, I don't think so - there are 2 use cases in test. One is where we don't care how the `CalmApiClient` works, and for that we use `TestCalmApiClient`, which is straightforward.

However, we have an implementor of `CalmApiClient` which is one that deals with `HttpResponse`s and `HttpRequest`s - that is therefore called `HttpCalmApiClient`.

It's non-trivial to do all the right things with this, so we want to test the client itself, but we don't really want to be doing actual HTTP requests because that becomes a headache.
So, we require that `HttpCalmApiClient`s must extend an `HttpClient`, which we can then use to provide responses without making requests in `TestHttpCalmApiClient` or we can use "for real" with `AkkaHttpCalmApiClient`.
This is actually very similar to how it was before (including naming), but using self-types like this makes it easier to mock in the right places than using constructor parameters like before.

jamesgorrie: This makes more sense to me - thanks for running through it.

HTTP isn't using HTTP but is dealing with the `HTTPRequest/Response` types.

jamesgorrie (approved):

PR #1375, opened 2021-02-15
Extract CalmRecord to source_model

jamieparkinson:
This gets used in 3+ places now, so it no longer makes sense to duplicate


alexwlchan (approved):
More power to the weco namespace!

PR #1377, opened 2021-02-16
Add deletion checker terraform

jamieparkinson:
I imagine this'll need some more work as everything is glued together but it's a start. Most of it is moving around / renaming stuff that's used by the adapter; I've already checked that I can `state mv` it into compliance without causing a mess.


calm_adapter/terraform/deletion_checker_worker.tf
```diff
@@ -0,0 +1,58 @@
...
+  source = "../../infrastructure/modules/worker"
+
+  name = local.namespace
+
+  image = local.calm_adapter_image
+
+  env_vars = {
+    calm_api_url             = local.calm_api_url
+    deletion_checker_sqs_url = module.calm_deletions_checker_queue.url
+    adapter_sns_topic        = module.calm_adapter_topic.arn
```
alexwlchan, line 20: I would use a separate topic for the adapter and deletion worker. It doesn't cost any extra, and being able to tap the outputs separately can be useful.

calm_adapter/terraform/deletion_checker_worker.tf
```diff
@@ -0,0 +1,58 @@
+module "calm_deletions_checker_queue" {
+  source                     = "git::github.com/wellcomecollection/terraform-aws-sqs//queue?ref=v1.1.2"
+  queue_name                 = "calm-deletions-checker"
```
alexwlchan, line 3: minor:

```suggestion
  queue_name                 = "calm-deletions-checker-input"
```

I know it being an input is implicit from being a queue, but that's not always the case, and I find it useful to have them distinguished when I'm looking through the console.

calm_adapter/terraform/deletion_checker_worker.tf
```diff
@@ -0,0 +1,58 @@
...
+    vhs_dynamo_table_name    = module.vhs.table_name
+    batch_size               = 1000
+  }
+  secret_env_vars = {
+    calm_api_username = "calm_adapter/calm_api/username"
+    calm_api_password = "calm_adapter/calm_api/password"
+  }
+
+  min_capacity = 0
+  max_capacity = 2
```
alexwlchan, line 30: How did you choose this limit? Is it to reduce load on the Calm API, or some other reason?

calm_adapter/terraform/vhs.tf
```diff
@@ -7,7 +7,12 @@ module "vhs" {
...
 }
 
-resource "aws_iam_role_policy" "vhs_readwrite" {
-  role   = module.worker.task_role_name
+resource "aws_iam_role_policy" "vhs_adapter_readwrite" {
+  role   = module.adapter_worker.task_role_name
   policy = module.vhs.full_access_policy
 }
+
+resource "aws_iam_role_policy" "vhs_delection_checker_dynamo_update" {
```
alexwlchan, line 15:
```suggestion
resource "aws_iam_role_policy" "vhs_deletion_checker_dynamo_update" {
```

alexwlchan (approved):

calm_adapter/terraform/deletion_checker_worker.tf
jamieparkinson: Good point (that will require another subscription in the transformer though)

calm_adapter/terraform/deletion_checker_worker.tf
jamieparkinson: No choice, just copy and paste - I may actually see how it performs with only 1

PR #1379, opened 2021-02-17
Add a script for fetching a snapshot of Sierra records

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5025

This script saves a snapshot of our Sierra records to S3. This is what an individual line of that snapshot looks like (pretty-printed):

```json
{
  "id": "1806550",
  "location": {
    "bucket": "wellcomecollection-vhs-sierra-sierra-adapter-20200604",
    "key": "1806550/0/e19cd2db-fcff-4043-a496-aee96cbf7b4a"
  },
  "record": {
    "bib": {
      "data": {
        "deleted": true,
        "deletedDate": "2016-05-11",
        "fixedFields": {
        },
        "id": "1806550",
        "locations": [

        ],
        "orders": [

        ],
        "varFields": [

        ]
      },
      "id": {
        "recordNumber": "1806550"
      },
      "modifiedDate": "2016-05-11T23:59:59.999999999Z"
    },
    "items": {
    }
  }
}
```

Actually building the whole thing takes a very long time, so it includes the location of the S3 object pointed to by DynamoDB. If that hasn't changed since the last time the snapshot was fetched, we can skip re-fetching it from S3.

Long-term I think we should consider moving away from using Dynamo/S3 pointers in this way. We could fit 99% of Sierra records into a DynamoDB row, and just move out the largest ones (with a conservative limit, say 200KB so the record would fit in SNS/SQS and DynamoDB). That would reduce costs and make them much easier to manage.

### DIFF
--- a/sierra_adapter/.gitignore
+++ b/sierra_adapter/.gitignore
@@ -1,0 +1,1 @@
+_snapshots
--- a/sierra_adapter/dynamo.py
+++ b/sierra_adapter/dynamo.py
@@ -1,0 +1,26 @@
+import decimal
+import json
+
+
+class DecimalEncoder(json.JSONEncoder):
+    def default(self, value):
+        if isinstance(value, decimal.Decimal) and int(value) == value:
+            return int(value)
+
+
+def get_table_item_count(session, *, table_name):
+    """
+    Counts the items in a DynamoDB table.
+    """
+    dynamo = session.client("dynamodb")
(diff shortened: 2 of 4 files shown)


alexwlchan:
This is broken in some way – records are being presented in mismatched formats. I need to investigate further.

jamesgorrie (approved):
Nice

alexwlchan:
Closing in favour of having a proper Sierra indexer in the pipeline.

PR #1384, opened 2021-02-18
Handle an HTTP 413 Request Too Large error from Elasticsearch

alexwlchan:
This error occurs when we try to index more documents than Elasticsearch can handle in a single request. Unfortunately, we can't just trust the queue logic to redrive these and hope we'll luck out – we have single messages that can hit this error.

Thus, we divide and conquer -- split the list of documents in two, and index both halves separately.  Eventually we will either index everything, or find the single document that can't be indexed (and everything else will index successfully).

In theory this could cause recursion issues; unfortunately I couldn't see a good way to make this tail-recursive.

In practice I don't think that'll be too much of an issue -- we only see 413s a handful of times during a reindex, and I think a single recursion will fix it.  If not, we'll find out quickly!

Closes https://github.com/wellcomecollection/platform/issues/5036

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
@@ -61,19 +61,20 @@ class ElasticIndexer[T: Indexable](
             // Either we'll narrow it down to a single document that's to big to be
             // indexed, or we'll index everything successfully.
             case response if response.status == 413 =>
-
               // If there's only one document left, there's nothing to do --
               // this document is Just Too Big.
               // https://twitter.com/smolrobots/status/1001226918107246592
               if (documents.size == 1) {
-                warn(s"HTTP 413 from Elasticsearch for a single document (${indexable.id(documents.head)})")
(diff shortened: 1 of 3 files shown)


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
```diff
@@ -52,11 +52,47 @@ class ElasticIndexer[T: Indexable](
...
+            // An HTTP 413 Request Too Large error means we tried to index too
+            // many documents at once.  In this case, we split the list in half
+            // and try to index both halves separately.
+            //
+            // Either we'll narrow it down to a single document that's to big to be
+            // indexed, or we'll index everything successfully.
+            case response if response.status == 413 =>
+              // If there's only one document left, there's nothing to do --
+              // this document is Just Too Big.
+              // https://twitter.com/smolrobots/status/1001226918107246592
```
jamesgorrie, line 66: 🤖

jamesgorrie (approved):
Really nice solution

jamieparkinson:
fwiw I don't think you need to worry about tail recursion as - because we're traversing futures - stacks won't be shared between the different invocations

PR #1386, opened 2021-02-18
Fixes for reliability in the 2021-02-17 pipeline

alexwlchan:


common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
```diff
@@ -72,13 +73,26 @@ class ElasticIndexer[T: Indexable](
 
               // Slice the documents in two, and index them both separately.
               // We'll combine the results.
+              //
+              // The "trace ID" allows us to follow multiple requests through the logs.
```
jamieparkinson, line 77: Good idea

jamieparkinson (approved):

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/ElasticIndexer.scala
alexwlchan: I'm honestly tempted to rip all of this logic out and just have it bubble up as an error, but this'll do for now.

PR #1388, opened 2021-02-18
Add Calm deletion check initiator (CLI + Lambda)

jamieparkinson:


alexwlchan (approved):
👍 

I feel like we should really move the reindexer logic into a Lambda at some point so it can be shared, but this is good enough for now.

PR #1393, opened 2021-02-22
The Sierra adapter should behave in an idempotent way

alexwlchan:
Spotted while working on wellcomecollection/platform#5003

The Sierra adapters have the following failure mode:

* $service receives an update A
* it stores update A
* it tries to send update A to the next $service, but fails
* $service re-receives update A
* it sees update A is already stored, and discards it

This is not ideal – it's better to always resend an update if it looks the same as what we've already got, just to be sure. (We still discard updates that are strictly older than what's stored.)

This means we can force a service to resend the onward message by resending the input message.


jamieparkinson (approved):
Do we know whether this might cause a load more redundant work?

alexwlchan:
> Do we know whether this might cause a load more redundant work?

It will cause some, because we deliberately overlap the Sierra update windows.

PR #1394, opened 2021-02-22
Write the Scala for a new Sierra linker

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/5003

This is way too big to be reviewed in its current state, but it illustrates the general trend: refactor the Sierra items to Dynamo into a general-purpose Sierra “linker” which can link holdings/items (similar to the reader).

Quick reminder: bibs become Works, and both holdings and items link to bibs. Holdings also have an itemIds field, but it looks like we can ignore that – if holdings `H` links to item `I`, and item `I` links to bib `B`, then holdings `H` explicitly links to bib `B`. (i.e. links are transitive)

Screenshot 2021-02-22 at 14 15 35

I'm going to break it up into smaller pieces.


alexwlchan:
All the relevant pieces of this have been merged now.

PR #1399, opened 2021-02-23
Make it easier to understand why a Work is Deleted/Invisible

alexwlchan:
Follows #1398. This is the patch I was actually writing.

A conversation in the #digirati channel this morning had me looking in the API index at invisible works. We do have a field where we can record why a particular work was deleted/suppressed/marked as invisible, but we weren't using it very consistently. This patch makes more use of that field, so it will be easier to understand work visibility in future.

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
@@ -65,7 +65,7 @@ object Work {
     version: Int,
     data: WorkData[State#WorkDataState],
     state: State,
-    deletedReason: Option[DeletedReason] = None,
+    deletedReason: DeletedReason,
   ) extends Work[State]
 }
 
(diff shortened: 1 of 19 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -65,7 +65,7 @@ object Work {
     version: Int,
     data: WorkData[State#WorkDataState],
     state: State,
-    deletedReason: Option[DeletedReason] = None,
+    deletedReason: DeletedReason,
```
jamieparkinson, line 68: I think this'll upset the api wrt the existing index fyi

jamieparkinson (approved):

common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
alexwlchan: Yeah, I'm freezing the current pipeline.

kenoir (approved):

PR #1400, opened 2021-02-23
Speculative Calm API client improvements

jamieparkinson:
I suspect that a major factor in the fact that the deletion checker kicked over the Calm API was not abandoning sessions created by searches - I'd hoped that it'd be OK with this, but perhaps not. Depending on how lazy its data fetching is, it may be holding the search results (on the order of thousands of records _per batch_) for each created session until expiry. That would be bad.

So, in the absence of any more info, I've made it abandon each session and also log a bit more. I won't be trying this out until I've managed to talk to someone from D&T about the Calm server.

### DIFF
--- a/calm_adapter/calm_api_client/src/main/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClient.scala
+++ b/calm_adapter/calm_api_client/src/main/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClient.scala
@@ -1,5 +1,6 @@
 package uk.ac.wellcome.platform.calm_api_client
 
+import akka.Done
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model._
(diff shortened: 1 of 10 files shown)


jamesgorrie (approved):
Nice speculation! Hopefully it's right.

calm_adapter/calm_api_client/src/main/scala/uk/ac/wellcome/platform/calm_api_client/CalmApiClient.scala
```diff
@@ -32,6 +33,9 @@ trait CalmApiClient {
     implicit p: CalmHttpResponseParser[CalmSummaryRequest])
     : Future[CalmRecord] =
     request(CalmSummaryRequest(pos), cookie)
+
+  def abandon(cookie: Cookie): Future[Done] =
+    request(CalmAbandonRequest, Some(cookie))
```
alexwlchan, line 38: keyword arguments pls!

calm_adapter/calm_deletion_checker/src/main/scala/uk/ac/wellcome/platform/calm_deletion_checker/DefectiveChecker.scala
```diff
@@ -65,16 +66,23 @@ trait DefectiveChecker[Item] {
...
-      )
-      .map {
+    for {
+      session <- calmApiClient
+        .search(
+          records
+            .map(record => CalmQuery.RecordId(record.id))
+            .reduce[CalmQuery](_ or _)
+        )
+      _ <- calmApiClient.abandon(session.cookie).recover {
```
alexwlchan, line 80: Probably worth a comment explaining why we need to Abandon sessions here? (Or think we need to.)

PR #1403, opened 2021-02-23
Remove .jpg from iiif digital locations in miro transformer

kenoir:
Due to the recent changes to register miro images with DLCS and update the IIIF CloudFront distribution, the ".jpg" in IIIF identifiers is no longer required.

e.g. https://iiif.wellcomecollection.org/image/B0009851/full/1338%2C/0/default.jpg

This change removes the additional `.jpg` required by images registered with Loris.


alexwlchan (approved):
Closes https://github.com/wellcomecollection/platform/issues/5023

PR #1404, opened 2021-02-23
Add availabilities property to work

jamieparkinson:
What this does:
- Adds a type `Availability` for representing works that are available `Online` or `InLibrary`.
- Adds a list of these, `availabilities`, to a work's `derivedData`.
- Adds these `availabilities` to the display model.
- Removes some logic related to en/decoding enumeratum enums which enumeratum could already do for us.

What this does not do:
- `availableOnline` is still there, it'll be removed when this feature is done.
- There is no filtering or aggregating on `availabilities` - this comes next.
- Archives do not yet inherit availability from their `parts`

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/License.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/License.scala
@@ -2,49 +2,27 @@ package uk.ac.wellcome.models.work.internal
 
 import enumeratum.EnumEntry
 import enumeratum.Enum
-import io.circe.{Decoder, Encoder, Json}
+import io.circe.{Decoder, Encoder}
 
 sealed trait License extends EnumEntry {
   val id: String
   val label: String
   val url: String
+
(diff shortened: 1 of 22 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/License.scala
```diff
@@ -2,49 +2,27 @@ package uk.ac.wellcome.models.work.internal
...
-      for {
-        id <- cursor.downField("id").as[String]
-      } yield {
-        createLicense(id)
-      }
-  }
+  implicit val licenseEncoder: Encoder[License] =
+    Encoder.forProduct1("id")(_.id)
 
-  private val licenseIdIndex = {
```
jamieparkinson, line 32: https://github.com/lloydmeta/enumeratum/blob/master/enumeratum-core/src/main/scala/enumeratum/Enum.scala#L41

alexwlchan (approved):

PR #1406, opened 2021-02-24
Rename "items to Dynamo" to "Sierra linker"

alexwlchan:
For wellcomecollection/platform#5003, follows #1405, pulled out of #1394

This uses the new name in things like the build system, ECR repos, and our Makefiles.

The old name still persists in our Terraform services; I'll get to that separately.


kenoir (approved):
Nice work.

PR #1407, opened 2021-02-24
Create items from field 856 on bibs

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5050, based on the current state of https://github.com/wellcomecollection/docs/pull/48

This adds an initial implementation of field 856 on bibs, so we can deploy something that the Experience team can use and experiment with. We may come back and tweak some of these rules later.

The parsing code for field 856 is split out as separate from the bib/item parsing logic because I expect we'll want to reuse it when we pull field 856 from holdings records.


kenoir (approved):
👏 well tested. Nice.

PR #1408, opened 2021-02-24
Start to make the Sierra adapter "merger" more generic

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/5003

We need a service that can attach a SierraHoldingsRecord to our existing SierraTransformable model.

We could take the existing items merger, copy all the code, and do a find/replace – but actually I'd like to take us in a different direction, and make the "Sierra merger" a generic application that can do bibs/items/holdings depending on config.

The two new type classes (renamed versions of existing classes) are:

* RecordOps takes a SierraRecord and works out what bib IDs it's linked to/unlinked to.

    For item/holdings records we'll get this from the linker; for bib records this is somewhat trivial.

* TransformableOps can combine/create a SierraTransformable from a SierraRecord.

And this allows the merger behaviour to be very generic:

* Get a record from SQS
* For every bib in RecordOps.getLinkedBibIds(r), call TransformableOps.add(transformable, r) and store the result
* For every bib in RecordOps.getUnlinkedBibIds(r), call TransformableOps.remove(transformable, r) and store the result

Although it's a big diff, most of it is moving around existing tests and updating names, rather than changing anything substantial.

The fun part will be the next patch, when I add bibs/holdings support with very little code added.


jamieparkinson (approved):
Neat

PR #1409, opened 2021-02-24
Add availabilities filter and aggregation

jamieparkinson:
Think the contents of this should be fairly predictable. Did some more cleanup of enumeratum + circe logic while I was at it.


alexwlchan (approved):
Looks good. We also have tests for filters+aggregations that you might want to add a test for, but otherwise fine.

PR #1401, opened 2021-02-23
Everything is DLCS now; always download thumbnail images

jamieparkinson:

PR #1412, opened 2021-02-24
Create the 2021-02-24 pipeline

alexwlchan:
This brings in the following changes:

- We get titles for a few more images by looking at subfield $a on field tag $v
- Make it easier to understand why a Work is Deleted/Invisible. We expect to see a lot of Works move from Invisible to Deleted in this reindex. #1399
- The image inferrer will fetch images from DLCS (#1401), and the IIIF Image URLs will point exclusively to the DLCS images (#1403).
- Works now have the "availabilities" property (#1404)
- We get digital items from field 856 on bibs (#1407)


alexwlchan:
This index is deployed! 🎉

jamieparkinson (approved):

PR #1414, opened 2021-02-24
Add access conditions for images from Miro

alexwlchan:
This means Miro images will get populated with Jamie's new "availabilities" field.


jamieparkinson (approved):
Thank you @alexwlchan 🎉

PR #1416, opened 2021-02-25
Refactor the SierraWindowManager class

alexwlchan:
Non-urgent refactoring I spotted while looking at #1415

This makes more use of our shared storage libraries, and more use of Try/Either monads to pass around results – rather than throwing unhandled exceptions. 😱

### DIFF
--- a/sierra_adapter/sierra_reader/src/main/scala/uk/ac/wellcome/platform/sierra_reader/services/SierraReaderWorkerService.scala
+++ b/sierra_adapter/sierra_reader/src/main/scala/uk/ac/wellcome/platform/sierra_reader/services/SierraReaderWorkerService.scala
@@ -37,16 +37,16 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class SierraReaderWorkerService(
   sqsStream: SQSStream[NotificationMessage],
-  s3client: AmazonS3,
   s3Config: S3Config,
   readerConfig: ReaderConfig,
   sierraAPIConfig: SierraAPIConfig
-)(implicit actorSystem: ActorSystem, ec: ExecutionContext)
+)(implicit
+  actorSystem: ActorSystem,
+  ec: ExecutionContext,
+  s3Client: AmazonS3)
     extends Logging
     with Runnable {
-  implicit val s = s3client
   val windowManager = new WindowManager(
-    s3client = s3client,
     s3Config = s3Config,
     readerConfig = readerConfig
   )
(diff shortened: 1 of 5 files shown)


sierra_adapter/sierra_reader/src/main/scala/uk/ac/wellcome/platform/sierra_reader/services/SierraReaderWorkerService.scala
```diff
@@ -37,16 +37,16 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class SierraReaderWorkerService(
   sqsStream: SQSStream[NotificationMessage],
-  s3client: AmazonS3,
   s3Config: S3Config,
```
jamieparkinson, line 40: Why do we have to pass around a client and a config? Can we not build one from the other?

sierra_adapter/sierra_reader/src/test/scala/uk/ac/wellcome/platform/sierra_reader/services/WindowManagerTest.scala
```diff
@@ -139,4 +137,15 @@ class WindowManagerTest
...
     }
   }
+
+  it("throws an error if it can't list the contents of the bucket") {
+    withWindowManager(createBucket) { windowManager =>
+      val future =
+        windowManager.getCurrentStatus(s"[$startDateTime,$endDateTime]")
+
+      whenReady(future.failed) {
+        _ shouldBe a[AmazonS3Exception]
```
jamieparkinson, line 147:
```suggestion
        _ shouldBe an[AmazonS3Exception]
```

jamieparkinson (approved):

sierra_adapter/sierra_reader/src/main/scala/uk/ac/wellcome/platform/sierra_reader/services/SierraReaderWorkerService.scala
alexwlchan: The config is telling us which bucket to use in S3, which is different from how actually connect to S3, although I can see it might not be immediately obvious.

jamieparkinson: Ah, is the config that the client is built from different to the one that gets passed in?

alexwlchan: Different to the one that gets passed in here, yep.

(I don't think we use any config directly for the client, actually, we just pull in creds from the environment.)

PR #1420, opened 2021-02-25
Inline the sierra-streams-source library

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5065

It turns out the Sierra API is a bit inconsistent in output: if you fetch bibs or items, IDs are returned as Strings. If you fetch holdings, IDs are returned as Numbers. This causes an issue for anything that deals with Sierra API responses, because until now it's been able to treat everything as a string.

This is unhelpful.

To fix this, we need to patch [sierra-streams-source](https://github.com/wellcomecollection/sierra-streams-source), the library that we use to read from Sierra. This has drifted spectacularly from our current best practices, and the state of its logging made it quite hard to diagnose this issue. Rather than try to resurrect a deployment process that's withered, I'm going to inline it into this repo and update it in-place – at this point there seems little value in keeping it separate.

This patch just brings in the existing code and gets the tests passing. Actually bringing it up-to-date and fixing it for what we need will come in a separate patch.


kenoir (approved):
👍 think it's fine to drop the external lib TBH, I don't think it's get any consumers other than us , and at this point it's a bit hopeful to assume there might be.

alexwlchan:
> 👍 think it's fine to drop the external lib TBH, I don't think it's get any consumers other than us , and at this point it's a bit hopeful to assume there might be.

I've archived the repo just to be sure!

PR #1422, opened 2021-02-26
Handle Sierra IDs as strings or ints in API responses

alexwlchan:
Tentative fix for https://github.com/wellcomecollection/platform/issues/5065, follows #1420 and #1421

This patch updates the Sierra adapter to parse IDs as strings or ints, because the API responses are inconsistent. We stringify internally, but we'll need to be able to keep reading them as both because we pass around the raw Sierra API responses.

### DIFF
--- a/sierra_adapter/common/src/main/scala/weco/catalogue/sierra_adapter/json/JsonOps.scala
+++ b/sierra_adapter/common/src/main/scala/weco/catalogue/sierra_adapter/json/JsonOps.scala
@@ -8,9 +8,13 @@ object JsonOps {
   class StringOrInt(val underlying: String)
 
   implicit val decoder: Decoder[StringOrInt] =
-    (c: HCursor) => c.as[String] match {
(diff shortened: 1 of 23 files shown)


sierra_adapter/common/src/main/scala/weco/catalogue/sierra_adapter/json/JsonOps.scala
```diff
@@ -0,0 +1,22 @@
+package weco.catalogue.sierra_adapter.json
+
+import io.circe.{Decoder, Encoder, HCursor, Json}
+
+object JsonOps {
+  // Sierra API responses can return some IDs as a String or an Int.
+  // Parsing the value as an instance of this class will handle it for you.
+  class StringOrInt(val underlying: String)
+
+  implicit val decoder: Decoder[StringOrInt] =
```
kenoir, line 10: 🎩

kenoir (approved):

PR #1423, opened 2021-02-26
Add the Sierra indexer

alexwlchan:
This is a service that will index the contents of the Sierra catalogue in the reporting cluster. It builds on a Python prototype I built, which is in the reporting account, but we need more control over the concurrency to avoid breaking the reporting cluster.

It's not as well-tested as I'd like (in particular I can't test it removes outdated varfields/fixed fields, because I get 409 Conflict errors when trying to delete a just-indexed document locally), but it's good enough to deploy as a prototype and play with.

The trick is to index the top-level fields/varfields/fixed fields in separate indexes.

For https://github.com/wellcomecollection/platform/issues/5052


sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/JsonOps.scala
```diff
@@ -0,0 +1,18 @@
+package weco.catalogue.sierra_indexer.services
+
+import io.circe.Json
+
+object JsonOps {
```
jamieparkinson, line 5: `SierraJsonOps`?

sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
```diff
@@ -0,0 +1,156 @@
...
+                  index = fixedFieldsIndex,
+                  id = Some(s"${parent.id}-$code"),
+                  source = Some(
+                    IndexedFixedField(parent, code, fixedField).asJson.noSpaces
+                  )
+                )
+            }
+      }
+
+      fixedFieldDeletions = apiData.map {
```
jamieparkinson, line 109: Can the functions inside these `apiData.map/flatMap`s be extracted into methods so we can write something like:
```scala
for {
  apiData <- getSierraApiData(t)
  mainRecords = apiData.flatMap(IndexRequest.mainRecords)
  varFields = apiData.flatMap(IndexRequest.varFields)
  varFieldDeletions = apiData.map(IndexRequest.varFieldDeletions)
  fixedFields = apiData.flatMap(IndexRequest.fixedFields)
  fixedFieldDeletions = apiData.map(IndexRequest.fixedFieldDeletions)
} yield (
  mainRecords ++ varFields ++ fixedFields,
  varFieldDeletions ++ fixedFieldDeletions
)
```
They're a bit hard to follow and it would clarify that they are not interdependent (unless they are and I've missed that, in which case they could do with being more explicit anyway)

sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Worker.scala
```diff
@@ -0,0 +1,67 @@
...
+      .fromTry(ops)
+      .flatMap {
+        case (indexRequests, deleteByQueryRequests) =>
+          val futures = List(
+            elasticClient.execute(
+              bulk(indexRequests)
+            )
+          ) ++ deleteByQueryRequests.map { elasticClient.execute(_) }
+
+          Future.sequence(futures)
```
jamieparkinson, line 61: Bit of a nit but constructing single-element lists always feels uncomfortable to me if it's not done to emphasise the cardinality
```suggestion
          val futures = elasticClient.execute(bulk(indexRequests))
           +: deleteByQueryRequests.map { elasticClient.execute(_) }
           
          Future.sequence(futures)
```

sierra_adapter/sierra_indexer/src/test/scala/weco/catalogue/sierra_indexer/fixtures/IndexerFixtures.scala
```diff
@@ -0,0 +1,88 @@
...
+
+        val worker = new Worker(sqsStream, typedStore, indexPrefix)
+
+        worker.run()
+
+        testWith(worker)
+      }
+    }
+
+  def withIndexes[R](testWith: TestWith[String, R]): R = {
```
jamieparkinson, line 46: Indices!

jamieparkinson: This isn't _just_ pedantry because elastic do pluralise like that too, unlike (eg) most database vendors' indexes

alexwlchan: 👍

kenoir:
Bit more higher level context would be useful, i.e. why do we need this & where does this fit into the pipeline?

More specifically how can we reconstruct a record if it's spread across multiple indices?

sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
alexwlchan: Good shout, changed. Hopefully it’s a bit clearer now (and easier to test, too!).

sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Worker.scala
alexwlchan: Ah yeah, that's much nicer.

alexwlchan:
> Bit more higher level context would be useful, i.e. why do we need this

The goal is to make it easier to query Sierra data, so that we can work out how we want to model data, or find records matching a particular pattern.

e.g. for the recent digitised-items-on-bibs work, I wanted a list of all 856 varfields. There's no good way to get that with the adapter as it exists today, but pulling the varfields into a dedicated index makes that easy to get.

This also opens the door to doing more targeted indexes of Sierra data – if you know you're only changing records with a particular property in their MARC, you can find and reindex just those.

> where does this fit into the pipeline?

It listens to updates from the Sierra adapter, or can be triggered directly by the reindexer.

> More specifically how can we reconstruct a record if it's spread across multiple indices?

Fetch it directly from the Sierra adapter store.

jamieparkinson (approved):

PR #1428, opened 2021-03-01
Publish internal model

kenoir:
This is part of https://github.com/wellcomecollection/platform/issues/5041.
The changes in this PR will require to add a RELEASE.md file for every PR with changes that affect the `internal_model`.
A merge in master with changes to the internal model will result in:
- an `internal_model` artifact being published with a version like <buildkite_build_number>.<commit hash>
- a new line being written in CHANGELOG.md with the contents of RELEASE.md 
- a new tag being pushed with the new version

### DIFF
--- a/.buildkite/scripts/release.py
+++ b/.buildkite/scripts/release.py
@@ -8,8 +8,7 @@ import sys
 
 from commands import git
 from git_utils import (
-    get_changed_paths,
-    remote_default_branch,
+    get_all_tags,
     local_current_head,
     get_sha1_for_tag,
     remote_default_head,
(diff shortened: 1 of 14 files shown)


alexwlchan:
> The changes in this PR will require to add a RELEASE.md file for every merge in master with a summary of the changes.

Is cutting a release for every Scala change the best approach? Would it be better to scope it to changes that just affect internal_model?

(I stack up a lot of small PRs that really don't need an internal_model change cut, because they're fixing a test or modifying a service or whatever.)

.buildkite/scripts/release.py
```diff
@@ -0,0 +1,177 @@
...
+def has_release():
+    """
+    Returns True if there is a release file, False if not.
+    """
+    return os.path.exists(RELEASE_FILE)
+
+
+def read_release_file():
+    """
+    Parses the release file, returning a tuple (release_type, release_contents)
```
jamesgorrie, line 126: Is there some sort of pydoc that allows you to define return types?

build.sbt
```diff
@@ -2,278 +2,289 @@ import java.io.File
...
-  project
-    .in(new File(folder))
-    .settings(Common.settings: _*)
-    .settings(DockerCompose.settings: _*)
-    .enablePlugins(DockerComposePlugin)
-    .enablePlugins(JavaAppPackaging)
-    .dependsOn(dependsOn: _*)
-    .settings(libraryDependencies ++= externalDependencies)
-}
+val projectVersion = "0.0.0"
```
jamesgorrie, line 5: Why does every project need a version?

alicefuzier: yeah, they don't. Will change

PR #1429, opened 2021-03-01
Merge digcode identifiers from source works

jamieparkinson:
Resolves https://github.com/wellcomecollection/catalogue/issues/1427, see https://wellcome.slack.com/archives/C8X9YKM5X/p1614249633039500 for both context and justification for this approach to resolving the issue


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/OtherIdentifiersRule.scala
```diff
@@ -23,6 +23,8 @@ object OtherIdentifiersRule extends FieldMergeRule with MergerLogging {
 
   type FieldData = List[SourceIdentifier]
 
+  private val otherIdentifiersTypeAllowList = Set("wellcome-digcode")
```
alexwlchan, line 26: We should have a comment here that explains:

1. What this list is
2. Why each member of it exists

alexwlchan (approved):
👾

PR #1430, opened 2021-03-01
Include the item ID in a warning log about unrecognised locations

alexwlchan:
We get warnings in the Sierra transformer logs like:

> Unable to map Sierra location name to LocationType: 215 Information Point

For debugging, it would be useful if this warning included some context – which record does this value come from?

This patch adds that ID to this warning log.


jamesgorrie (approved):
Neat

PR #1431, opened 2021-03-01
Add access conditions to 856 and availability filters

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5062


alexwlchan:
I should add a test that a work with LicensedResource actually gets the available online filter in the API, because that's what this is all about.

jamesgorrie (approved):

PR #1432, opened 2021-03-02
Deploy the Sierra indexer

alexwlchan:
This adds the infra and weco-deploy config to actually get it deployed, and it's running now with a reindex to populate the initial data set. For https://github.com/wellcomecollection/platform/issues/5052

### DIFF
--- a/sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
+++ b/sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
@@ -36,10 +36,19 @@ class Splitter(indexPrefix: String) {
 
   private def getSierraApiData(t: SierraTransformable)
     : Either[Seq[(Parent, ParsingFailure)], Seq[(Parent, Json)]] = {
+    val itemIds = t.itemRecords.keys.map { _.withoutCheckDigit }.toList.sorted
+    val holdingsIds = t.holdingsRecords.keys.map { _.withoutCheckDigit }.toList.sorted
+
     val bibData = t.maybeBibRecord match {
       case Some(bibRecord) =>
         Seq(
-          Parent(SierraRecordTypes.bibs, bibRecord.id) -> parse(bibRecord.data)
+          Parent(SierraRecordTypes.bibs, bibRecord.id) ->
+            parse(bibRecord.data)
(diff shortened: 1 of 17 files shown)


sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
```diff
@@ -36,10 +36,25 @@ class Splitter(indexPrefix: String) {
...
     : Either[Seq[(Parent, ParsingFailure)], Seq[(Parent, Json)]] = {
+    val itemIds = t.itemRecords.keys.map { _.withoutCheckDigit }.toList.sorted
+    val holdingsIds =
+      t.holdingsRecords.keys.map { _.withoutCheckDigit }.toList.sorted
+
     val bibData = t.maybeBibRecord match {
       case Some(bibRecord) =>
         Seq(
-          Parent(SierraRecordTypes.bibs, bibRecord.id) -> parse(bibRecord.data)
+          Parent(SierraRecordTypes.bibs, bibRecord.id) ->
```
jamesgorrie, line 46: Could we use a case class to do this?

jamesgorrie (approved):
Great.

Wondering if we should be adding this to a shared architecture diagram somewhere?

If we don't have one (I can't find a canonical one), I would really like top pair on getting one setup.

sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
alexwlchan: Not really, or at least not without reharvesting all the data – I'll add a comment to clarify what it's doing here.

This app works mostly on `Json` instances, so we're passing around the data exactly as we get it from the Sierra API. If you turn that into a case class, you only get the fields you explicitly include on the case class.

This is injecting the "itemIds" and "holdingIds" fields alongside the bib data fields. It's slightly cheating, but such a useful piece of info I figured it worth adding. It mimics similar fields on the items/holdings records.

alexwlchan:
I've created a separate ticket to track adding the architecture diagram properly: https://github.com/wellcomecollection/platform/issues/5071

Here's a quick sketch of how the indexer works:

Screenshot 2021-03-03 at 10 49 07

PR #1436, opened 2021-03-04
Keep track of all the other Works that redirect to a given Work

alexwlchan:
This means that redirects are tracked symmetrically in the index: if `A -> B`, then we'll record this on both A and B.

This is just useful for debugging in general, and closes https://github.com/wellcomecollection/platform/issues/5043. We should be able to use this to find out if merged pairs haven't been indexed consistently when we do redirects.

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
@@ -111,7 +111,7 @@ trait Merger extends MergerLogging {
         source.sourceIdentifier,
         source.state.canonicalId,
         source.state.modifiedTime),
-      redirect =
+      redirectTarget =
         IdState.Identified(target.state.canonicalId, target.sourceIdentifier)
     )
 
@@ -148,8 +148,8 @@ object Merger {
           Work.Visible(version, f(data), state)
         case Work.Invisible(version, data, state, reasons) =>
           Work.Invisible(version, f(data), state, reasons)
-        case Work.Redirected(version, redirect, state) =>
-          Work.Redirected(version, redirect, state)
+        case Work.Redirected(version, redirectTarget, state) =>
+          Work.Redirected(version, redirectTarget, state)
         case Work.Deleted(version, data, state, reason) =>
           Work.Deleted(version, f(data), state, reason)
       }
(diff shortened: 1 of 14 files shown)


pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
```diff
@@ -96,8 +96,21 @@ trait Merger extends MergerLogging {
...
+            redirectedSources.map { s =>
+              IdState.Identified(s.id, s.sourceIdentifier)
+            }.toSeq
+
+          val targetWork: Work.Visible[Identified] =
+            Work.Visible[Identified](
+              version = result.mergedTarget.version,
+              data = result.mergedTarget.data,
+              state = result.mergedTarget.state,
+              redirectSources = result.mergedTarget.redirectSources ++ redirectedIdentifiers
```
jamieparkinson, line 109: Will this contain duplicates?

pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/services/MergerTest.scala
```diff
@@ -111,4 +129,33 @@ class MergerTest
...
+    val sourceWorks = (1 to 5).map { _ =>
+      identifiedWork()
+    }
+
+    val mergerOutput = FirstWorkMerger.merge(targetWork +: sourceWorks)
+
+    val mergedWork: Work.Visible[Merged] = mergerOutput
+      .mergedWorksWithTime(now)
+      .collect { case w: Work.Visible[Merged] => w }
+      .filter { w =>
```
jamieparkinson, line 150: You can do this filtering in the `collect`

jamieparkinson (approved):
The test improvements are good too

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
alexwlchan: I think it's theoretically possible, if the same target/source pair got merged with different rules?

I'm going to leave it as-is, because I don't think there's any problem with having duplicates here – this field shouldn't be displayed externally – and seeing lots of duplicates here might be a clue about weird merger behaviour.

PR #1439, opened 2021-03-04
Add an implementation of the Sierra 853/863 enumerations field

alexwlchan:
This is part of the work to add holdings to the API.

When you look at a journal on the current library site, you see a list of the issues that we have:

Screenshot 2021-03-08 at 09 40 42

These strings aren't stored directly in the MARC record. Instead, the record splits the display labels and values, and you have to reconstruct them into human-readable strings. For example, the record is something like:

```
853 v.|bno.|i(year)|j(month)  # labels - v., no., year/month
863 8-69|b-1|i1959-2020|j-21  # v.8 (1959) - v.69:no.1 (Spring 2020)
863 69|b2-3|i2020|j22-23      # v.69:no.2 (Summer 2020) - v.69:no.3 (Autumn 2020)
```

This gives you a lot of flexibility in the MARC, but we shouldn't expose this degree of granularity to API consumers. The catalogue pipeline should parse these MARC records, and turn them into a human-readable string.

This patch adds some code to do this transformation.

Unfortunately, I couldn't find a good set of instructions for how to do this, so I did it by observing the behaviour on the current Wellcome Library site. I saved the complete list of 85X/86X fields, the corresponding strings on the Library website, and I kept comparing the two. They're not an exact match, but I think they're good enough. We can always tweak that logic later (or in some cases, tweak the underlying Sierra record).

I've turned all the “interesting” examples from Sierra into explicit test cases, because I’m not sure there isn’t private information in my file full of Library examples.

Part of https://github.com/wellcomecollection/platform/issues/5076

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraHoldingsEnumeration.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraHoldingsEnumeration.scala
@@ -194,7 +194,7 @@ object SierraHoldingsEnumeration extends SierraQueryOps with Logging {
           // I haven't worked out the exact rules around this yet.
           // In some cases, the old Wellcome Library site would join parts with
           // a space.  In others (e.g. "v.130:no.3"), it uses a colon.
-          if (accum.startsWith("no.") && nextPart.startsWith("v.")) {
+          if (accum.startsWith("no.") && nextPart.startsWith("v")) {
             nextPart + ":" + accum
           } else {
             nextPart + " " + accum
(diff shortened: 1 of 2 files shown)


kenoir (approved):
MARC is absolutely bananas. Your test cases here are 👌 .

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraHoldingsEnumeration.scala
```diff
@@ -0,0 +1,325 @@
...
+        .flatMap { sf =>
+          label.varField.subfieldsWithTag(sf.tag).headOption match {
+            case Some(MarcSubfield(_, subfieldLabel)) =>
+              Some((subfieldLabel, sf.content))
+            case None => None
+          }
+        }
+        .filterNot { case (_, value) => value.trim == "-" }
+
+    // TODO: Explain why this is the case
```
kenoir, line 124: Is this missing?

alexwlchan: Yup, added a proper comment now.

alexwlchan:
🍌 🍌 🍌 

Thanks for reviewing!

PR #1440, opened 2021-03-05
Pipeline 2021-03-05

jamieparkinson:


pipeline/terraform/main.tf
```diff
@@ -1,8 +1,75 @@
...
+
+  pipeline_date = "2021-03-05"
+  release_label = "stage"
+
+  # Transformer config
+  #
+  # If this pipeline is meant to be reindexed, remember to uncomment the
+  # reindexer topic names.
+
+  is_reindexing = false
```
alexwlchan, line 12: Typically I reindex with this set to “true” – it adds more tasks and turns up RDS and the pipeline storage cluster.

might be interesting to see wa reindex with rate limiting, but it might be very slow

jamieparkinson: I have done that locally, wasn't going to commit it but actually I guess I may as well

alexwlchan: Ah, doesn't need to be committed, as long as you've got it.

PR #1441, opened 2021-03-08
Normalise genres - strip punctuation/whitespace, fix capitalisation

alexwlchan:
None of these transformations affect the meaning of this data, but it should display more consistently, e.g. users won't see two entries for "Electronic [Bb]ooks" in the genre filter.

Closes https://github.com/wellcomecollection/platform/issues/5072

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,1 @@
+Add more options for normalising `Genre` data.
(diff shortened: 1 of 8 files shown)


jamesgorrie (approved):
Just to check I'm clear this will be in the data, rather than just the display?

alexwlchan:
> Just to check I'm clear this will be in the data, rather than just the display?

Yeah, this will be applied in the transformer – so things like aggregations and filters will use the normalised values.

PR #1447, opened 2021-03-08
Try to fix the tags used in CI

alexwlchan:
We've overloaded the meaning of "latest" tag, and I think that's caused the current issues.

It simultaneously means the commit that was used to create images pushed by Buildkite _and_ is being used as a point of comparison for the sbt release mechanism.

This patch introduces a new tag `latest-sbt-release` to pick up the latter meaning, which can be pushed/released independently of the Docker images.

🤞 this sorts out CI.


kenoir (approved):
Nice

jamieparkinson (approved):

PR #1450, opened 2021-03-09
Re-check that we have relevant changes before looking for a release file

alexwlchan:
This is yet another attempt to stabilise CI with the internal_model deployment.

Here's my current best guess for what's going on, based on looking at https://buildkite.com/wellcomecollection/catalogue/builds/3342#0b14305c-7c3c-4b71-82e6-8309380ba8e2

Suppose we have the following sequence of commits:

Screenshot 2021-03-09 at 09 40 14

When you merge, we compare "merge commit" and "where you forked" – which shows up all the changes in both the yellow/square and green/triangle commits. This means that if there are changes to `internal_model` in the yellow/square commits, we'll run the `internal_model` task.

But the diff we really care about for checking a release note is between "current state of master" and "merge commit" – so let's re-check for relevant changes on the commit range we care about, before breaking a build for non-release note.

### DIFF
--- a/.buildkite/scripts/run_job.py
+++ b/.buildkite/scripts/run_job.py
@@ -79,7 +79,7 @@ if __name__ == "__main__":
     local_head = local_current_head()
 
     if is_default_branch():
-        latest_sha = get_sha1_for_tag("latest")
+        latest_sha = get_sha1_for_tag("latest-sbt-release")
         commit_range = f"{latest_sha}..{local_head}"
     else:
         remote_head = remote_default_head()
@@ -120,14 +120,7 @@ if __name__ == "__main__":
             sys.exit(0)
 
     if should_check_release(args.project_name):
-        latest_sha = get_sha1_for_tag("latest-sbt-release")
-        commit_range = f"{latest_sha}..{local_head}"
-        changed_paths = get_changed_paths(commit_range, globs=change_globs)
-
-        if should_run_sbt_project(sbt_repo, args.project_name, changed_paths):
-            check_release_file(commit_range)
-        else:
-            print(f"Not checking release file because there are no changes between latest-sbt-release and this commit")
+        check_release_file(commit_range)
 
     # Perform make tasks
     make(f"{args.project_name}-test")
--- a/.buildkite/scripts/run_job.py
+++ b/.buildkite/scripts/run_job.py
@@ -122,7 +122,12 @@ if __name__ == "__main__":
     if should_check_release(args.project_name):
         latest_sha = get_sha1_for_tag("latest-sbt-release")
         commit_range = f"{latest_sha}..{local_head}"
-        check_release_file(commit_range)
+        changed_paths = get_changed_paths(commit_range, globs=change_globs)
+
+        if should_run_sbt_project(sbt_repo, args.project_name, changed_paths):
+            check_release_file(commit_range)
+        else:
+            print(f"Not checking release file because there are no changes between latest-sbt-release and this commit")
 
     # Perform make tasks
     make(f"{args.project_name}-test")


.buildkite/scripts/run_job.py
```diff
@@ -122,7 +122,14 @@ def should_check_release(project_name):
     if should_check_release(args.project_name):
         latest_sha = get_sha1_for_tag("latest-sbt-release")
         commit_range = f"{latest_sha}..{local_head}"
-        check_release_file(commit_range)
+        changed_paths = get_changed_paths(commit_range, globs=change_globs)
+
+        if should_run_sbt_project(sbt_repo, args.project_name, changed_paths):
```
alicefuzier, line 127: I think I'm missing something... Why would the changed paths be different here than on line 112?

jamieparkinson: Agreed - I think the problem is that `commit_range` (line 83) is using the `latest` tag rather than the new `latest-sbt-release` tag (?)

jamieparkinson: And as we found yesterday, that tag is on the commit _before_ the version bump

alexwlchan: So the fix might be to compare to `latest-sbt-release` on line 83 instead?

alexwlchan: > Why would the changed paths be different here than on line 112?

The changed paths on line 112 are `latest..HEAD`. The changed paths here are `latest-sbt-release..HEAD`. Those two tags are set and managed differently – as Jamie says, one is before the version bump, the other is after.

(We created the `latest-sbt-release` tag yesterday to specifically point to the version bump commit.)

jamieparkinson: I think that's right, which makes me wonder if we do need 2 separate tags - where would we be using `latest` if that's the comparison we were making?

alexwlchan: We use `latest` in weco-deploy, right?

alexwlchan: Is `latest` something we copied across from scala-libs but doesn't make sense here sans version bump?

jamieparkinson: I _think_ that the `latest` in weco-deploy is an unrelated one to this git tag

alexwlchan: Note that even if we sort out the `latest` tag, I don't think we're out of the woods – there's about a 20–30 min delay between you merging to master and the internal_model being released/tagged, which could potentially break subsequent builds.

(I thought this was the issue with your PR yesterday.)

jamieparkinson: I think you're right about that, yeah (although maybe it wasn't the case for the build yesterday, I waited until everything was done to merge that PR). Trying to think of a solution for that...

alexwlchan: I am tempted to comment this out until we can come up with a proper fix…

alicefuzier: ok thanks for the explanation...I did copy across from scala-libs and yeah, I don't think the `latest` tag was ever used before that in catalogue.
I think my understanding is still a bit fuzzy, but why do we want to hold off on this fix?I would still put this solution out if it works. it's still better even if we have a problem while the internal_model is being released irghth?

alexwlchan: I am possibly a little frustrated, having fought it all of yesterday. 😓 

I've pushed a change to compare to `latest-sbt-release`, are we happy for that to go in as a stopgap?

alicefuzier: sounds ok to me. Sorry about this

alicefuzier (approved):

PR #1453, opened 2021-03-09
Aggregate on image source contributors

jamieparkinson:
This required the same concatenation of agent label & type as is done on the works endpoint


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/ImagesParams.scala
```diff
@@ -90,6 +90,7 @@ object MultipleImagesParams extends QueryParamsUtils {
 
   implicit val aggregationsDecoder: Decoder[List[ImageAggregationRequest]] =
     decodeOneOfCommaSeparated(
-      "locations.license" -> ImageAggregationRequest.License
+      "locations.license" -> ImageAggregationRequest.License,
+      "source.contributors.agent.label" -> ImageAggregationRequest.SourceContributors
```
jamesgorrie, line 94: Not to spark an annoying amount of conversation, but my issue with calling it this is:
* We are returning `source.contributors`, and choosing to aggregate on the derived key, which is why we opted for `contributors` in works. 
* Even if this is crap reasoning, it would be consistent to keep it as `source.contributors`

[I tried to capture this here](https://github.com/wellcomecollection/platform/issues/5078) - maybe not as clearly as I would have liked.

jamieparkinson: Actually would revise the first point and say we should call the aggregation `source.contributors.agent` and only return the agents (because they're the actual aggregation) - do you remember the reasoning that led to constructing the contributors like that?

jamieparkinson: See discussion following https://wellcome.slack.com/archives/C016NQB58N4/p1615376108095600

jamesgorrie: A really really interesting point on `Agents`, as this is actually what we are returning - a `Contributor` is something that has context specific to the `Work` i.e. `roles`.

On that note, I would almost definitely say we should be calling this aggregation something to do with `Agents` on both `works` and `images`.

Linking the filter / aggregation sort of makes sense, but I think something we are overlooking is eventually we will want to filter on ID, not label. This type of filter is a half way house.

If we don't want to  consider that this aggregation will eventually aggregate on `contributors.agent.id` and rather create a new `contributors.agent.label` aggregation and deprecate it with a `contributors.agent.id` when we have that info that is OK, but it feels like iterating on it until we get there means we can forget about the deprecation step.

I think this thinking is applicable to genres and subjects to when the filter linked to the aggregation will eventually be an ID filter, not a label filters.

jamieparkinson: Note for future: discussed this in person and will sidestep id/label/string/type discussions for the time being, merge this and https://github.com/wellcomecollection/catalogue/pull/1457, and propagate these changes to the equivalent works aggregations asap

jamesgorrie (approved):
Nice

PR #1459, opened 2021-03-10
Talk to the logging cluster over VPC endpoints

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/5002. This should be read in conjunction with  and 

This uses a new version of our terraform-aws-ecs-service module (see https://github.com/wellcomecollection/terraform-aws-ecs-service/pull/31), which uses a VPC endpoint to talk to the new logging cluster (see https://github.com/wellcomecollection/platform-infrastructure/pull/123). I've checked a couple of services and seen them coming through to the new cluster, so I think this is heading in the right direction.

This won't make a huge impact on our bill (NAT Gateway is ~$300/month for the catalogue services, and VPC endpoints aren't free) – but it will reduce the spikiness of reindexes.


snapshots/terraform/stack/snapshot_generator/service.tf
```diff
@@ -28,9 +28,10 @@ module "snapshot_generator" {
   cluster_arn  = var.cluster_arn
 
   security_group_ids = [
-    var.elastic_cloud_vpce_sg_id,
+    # TODO: Do we need this egress security group?
```
kenoir, line 31: Easy to test by manually removing this in the console and seeing what happens I guess?

kenoir (approved):

snapshots/terraform/stack/snapshot_generator/service.tf
alexwlchan: Yeah, I just didn't want to do it right now – I've broken enough stuff!

PR #1464, opened 2021-03-12
Case-insensitive matching on the UkLW prefix in 776

jamieparkinson:
Context: https://wellcome.slack.com/archives/C8X9YKM5X/p1615483335044200

The UkLW match is case insensitive because there are sufficient inconsistencies in the source data that it's easier to handle that here than to fix it in the source data. This was causing physical/digital Sierra bibs not to be linked due to the mergeCandidates not being picked up.


jamesgorrie (approved):
Do we link just attaching slack as context, or is there value on writing a little something up?

jamieparkinson:
You're right, I'm being lazy

PR #1465, opened 2021-03-12
Big boost manifesto: bump boost bifold

jamieparkinson:
Following up from https://github.com/wellcomecollection/catalogue/pull/1444, we thought it might be worth making the colour term dominate even other boosted query components


harrisonpim (approved):
big boosts are the best, score high all the time

PR #1467, opened 2021-03-12
Use one date parser for everything

jamieparkinson:
This PR means we use what was the Calm date parser for Sierra too (and improves it a bit based on what I saw in the Sierra data).

Of all of the sierra records with 260/264 fields present:
- The old parser handled **38.13%** (and didn't fall back to 008 for the rest https://github.com/wellcomecollection/catalogue/pull/1463)
- The calm parser handled **86.75%**
- The parser here - which knows to ignore Roman numerals - handles **_96.12%_** of them

🎉


jamieparkinson:
Ah, I should add - this removes the `inferred` flag from `InstantRange`s. This was never used anywhere and I think is related to an aspect of the Sierra syntax (namely, square brackets) that relates to whether the date is taken directly from the source material. Does anyone know a reason that it should remain?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
```diff
@@ -1,61 +0,0 @@
...
-      case (y, m, d) => CalendarDate(d, m, y)
-    }
-
-  def monthAndYear[_: P] =
-    (monthFollowedByYear | yearFollowedByMonth)
-      .map { case (m, y) => MonthAndYear(m, y) }
-
-  def month[_: P] = writtenMonth map (Month(_))
-
-  def day[_: P] = writtenDay map (Day(_))
```
alexwlchan, line 60: Note to self: these methods have all been moved unmodified to PeriodParser.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/PeriodParser.scala
```diff
@@ -26,13 +24,27 @@ object PeriodParser extends Parser[InstantRange] with DateParserUtils {
...
+
+  // This is from:
+  // https://www.oreilly.com/library/view/regular-expressions-cookbook/9780596802837/ch06s09.html
+  // It is additionally restricted to numerals 3 characters or longer,
+  // to avoid interfering with things like (eg) 'c.1920'", and includes
+  // period and comma separators to handle numerals like "M.DCC.XXXVII."
+  private val romanNumeralRegexString = {
+    val sep = "[\\.\\,]?\\s?"
+    s"(?=[MDCLXVI\\.\\,\\s]{3,})M*$sep(C[MD]|D?C*)$sep(X[CL]|L?X*)$sep(I[XV]|V?I*)".toLowerCase
+  }
```
alexwlchan, line 44: Useful comment, although a quick glance at that page suggests this regex is only for matching valid Roman numerals. Do you know if we have any invalid Roman numerals in the catalogue?

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/PeriodParser.scala
```diff
@@ -166,6 +178,34 @@ object PeriodParser extends Parser[InstantRange] with DateParserUtils {
...
+      case (y, m, d) => CalendarDate(d, m, y)
+    }
+
+  def monthAndYear[_: P]: P[MonthAndYear] =
+    (monthFollowedByYear | yearFollowedByMonth)
+      .map { case (m, y) => MonthAndYear(m, y) }
+
+  def month[_: P]: P[Month] = writtenMonth map Month
+
+  def day[_: P]: P[Day] = writtenDay map Day
```
alexwlchan, line 207: Note to self: these methods have all been moved unmodified from FreeformDateParser.

common/internal_model/src/test/scala/uk/ac/wellcome/models/parse/PeriodParserTest.scala
```diff
@@ -348,6 +348,14 @@ class ParserTest extends AnyFunSpec with Matchers with Inspectors {
           "fl. 1999-2001 [gaps]"))
     }
 
+    it("strips Roman numerals") {
```
alexwlchan, line 351: nit: maybe this should be

```suggestion
    it("ignores Roman numerals") {
```

The word "strips" makes me expect it to be removed from the display string (more like whitespace). Yes, I might be splitting hairs.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
```diff
@@ -1,61 +0,0 @@
...
-      year.toInstantRange |
-      calendarDate.toInstantRange |
-      monthAndYear.toInstantRange
-
-  def dateRange[_: P] =
-    (year to year).toInstantRange |
-      (calendarDate to calendarDate).toInstantRange |
-      (monthAndYear to monthAndYear).toInstantRange |
-      (month to monthAndYear).toInstantRange |
-      (day to calendarDate).toInstantRange
```
alexwlchan, line 32: Am I right in thinking you binned all these because they weren't used anywhere prior to this patch?

alexwlchan (approved):
LGTM. A couple of suggestions, but nothing that requires another review from me. 🗓️ 

Given that the parser is all about dealing with source data, I wonder if the parsing code might be better placed in the `source_model` library? Or the transformer common lib? I’m not sure it really feels like an "internal model" thing.

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/PeriodParser.scala
jamieparkinson: I don't, but I had problems with a more permissive regex blatting other things (eg `mid-19th century`). Unscientific but in a quick look through the stuff that couldn't be parsed, I didn't see any Roman numerals...

alexwlchan: Conclusion: either there aren't any, or they're extremely malformed. 😉

common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/FreeformDateParser.scala
jamieparkinson: Not quite - these were the old Sierra parsing rules which are a strict subset of the Calm rules, so binning them was basically just removing duplication. See comment above re the "inferred" stuff

jamieparkinson:
Good shout re `source_model` - I had also not been happy with `internal_model` although that was where the existing parser lived (it predates any of the other more suitable locations). I'll move it.

jamieparkinson:
Turns out it's a real headache to do that refactoring because of dependencies on `InstantRange`, going to leave it for now

PR #1468, opened 2021-03-15
Tweak how we transform the holdings data

alexwlchan:
* Remove the `description`; add its value as another entry in `enumeration`. Often they contain similar-looking data, and this makes it easier for the front-end to render. Closes https://github.com/wellcomecollection/platform/issues/5082
* De-duplicate holdings records. A lot of the holdings data is private info we don't expose in the transformed Work, so we can de-duplicate safely. Closes https://github.com/wellcomecollection/platform/issues/5083


alexwlchan:
This feels like a change that can be applied in-place – because the `description` field is already optional, we can reindex affected works and the already-deployed API will just stop showing that field.

A catalogue snapshot suggests it's ~26k works.

jamesgorrie (approved):

PR #1473, opened 2021-03-16
Trying to fix this flaky test again

jamieparkinson:
It somehow seems to have got worse


alexwlchan (approved):

jamesgorrie:
Do we know why, and why this might help?

jamieparkinson:
@jamesgorrie the tests generate fake LSH vectors which are _supposed_ to be in order of descending similarity - the problem is that it's really quite hard to do that and sometimes the results are returned in a different order to expected

Issue #1471, opened 2021-03-16
Ensure consistent normalisation on labels for subjects / genres / concepts


jamesgorrie:
Similar to [this](https://github.com/wellcomecollection/catalogue/pull/1441/files), we should be applying that to the subject label, as well as concepts within that (which I think might actually happen already).

alexwlchan:
We should start by trimming whitespace and full stops; that consolidates ~7k subjects (or ~250k total).

PR #1477, opened 2021-03-17
Trim a single trailing period from Sierra subjects

alexwlchan:
Closes https://github.com/wellcomecollection/catalogue/issues/1471

This does a bit more tidying up – removing a single trailing period from subject labels and the associated concepts. It's based on two examples (one from Nicola in the platform-feedback channel, one I found while looking through a snapshot).

I imagine we'll find more examples of things that need fixing, so this is "one more step" rather than "completely done".

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraConcepts.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraConcepts.scala
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.platform.transformer.sierra.transformers
 
 import uk.ac.wellcome.models.work.internal._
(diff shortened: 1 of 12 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraConcepts.scala
```diff
@@ -15,7 +16,7 @@ trait SierraConcepts extends SierraQueryOps {
   protected def getLabel(primarySubfields: List[MarcSubfield],
                          subdivisionSubfields: List[MarcSubfield]): String = {
     val orderedSubfields = primarySubfields ++ subdivisionSubfields
-    orderedSubfields.map { _.content }.mkString(" - ")
+    orderedSubfields.map { _.content }.mkString(" - ").trimTrailingPeriod
```
kenoir, line 19: ✂️

kenoir (approved):

PR #1479, opened 2021-03-17
Better handling of non b numbers in 776 subfield $w

alexwlchan:
Victoria spotted a bug where two works weren't being merged correctly; it's because they had two values in 776 subfield $w, even though only one of them is the UkLW that we use for merging:

```
776 1  |iPrint version:|tThesaurus anatomicus primus [-decimus] 
       ... Het eerste [-tiende] anatomisch cabinet
       |w(OCoLC)28467848|w(UkLW)b1121837x 
```

This patch will cause the transformer to identify the single b number in this field, and also warn if we see values in this field that don't look like valid b numbers, because those need to be raised with Collections to be fixed at source.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidates.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidates.scala
@@ -40,18 +40,21 @@ object SierraMergeCandidates extends SierraDataTransformer with SierraQueryOps {
     * If the identifier starts with (UkLW), we strip the prefix and use the
     * bib number as a merge candidate.
     *
+    * We ignore any values in 776 subfield ǂw that don't start with (UkLW),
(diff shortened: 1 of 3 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraMergeCandidates.scala
```diff
@@ -34,36 +36,54 @@ object SierraMergeCandidates extends SierraDataTransformer with SierraQueryOps {
...
+        .contents
+        .flatMap {
+          case uklwPrefixRegex(bibNumber) => Some(bibNumber.trim)
+          case _                          => None
+        }
+        .filter { s =>
+          bnumberRegex.findFirstIn(s) match {
+            case Some(_) => true
+            case None =>
+              warn(s"$bibId: not a b number in field 776 subfield ǂw: $s")
```
jamieparkinson, line 70: Love a verbose warning, you've converted me to the cult

jamieparkinson (approved):

PR #1480, opened 2021-03-18
Split the `uk.ac.wellcome.models.work.internal` namespace

alexwlchan:
This namespace has become increasingly crowded, and it contains lots of stuff that's only tangentially related to Works, including Images and models that are shared between Images/Works.

This PR breaks it up into smaller namespaces, so that Work and Image models are stored in separate namespaces now:

Screenshot 2021-03-18 at 13 10 21

This is some preliminary refactoring for https://github.com/wellcomecollection/platform/issues/5090


jamieparkinson (approved):
All hail the `weco.catalogue` namespace

alexwlchan:
I've also pulled the MarcLanguageCodeList out of internal_model and into source_model – it makes a bit more sense there. The parsers might follow.

jamieparkinson:
Good luck with the parsers 😬  - they are 🐍 entangled🐍 . I think they belong in source_model but transformer_common might be easier...

alicefuzier:
I think I tried moving some of that out of internal model, but at least something is used by the API (can't remember what). You might need some disentagling or it might be difficult to import your version of the internal model in the API.

jamieparkinson:
@alicefuzier I think it's `InstantRange` that causes all the problems

alexwlchan:
> I think they belong in source_model but transformer_common might be easier...

The distinction between those two is a fine line…

PR #1486, opened 2021-03-18
Use an enumeration for IdentifierType, rather than a CSV list

alexwlchan:
This makes it more consistent with some of our other models (LocationType, AccessStatus, etc.) and ensures a more efficient JSON serialisation, e.g.

```
{"id":"miro-image-number"}
```

rather than

```
{"id":"miro-image-number","label":"Miro image number","ontologyType":"IdentifierType"}
```

which is just generally efficient, and I'm keeping 🤞 might sort out the relation embedder by trimming the size of the Work JSON.

Closes https://github.com/wellcomecollection/platform/issues/5093


jamesgorrie (approved):
Superb! Something that's been on the pet peeve list for too long. Nice work.

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
```diff
@@ -1,47 +1,106 @@
...
+  val values = findValues
+
+  implicit val locationTypeEncoder: Encoder[IdentifierType] =
+    Encoder.forProduct1("id")(_.id)
+
+  implicit val locationTypeDecoder: Decoder[IdentifierType] =
+    Decoder.forProduct1("id")(IdentifierType.withName)
+
+  def apply(id: String): IdentifierType =
+    IdentifierType.withName(id)
```
jamieparkinson, line 25: How is it working without `entryName` like the other types use? Does it secretly support an `id` field as well as `entryName`?(!) https://github.com/wellcomecollection/catalogue/blob/51ab8cf0288a9362ceec65a318cfe6766534142a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Format.scala#L10

alexwlchan: The answer might be… it doesn't 🙈 

I wanted to get this up for review quickly, so I haven't run the tests yet

jamieparkinson: 😬 well, there's your solution if it doesn't

alexwlchan: Indeed, that seems to be important…

PR #1489, opened 2021-03-18
Bump the version of internal_model used by the API/display libs

alexwlchan:
* Use the new `weco.catalogue` namespace (https://github.com/wellcomecollection/catalogue/pull/1480)
* Uses the new IdentifierType serialisation (https://github.com/wellcomecollection/platform/issues/5093)
* Supports a singular location on holdings (https://github.com/wellcomecollection/platform/issues/5092)
* Use the new `internal_model.languages` namespace (https://github.com/wellcomecollection/catalogue/pull/1491)


.sbt_metadata/api.json
```diff
@@ -2,6 +2,7 @@
   "id" : "api",
   "folder" : "api/api",
   "dependencyIds" : [
-    "display"
+    "display",
+    "source_model"
```
alicefuzier, line 6: It sound very iffy to me to add a dependency on source_model to the api. Firstly, it shouldn't know anything about the source data; secondly it seems like it will potentially reintroduce the problems with the internal_model from a different module and thirdly, as the source model is used in the pipeline, we could end up with two internal models in the api class path if we make changes to the depencies. Is there a way to port in internal model what causes the need for this instead of adding this dependency?

alexwlchan: Ah, this is the MarcLanguageCodeList stuff. I moved it out earlier because I didn't realise the API depended on it – one slight annoyance of the separate packages is that "Find Usages" no longer looks in the API, so I thought it was only used by the transformers.

(I missed it when working locally – I added a local dependency on `internal_model` to make developing easier, and thought this was that.)

The correct fix is probably to grit my teeth and move MarcCodeLanguageList back into internal_model.

alexwlchan: Boom: https://github.com/wellcomecollection/catalogue/pull/1490

Merge PR, bump version, come back and undo this change in the API.

alicefuzier: Btw the way I do this to see if the api compiles with internal model changes is adding a local dependency, as you did, comment out the published internal model dependency from `displayDependencies`, sbt reload and force rebuild of everything.

jamieparkinson (approved):
I haven't been paying enough attention - is this PR the first instance of using @alicefuzier's versioning work to prevent breaking compatibility "in the wild"?

PR #1496, opened 2021-03-19
Matcher graph visualisation

jamieparkinson:
I've wanted to do this for ages, hopefully it'll be very useful for debugging complex merging issues.

```
./get_graph.py <index date> <work id>
```
Will construct the directed matcher graph containing the work with id `<work id>` from the index date `<index date>`, where the edge directions represent "links to" relationships. It will create and open a PDF visualisation of this graph, and print the [`.DOT`/graphviz](https://graphviz.org/doc/info/lang.html) representation to stdout.


pipeline/matcher/scripts/get_graph.py
```diff
@@ -0,0 +1,61 @@
...
+}
+
+
+@click.command()
+@click.argument("index_date")
+@click.argument("work_id")
+def main(index_date, work_id):
+    session = boto3.Session(profile_name="platform-read_only")
+    # Necessary for reading secrets
+    dev_session = boto3.Session(profile_name="platform-developer")
```
alexwlchan, line 24: Minor: I would prefer these use the full role ARNs; we have helpers for getting these. It's a bit more work, but it means you don't need `~/.aws/credentials` set up.

Separately: I wonder if we should have a `-secrets` role, that gives you read_only plus secrets access?

alexwlchan (approved):

PR #1503, opened 2021-03-23
Refactoring in service of a new CanonicalID type

alexwlchan:
This is some refactoring I spotted while working on https://github.com/wellcomecollection/platform/issues/3659

This should be low-stakes, uncontroversial stuff – just general clean ups and tidies.

### DIFF
--- a/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
+++ b/common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
@@ -105,7 +105,7 @@ object PipelineStorageStream extends Logging {
     Flow[(Message, NotificationMessage)]
       .map {
         case (message, notificationMessage) =>
-          Bundle(message, notificationMessage.body, 1)
+          Bundle(message, notificationMessage.body, numberOfItems = 1)
       }
       .groupedWithin(config.batchSize, config.flushInterval)
       .mapAsyncUnordered(parallelism = config.parallelism) { bundles =>
--- a/common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticIndexerTest.scala
+++ b/common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticIndexerTest.scala
@@ -58,8 +58,8 @@ class ElasticIndexerTest
     testWith(indexer)
   }
 
-  override def createDocumentWith(id: String, version: Int): SampleDocument =
-    createSampleDocumentWith(canonicalId = id, version = version)
+  override def createDocument: SampleDocument =
+    createDocumentWith()
 
   override def assertIsIndexed(doc: SampleDocument)(
     implicit index: Index): Assertion =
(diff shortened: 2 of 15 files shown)


kenoir (approved):

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -105,7 +105,7 @@ object PipelineStorageStream extends Logging {
     Flow[(Message, NotificationMessage)]
       .map {
         case (message, notificationMessage) =>
-          Bundle(message, notificationMessage.body, 1)
+          Bundle(message, notificationMessage.body, numberOfItems = 1)
```
jamesgorrie, line 108: [Bah](https://github.com/scalameta/scalafmt/issues/755)

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticIndexerTest.scala
```diff
@@ -160,6 +164,7 @@ class ElasticIndexerTest
           val future = indexer(documents)
 
           whenReady(future) { result =>
+            println(result)
```
jamesgorrie, line 167: 👀

pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/fixtures/MatcherFixtures.scala
```diff
@@ -132,8 +132,8 @@ trait MatcherFixtures
     sendNotificationToSQS(queue, body = links.workId)
   }
 
-  def ciHash(str: String): String =
-    DigestUtils.sha256Hex(str)
+  def ciHash(ids: String*): String =
+    DigestUtils.sha256Hex(ids.sorted.mkString("+"))
```
jamesgorrie, line 136: I know this PR is just generalising this, but why do we do this?

common/pipeline_storage/src/test/scala/uk/ac/wellcome/pipeline_storage/elastic/ElasticIndexerTest.scala
alexwlchan: Oops, removed in the follow-up PR: https://github.com/wellcomecollection/catalogue/pull/1504/commits/74ea4df83607ec1a9f0275e6de0718277eed3695

pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/fixtures/MatcherFixtures.scala
alexwlchan: This is one of the kludges we do for the matcher implementation.

Quick matcher recap: it looks at all the MergeCandidates on Works, and uses those to construct a graph of all the Works. Works in the same graph component can potentially be merged together.

Screenshot 2021-03-23 at 16 05 24

We store these graphs in DynamoDB, with one Dynamo item per Work / graph node.

When you update one Work, you want to find all the other Works in the same component, so we use the "component ID" as an opaque string. That goes in a DynamoDB GSI, so you can do something like `index.findAll(componentId = …)` to find all the Works in that component.

I can't remember why exactly we did it this way, but I think the idea is we want these component IDs to be relatively stable for easier debugging across pipelines.

Does that make sense?

PR #1504, opened 2021-03-23
Add a type for Canonical ID

alexwlchan:
A persistent source of confusion (for me, at least) is the number of things in the catalogue codebase called "identifier" or "id" with type `String`. This patch adds a new `CanonicalID` type, which looks like a string externally (in JSON/Dynamo serialisations), but gives a human a bit more of a hint that they really are dealing with the canonical IDs we see in API URLs and the like.

Follows https://github.com/wellcomecollection/catalogue/pull/1503, part of wellcomecollection/platform#3659

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/CanonicalID.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/CanonicalID.scala
@@ -54,4 +54,3 @@ object CanonicalID {
   implicit val ordering: Ordering[CanonicalID] =
     (x: CanonicalID, y: CanonicalID) => x.underlying.compare(y.underlying)
 }
-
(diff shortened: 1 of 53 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/CanonicalID.scala
```diff
@@ -0,0 +1,56 @@
+package weco.catalogue.internal_model.identifiers
+
+import io.circe.{Decoder, Encoder, HCursor, Json}
+import org.scanamo.DynamoFormat
+
+class CanonicalID(val underlying: String) {
```
jamieparkinson, line 6: This is such a bikeshed but I'm going to do it - I think we use `CamelCaseId` everywhere else, is there a reason for this being `CamelCaseID`?

alexwlchan: 🚲 🏠 

Because I forgot we use camel case everywhere.

Also, we use `IngestID` in the storage service, which is the code I was using as a reference point for this work. I'll change it. https://github.com/wellcomecollection/storage-service/blob/122c03d0e90007fcc0781376dcf79e77bb5c388a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/ingests/models/IngestID.scala

jamieparkinson (approved):
http://dotty.epfl.ch/docs/reference/other-new-features/opaques.html

PR #1513, opened 2021-03-25
Half-bounded dates and descriptive labels

jamieparkinson:
This adds support for half bounded dates (eg `1990-`) and also makes it so that when we can't parse a 260/4 date field, we still use it as a label (they're usually much more descriptive than the 008 codes)

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/PeriodParser.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/PeriodParser.scala
@@ -51,18 +51,29 @@ object PeriodParser extends Parser[InstantRange] with DateParserUtils {
   // Try to parse as many `timePeriod`s as we can and then add them to find
   // the interval that covers all of them
   def parser[_: P]: P[InstantRange] =
-    (Start ~ timePeriod.rep(sep = multiPeriodSeparator, min = 1) ~ End) map {
+    ((Start ~ timePeriod.rep(sep = multiPeriodSeparator, min = 1) ~ End) map {
       periods =>
         periods.reduceLeft(_ + _)
-    }
+    }) |
+      (Start ~ halfBoundedDate ~ End)
 
   def multiPeriodSeparator[_: P]: P[Unit] =
     ws.? ~ StringIn(";", ",", "and") ~ ws.?
 
-  def timePeriod[_: P] =
+  def timePeriod[_: P]: P[InstantRange] =
     dateRange |
       singleDate |
       (noopQualifier ~ ws.? ~ singleDate) // Try not to fail for un-spec'd qualifiers
+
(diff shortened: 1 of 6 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/parse/PeriodParser.scala
```diff
@@ -51,19 +51,30 @@ object PeriodParser extends Parser[InstantRange] with DateParserUtils {
...
 
+  def halfBoundedDate[_: P]: P[InstantRange] =
+    (leftBoundedDate map InstantRange.after) |
+      (rightBoundedDate map InstantRange.before)
+
+  def rightBoundedDate[_: P]: P[InstantRange] =
+    ("-" ~ ws.? ~ singleDate) | ("before" ~ ws.? ~ singleDate)
+
+  def leftBoundedDate[_: P]: P[InstantRange] =
+    (singleDate ~ ws.? ~ "-") | ("after" ~ ws.? ~ singleDate)
```
alexwlchan, line 76: Minor: I would find it helpful if there were some comments with examples of what you expected these to match.

Nit: rename `right`/`left` to `before`/`after` for consistency, and make them private.

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/InstantRange.scala
```diff
@@ -39,6 +39,30 @@ object InstantRange {
...
+    .of(9999, 12, 31)
+    .atStartOfDay()
+    .plusDays(1)
+    .minusNanos(1)
+    .toInstant(ZoneOffset.UTC)
+
+  val negativeInfinity = LocalDate
+    .of(-9999, 1, 1)
+    .atStartOfDay()
+    .toInstant(ZoneOffset.UTC)
```
alexwlchan, line 52: Minor: possibly worth a comment that Sierra uses 9999, and we didn't just pick it arbitrarily?

Also make these values private unless you're using them elsewhere?

common/internal_model/src/test/scala/uk/ac/wellcome/models/parse/PeriodParserTest.scala
```diff
@@ -396,6 +396,22 @@ class PeriodParserTest extends AnyFunSpec with Matchers with Inspectors {
...
+          "1897-"))
+    }
+
+    it("handles right-bounded intervals") {
+      PeriodParser("-1897") shouldBe Some(
+        InstantRange(
+          LocalDate of (-9999, 1, 1),
+          LocalDate of (1897, 12, 31),
+          "-1897"))
+    }
```
alexwlchan, line 413: Perhaps also tests for `before 1999` and `after 2001`?

alexwlchan (approved):

PR #1520, opened 2021-03-30
Replace post-filtering & min_doc_count: 0 with global aggregations

jamieparkinson:
As per the advice in https://github.com/elastic/elasticsearch/issues/62084#issuecomment-689507644

Using the principles in https://github.com/wellcomecollection/docs/blob/master/adr/api_faceting.md I could identify where we did want to return empty buckets and where we didn't - this PR _does_ change some behaviours but I believe that these changes won't hit the UI as the now-removed buckets were already being filtered out. It also turns out there were some inconsistencies between image and works aggregations (oops!) which this fixes.

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/models/Aggregation.scala
@@ -2,8 +2,10 @@ package uk.ac.wellcome.platform.api.models
 
 import io.circe.generic.extras.JsonKey
 import io.circe.{Decoder, Json}
+import io.circe.parser._
+import io.circe.optics.JsonPath._
 
-import scala.util.Try
+import scala.util.{Success, Try}
 
 // This object maps an aggregation result from Elasticsearch into our
 // Aggregation data type.  The results from ES are of the form:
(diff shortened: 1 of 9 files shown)


jamesgorrie (commented):
It looks good, but just checking my understanding just trying to get an understanding of what the recommendation is.

Is it:
* you have an aggregation, unfiltered, with a sub aggregation filtered by a paired filter
* if there is a paired filter, we show the filtered aggregation, else the normal one

jamieparkinson:
What do you mean by a "normal" aggregation?

jamieparkinson:
We need to have filtered subaggregations in order _not_ to apply the same filters to the search and the aggregation buckets: we don't want aggregation buckets to be filtered by their paired filter.

This is achieved by doing a `global` aggregation in the case that a paired filter is present and applying all of the filters _except_ the paired one as subaggregations of the global parent aggregation.

If filtered aggregations are present then we always use them - the decisions about this are applied when the query is made, not when the response is parsed. The extra parsing logic here is just due to the fact that changing the queries changes the response format; you can summarise it as "use the most specific counts you can find in the response".

jamesgorrie (approved):

jamesgorrie:
Thanks for explaining that @jamieparkinson.

PR #1527, opened 2021-03-31
Expunge old facets: license, genres, subjects, contributors

jamieparkinson:
Tidying up after https://github.com/wellcomecollection/catalogue/pull/1518 and https://github.com/wellcomecollection/wellcomecollection.org/pull/6318

### DIFF
--- a/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksParams.scala
+++ b/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksParams.scala
@@ -136,7 +136,6 @@ object MultipleWorksParams extends QueryParamsUtils {
         "genres.label".as[GenreFilter].?,
         "subjects.label".as[SubjectFilter].?,
         "contributors.agent.label".as[ContributorsFilter].?,
-        "license".as[LicenseFilter].?,
         "items.locations.license".as[LicenseFilter].?,
         "include".as[WorksIncludes].?,
         "aggregations".as[List[WorkAggregationRequest]].?,
@@ -146,7 +145,8 @@ object MultipleWorksParams extends QueryParamsUtils {
         "identifiers".as[IdentifiersFilter].?,
         "items.locations.locationType".as[ItemLocationTypeIdFilter].?,
         "items.locations.accessConditions.status".as[AccessStatusFilter].?,
-        "type".as[WorkTypeFilter].?
+        "type".as[WorkTypeFilter].?,
+        "partOf".as[PartOfFilter].?
       )
     ).tflatMap {
       case (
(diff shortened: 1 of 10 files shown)


api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksParams.scala
```diff
@@ -169,17 +168,17 @@ object MultipleWorksParams extends QueryParamsUtils {
           identifiers,
           itemLocationTypeId,
           accessStatus,
-          workType) =>
+          workType,
+          partOf) =>
         // Scala has a max tuple size of 22 so this is nested to get around this limit
         parameter(
           (
-            "partOf".as[PartOfFilter].?,
```
jamesgorrie, line 176: Why is this part of this PR?

jamesgorrie (approved):

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksParams.scala
jamieparkinson: Because of the tuple size restriction thing - I could move it back into the "other" tuple having removed some filters. Annoying!

jamesgorrie: Yea this bit of the params always makes me feel a little uneasy.

jamieparkinson: https://dotty.epfl.ch/docs/reference/dropped-features/limit22.html

alexwlchan: Looking at this briefly, I wonder if changing the groups would be better?

In particular you get page/pageSize/query/_index as parameters that are common to both MultipleWorkParams and MultipleImageParams. Have a shared function that extracts those (so they're consistent across images/work endpoints), and then you can fit the remaining parameters into a 22-tuple.

It doesn't get you around the problem, exactly, but it gives a slightly clearer distinction of which parameter is unpacked where.

jamieparkinson:
Going to merge & deploy this on Tuesday 6th April

api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksParams.scala
jamesgorrie: I don't massively see the point of coupling certain parameters, but this is my personal thing of keeping things until there is a very clear benefit to do so.

PR #1528, opened 2021-04-01
Fix: ensure that aggregation buckets are always ordered by count

jamieparkinson:
afaict this has always been an issue but it's just more obvious after https://github.com/wellcomecollection/catalogue/pull/1520. It presents in the UI as the aggregations changing order when a filter is applied.


jamesgorrie (approved):
Do we need a test to reflect this?

jamieparkinson:
Probably not a bad idea to be explicit...

PR #1529, opened 2021-04-06
Fix merging on Works that have electronic items from field 856

alexwlchan:
It's possible for e-video Works to pick up unidentified items from MARC field 856, which confuses the merger and causes it to attach the IIIF manifest to the wrong Work.  By filtering only for *identified* items (i.e. those that came directly from Sierra), the manifest gets attached to the correct work.

This fixes an issue raised by Harkiran in Slack: https://wellcome.slack.com/archives/C8X9YKM5X/p1617705467131600


jamieparkinson (approved):
Nice catch, nice test

PR #1535, opened 2021-04-12
Fix an issue with the Swagger model for DisplayWorkAggregations

alexwlchan:
Previously, Swagger was serialising the field names with the slightly ugly `$u002E` in the name (a Unicode full stop), e.g.

```json
"Aggregations": {
  "type": "object",
  "properties": {
    "genres$u002Elabel": {
      "$ref": "#/components/schemas/GenreAggregation"
    },
    "genres.label": {
      "$ref": "#/components/schemas/GenreAggregation"
    },
    ...
  },
  "description": "A map containing the requested aggregations."
}
```

This is clearly wrong -- setting the `name` parameter in the `@Schema()` annotation means we see the properties as they actually appear in the serialised JSON.

This is the only currently visible model in the Swagger that displays this issue.


jamesgorrie (approved):
nice catch

PR #1545, opened 2021-04-16
add relations.ancestors.collectionPath to IndexedWorkIndexConfig

jamesgorrie:
ref https://github.com/wellcomecollection/platform/issues/5123

We need to be able to query on the ancestors label to be able to retrieve **all works that are children of a collection**.

This adds the mappings to make the data queryable.

We will then add to the [current query](https://github.com/wellcomecollection/catalogue/blob/master/api/api/src/test/resources/WorksMultiMatcherQuery.json) to satisfy this data need.

Example of what the query might look like.

```JSON
{
    "bool": {
      "should": [
        {
          "term": {
            "state.relations.ancestors.collectionPath.label": {
              "value": "WA/HMM"
            }
          }
        }
      ]
    }
  }
```

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/index/WorksIndexConfig.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/index/WorksIndexConfig.scala
@@ -56,7 +56,7 @@ object IdentifiedWorkIndexConfig extends WorksIndexConfig {
   val fields =
     Seq(
       objectField("data").fields(
-        objectField("otherIdentifiers").fields(lowercaseKeyword("value")),
+        objectField("otherIdentifiers").fields(lowercaseKeyword("value"))
       ),
       objectField("state")
         .fields(sourceIdentifier)
@@ -90,7 +90,7 @@ object IndexedWorkIndexConfig extends WorksIndexConfig {
       englishTextKeywordField("physicalDescription"),
       multilingualField("lettering"),
       objectField("contributors").fields(
-        objectField("agent").fields(label),
+        objectField("agent").fields(label)
       ),
       objectField("subjects").fields(
         label,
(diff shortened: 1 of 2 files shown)


common/internal_model/src/main/scala/uk/ac/wellcome/models/index/WorksIndexConfig.scala
```diff
@@ -56,7 +56,7 @@ object IdentifiedWorkIndexConfig extends WorksIndexConfig {
   val fields =
     Seq(
       objectField("data").fields(
-        objectField("otherIdentifiers").fields(lowercaseKeyword("value")),
+        objectField("otherIdentifiers").fields(lowercaseKeyword("value"))
```
jamesgorrie, line 59: I'll need to figure out why my `scalafmt` is being so strict.

alexwlchan (commented):
Could you add some API tests that demonstrates this change actually produces the expected change in query results?

jamesgorrie:
> Could you add some API tests that demonstrates this change actually produces the expected change in query results?

@alexwlchan This change doesn't have any changes to the query, but makes a field available. We could have a test for each field, but I feel that might become very cumbersome very quickly.

There are already tests that ensure the the mapping is valid, and that the general queries aren't broken.

When I add the query part I will add tests.

alexwlchan:
Ah, I got confused about what part of the work this was. I think I understand now, thanks.

jamesgorrie:
@alexwlchan updated the description to outline the work more clearly.

alexwlchan:
\*looks again*

Okay, no, I’m confused again. It looks like you’re suggesting you’d do a specific query on the `ancestors.collectionPath`, presumably in the API – but how would you know which part of a user's query to match on there?

I thought the idea of adding the field to the mapping was to index the collectionPath field, so that values in that field are queryable in the free text search. Maybe the feature is bigger than what I read from the ticket?

jamesgorrie:
@alexwlchan you're right with your latter part of what you said. We will be adding to the general query we have so far - the suggestion above is potentially a **part** we might add to it.

* Add mapping for field we want queryable (`state.relations.ancestors.collectionPath.label`)
* Add to current query to match appropriately from the examples we have

harrisonpim (approved):

PR #1547, opened 2021-04-16
Add a basic aspect ratio inferrer

harrisonpim:
Uses the usual python inferrer structure. 

We could probably avoid fetching full images in most cases by using the `info.json` when given a IIIF URL, but this basic version should work universally, if a bit slowly.

## Example response
Using [a lovely cat poster from wikipedia](https://upload.wikimedia.org/wikipedia/commons/thumb/0/0b/Cat_poster_1.jpg/1280px-Cat_poster_1.jpg), we get

```json
{
  "aspect_ratio": 1.521997621878716
}
```


pipeline/inferrer/aspect_ratio_inferrer/README.md
```diff
@@ -0,0 +1,14 @@
...
+docker run -p 80:80 aspect_ratio_api
+```
+
+```
+curl "http://0.0.0.0/aspect_ratio/?image_url=SOME_ENCODED_IMAGE_URL"
+```
+
+## TODO
+
+Could avoid fetching the full image by using the info.json when given a IIIF URL.
```
jamesgorrie, line 14: Another solution might be to get a smaller version of the image - although I am not sure where we might loose accuracy. Or maybe read it from the metadata of the smaller image... not sure it's an issue, but just some thoughts.

pipeline/inferrer/aspect_ratio_inferrer/app/main.py
```diff
@@ -0,0 +1,49 @@
...
+from weco_datascience.image import get_image_from_url
+from weco_datascience.logging import get_logger
+from .encoder import Encoder
+
+logger = get_logger(__name__)
+
+
+# initialise API
+logger.info("Starting API")
+app = FastAPI(title="Aspect ratio extractor", description="extracts aspect ratios")
```
jamesgorrie, line 13: deep.

pipeline/inferrer/aspect_ratio_inferrer/app/main.py
```diff
@@ -0,0 +1,49 @@
...
+logger.info("API started, awaiting requests")
+
+encoder = Encoder()
+batch_inferrer_queue = BatchExecutionQueue(encoder, batch_size=8, timeout=1)
+
+
+@app.get("/aspect_ratio/")
+async def main(query_url: str):
+    try:
+        image = await get_image_from_url(query_url)
```
jamesgorrie, line 23: Do we validate on URLs?

jamesgorrie (approved):
It'd be good to have a run down of how this gets into the stack.

jtweed (approved):

pipeline/inferrer/aspect_ratio_inferrer/app/main.py
harrisonpim: Yeah, `weco_datascience` figures out whether it's a IIIF URL or not, what to do in either case, resizing, throwing errors etc https://github.com/wellcomecollection/data-science/blob/master/weco_datascience/image.py

That said, requests can only come from the inference manager at the moment so we don't expect to see many funky URLs.

pipeline/inferrer/aspect_ratio_inferrer/README.md
harrisonpim: Yeah, I think a combination of smaller-image-fetching and IIIF would do a lot to speed this up without meaningful decrease in accuracy. The current state feels like a reasonable MVP for now though.

pipeline/inferrer/aspect_ratio_inferrer/app/encoder.py
```diff
@@ -0,0 +1,26 @@
...
+        self.delayed_process_image = delayed(self.process_image)
+
+    def process_image(self, image):
+        return image.width / image.height
+
+    def __call__(self, images):
+        """
+        process images in parallel
+        """
+        return Parallel(n_jobs=-2)(
```
jamieparkinson, line 24: I would hazard a guess that this will be less performant than a single-threaded version because aspect ratio calculation is so trivial and the overhead of thread management will dominate compute time - in the other inferrers we are doing quite complex numerics which benefit from being parallel

pipeline/inferrer/aspect_ratio_inferrer/README.md
jamieparkinson: Remember that we're fetching the large-ish images anyway and the URLs will, in reality, be filesystem paths. Obviously HTTP requests are waaaayyyy slower than file IO - so I don't think fetching a smaller image or a JSON file would be faster in anger, this is probably the most performant at the moment.

jamieparkinson (approved):

PR #1550, opened 2021-04-19
Bump internal model for today's reindex

alexwlchan:


alexwlchan:
@jamesgorrie @harrisonpim API tests are failing for something that looks query related. Any ideas?

harrisonpim:
looks like this could be the order of the language fields playing up again...

alexwlchan:
> looks like this could be the order of the language fields playing up again...

Ah, I see the issue now. I think we have something for this already – will have a look.

alexwlchan:
Okay, I misremembered – we used to, but not any more. The actual order is alphabetical, which seems more sensible, so I’ve just changed the JSON for now.

jamesgorrie:
@harrisonpim  this would make sense why we didn't see the sort method take affect as we hadn't bumped the model to do so.

Now we have, it has taken affect and I've updated the test JSON, to reflect that. Sorry @alexwlchan  I think this was my misunderstanding the model versioning.

PR #1554, opened 2021-04-19
Add Terraform for the 2021-04-19 pipeline

alexwlchan:


reindexer/get_reindex_status.py
```diff
@@ -9,6 +9,7 @@
 from elasticsearch.exceptions import NotFoundError
 import humanize
 import tabulate
+from tenacity import retry, stop_after_attempt, wait_exponential
```
jamesgorrie, line 12: Does this need to be added as a dependency somewhere?

jamesgorrie (approved):

pipeline/terraform/stack/elastic_pipeline_storage.tf
```diff
@@ -6,7 +6,7 @@ resource "ec_deployment" "pipeline_storage" {
   name = "catalogue-pipeline-storage-${var.pipeline_date}"
 
   region                 = "eu-west-1"
-  version                = "7.10.2"
+  version                = "7.12.0"
```
jamesgorrie, line 9: I have done a check, and we don't seem to have any breaking problems... but probably worth double checking if you haven't.

reindexer/get_reindex_status.py
alexwlchan: I don't think we have a requirements.txt for this.

pipeline/terraform/stack/elastic_pipeline_storage.tf
alexwlchan: We only use pretty basic Elasticsearch features in pipeline storage (index/get), so I'd be quite surprised if a minor update broke anything.

jamesgorrie: Ah, great - I suppose any slight fluctuation in querying etc would be OK too.

alexwlchan: I don't think we even make any queries tbh.

PR #1555, opened 2021-04-20
Implement a basic form of lazy indexing in the Sierra indexer

alexwlchan:
The Sierra indexer has a queue that will, at current processing rates, clear in 70 days. A large pile of stuff landed on the queue about four days ago, and it can't keep up.

We could increase the CPU in the reporting cluster, but the next step up is $400/mo, and the next step up again is an extra $800/mo on top of that. That's not great. We're also burning money on Fargate tasks.

This is an attempt at a quick fix – we look up the top-level record, and if it matches what's already stored we'll skip indexing it again (assuming the varFields and fixedFields are similarly the same). This won't fix the problem fully, but I'm hoping it releases some of the pressure, and gets it "good enough" until we have time to look at this properly.

### DIFF
--- a/sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
+++ b/sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
@@ -1,29 +1,64 @@
 package weco.catalogue.sierra_indexer.services
 
+import com.sksamuel.elastic4s.{ElasticClient, Index}
+import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.requests.delete.DeleteByQueryRequest
+import com.sksamuel.elastic4s.requests.get.GetResponse
 import com.sksamuel.elastic4s.requests.indexes.IndexRequest
+import grizzled.slf4j.Logging
 import io.circe.parser._
 import io.circe.{Json, ParsingFailure}
 import weco.catalogue.sierra_indexer.models.{IndexerRequest, Parent}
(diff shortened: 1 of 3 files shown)


sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
```diff
@@ -1,38 +1,99 @@
...
+      // If you wanted to go one step further, you could diff individual
+      // varFields/fixedFields, but that adds more complexity.
+      //
+      // I'm hoping this is a quick fix that makes the indexer performance "good enough"
+      // without us having to pour money into the reporting cluster.
+      stored <- getStored(apiData)
+      modifiedApiData = stored
+        .filter {
+          case ((parent, json), getResponse)
+              if jsonStringsMatch(
```
jamesgorrie, line 45: You comment above says it's checking the date, but it looks like it's checking the JSON?

jamesgorrie (approved):
Makes sense, let's 🤞 - one note on the comment <=> code maybe not aligning?

sierra_adapter/sierra_indexer/src/main/scala/weco/catalogue/sierra_indexer/services/Splitter.scala
alexwlchan: Ah, I've been a bit coy – the date is part of the JSON., I've updated the comment to clarify.

alexwlchan:
It's hard to know how much is these changes, how much is the increased size of the reporting cluster – but the queue is shifting pretty dramatically. 🎉 

Screenshot 2021-04-20 at 10 19 59

PR #1558, opened 2021-04-20
Add stacks services to catalogue repo

kenoir:
Adding the stacks code to the catalogue repo in preparation for consuming catalogue models / logic.

This change just adds the code / verifies it will compile & pass tests.

Future PRs will address infrastructure and deployment.


alexwlchan (approved):
I can't read 90 Scala files, so I'm going to trust you haven't futzed with them in the migration from the old repo. Let's start to pull in bits of the catalogue repo in subsequent PRs.

PR #1560, opened 2021-04-21
Use the matcher result time for works after the merger

alexwlchan:
The second half of https://github.com/wellcomecollection/platform/issues/5132

This PR implements the merger versioning scheme described in [RFC 038](https://github.com/wellcomecollection/docs/tree/master/rfcs/038-matcher-versioning). I've already deployed a matcher that sends messages with a `createdTime`; this PR updates the merger to pay attention to it.

Previously: the merger would use `max(w.modifiedTime for w in works)` as the modifiedTime when it created works and images. This caused some weird edge cases when works get unlinked and the pipeline became unable to update them because it "remembers" what they used to be linked to.

Now: the merger uses the time when the matcher processed these works.

I've included a feature test that reproduces the worked example from the RFC, and confirmed that it fails without this patch.


kenoir (approved):
Super clear and useful tests 👍

PR #1563, opened 2021-04-21
Use the British spelling of "Licence" for specific named licences in the API

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/6409

If you follow the URLs, you see these licences both use the British spelling.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/locations/License.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/locations/License.scala
@@ -79,16 +79,14 @@ object License extends Enum[License] {
 
   case object OGL extends License {
     val id = "ogl"
-    val label =
-      "Open Government License"
+    val label = "Open Government Licence"
     val url =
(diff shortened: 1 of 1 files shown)


jamesgorrie (commented):
Is the rule docs use UK, and everything else US?

As we use this in quite a [few exposed places](https://github.com/wellcomecollection/catalogue/blob/30d7ab3a76d2e3f5d0b9333b867d10891f2963e9/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksParams.scala#L64), including parameters for the API.

alexwlchan:
> Is the rule docs use UK, and everything else US?

I'm honestly not sure – Jenn opened a ticket (linked in the PR description) to correct some UK spellings on the site, but we have tended to use US spellings in most parts of the API.

I don't really want to start a discussion on the use of UK/US spelling in the API as a whole, just fix the names of these two specific licences, where we are unambiguously using proper names that are spelt incorrectly:

uk-gov-licence-large-banner

Screenshot 2021-04-21 at 19 16 51

kenoir (approved):

PR #1568, opened 2021-04-22
Expose copies on order as an item with a new LocationType in the catalogue API

alexwlchan:
This implements the rules described in https://github.com/wellcomecollection/platform/issues/5139

* If a bib has no items, but it has an order attached to the bib record with STATUS of `o` ("on order"), we create an item with LocationType `on-order` and label _"X copies ordered for Wellcome Collection on 20 April 2021"_
* If a bib has no items, but it has an associated order with STATUS of `a` ("fully paid") and a RDATE ("received date"), we create an item with LocationType `on-order` and label _"X copies awaiting cataloguing for Wellcome Collection"
* If a bib has any items, we ignore all the attached order records.

I haven't done too much planning for this patch – realistically it's only a few thousand items at most, so I just want to get _something_ into the API, and we can tweak it later if necessary.

I've run this code against the six examples @jtweed put in the original ticket, and it does the right thing for all of them. The only difference is that in Encore, one of them is labelled "Processing for the Asian Collections", whereas in this patch everything is "for Wellcome Collection".

Reviewers: @wellcomecollection/scala-devs for code, @jtweed to validate the approach.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
@@ -115,7 +115,7 @@ class SierraTransformer(sierraTransformable: SierraTransformable, version: Int)
       notes = SierraNotes(bibData),
       duration = SierraDuration(bibData),
       items =
-        SierraItemsOnOrder(bibId, itemDataMap, orderDataMap) ++
+        SierraItemsOnOrder(bibId, hasItems = hasItems, orderDataMap) ++
           SierraItems(itemDataMap)(bibId, bibData) ++
           SierraElectronicResources(bibId, varFields = bibData.varFields),
       holdings = SierraHoldings(bibId, holdingsDataMap)
@@ -128,6 +128,9 @@ class SierraTransformer(sierraTransformable: SierraTransformable, version: Int)
     ontologyType = "Work",
     value = bibId.withCheckDigit
   )
+
+  lazy val hasItems: Boolean =
+    sierraTransformable.itemRecords.nonEmpty
 
   lazy val itemDataMap: Map[SierraItemNumber, SierraItemData] =
     sierraTransformable.itemRecords
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
@@ -116,7 +116,7 @@ class SierraTransformer(sierraTransformable: SierraTransformable, version: Int)
       duration = SierraDuration(bibData),
       items =
         SierraItemsOnOrder(bibId, hasItems = hasItems, orderDataMap) ++
-          SierraItems(itemDataMap)(bibId, bibData) ++
+          SierraItems(bibId, bibData, itemDataMap) ++
           SierraElectronicResources(bibId, varFields = bibData.varFields),
       holdings = SierraHoldings(bibId, holdingsDataMap)
     )
--- a/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
@@ -114,8 +114,10 @@ class SierraTransformer(sierraTransformable: SierraTransformable, version: Int)
       edition = SierraEdition(bibData),
       notes = SierraNotes(bibData),
       duration = SierraDuration(bibData),
-      items = SierraItems(itemDataMap)(bibId, bibData) ++
-        SierraElectronicResources(bibId, varFields = bibData.varFields),
+      items =
+        SierraItemsOnOrder(bibId, itemDataMap, orderDataMap) ++
+          SierraItems(itemDataMap)(bibId, bibData) ++
+          SierraElectronicResources(bibId, varFields = bibData.varFields),
       holdings = SierraHoldings(bibId, holdingsDataMap)
     )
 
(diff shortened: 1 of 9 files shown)


kenoir (approved):

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItemsOnOrder.scala
```diff
@@ -0,0 +1,182 @@
...
+            )
+          )
+        )
+
+      // status 'a' = "Fully paid", RDATE = "when we actually received the thing"
+      //
+      // We create an item with a message like "Awaiting cataloguing for Wellcome Collection"
+      // We don't expose the received date publicly (in case an item has been in the queue
+      // for a long time) -- but we do expect it to be there for these records.
+      case (Some(status), _, Some(_), copies) if status == "a" =>
```
jamieparkinson, line 91: I know IntelliJ likes it to be like this if the variable would be unused but `Some(_)` is quite hard to understand, could it be either
```scala
case (Some(status), _, receivedDate, copes) if status == "a" && receivedDate.isDefined =>
```
or
```scala
case (Some(status), _, Some(receivedDate), copes) if status == "a" =>
```

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItemsOnOrder.scala
```diff
@@ -0,0 +1,182 @@
...
+                label = createAwaitingCataloguingMessage(copies)
+              )
+            )
+          )
+        )
+
+      // We're deliberately quite conservative here -- if we're not sure what an order
+      // means, we ignore it.  I don't know how many orders this will affect, and how many
+      // will be ignored because they're suppressed/there are other items.
+      case (Some(status), _, None, _) if status == "a" =>
```
jamieparkinson, line 107: ditto, `receivedDate @ None` perhaps

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItemsOnOrder.scala
```diff
@@ -0,0 +1,182 @@
...
+  * https://documentation.iii.com/sierrahelp/Default.htm#sril/sril_records_fixed_field_types_order.html%3FTocPath%3DSierra%2520Reference%7CHow%2520Innovative%2520Systems%2520Store%2520Information%7CFixed-length%2520Fields%7C_____11
+  *
+  */
+object SierraItemsOnOrder extends Logging {
+  def apply(
+    id: TypedSierraRecordNumber,
+    itemDataMap: Map[SierraItemNumber, SierraItemData],
+    orderDataMap: Map[SierraOrderNumber, SierraOrderData]
+  ): List[Item[IdState.Unidentifiable.type]] =
+    if (itemDataMap.isEmpty) {
```
jamieparkinson, line 50: I wonder if it would be clearer for this conditional to live at https://github.com/wellcomecollection/catalogue/pull/1568/files#diff-9f82908b0b203caf49a1bf7baed59891b25af2898e2ecbf6b215f1fa57c4f916R118 so we don't have to pass `itemDataMap` so opaquely when we don't really use it

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
```diff
@@ -114,8 +114,10 @@ class SierraTransformer(sierraTransformable: SierraTransformable, version: Int)
       edition = SierraEdition(bibData),
       notes = SierraNotes(bibData),
       duration = SierraDuration(bibData),
-      items = SierraItems(itemDataMap)(bibId, bibData) ++
-        SierraElectronicResources(bibId, varFields = bibData.varFields),
+      items =
+        SierraItemsOnOrder(bibId, itemDataMap, orderDataMap) ++
+          SierraItems(itemDataMap)(bibId, bibData) ++
```
jamieparkinson, line 119: We should probably be consistent about whether these methods are curried

pipeline/transformer/transformer_sierra/src/test/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraTransformerTest.scala
```diff
@@ -992,6 +992,47 @@ class SierraTransformerTest
     }
   }
 
+  it("includes items based on the order records") {
```
jamieparkinson, line 995: Worth testing some invalid cases?

alexwlchan: Was there anything you think I've missed?

jamieparkinson: Ah hang on, github was helpfully hiding the "large diff" on the tests file 🙃

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItemsOnOrder.scala
alexwlchan: I've swapped it out for a `hasItems` boolean.

pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/SierraTransformer.scala
alexwlchan: Fixed; I have no idea why SierraItems was being curried.

jtweed:
I think this needs revisited ever so slightly. Both the currently agreed messages in the OPAC (which are NOT the same as Encore) and the consensus in that Slack thread is that the number of copies should NOT be included in the message.

So this is fine overall, but the "X copies" should be dropped from the start of both messages.

alexwlchan:
> Both the currently agreed messages in the OPAC (which are NOT the same as Encore) and the consensus in that Slack thread is that the number of copies should NOT be included in the message.

Ah, oops. Will fix.

alexwlchan:
Tada: https://github.com/wellcomecollection/catalogue/pull/1569

PR #1573, opened 2021-04-23
Images query

harrisonpim:
Applies some of what we've learned about catalogue search to image search like language analysis, exact vs approximate spelling, prefix matching, etc. 

Because image search is a different domain and its users have different intentions, the lessons are applied to a slightly different set of fields, though the structure is almost exactly the same.

Follows https://github.com/wellcomecollection/catalogue/pull/1549
Linked to https://github.com/wellcomecollection/rank/pull/35
Closes https://github.com/wellcomecollection/rank/issues/33


jamesgorrie (approved):
Having the output here is really _really_ useful.

Amazing work!

harrisonpim:
> Having the output here is really really useful.

This got my brain whirring yesterday - was thinking about a future with `rank` and `api` in the same repo, and the ways that this kind of introspection and testing could be even smoother

jamesgorrie:
@harrisonpim that's definitely the plan.

Issue #1572, opened 2021-04-23
Extract api services to another repository


kenoir:
There are many services now being built in the catalogue repo, and the addition of stacks services increases that further.

To enable more easily working on services that make use of the display models, we will move those shared services to a new repository and consume the internal models published in the catalogue repo.

PR #1576, opened 2021-04-23
Consolidate things to be moved into catalogue_api

kenoir:
To facilitate https://github.com/wellcomecollection/catalogue/issues/1572 this change moves the snapshot generator into the api folder.

We will eventually do a filter-branch operation specifying that folder to retain git history.


api/common/catalogue_python_fixtures/fixtures/aws.py
```diff
@@ -0,0 +1,66 @@
+import boto3
```
kenoir, line 1: This is duplicated not moved as some of the adapters rely on this as well as the snapshot generator (and it didn't look complicated enough to extract and share).

jamesgorrie (approved):

PR #1580, opened 2021-04-26
Remove some unnecessary VPC endpoints

alexwlchan:
Running a dedicated VPC endpoint is only cheaper than using the managed NAT Gateway when you're pushing 200GB+ of traffic per month*.

Since we created these endpoints, we've substantially removed the amount of traffic:

* We don't use API Gateway in this VPC at all now
* Most of our logging is going to Kibana, not CloudWatch
* Rather than passing around the whole Work in SNS/SQS, we pass around individual IDs, which are much smaller

At peak, the Sierra adapter only pushes ~3GB a day, which is a far cry from the 200GB we'd need to make this work.

*NAT Gateway costs $0.045/GB, VPC endpoint is $0.01/hr + $0.01/GB Break-even point is when:

        $0.045/GB * X GB ≥ $0.01/hr * 24 hrs/day * 30 days + $0.01/GB * X GB
    =>  X >= $226.29

I happened to be staring at this code, and the obvious optimisation fell out.


jamesgorrie (approved):
💸

PR #1585, opened 2021-04-28
add ancestors fields and tests for how mappings should work

jamesgorrie:
* Renames `multilingualKeywordField` => `multilingualFieldWithKeyword` as it was implying it was a `keyword` field
* Adds some tests to help explain what our abstracted index fields are for
* add lowercaseKeyword to allow us to search keywords as `VALUE` or `value` but not loose the ability to aggregate maintaining their case. The aggregation stuff will probably become redundant when we move to IDs for things, but is required for now
* Some bloopiness

### DIFF
--- a/common/internal_model/src/test/scala/uk/ac/wellcome/models/index/IndexConfigFieldsTest.scala
+++ b/common/internal_model/src/test/scala/uk/ac/wellcome/models/index/IndexConfigFieldsTest.scala
@@ -1,0 +1,105 @@
+package uk.ac.wellcome.models.index
+
+import org.scalatest.concurrent.Eventually
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+import uk.ac.wellcome.json.utils.JsonAssertions
+import uk.ac.wellcome.models.work.generators.WorkGenerators
+import weco.catalogue.internal_model.generators.ImageGenerators
+import uk.ac.wellcome.elasticsearch.IndexConfig
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.requests.mappings.dynamictemplate.DynamicMapping
+import io.circe._, io.circe.generic.semiauto._, io.circe.syntax._
+import com.sksamuel.elastic4s._
+import com.sksamuel.elastic4s.requests.searches._
+
+case class TestDoc(label: String)
+object TestDoc {
+  implicit val fooDecoder: Decoder[TestDoc] = deriveDecoder
+  implicit val fooEncoder: Encoder[TestDoc] = deriveEncoder
+}
+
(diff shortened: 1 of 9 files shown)


harrisonpim (approved):

common/internal_model/src/test/scala/uk/ac/wellcome/models/index/IndexConfigFieldsTest.scala
```diff
@@ -0,0 +1,105 @@
...
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.requests.mappings.dynamictemplate.DynamicMapping
+import io.circe._, io.circe.generic.semiauto._, io.circe.syntax._
+import com.sksamuel.elastic4s._
+import com.sksamuel.elastic4s.requests.searches._
+
+case class TestDoc(label: String)
+object TestDoc {
+  implicit val fooDecoder: Decoder[TestDoc] = deriveDecoder
+  implicit val fooEncoder: Encoder[TestDoc] = deriveEncoder
```
jamieparkinson, line 21:
```suggestion
  implicit val testDocDecoder: Decoder[TestDoc] = deriveDecoder
  implicit val testDocEncoder: Encoder[TestDoc] = deriveEncoder
```
If you import JsonUtils I think these will be fully automatically derived

common/internal_model/src/test/scala/uk/ac/wellcome/models/index/IndexConfigFieldsTest.scala
```diff
@@ -0,0 +1,105 @@
...
+}
+
+object TestIndexConfig extends IndexConfig with IndexConfigFields {
+  val analysis = WorksAnalysis()
+  val mapping = properties(
+    label
+  ).dynamic(DynamicMapping.Strict)
+}
+
+class IndexConfigFieldsTest
```
jamieparkinson, line 31: This is v useful documentation!

project/project/project/metals.sbt
```diff
@@ -1,4 +1,4 @@
 // DO NOT EDIT! This file is auto-generated.
 // This file enables sbt-bloop to create bloop config files.
 
-addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.4.8")
+addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.4.8-19-4d9f966b")
```
jamieparkinson, line 4: Out of curiosity what causes these to change?

jamieparkinson (approved):

common/internal_model/src/test/scala/uk/ac/wellcome/models/index/IndexConfigFieldsTest.scala
jamesgorrie: Over time auto-derivation slows things down like crazy so trying not to use it.

project/project/project/metals.sbt
jamesgorrie: My IDE, which bloop is used for local dev, so I just left it. I'll look into it deeper, and probably pin my IDE version to make sure it's not a common thing.

PR #1586, opened 2021-04-30
Tei windows lambda

alicefuzier:
This is the first bit of https://github.com/wellcomecollection/catalogue-pipeline/issues/1536


common/window_generator/src/build_windows.py
```diff
@@ -0,0 +1,118 @@
...
+
+Options:
+    --start=<START>         When to start polling
+    --end=<END>             The earliest point to stop polling
+    --window_length=<WINDOW_LENGTH>
+                            The number of minutes per window
+    --resource=<RESOURCE>   Should the pipeline fetch bibs or items?
+
+
+This script generates windows to poll for updates from Sierra, and sends
```
alexwlchan, line 18: nit: don't mention Sierra here

alicefuzier: thanks 👍

common/window_generator/src/build_windows.py
```diff
@@ -0,0 +1,118 @@
...
+    --start=<START>         When to start polling
+    --end=<END>             The earliest point to stop polling
+    --window_length=<WINDOW_LENGTH>
+                            The number of minutes per window
+    --resource=<RESOURCE>   Should the pipeline fetch bibs or items?
+
+
+This script generates windows to poll for updates, and sends
+them to the SNS topic that triggers our pipeline.  You pass it an interval
+(start, end), and then it generates windows of length ``window_length`` that
```
jamesgorrie, line 20:
```suggestion
(start, end), and then it generates windows of length `window_length` that
```

PR #1590, opened 2021-05-04
Only run the merger in CI, and with logging enabled

alexwlchan:
There's a persistently flaky merger test, and I can't make it fail locally.

This PR removes all but that test, and enables stdout logging in CI – I'm hoping this will give a clue about why it's failing, and help us un-fail it.


alexwlchan:
Replaced by #1592.

PR #1594, opened 2021-05-05
remove batching from the aspect ratio inferrer

harrisonpim:
Failures in the new image inferrer don't seem to be caused by insufficient resource, which makes us think it might be an issue with the number of available threads. All of the inferrers currently run on the same host, and all make use of some clever but complicated batching functionality. 

This PR removes the batching functionality from the aspect ratio inferrer, where it's probably unnecessary.


jamieparkinson:
Oh, I didn't realise you'd already removed the `joblib.parallel` usage - you may as well try this, but I'm not sure it will make very much difference as the batcher isn't multithreaded, it's asynchronous ("real" threads come from the use of `joblib.parallel`)

pipeline/inferrer/aspect_ratio_inferrer/app/encoder.py
```diff
@@ -1,26 +0,0 @@
...
-    def process_image(self, image):
-        return image.width / image.height
-
-    def __call__(self, images):
-        """
-        process images in parallel
-        """
-        return Parallel(n_jobs=-2)(
-            self.delayed_process_image(image) for image in images
-        )
```
harrisonpim, line 26: `joblib.Parallel` was used in `encoder.py`, which is removed by this PR

jamieparkinson: ah yes sorry - think that's the more important change. Probably no harm in removing the batching though

jamieparkinson (approved):

alicefuzier (approved):

PR #1595, opened 2021-05-05
Index all the fields on a SourceIdentifier

alexwlchan:
In the requesting service, we want to be able to find items with a specific SourceIdentifier.  We need to be able to match on all three fields, not just the value (which isn't guaranteed unique).

This patch changes the index config so we'll index all three fields, and will allow us to start querying against them.

For https://github.com/wellcomecollection/catalogue-api/issues/8

### DIFF
--- a/common/internal_model/src/main/scala/uk/ac/wellcome/models/index/IndexConfigFields.scala
+++ b/common/internal_model/src/main/scala/uk/ac/wellcome/models/index/IndexConfigFields.scala
@@ -2,6 +2,7 @@ package uk.ac.wellcome.models.index
 
 import com.sksamuel.elastic4s.ElasticDsl._
 import WorksAnalysis._
+import com.sksamuel.elastic4s.requests.mappings.KeywordField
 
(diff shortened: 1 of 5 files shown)


jamesgorrie (approved):
What makes it not unique? The fact we need to combine `id.type` and `id.value`?

alexwlchan:
> What makes it not unique?

If you look in the ID minter table, we key on a combination of (value, identifierType, ontologyType). This means if you search for "value" (either in the table or in the API index), you may find multiple identifiers.

e.g. this is totally possible:

```
value     | identifierType       | ontologyType
----------+----------------------+--------------
b12345678 | sierra-record-number | Work
b12345678 | mets                 | Work
```

If you wanted to find, say, (b12345678, sierra-record-number, Work), there's no easy way to do it.

jamesgorrie:
Do you then want this to be a [nested field](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html)? As you won't be able to do that query here due to object flattening explained above.

PR #1603, opened 2021-05-11
Add a script for fetching all the source data for a work

alexwlchan:
Sometimes it's really useful to see all the source data that went into a particular Work – in particular, I often grab records and run them in the transformer with IntelliJ, so I can see exactly why a particular transformation was applied.

Actually finding the source records is moderately fiddly and annoying, so I've written a script that downloads them all for you:

```console
$ python3 get_source_records.py a2242545
/var/folders/_c/h0g4c9_n6rg8d2v509hztzdc0000gn/T/a2242545o7gyqyo3/calm-record-id_129b0826-3ac4-4aea-8b0c-490d5889df46.json
/var/folders/_c/h0g4c9_n6rg8d2v509hztzdc0000gn/T/a2242545o7gyqyo3/sierra-system-number_b19871831.json
```


harrisonpim (approved):
so useful!

harrisonpim:
out of interest, are there any situations where one _wouldn't_ want to `apply_cleanups`?

alexwlchan:
> out of interest, are there any situations where one wouldn't want to `apply_cleanups`?

I use it if I want to test something in the Sierra transformer.

We store Sierra source records like:

```
{
  "id": "1234",
  "bibRecord": "{\"id\":\"1234\",\"fixedFields\":{\"108\":…}}",
  …
}
```

By default, the script will unpack that second layer of JSON encoding, so it can be pretty-printed and made readable.

If you pass `apply_cleanups=False`, it doesn't unpack that string, and you get something you can pass directly into the Sierra transformer code – because it expects to see the record as it's actually stored.

harrisonpim:
makes sense 👍  thanks!

PR #1607, opened 2021-05-13
Use the iconographic number as a reference number, not a shelfmark

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5172. Follows #1606 

If a bib is part of the Visual Collections (=material type `k`/Pictures or `r`/3D Objects) and it has something that looks like an i-number in field 001, we:

* Use that value as the `referenceNumber`
* Add an identifier with identifier type `iconographic-number`
* Hide the shelfmark on any associated items

This might need further refinement (e.g. if a bib has two items, and we only want to hide the shelfmark on one of them) – but this takes us in the right direction.

I've also separated the `referenceNumber` type on the WorkData model from `collectionPath` – I don't want works with an iconographic number to wait for the relation embedder, which will do nothing for them.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
@@ -2,6 +2,18 @@ package weco.catalogue.internal_model.identifiers
 
 import io.circe.{Decoder, Encoder, HCursor, Json}
 
+/** A reference number is a special type of identifier.
+  *
+  *     when [the identifier is] the reference we have written on the actual item, they
+  *     should be the referenceNumber
+
+  *     Ie there is a number written in pencil on the old piece of paper and it is XXXXX -
+  *     that's the reference number, as it's what people will see when they have the
+  *     object in their hands
+  *
+  * See https://wellcome.slack.com/archives/C294K7D5M/p1620733997159000 for more
+  *
+  */
 class ReferenceNumber(val underlying: String) {
   override def toString: String = underlying
 
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
@@ -1,0 +1,35 @@
+package weco.catalogue.internal_model.identifiers
+
+import io.circe.{Decoder, Encoder, HCursor, Json}
+
+class ReferenceNumber(val underlying: String) {
+  override def toString: String = underlying
+
+  // Normally we use case classes for immutable data, which provide these
+  // methods for us.
+  //
+  // We deliberately don't use case classes here so we skip automatic
+  // case class derivation for JSON encoding/DynamoDB in Scanamo,
+  // and force callers to intentionally import the implicits below.
+  def canEqual(a: Any): Boolean = a.isInstanceOf[ReferenceNumber]
+
+  override def equals(that: Any): Boolean =
+    that match {
+      case that: ReferenceNumber =>
+        that.canEqual(this) && this.underlying == that.underlying
+      case _ => false
+    }
+
+  override def hashCode: Int = underlying.hashCode
+}
+
+object ReferenceNumber {
+  def apply(underlying: String): ReferenceNumber =
+    new ReferenceNumber(underlying)
+
+  implicit val encoder: Encoder[ReferenceNumber] =
+    (value: ReferenceNumber) => Json.fromString(value.toString)
+
+  implicit val decoder: Decoder[ReferenceNumber] = (cursor: HCursor) =>
+    cursor.value.as[String].map(ReferenceNumber(_))
+}
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
@@ -4,6 +4,7 @@ import weco.catalogue.internal_model.identifiers.{
   CanonicalId,
   DataState,
   IdState,
+  ReferenceNumber,
   SourceIdentifier
 }
 import weco.catalogue.internal_model.image.ImageData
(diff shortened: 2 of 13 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
```diff
@@ -0,0 +1,35 @@
...
+import io.circe.{Decoder, Encoder, HCursor, Json}
+
+class ReferenceNumber(val underlying: String) {
+  override def toString: String = underlying
+
+  // Normally we use case classes for immutable data, which provide these
+  // methods for us.
+  //
+  // We deliberately don't use case classes here so we skip automatic
+  // case class derivation for JSON encoding/DynamoDB in Scanamo,
```
jamesgorrie, line 12: Could you explain why this is useful?

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
```diff
@@ -107,9 +108,23 @@ case class WorkData[State <: DataState](
...
   workType: WorkType = WorkType.Standard,
-)
+) {
+
+  // We have separate fields for collectionPath and referenceNumber so we
+  // can set a referenceNumber which isn't part of a collection.
+  // (e.g. an Iconographic number on Sierra records.)
+  //
+  // If a collectionPath is set, we should always require a referenceNumber
+  // to be set, based on its value.
```
jamesgorrie, line 121: Why is this a requirement?

jamesgorrie (commented):
A few questions on some information I can't seem to fathom from the comments.

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
alexwlchan: This is a pattern copied from a bunch of other places where we have types that are fancy names for strings – e.g. CanonicalId, SierraRecordNumber, ExternalIdentifier in the storage service. We want them to be distinct types in Scala so you don't mix them up.

If you used a case class, you'd get a JSON serialisation something like:

```json
{"underlying": "12345i"}
```

whereas our custom encoder/decoder will turn it into a string:

```
"12345i"
```

And the latter is preferable because it's a bit easier for any other tools working with the data (one less layer of indirection), and it's backwards-compatible with any existing data stores that used the old string-based serialisation.

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
alexwlchan: All collection paths are reference numbers (e.g. `PP/ALM/1`), but not all reference numbers are collection paths (e.g. `12345i`). This assertion was an attempt to encode that – if we've got a collection path, we should always set a reference number – but it breaks the Scalacheck tests and is not that important, so I've taken it out.

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
jamesgorrie: Ah, OK I see - so to avoid unintentional automatic derivation, we just don't let it be possible. 

I do feel having the auto stuff is _occasionally_ useful, but also creates compilation performance errors, and creates issues like this, where we have to create a special class for something that already exists.

Should we remove auto-derivation and either expect all case classes to define their encoders, or have a mega-encoder like [this](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/common/internal_model/src/main/scala/uk/ac/wellcome/models/Implicits.scala)?

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
jamesgorrie: Could you define what a reference number is?

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
jamesgorrie: bikeshedding - Not for this PR

jamesgorrie (approved):

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
alexwlchan: There's a longer thread in Slack here: https://wellcome.slack.com/archives/C294K7D5M/p1620733997159000

> when they are the reference we have written on the actual item, they should be the referenceNumber
>
> Ie there is a number written in pencil on the old piece of paper and it is XXXXX - that’s the reference number, as it’s what people will see when they have the object in their hands

jamesgorrie: Could have that on the class as it's more accessible to future devs.

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/ReferenceNumber.scala
alexwlchan: That mega-encoder is still using the auto-derivation, it’s just caching the results so they don't have to be re-derived every time.

jamesgorrie: Yea, true. I mean that we could use a `derivedDecoder`, but explicitly set it on a case class, in the case of the above, have a special one.

alexwlchan: I'm going to go ahead and extract some of this logic in a separate PR, because we repeat it quite a lot – and then we can add a thorough set of comments and tests around it.

jamesgorrie: Sounds great!

alexwlchan: ✨ https://github.com/wellcomecollection/scala-libs/pull/98 ✨

PR #1611, opened 2021-05-14
The field name is CustodialHistory, not CustodHistory

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5173

If you look at the data in the reporting cluster, you'll see there's no field `CustodHistory`, but there is one called `CustodialHistory`:

Screenshot 2021-05-14 at 13 05 44

That's what we should be using in the Calm transformer.


kenoir (approved):
🍮

PR #1617, opened 2021-05-19
Plumb the Miro source overrides into the reindexer and the Miro transformer

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5176, the first part of https://github.com/wellcomecollection/docs/pull/61

The broad strokes: this PR adds a new model `MiroSourceOverrides`, which will be stored directly in the DynamoDB table of the Miro VHS. Values set here will override anything in the source data, e.g. if we want to mark a particular image as "in copyright" instead of CC BY.

This patch modifies:

* the reindexer to send these overrides to the transformer, if there are any
* the transformer to pay attention to these overrides, if there are any

I'll get these deployed, then write the Python that actually updates these values.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -49,6 +49,7 @@ lazy val flows = setupProject(
 lazy val source_model = setupProject(
   project,
   folder = "common/source_model",
+  localDependencies = Seq(internal_model),
   externalDependencies = CatalogueDependencies.sourceModelDependencies
 )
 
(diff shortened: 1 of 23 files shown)


build.sbt
```diff
@@ -49,6 +49,7 @@ lazy val flows = setupProject(
 lazy val source_model = setupProject(
   project,
   folder = "common/source_model",
+  localDependencies = Seq(internal_model),
```
jamieparkinson, line 52: Is this a bad coupling? It feels uncomfortable

jamieparkinson (approved):

build.sbt
alexwlchan: Yeah, I don't love it – but I couldn't think of a better way to do this that doesn't involve passing implicit knowledge of the internal models around. 😕

PR #1618, opened 2021-05-19
A handful of quality-of-life fixes for pipeline developers

alexwlchan:
Spotted while looking at the Miro licenses, but not specifically part of that.


jamieparkinson (approved):
🎉

PR #1624, opened 2021-05-21
Add flag to dump all work data for a matcher graph

jamieparkinson:
This is useful for debugging merging issues!


alexwlchan:
I would be careful of calling this "source"; we have another script that gets the original records from Miro/Sierra/whatever – https://github.com/wellcomecollection/catalogue-pipeline/blob/main/scripts/get_source_records.py

Maybe "--save-works"?

jamieparkinson:
Good point, fiddly language

kenoir (approved):

pipeline/matcher/scripts/get_graph.py
```diff
@@ -56,7 +58,12 @@ def main(index_date, work_id):
 
     node_ids = [node["id"] for node in graph_component]
     node_links = [node["linkedIds"] for node in graph_component]
-    nodes = get_nodes_properties(es, index_date=index_date, work_ids=node_ids)
+    nodes = get_nodes_properties(
+        es,
+        index_date=index_date,
+        work_ids=node_ids,
+        fetch_complete_source=emit_work_data,
```
alexwlchan, line 65: Change the name of this parameter to use "work"?

alexwlchan (commented):
I

PR #1625, opened 2021-05-21
Redirect digaids Miro images even when their items aren't merged

jamieparkinson:
See discussion following https://wellcome.slack.com/archives/C8X9YKM5X/p1621598033034300?thread_ts=1620662731.355000&cid=C8X9YKM5X for some context; comments should cover the reasoning for the chosen implementation. Scalafmt did some weird stuff with this PR for some reason.

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ImageDataRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ImageDataRule.scala
@@ -18,11 +18,9 @@ object ImageDataRule extends FieldMergeRule {
   ): FieldMergeResult[FieldData] =
     FieldMergeResult(
       data = getMetsPictureAndEphemeraImages(target, sources).getOrElse(Nil) ++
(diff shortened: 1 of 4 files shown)


alexwlchan (commented):
Let me double check: does this cause the Miro work to redirect to the Sierra work, even though we haven't copied any of the data across? Or are they still separate works?

jamieparkinson:
Exactly correct, that is the desired behaviour which wasn't happening before this PR

jamieparkinson:
lol, I accidentally deleted the change that was the point of the whole PR 🙃

alexwlchan (commented):
Could we get a scenario test for this? I'd find it easier to see what's going on if I wasn't interpreting through the medium of FieldMergeResult effects.

jamieparkinson:
The scenario test is already updated to cover this case - the change is to make sure that the desired behaviour (of not creating Miro works/images) covers all cases of digaids works, not just those with single items. I will add a comment making that explicit.

alexwlchan:
Ah, I see you've repurposed the single-item Sierra test. Got it.

alexwlchan (approved):

PR #1626, opened 2021-05-21
Implement our own copy of the Sierra Rules for Requesting

alexwlchan:
This is a key component for getting access status into the API for https://github.com/wellcomecollection/platform/issues/5171, as described in https://github.com/wellcomecollection/docs/pull/60

---

One key part of "can I request this thing" is the Sierra Rules for Requesting, a set of rules that prevent an item:

1. From being requestable in the OPAC/Encore, i.e., hiding the request button
2. From being requested in the Sierra API

In some cases, these rules are the only thing that prevent an item from being requestable – e.g. if you look at [b1818948 in Encore](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1818948?lang=eng), the status is "Available" and the opacmsg is "Online request", but you can't actually request it because the R4R blocks it. (Data Protection Act.)

These are specified in a MARC-like format, e.g.

```
q|i||79||=|mfulc||Item cannot be requested online. Please contact Medical Film & Audio Library.   Email: mfac@wellcome.ac.uk. Telephone: +44 (0)20 76118596/97.
```

which means

> If an item has "mfulc" in fixed field 79 (location), then block requesting this item and show the message "Item cannot be requested online. Please contact Medical Film & Audio Library.   Email: mfac@wellcome.ac.uk. Telephone: +44 (0)20 76118596/97."

These rules map remarkably well onto Scala-style pattern matching. I've reimplemented the rules in Scala in this patch, with comments that explain the original rules they came from.

I've asked Louise S to send us new rules whenever they get updated, so we can keep out code up-to-date.

---

This patch is just about getting the logic into our codebase. I'll wire it up in a separate message. Here's a rough guess of how it'll go:

* In the Sierra transformer, we'll use this to hide the "Online request" message if it isn't appropriate for an item. This means the front-end can use "Online request" as a strong indicator that something is, in fact, requestable.
* In the items API, we'll use this to return the most up-to-date information about an item – so if an item has just been scanned on the hold shelf, it'll immediately display as unavailable on wellcomecollection.org.
* In the requests API, we could use this to explain why we can't make a request. I want to double check that Sierra really will block these requests in its API – if so, we could use it for better error reporting if we think that's important. (Although ideally no user should hit the requests API if an item fails these rules).

---

I've put this new object in the source_model library, and moved some other Sierra models to match. I think we'll soon want to share those models with the catalogue-api repo, and this is a step towards that.


common/source_model/src/test/scala/weco/catalogue/source_model/sierra/SierraRulesForRequestingTest.scala
```diff
@@ -0,0 +1,331 @@
...
+
+      forAll(testCases) {
+        assertBlockedWith(
+          _,
+          expectedMessage =
+            "Please complete a manual request slip.  This item cannot be requested online.")
+      }
+    }
+
+    it("if it's sepep, rm001 or rmdda") {
```
kenoir, line 208: What do these mean?

alexwlchan: Not sure, they're just the codes from the R4R file.

kenoir (approved):
👏 Super clear comments and test cases.

Presumably we'll need to build an export a source models package as with the internal model?

alexwlchan:
> Presumably we'll need to build an export a source models package as with the internal model?

Yup, but I'll do that later.

common/source_model/src/test/scala/weco/catalogue/source_model/sierra/SierraRulesForRequestingTest.scala
kenoir: Only stands out because the other tests have human parseable reasons.

PR #1631, opened 2021-05-25
Start to model the access conditions from Sierra

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5171

This patch starts to map various bits of Sierra access data into our model. The full explanation is in the RFC, but hopefully this code is understandable on its own.

This is an initial implementation that covers the "easy" cases (782,533 of 786,987 items, or ~99.5%), but there's a long tail of weird ones that need more investigation. I'm deliberately not using it anywhere except the tests yet – I don't want this on the critical path for the transformer until I've dug into that long tail some more.

---

The crux of this patch is a new class:

```scala
object SierraAccessCondition {
  def apply(bibId, bibData, itemId, itemData): (List[AccessCondition], ItemStatus)
}
```

It returns two things:

- The access conditions of the item, to be presented through the catalogue API. This will take time to react to changes made in Sierra, e.g. if somebody places a hold on an item.
- The status of the item, to be used in the items API. Here, we'd query Sierra to get the latest item information, and use that to update the page.

At some point we might split these two bits of logic, but they're so closely intertwined that it made sense to develop them together.

---

For the purposes of the front end, the string "Online request" in the access condition terms will indicate whether an item can ever be requested online, and whether it's worth querying the items API to get up-to-date availability.

This code is a bit more conservative than Encore/the OPAC – in general I think a false negative (not showing the button when we could show it) is preferable to a false positive (showing the button when we shouldn't). It's also more consistent – a lot of the issues in the current interface are caused by disagreeing sources of data. This code tries to create a more self-consistent output.

---

If this approach doesn't make sense or people can't follow it, then it'd be good to know now, before I start working on the long tail. I'll deal with the long tail either by adding extra cases, or getting the data changed in Sierra.


alexwlchan:
Closing as superseded by various other PRs.

PR #1632, opened 2021-05-26
Some refactoring to add access conditions on items

alexwlchan:
This patch contains some pieces spun out of https://github.com/wellcomecollection/catalogue-pipeline/pull/1631 to be a bit more reviewable, with some refactoring to make it easier to add access conditions in a subsequent patch.

Specifically, this patch:

* Adds a type hierarchy to Rules for Requesting, so we can pattern match on them without looking at the exact strings
* Makes the correspondence between the original rules and our code a bit clearer
* Adds a method for finding the display note on an item, which is one of the fields we use for access conditions

### DIFF
--- a/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
+++ b/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
@@ -4,10 +4,31 @@ sealed trait RulesForRequestingResult
 
 case object Requestable extends RulesForRequestingResult
 
-case class NotRequestable(message: Option[String] = None)
-  extends RulesForRequestingResult
+sealed trait NotRequestable extends RulesForRequestingResult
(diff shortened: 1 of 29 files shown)


common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
```diff
@@ -0,0 +1,34 @@
...
+  case class OnNewBooksDisplay(message: String) extends NotRequestable
+
+  case class AtDigitisation(message: String) extends NotRequestable
+  case class AtConservation(message: String) extends NotRequestable
+
+  case class RequestTopItem(message: String) extends NotRequestable
+
+  case class BelongsInStrongroom(message: String) extends NotRequestable
+
+  case object NoReason extends NotRequestable
```
kenoir, line 33: Love the clarity of these case class names.

PR #1634, opened 2021-05-26
Simplify using run_terraform.tf

kenoir:
As in the [identity repo](https://github.com/wellcomecollection/identity/blob/main/infra/scoped/run_terraform.sh), create a custom profile so as not to rely on a developer having this configuration correct in advance.

### DIFF
--- a/infrastructure/critical/run_terraform.sh
+++ b/infrastructure/critical/run_terraform.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
 AWS_CLI_PROFILE="catalogue-pipeline-terraform"
-IDENTITY_DEVELOPER_ARN="arn:aws:iam::760097843905:role/platform-developer"
+PLATFORM_DEVELOPER_ARN="arn:aws:iam::760097843905:role/platform-developer"
 
 aws configure set region eu-west-1 --profile $AWS_CLI_PROFILE
-aws configure set role_arn $IDENTITY_DEVELOPER_ARN --profile $AWS_CLI_PROFILE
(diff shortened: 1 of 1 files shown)


infrastructure/critical/run_terraform.sh
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 
-AWS_CLI_PROFILE=${AWS_PROFILE:-platform-dev}
+AWS_CLI_PROFILE="catalogue-pipeline-terraform"
```
jamesgorrie, line 3: Where is this profile from? Am I expected to have it in my `.aws/credentials`?

jamesgorrie (commented):
Does this set it forever in was locally?

infrastructure/critical/run_terraform.sh
kenoir: This change creates that profile for you. Currently this works by expecting you to have a profile of a particular name, which is annoying because it expects everyone to set this up themselves and follow an implicit naming scheme. 

This uses the AWS CLI to configure that profile for you so you don't have to know or care.

kenoir: > Does this set it forever in was locally?

It'll try and recreate it every time you run it, it will overwrite a profile of the same name if it already exists. Overwriting is only a problem if someone has already created a profile of this name and is using it for something else, which I think is unlikely.

jamesgorrie: I'm not sure if I am being silly - but where is it creating the profile?

PR #1635, opened 2021-05-26
Parse access conditions for 98.7% of Sierra items

alexwlchan:
It's not quite 99.5%, but I'm close!

This PR adds a new object `SierraItemAccess`, which creates an access condition for a Sierra item to use in the catalogue pipeline. It also creates an ItemStatus to use in the items API (Available/Unavailable/TemporarilyUnavailable). Although these are different services, the logic is similar enough that it's worth developing them in parallel.

There are ~10k items left that this code can't handle – either I need to add more test cases, or I need to get the data fixed in Sierra. I haven't wired this code in anywhere, because I want to do a bit more analysis before I put it on the critical path.

I'd like to get this much reviewed and merged as-is, and come back to add extra cases later. I think this is 80% complete, but it'll take another 80% of the time (and code!) to get the remaining 20%, so to speak.

### DIFF
--- a/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
+++ b/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
@@ -146,6 +146,25 @@ object SierraItemAccess extends SierraQueryOps {
               note = itemData.displayNote)),
           ItemStatus.TemporarilyUnavailable)
 
+      // An item which is restricted can be requested online -- the user will have to fill in
+      // any paperwork when they actually visit the library.
+      //
+      // Example: b29459126 / i19023340
+      case (
+        Some(AccessStatus.Restricted),
+        Some(0),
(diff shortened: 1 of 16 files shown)


common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
```diff
@@ -0,0 +1,267 @@
...
+    val status = itemData.status
+    val opacmsg = itemData.opacmsg
+    val isRequestable = SierraRulesForRequesting(itemData)
+
+    (bibStatus, holdCount, status, opacmsg, isRequestable, location) match {
+
+      // Items on the closed stores that are requestable get the "Online request" condition.
+      //
+      // Example: b18799966 / i17571170
+      case (
```
kenoir, line 46: Expressing this logic using pattern matching is 👌

kenoir (approved):

PR #1646, opened 2021-05-28
Actually use the new item access in the transformer

alexwlchan:
This is a pretty critical step to getting the data on the site!

I've also added a message directing people to Encore if we can't map an access status yet, so they'll be a bit more visible and we can direct our efforts to get them cleaned up.

This might need us to bring back that link in certain circumstances, annoyingly.


alexwlchan:
> Adding the link back when we don't know if we can offer access will need co-ordinating with @gestchild & @davidpmccormick.

Yeah, once the data is in the API then I'll put together a little guide to what the data means for the website.

PR #1647, opened 2021-05-28
Continuously fetch updates for electronic holdings that are under embargo

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5133

That ticket explains why this is necessary – essentially, it's possible for Sierra holdings records to change with the `updatedDate` being modified, so our current adapter won't pick them up. This Lambda will poll to see which holdings records should be updated, then send a window that will ask the adapter to re-fetch them. I've already deployed this and verified that it has the desired effect (i.e. the Catalogue API shows the correct date).

I'll leave the original issue open until Monday, then check I got the CloudWatch rule right and it actually is running at 2am each morning.


jamesgorrie (approved):

kenoir:
This adds a dependency on the reporting cluster to the pipeline. It might be easy to forget that. The lamda is well documented which is great, but perhaps it would be good to document this dependency in: https://github.com/wellcomecollection/catalogue-pipeline/tree/main/sierra_adapter?

PR #1653, opened 2021-06-02
bump elastic version

jamesgorrie:
[As per the scala libs](https://github.com/wellcomecollection/scala-libs/blob/main/project/Dependencies.scala#L9)

Strangely `elastic4s` is `7.12.2`, where that doesn't really exist, elastic only goes to `7.12.1` - but `elastic4s` is weird.


alicefuzier (approved):

jamesgorrie:
Actually this needs a whole lot of update, going to get the collections work in first, then update everything that uses this. elastic4s has decided to use a whole different DSL.

PR #1660, opened 2021-06-07
Link to wellcomelibrary.org directly from the access terms

alexwlchan:
There are access terms that contain HTML that we'll need to render on /works (e.g. links to the Access Policy).

This patch modifies the message _"Please check this item on the Wellcome Library website for access information"_ (which gets added when we can't yet determine the access status) to include a hyperlink to the page in Encore. This lets us keep all the logic for including this link in a single place, rather than having data quality issues leak into the front-end code.

### DIFF
--- a/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
+++ b/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
@@ -9,7 +9,11 @@ import weco.catalogue.internal_model.locations.{
   LocationType,
   PhysicalLocationType
 }
-import weco.catalogue.source_model.sierra.{SierraItemData, SierraItemNumber}
+import weco.catalogue.source_model.sierra.{
+  SierraBibNumber,
+  SierraItemData,
+  SierraItemNumber
+}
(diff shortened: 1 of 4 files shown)


kenoir:
I'd check with @gestchild whether this is desirable as it makes decisions about front-end formatting that might be better off to remain in the front-end e.g. maybe that anchor tags should have some other attributes on it, is the link text accessible? TBH I think in this case it's probably fine, but worth checking.

jamesgorrie:
Just for info - we do [render HTML from notes](https://github.com/wellcomecollection/wellcomecollection.org/blob/22f8aae5952a38161c9d4894a0699423d3a0f645/catalogue/webapp/components/WorkDetailsText/WorkDetailsText.tsx#L25)(note `dangerouslySetHtml`), so that's handy.

If we wanted to test other text, we could by just checking the note and not rendering the content from the API.
```tsx
if (notes.accessCondition) {
  <a href="thingsthanks">CustomText</a>
}
```

2¢ - definitely worth checking with the experience team as there might have been a reason _not_ to have included the link this way.

gestchild:
Looks good to me

jamesgorrie (approved):

PR #1664, opened 2021-06-07
created adapter shaped config so we can stop commenting

jamesgorrie:
Got bored of commenting / uncommenting. 

Thought I'd try define an adapter config shape. We could probably make the stack module a little smarter, but didn't feel it was worth changing for now, this is just to make creating a new stack to reindex to a little less faffy.


kenoir (approved):
Neat

PR #1667, opened 2021-06-08
Don't add Availability.InLibrary if the location is offsite

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5190

This isn't perfect, but it takes us in the right direction and removes a bunch of items from the "In the library" filter that are actually in somebody else's library.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Availability.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Availability.scala
@@ -64,7 +64,8 @@ object Availabilities {
         // can actually see, so we don't include items that have been ordered
         // but aren't available to view yet.
         case physicalLoc: PhysicalLocation
-          if physicalLoc.locationType == LocationType.OnOrder => false
+            if physicalLoc.locationType == LocationType.OnOrder =>
+          false
 
         // Don't include items if they have a physical location but are actually
         // held in another institution.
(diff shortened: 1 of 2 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Availability.scala
```diff
@@ -78,6 +93,27 @@ object Availabilities {
...
+    def termsAreOtherInstitution: Boolean =
+      ac.terms match {
+        case Some(t) if t.toLowerCase.contains("available at") => true
+        case Some(t) if t.toLowerCase.contains("available by appointment at") =>
+          true
+
+        case Some(t) if t.contains("Churchill Archives Centre") => true
+        case Some(t) if t.contains("UCL Special Collections and Archives") =>
+          true
+        case Some(t) if t.contains("at King's College London") => true
```
jamesgorrie, line 106: Wondering why we have `at` here, but not with the `Churchill Archives Centre`? Not that it's an issue, just made me wonder.

jamesgorrie (approved):

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Availability.scala
alexwlchan: Possibly the Churchill Archives string is "in the", or something? I don't remember.

PR #1668, opened 2021-06-08
Publish the source_model library to S3

alexwlchan:


kenoir (approved):
I think the permissions here might also need updating: https://github.com/wellcomecollection/platform-infrastructure/blob/main/accounts/platform/iam_platform_ci.tf#L24

PR #1671, opened 2021-06-08
Fix BuildKite publishing of source_model

alexwlchan:
The previous commit to main published source_model 4268, but didn't publish any version of internal_model.

Because source_model has a dependency on internal_model, when you try to load that in the catalogue-api repo, you get an error:

> I need internal_model 4268 to go with source_model 4268, but I can't find it. 💥 

See the BuildKite failures for https://github.com/wellcomecollection/catalogue-api/pull/105. Publishing both libraries if either of them has changed will get around this for now.

This isn't an issue in scala-libs because publishing there is an all-or-nothing process – either everything gets published, or nothing does.

For https://github.com/wellcomecollection/catalogue-api/issues/103

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/locations/AccessStatus.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/locations/AccessStatus.scala
@@ -24,10 +24,10 @@ sealed trait AccessStatus extends EnumEntry { this: AccessStatus =>
   }
 }
 
-object AccessStatus extends Enum[License] {
+object AccessStatus extends Enum[AccessStatus] {
   val values = findValues
   assert(
-    values.size == values.map { _.id }.toSet.size,
+    values.size == values.map { _.name }.toSet.size,
     "IDs for AccessStatus are not unique!"
   )
 
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -71,21 +71,7 @@ steps:
   - label: publish internal model
     if: build.branch == "main"
     env:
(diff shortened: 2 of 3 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/locations/AccessStatus.scala
```diff
@@ -24,10 +24,10 @@ sealed trait AccessStatus extends EnumEntry { this: AccessStatus =>
   }
 }
 
-object AccessStatus extends Enum[License] {
+object AccessStatus extends Enum[AccessStatus] {
```
jamesgorrie, line 27: Was this meant to slip in here?

alexwlchan: Oops. It was not. I can pull it into a separate PR if you think it's important, but it's just a typo I spotted while working on this and I forgot to pull into a separate PR.

jamesgorrie: It looks like more than a typo? If it's a noop, all good to stay here.

jamesgorrie (approved):

common/internal_model/src/main/scala/weco/catalogue/internal_model/locations/AccessStatus.scala
alexwlchan: typo or copy/paste error. Should be a no-op, but CI will confirm.

PR #1678, opened 2021-06-14
add eager ordinals to works config

jamesgorrie:


kenoir:
I'd like to be sure that this has actually improved performance, as it has a potential negative impact on indexing.

jamesgorrie:
> I'd like to be sure that this has actually improved performance, as it has a potential negative impact on indexing.

Good shout, I'll swap back to WIP

PR #1680, opened 2021-06-15
Add time to messages

alicefuzier:
the lambda needs to send a timestamp in the message so that the next service can order messages


tei_adapter/tei_updater/src/tei_updater.py
```diff
@@ -57,19 +59,28 @@ def diff_trees(old_tree, new_tree):
...
+            }
+            for path in paths_changed
         ]
     if items_added:
         messages += [
             {
                 "path": get_path_from_diff(added.path()),
-                "url": new_tree[get_path_from_diff(added.path())]["url"],
+                "uri": new_tree[get_path_from_diff(added.path())]["uri"],
+                "timeModified": time.isoformat(),
```
jamesgorrie, line 74: Bikeshed - would it be better to have `"status": "deleted | modified | etc"` that you can differentiate on, and just have a "time" attribute?

jamesgorrie (approved):

tei_adapter/tei_updater/src/tei_updater.py
alicefuzier: I don't really mind to be honest. I have a slight preference for different messages rather than statuses with different values because in the scala side you can leave it up to circe to decode into one or the other and then pattern match on the type. But you can pattern match on the value of status as well so 🤷‍♀️

PR #1685, opened 2021-06-16
Tei id extractor handles new files

alicefuzier:
For reference as to what the TEI Id extractor is https://github.com/wellcomecollection/docs/pull/62
This PR makes the TEI id extractor:
* receive a message about a path in the tei repo that has changed
* read the tei file from github
* get the id from the file
* store the file in s3
* send a message with the location in s3 of the file and the id

It doesn't store the path and the id anywhere yet, and it doesn't handle files deleted or moved. That will be added in a subsequent pr


alexwlchan:
I'm going to review this now.

alexwlchan (changes requested):
This mostly looks good, and was less of a tangle to review than I was expecting, but I'd like more in the way of signposting and comments for future readers to understand why it's implemented this way. Specific suggestions and comments inline.

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/Main.scala
```diff
@@ -13,10 +15,14 @@ object Main extends WellcomeTypesafeApp {
...
-    TeiIdExtractorWorkerService(
+    new TeiIdExtractorWorkerService(
       messageStream = SQSBuilder.buildSQSStream(config),
-      messageSender = SNSBuilder
-        .buildSNSMessageSender(config, subject = "TEI id extractor")
+      messageSender =
+        SNSBuilder.buildSNSMessageSender(config, subject = "TEI id extractor"),
+      gitHubBlobReader =
+        new GitHubBlobReader(config.requireString("tei.github.token")),
+      store = S3TypedStore[String](???, ???),
```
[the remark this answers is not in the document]
alicefuzier: This is fixed in #1687 .  I haven't bothered to do it properly here cause this version is not meant to be deployed ( In fact the deployed version is from #1687 )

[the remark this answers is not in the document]
alexwlchan: Gotcha.

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobReader.scala
```diff
@@ -0,0 +1,71 @@
...
+    val request = HttpRequest(
+      uri = Uri(uri.toString),
+      headers = List(
+        Accept(
+          MediaType.applicationWithFixedCharset(
+            "vnd.github.v3+json",
+            HttpCharsets.`UTF-8`)),
+        `Cache-Control`(CacheDirectives.`no-cache`),
+        Connection("keep-alive"),
+        Authorization(OAuth2BearerToken(token))
```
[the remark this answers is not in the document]
alicefuzier: Actually I don't think we do. The github documentation recommends - but doesn't require - the Accept header with the api version, so that things don't shift under your feet. EVerything else was me trying to understand why I was getting 403 and it turned out it was rate limiting.
I will delete the ones we don't need and comment the one we do need

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobReader.scala
```diff
@@ -0,0 +1,71 @@
...
+      )
+    )
+    for {
+      response <- Http().singleRequest(request)
```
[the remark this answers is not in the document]
alicefuzier: Thanks sounds sensible. I will keep using the AkkaHttpClient for testing too though, cause I don't trust my understanding of the returned HttpResponse. I'd rather leave it to Wiremock to record and send back without me having to create a response from that and assume they are the same.

One side question while I'm adding it: why does [AkkaHttpClient](https://github.com/wellcomecollection/scala-libs/blob/main/http/src/main/scala/weco/http/client/AkkaHttpClient.scala) want both an actor system and an execution  context in the constructor? can it not derive the execution context from the actor system?

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobReader.scala
```diff
@@ -0,0 +1,71 @@
...
+  private val base64 = java.util.Base64.getMimeDecoder
+
+  private def decodeBase64(s: String) = Try {
+    val bytes =
+      base64.decode(s.getBytes(java.nio.charset.StandardCharsets.UTF_8))
+    new String(bytes, java.nio.charset.StandardCharsets.UTF_8)
+  }
+
+  private def unmarshalAs[T](response: HttpResponse, request: HttpRequest)(
+    implicit um: Unmarshaller[ResponseEntity, T]): Future[T] = {
```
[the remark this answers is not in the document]
alicefuzier: why are the other ones different? maybe I'm not understanding how it works, but I though the circe support was just adding marshallers, not encodes/decoders, and if you have in scope your own decoder/encoders it will use those, or so I though

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/TeiIdExtractorWorkerService.scala
```diff
@@ -1,13 +1,105 @@
...
+                                bucket: String,
+                                teiDirectories: Set[String] = Set(
+                                  "Arabic",
+                                  "Batak",
+                                  "Egyptian",
+                                  "Greek",
+                                  "Hebrew",
+                                  "Indic",
+                                  "Javanese",
+                                  "Malay"))
```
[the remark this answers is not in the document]
alicefuzier: I can put a link to the tei repo, i think looking at that is pretty self explanatory.
The only approach that I can think of to deduce programmatically is going via a blacklist instead of a whitelist approach like here, but I think it might be a bit more fiddly? Also, I don't think new languages are added very often. 
This field could be loaded from config and everything could be moved in terraform which at least skips the pr/release/deploy bit. I don't think it's very useful doing that right now though.
I think this is something to keep in mind but IMO this is good enough for now

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobReader.scala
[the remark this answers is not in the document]
alexwlchan: > I will keep using the AkkaHttpClient for testing too though, cause I don't trust my understanding of the returned HttpResponse. I'd rather leave it to Wiremock to record and send back without me having to create a response from that and assume they are the same.

👍

> why does AkkaHttpClient want both an actor system and an execution context in the constructor? can it not derive the execution context from the actor system?

Good spot! That looks like a leftover from an older version of the code, I'll fix that now.

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobReader.scala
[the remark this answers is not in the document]
alicefuzier: Unless there's a strong opposition I think I'm going to keep the marshalling library. It still allows to pass an implicit decoder instead of an implicit unmarshaller, but it can create the unmarshaller from the decoder implicitly. Also, I like the control with the level of error details.
I'm going to change the method to have an implicit decoder instead of an implicit unmarshaller though

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/TeiIdExtractorWorkerService.scala
```diff
@@ -1,13 +1,105 @@
...
-case class TeiIdExtractorWorkerService(messageStream: SQSStream[Nothing],
-                                       messageSender: SNSMessageSender)
-    extends Runnable {
-  override def run(): Future[Any] = ???
+case class TeiFileChangedMessage(path: String,
+                                 uri: URI,
+                                 timeModified: ZonedDateTime)
+case class TeiIdChangeMessage(id: String,
+                              s3Location: S3ObjectLocation,
+                              timeModified: ZonedDateTime)
```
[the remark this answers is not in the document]
alicefuzier: No reason..It allows for different timezones while Instant is always UTC but actually that's good and ZonedDateTime is an overkill. WIll change

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/IdExtractor.scala
```diff
@@ -0,0 +1,20 @@
...
+import scala.xml.XML
+
+object IdExtractor {
+  def extractId(blobContent: String, uri: URI): Try[String] = {
+    Try {
+      val xml = XML.loadString(blobContent)
+      xml.attributes
+        .collectFirst {
+          case metadata if metadata.key == "id" => metadata.value.text
+        }
```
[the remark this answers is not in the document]
alicefuzier: I have tested this and it fails if there's more than one xml:id attribute in the same node cause it's not valid xml. It also doesn't pick up other xml:id attributes in other nodes. So the code is fine. I've committed the extra tests

alicefuzier:
I think I've addressed all your comments @alexwlchan

common/flows/src/main/scala/weco/flows/FlowOps.scala
```diff
@@ -51,4 +53,20 @@ trait FlowOps extends Logging {
           (ctx, result)
       }
       .collect { case (ctx, Right(data)) => (ctx, data) }
+
+  /**
+    * Broadcasts the output of a flow to flows `a` and `b` and merges them again
+    */
```
alexwlchan, line 59: It might be helpful to explain what "merge" means in this context – you get a new flow with all the outputs from both flows, in any order.

e.g. if `a` produced `(1, 2, 3, …)` and `b` produced `(-1, -2, -3, …)`, you might get `(1, -1, 2, 3, 4, -2, 5, -3)`.

Or have I misunderstood?

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -23,8 +24,9 @@ class PipelineStorageStream[In, Out, MsgDestination](
   messageStream: SQSStream[In],
   indexer: Indexer[Out],
   messageSender: MessageSender[MsgDestination])(
-  val config: PipelineStorageConfig)(implicit ec: ExecutionContext)
-    extends Logging {
+  val config: PipelineStorageConfig)(implicit val ec: ExecutionContext)
```
alexwlchan, line 27: Why has `ec` become a `val`?

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/MergerWorkerService.scala
```diff
@@ -33,8 +33,9 @@ class MergerWorkerService[WorkDestination, ImageDestination](
   workMsgSender: MessageSender[WorkDestination],
   imageMsgSender: MessageSender[ImageDestination],
   config: PipelineStorageConfig
-)(implicit ec: ExecutionContext)
-    extends Runnable {
+)(implicit val ec: ExecutionContext)
```
alexwlchan, line 36: Ditto above.

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobContentReader.scala
```diff
@@ -0,0 +1,71 @@
...
+import weco.http.client.HttpClient
+
+class GitHubBlobContentReader(httpClient: HttpClient, token: String)(
+  implicit ac: ActorSystem) {
+  implicit val ec = ac.dispatcher
+  def getBlob(uri: URI): Future[String] = {
+    val request = HttpRequest(
+      uri = Uri(uri.toString),
+      headers = List(
+        // Send the version of GitHub API we expect as per https://docs.github.com/en/rest/overview/media-types
```
alexwlchan, line 23: 👍

Super, thanks.

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobContentReader.scala
```diff
@@ -0,0 +1,71 @@
...
+        }
+      case HttpResponse(code, _, entity, _) =>
+        Unmarshal(entity).to[String].transform {
+          case Success(responseEntity) =>
+            Failure(new RuntimeException(
+              s"The GitHub api returned an error: ${code.value} for url ${request.uri} with headers ${response.headers}: $responseEntity"))
+          case _ =>
+            Failure(new RuntimeException(
+              s"The GitHub api returned an error: ${code.value} for url ${request.uri.toString} with headers ${response.headers
+                .toString()}: ${entity.toString}"))
```
alexwlchan, line 64: nit: 
```suggestion
              s"The GitHub api returned an error: ${code.value} for url ${request.uri} with headers ${response.headers}: $entity (couldn't decode entity as string)"))
```

This will make it more obvious which path you went down in the logs, and if memory serves the string-ified representation of an `HttpEntity` isn't particularly helpful.

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobContentReader.scala
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+  }
+}
+
+// Represents a Blob as returned by the GitHub API https://docs.github.com/en/rest/reference/git#blobs
```
alexwlchan, line 70: ✨

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/IdExtractor.scala
```diff
@@ -0,0 +1,25 @@
+package weco.catalogue.tei.id_extractor
+
+import java.net.URI
+import scala.util.{Failure, Try}
+import scala.xml.XML
+
+object IdExtractor {
+  // Extracts an id from the root node of a TEI file.
+  // The root node looks like <TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="manuscript_15651">
+  // and we want to extract the value of the xml:id attribute, "manuscript_15651" in the example
```
alexwlchan, line 10: Super. ✨

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/IdExtractor.scala
```diff
@@ -0,0 +1,25 @@
...
+      val xml = XML.loadString(blobContent)
+      xml.attributes
+        .collectFirst {
+          case metadata if metadata.key == "id" => metadata.value.text
+        }
+        .getOrElse(
+          throw new RuntimeException(s"Could not find an id in XML at $uri"))
+    }.recoverWith {
+      case th =>
+        Failure(new RuntimeException(s"Error processing XL at $uri", th))
```
alexwlchan, line 22:
```suggestion
        Failure(new RuntimeException(s"Unable to extract ID from XML at $uri", th))
```

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/TeiIdExtractorWorkerService.scala
```diff
@@ -1,13 +1,99 @@
...
+import scala.concurrent.{ExecutionContext, Future}
 
-case class TeiIdExtractorWorkerService(messageStream: SQSStream[Nothing],
-                                       messageSender: SNSMessageSender)
-    extends Runnable {
-  override def run(): Future[Any] = ???
+// Represents a path change coming from the tei_updater lambda
+case class TeiFileChangedMessage(path: String, uri: URI, timeModified: Instant)
+
+// Represents a message for the tei_adapter with changes to id instead of file path
```
alexwlchan, line 21: Explaining that these come from the Lambda is helpful. 👍

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/TeiIdExtractorWorkerService.scala
```diff
@@ -1,13 +1,99 @@
...
+              IdExtractor.extractId(blobContent, message.uri))
+            stored <- Future.fromTry(
+              store
+                .put(S3ObjectLocation(
+                  config.bucket,
+                  s"tei_files/$id/${message.timeModified.getEpochSecond}.xml"))(
+                  blobContent)
+                .left
+                .map(error => error.e)
+                .toTry)
```
alexwlchan, line 78: Break this up to make it more readable:

```suggestion
            location = S3ObjectLocation(
              bucket = config.bucket,
              key = s"tei_files/$id/${message.timeModified.getEpochSecond}.xml")    
            )
            stored <- store.put(location, blobContent) match {
              case Right(stored) => Future.successful(stored)
              case Failure(error) => Future.failed(error.e)
            }
```

tei_adapter/tei_id_extractor/src/test/scala/weco/catalogue/tei/id_extractor/IdExtractorTest.scala
```diff
@@ -0,0 +1,69 @@
...
+      """<?xml version="1.0" encoding="UTF-8"?>
+    <?xml-model href="https://raw.githubusercontent.com/bodleian/consolidated-tei-schema/master/msdesc.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
+    <?xml-model href="https://raw.githubusercontent.com/bodleian/consolidated-tei-schema/master/msdesc.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+    <TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="manuscript_12345" xml:id="manuscript_5678"></TEI>""",
+      uri
+    )
+    triedId shouldBe a[Failure[_]]
+    triedId.failed.get shouldBe a[RuntimeException]
+    triedId.failed.get.getMessage should include(uri.toString)
+  }
```
alexwlchan, line 55: Nice test. ✨

alexwlchan (approved):
Looking good, thanks for addressing my comments. ✨ 

A bunch more minor suggestions, but nothing that needs another review from me.

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
alicefuzier: I'm not 100% sure.. After moving out the broadcastAndMerge thing it complained that it didn't have access to the ExecutionContext 🤷‍♀️

common/flows/src/main/scala/weco/flows/FlowOps.scala
alicefuzier: yes, I will add it

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
kenoir: Does the `logging` trait expect this perhaps?

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobContentReader.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import java.net.URI
+import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
+import de.heikoseeberger.akkahttpcirce.ErrorAccumulatingCirceSupport._
+import io.circe.Decoder
+import uk.ac.wellcome.json.JsonUtil._
+import weco.http.client.HttpClient
+
+class GitHubBlobContentReader(httpClient: HttpClient, token: String)(
```
kenoir, line 16: Has this got any TEI logic in it? If not would it work to change the namespace here so it's not part of the id_extractor?

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobContentReader.scala
```diff
@@ -0,0 +1,70 @@
...
+  implicit ac: ActorSystem) {
+  implicit val ec = ac.dispatcher
+  def getBlob(uri: URI): Future[String] = {
+    val request = HttpRequest(
```
kenoir, line 20: I wonder if this would benefit from the HttpClient pattern @alexwlchan abstracted recently (https://github.com/wellcomecollection/scala-libs/blob/main/http/src/main/scala/weco/http/client/HttpClient.scala) for reducing WireMock interactions?

kenoir (approved):

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobContentReader.scala
alicefuzier: There's a conversation with @alexwlchan above about this. I don't want to bypass wiremock cause that requires me understanding properly the response and converting in an HttpResponse object accurately, and I don't trust myself on this, I'd rather delegate it to wiremock

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/GitHubBlobContentReader.scala
alicefuzier: No tei. It could technically live elsewhere although I'm not sure it's worth it right now

common/pipeline_storage/src/main/scala/uk/ac/wellcome/pipeline_storage/PipelineStorageStream.scala
alicefuzier: I don't think it's the logging, it shouldn't need the ExecutionContext. IIRC, in case classes any param in the class is automatically converted to a val so it's available to any method in the class. In normal classes, they're not so, unless you declare them val, they're only available when instantiating the class (I think). What I don't understand really, is why it ws working before..

PR #1686, opened 2021-06-16
Tei id extractor deals with deleted and moved files

alicefuzier:
Depends on #1685 
For reference as to what the TEI Id extractor is wellcomecollection/docs#62


tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
```diff
@@ -0,0 +1,182 @@
...
+
+import java.time.Instant
+import scala.util.{Success, Try}
+
+class PathIdManager[Dest](pathIds: PathIdTable,
+                          store: Writable[S3ObjectLocation, String],
+                          messageSender: MessageSender[Dest],
+                          bucket: String) {
+
+  def handlePathChanged(pathId: PathId, blobContent: String): Try[Unit] = {
```
alexwlchan, line 24: nit: bin the extraneous braces

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
```diff
@@ -0,0 +1,182 @@
...
+      .put(location)(blobContent)
+      .left
+      .map(error => new RuntimeException(s"Error putting $location: ", error.e))
+      .toTry
+  }
+}
+
+object PathIdManager {
+
+  def selectById(pathIds: PathIdTable, id: String)(
```
alexwlchan, line 141: Minor thing: the use of `pathIds` as a variable name is really throwing me here, because it looks like a `Seq(…)` or `Map(…)` to me. A name like `table` would make this code clearer for me.

Some type hints/comments would also make it easier for me to work out what this functions are doing.

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
```diff
@@ -0,0 +1,182 @@
...
+          case (None, None) =>
+            insert(pathId, blobContent)
+          case (None, Some(row)) =>
+            deleteOldIdInsertNewId(pathId, blobContent, row)
+          case (Some(idRow), Some(pathRow)) =>
+            if (idRow == pathRow) {
+              update(pathId, blobContent, idRow)
+            } else {
+              deleteOldIdUpdateNewId(pathId, blobContent, idRow, pathRow)
+            }
```
alexwlchan, line 42: Minor: I would roll the `if` statement into the pattern match to keep it all at a consistent indentation level.

```suggestion
          case (Some(idRow), Some(pathRow)) if idRow == pathRow =>
            update(pathId, blobContent, idRow)
          case (Some(idRow), Some(pathRow)) =>
            deleteOldIdUpdateNewId(pathId, blobContent, idRow, pathRow)
```

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  private def storeTei(pathId: PathId, blobContent: String) = {
+    val location = S3ObjectLocation(
+      bucket,
+      s"tei_files/${pathId.id}/${pathId.timeModified.getEpochSecond}.xml")
+    store
+      .put(location)(blobContent)
+      .left
+      .map(error => new RuntimeException(s"Error putting $location: ", error.e))
+      .toTry
```
alexwlchan, line 135: FWIW, I prefer a single match rather than the combo of `.left.map` and `.toTry`.

```suggestion
    store.put(location)(blobContent) match {
      case Right(stored) => Success(stored)
      case Left(err) => Failure(new RuntimeException(s"Error putting $location: ", error.e))
    }
```

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
```diff
@@ -0,0 +1,182 @@
...
+    } else Success(())
+  private def sendDeleted(pathId: PathId) =
+    messageSender.sendT[TeiIdMessage](
+      TeiIdDeletedMessage(pathId.id, pathId.timeModified))
+
+  private def storeAndSendChange(pathId: PathId, blobContent: String) =
+    for {
+      stored <- storeTei(pathId, blobContent)
+      _ <- messageSender.sendT[TeiIdMessage](
+        TeiIdChangeMessage(pathId.id, stored.id, pathId.timeModified))
```
alexwlchan, line 124: minor, just to reduce the amount spread over different lines:

```suggestion
      stored <- storeTei(pathId, blobContent)
      message = TeiIdChangeMessage(pathId.id, stored.id, pathId.timeModified)
      _ <- messageSender.sendT[TeiIdMessage](message)
```

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
```diff
@@ -0,0 +1,182 @@
...
+  private def insert(pathId: PathId, blobContent: String)(
+    implicit session: DBSession) =
+    for {
+      _ <- storeAndSendChange(pathId, blobContent)
+      - <- insertPathId(pathIds, pathId)
+    } yield ()
+
+  private def update(pathId: PathId, blobContent: String, storedPathId: PathId)(
+    implicit session: DBSession) =
+    if (pathId.timeModified.isAfter(storedPathId.timeModified)) {
```
alexwlchan, line 110: If I've read this correctly, the comparison here is `pathId.timeModified > storedPathId.timeModified`. Should we change this to `≥` instead?

That way, if the same message is sent twice, we'll still do the processing here – so we won't have the issue we had in the transformer a while back, where a document would get to the database but not be sent to SNS.

(You may have copied this from the Sierra adapter; if so, apologies – I have a feeling we do it wrong there too, but I've not got round to investigating properly yet.)

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/TeiIdExtractorWorkerService.scala
```diff
@@ -48,33 +46,54 @@ class TeiIdExtractorWorkerService[Dest](
...
+      }
+      .via(broadcastAndMerge(processDeleted, processChange))
+
+  def processDeleted =
+    Flow[(Context, TeiPathMessage)]
+      .collect {
+        case (ctx, msg) if msg.isInstanceOf[TeiPathDeletedMessage] =>
+          (ctx, msg.asInstanceOf[TeiPathDeletedMessage])
+      }
+      .delay(config.deleteMessageDelay)
```
alexwlchan, line 70: Is this a normal thing, or is there a reason we delay deleting messages here?

alexwlchan (commented):
Some initial comments inline; I'll look at it more thoroughly in the morning.

One passing comment: I hadn't realised how much RDS-related stuff you had to copy out of the ID minter. Is it worth copying that into another lib in the "common" directory?

alicefuzier:
> One passing comment: I hadn't realised how much RDS-related stuff you had to copy out of the ID minter. Is it worth copying that into another lib in the "common" directory?

Maybe, but I would invoke the rule of 3 for now. I suspect there won't be many more instances where we use RDS and if that's the case I think duplicated code is not that bad and actually is more flexible than a library somewhere. But if we copy this code somewhere else then, yes definitely it should be taken out

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
alicefuzier: yes, looks much better, thanks

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/TeiIdExtractorWorkerService.scala
alicefuzier: yeah, there should be a comment here, I will add. The reason is that if something is moved in the repo, the updater lambda has no way that is the same thing moved, cause you only know exactly what it is when you inspect the XML and get the id and the lambda never sees the XML. So what happens in that case is that the lambda sends a deleted message for the old path and a changed message for the new path. If the delete message is processed before the change message we may send a deleted message downstream. It's not a massive issue cause the change message will arrive after and override the delete one but it may cause a 'blip' where something disappears and then reappears. The delay is to avoid that 'blip'

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
alicefuzier: I didn't copy from the sierra adapter this bit. The reason is that we send the message first and we store in RDS last. So if the message fails sending, there's nothing in RDS and the message will be reprocessed. If it fails storing in RDS, the message will be sent twice downstream, which is fine. 

I actually though this needed to be gte rather than gt and wrote a test for it and realised it just never failed so I left the code and the test.

The reason why it sends the message first and updates the db after is that with deleted messages we delete the row entirely so there's no record of it ever existing. So, if you delete the row first and send the message after you have no way of recovering if sending the message fails. That message will be forever lost. The approach of sending the message first and updating the db after solves both update and delete cases

alexwlchan: Ah, I see – and because the use of RDS is entirely contained to this app, it doesn't matter if we send a message that doesn't have a corresponding database entry. (Whereas in the pipeline storage, the opposite is true – a message with a DB entry is useless, so we have to send the message second.)

Thanks for explaining. 👍

tei_adapter/tei_id_extractor/src/main/scala/weco/catalogue/tei/id_extractor/PathIdManager.scala
alicefuzier: sorry only changed the variable names, but didn't add comments. Will do now

kenoir (approved):

PR #1687, opened 2021-06-16
Tei id extractor config and terraform

alicefuzier:
Depends on #1686 


tei_adapter/tei_id_extractor/src/main/resources/application.conf
```diff
@@ -1,3 +1,14 @@
 aws.sqs.queue.url=${?queue_url}
 aws.sns.topic.arn=${?topic_arn}
 tei.github.token=${?github_token}
+aws.rds.host=${?db_host}
+aws.rds.password=${?db_password}
+aws.rds.port=${?db_port}
+aws.rds.username=${?db_username}
+aws.rds.maxConnections=${?max_connections}
+tei.id_extractor.bucket=${?bucket}
+tei.id_extractor.database=pathid
```
kenoir, line 10: Why not also parameterize these?

tei_adapter/terraform/provider.tf
```diff
@@ -2,6 +2,6 @@ provider "aws" {
   region = local.aws_region
 
   assume_role {
-    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
+    role_arn = "arn:aws:iam::760097843905:role/platform-admin"
```
kenoir, line 5: Why must this be admin?

kenoir (approved):

tei_adapter/terraform/provider.tf
alicefuzier: argh thanks. it's a leftover when I tried to destroy the stack and recreate it. developer can't do it. Will change

tei_adapter/tei_id_extractor/src/main/resources/application.conf
alicefuzier: no reason really. They're not parametrised in the id minter and I just copied that I think. Will change

PR #1692, opened 2021-06-22
Bin the use of Wiremock in the Sierra adapter tests

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5192

A bunch of the functionality around getting/refreshing the Sierra access token has been moved into scala-libs so we can share it with the requesting/items API. We can now skip using/testing it in the Sierra adapter, and instead just use the in-memory HttpClient implementation.

I find this style of test more explicit and easier to follow, and we have significantly less code to boot.


jamesgorrie (approved):
Much much easier to see what the tests are doing, nice work.

PR #1693, opened 2021-06-22
add reindexing => refreshInterval integration

jamesgorrie:
Set's `refresh_interval: -1` when reindexing. This will hopefully help a lot of the performance issues we've seen with reindexing. 

Set's it back to `IndexConfig` value when not reindexing.

This is with the assumption that we will never really be reading from indices that we are reindexing into, which I think is a safe assumption given that the perf of an index being written into is pretty poor.

I haven't added the 30.seconds refresh interval to images, as we've not seen much of a problem with that index, so I'm not going to rock the boat there until we do more testing.

Wasn't sure if this is very scala-like, but extending objects felt a bit odd, but that's probably because I've been typescripting for ages.

Blocked by https://github.com/wellcomecollection/scala-libs/pull/117/files


pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/IndexConfigOps.scala
```diff
@@ -0,0 +1,22 @@
...
+        if (isReindexing) RefreshInterval.Off else indexConfig.refreshInterval
+
+      object NewConfig extends IndexConfig {
+        override def mapping = indexConfig.mapping
+        override def analysis = indexConfig.analysis
+        override def shards = indexConfig.shards
+        override def refreshInterval = interval
+      }
+      NewConfig
+    }
```
alexwlchan, line 20: This feels like something that should be handled by making `IndexConfig` a case class.

jamesgorrie: I wasn't sure how universally useful it was going to be - happy to move it to scala-libs if you think it would be?

alexwlchan: So my concern is not so much the specific method, more the blatting of distinct types – all we know is that we get an `IndexConfig` out of this method.

If we can do that here and it's fine, it feels like maybe we should be using case classes instead of trying to shoehorn a `copy`-like method onto a trait.

jamesgorrie: Yea, thought the same. My understanding is we're using objects as that's all they are, we never really use typechecking etc.

I'll give it a go to move it over to case classes - as I think that's the more relevant structure as it's the same type with different values.

jamesgorrie: Here's a first pass at the case class implementation:
https://github.com/wellcomecollection/scala-libs/pull/117

jamesgorrie:
Closed in favour of #1703

PR #1630, opened 2021-05-24
Tei adapter service

alicefuzier:
Part of https://github.com/wellcomecollection/catalogue-pipeline/issues/1536
Depends on #1687

Issue #1536, opened 2021-04-14
TEI Github adapter


jamesgorrie:

alicefuzier:
- [ ] RFC. (https://github.com/wellcomecollection/docs/pull/57)
- [ ] TEI Windows Lambda (PR https://github.com/wellcomecollection/catalogue-pipeline/pull/1586)
- [ ] TEI GitHub Adapter Service
- [ ] TEI Adapter Service

PR #1698, opened 2021-06-22
Tei adapter service config and terraform

alicefuzier:
Part of #1536
Depends on #1630

### DIFF
--- a/tei_adapter/tei_adapter/src/main/scala/weco/tei/adapter/Main.scala
+++ b/tei_adapter/tei_adapter/src/main/scala/weco/tei/adapter/Main.scala
@@ -23,8 +23,8 @@ object Main extends WellcomeTypesafeApp {
       DynamoBuilder.buildDynamoClient(config)
     new TeiAdapterWorkerService(
       messageStream = SQSBuilder.buildSQSStream(config),
-    messageSender =
-      SNSBuilder.buildSNSMessageSender(config, subject = "TEI adapter"),
+      messageSender =
+        SNSBuilder.buildSNSMessageSender(config, subject = "TEI adapter"),
       store = new DynamoSingleVersionStore(
         DynamoBuilder.buildDynamoConfig(config, namespace = "tei")
       ),
--- a/tei_adapter/tei_adapter/src/main/scala/weco/tei/adapter/Main.scala
+++ b/tei_adapter/tei_adapter/src/main/scala/weco/tei/adapter/Main.scala
@@ -28,8 +28,8 @@ object Main extends WellcomeTypesafeApp {
       store = new DynamoSingleVersionStore(
         DynamoBuilder.buildDynamoConfig(config, namespace = "tei")
       ),
-      config.requireInt("tei.adapter.parallelism"),
-      config.getDuration("tei.adapter.delete.delay").toScala
+      parallelism = config.requireInt("tei.adapter.parallelism"),
+      delay = config.getDuration("tei.adapter.delete.delay").toScala
     )
   }
 }
--- a/tei_adapter/tei_adapter/src/main/scala/weco/tei/adapter/Main.scala
+++ b/tei_adapter/tei_adapter/src/main/scala/weco/tei/adapter/Main.scala
@@ -1,9 +1,35 @@
 package weco.tei.adapter
 
+import akka.actor.ActorSystem
+import uk.ac.wellcome.messaging.typesafe.{SNSBuilder, SQSBuilder}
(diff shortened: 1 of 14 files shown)


tei_adapter/tei_adapter/src/main/scala/weco/tei/adapter/Main.scala
```diff
@@ -1,9 +1,35 @@
...
+      DynamoBuilder.buildDynamoClient(config)
+    new TeiAdapterWorkerService(
+      messageStream = SQSBuilder.buildSQSStream(config),
+      messageSender =
+        SNSBuilder.buildSNSMessageSender(config, subject = "TEI adapter"),
+      store = new DynamoSingleVersionStore(
+        DynamoBuilder.buildDynamoConfig(config, namespace = "tei")
+      ),
+      config.requireInt("tei.adapter.parallelism"),
+      config.getDuration("tei.adapter.delete.delay").toScala
```
alexwlchan, line 32: Keyword arguments please!

tei_adapter/terraform/tei_adapter.tf
```diff
@@ -0,0 +1,85 @@
...
+    parallelism              = 10
+    delete_delay             = "2 minutes"
+    tei_adapter_dynamo_table = aws_dynamodb_table.tei_adapter_table.id
+  }
+
+  secret_env_vars = {
+  }
+
+  // The total number of connections to RDS across all tasks from all ID minter
+  // services must not exceed the maximum supported by the RDS instance.
```
alexwlchan, line 30: This comment doesn't make sense here.

tei_adapter/terraform/tei_adapter.tf
```diff
@@ -0,0 +1,85 @@
...
+    metrics_namespace        = "${local.namespace}_tei_adapter"
+    queue_url                = module.tei_adapter_queue.url
+    topic_arn                = module.tei_adapter_topic.arn
+    parallelism              = 10
+    delete_delay             = "2 minutes"
+    tei_adapter_dynamo_table = aws_dynamodb_table.tei_adapter_table.id
+  }
+
+  secret_env_vars = {
+  }
```
alexwlchan, line 27: I think you can omit this if it's empty.

tei_adapter/terraform/topics.tf
```diff
@@ -15,3 +19,15 @@ data "aws_iam_policy_document" "publish_to_tei_id_extractor_topic" {
...
+  statement {
+    actions = [
+      "sns:Publish",
+    ]
+
+    resources = [
+      module.tei_adapter_topic.arn
+    ]
+  }
+}
```
alexwlchan, line 33: There's a `publish_policy` output on our SNS topic module you can use rather than declaring it again.

tei_adapter/terraform/locals.tf
```diff
@@ -13,7 +13,8 @@ locals {
   min_capacity                     = 0
   max_capacity                     = 15
   rds_max_connections              = 45
-  tei_id_extractor_max_connections = 9
+  tei_id_extractor_max_connections = 5
+  rds_lock_timeout_seconds         = 600
```
alexwlchan, line 17: I think you can do multiplication in Terraform expressions – let's make this a bit more obvious that it's 10 minutes:

```suggestion
  rds_lock_timeout_seconds         = 10 * 60
```

tei_adapter/terraform/rds_id_extractor.tf
```diff
@@ -48,18 +48,19 @@ resource "aws_security_group" "database_sg" {
...
-  cluster_identifier = "tei-adapter-cluster-delta"
+  cluster_identifier = "tei-adapter-cluster"
   database_name      = "pathid"
   username           = local.rds_username
   password           = local.rds_password
 
   # This instance class provides us with 45 connections per instance.
   instance_count = 1
-  instance_class = "db.t3.medium"
+  instance_class = "db.t3.small"
```
alexwlchan, line 58: Is the instance class comment still correct? (And does it matter?)

alexwlchan (approved):
✨ Looks great. ✨

tei_adapter/terraform/rds_id_extractor.tf
alicefuzier: Looking [here](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraMySQL.Managing.Performance.html), the comment is actually correct, which means the comments in the id minter terraform are not

PR #1701, opened 2021-06-23
Just use the Sierra/METS/Miro items when merging with Calm

alexwlchan:
The Calm items are just a stub with nothing useful on them, so rather than copying fields into the Calm items, let's just replace them with the items that have more data.

For wellcomecollection/platform#5204

### DIFF
--- a/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
+++ b/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
@@ -135,12 +135,20 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
       Nil
     }
 
-  /**
-    * Sierra records are created from the Sierra / Calm harvest.
-    * We merge the Sierra item ID into the Calm item.
+  /** When records are harvested from Calm, both a bib and an item record are
+    * created in Sierra.  The Sierra item will have more interesting information
+    * about access status, hold count, etc.  The Calm items are just stubs.
     *
-    * If an item is digitised via METS or Miro, it is linked to the Sierra record.
-    * We merge that into the Calm item.
+    * See CalmItems.scala -- the Calm item is only:
+    *
+    *     Item {
+    *       locations: [
+    *         Location { locationType = ClosedStores }
+    *       ]
+    *     }
+    *
+    * For this reason, we keep all the items *except* the Calm item.  This means
+    * we'll also pick up any items linked to the Sierra work from METS or Miro.
     */
   private val mergeIntoCalmTarget = new PartialRule {
     val isDefinedForTarget: WorkPredicate = singlePhysicalItemCalmWork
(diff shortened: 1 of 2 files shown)


pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/rules/ItemsRuleTest.scala
```diff
@@ -228,27 +229,33 @@ class ItemsRuleTest
     }
   }
 
-  it("Adds the METS item location to the Calm item") {
```
alexwlchan, line 231: I don't think METS items can be linked to Calm items in practice without an intermediate Sierra item, so I removed this test. If this were possible, you'd want some different behaviour – only using the METS item would lose the physical location.

jamieparkinson: Agreed

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
```diff
@@ -152,27 +160,8 @@ object ItemsRule extends FieldMergeRule with MergerLogging {
...
-
-      List(
-        calmItem.copy(
-          locations = calmItem.locations ++ digitisedLocations,
-          id = sierraItemId.getOrElse(IdState.Unidentifiable)
-        )
-      )
-    }
+    ): FieldData =
+      sources.map(_.data.items).toList.flatten
```
jamieparkinson, line 164: Trying to work out what the side effects of the change to how digitised locations are merged are here...

jamieparkinson (approved):

pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ItemsRule.scala
alexwlchan: I've done a quick test locally – if I merge Calm+Sierra+METS, it looks like the final work has two items, with the physical and digitised locations separate.

Is there an easy way to say "re-run the ItemRules" on the remaining set of works?

jamieparkinson: Interesting - there is not an easy way to say that, but this reminds me of what I did recently here https://github.com/wellcomecollection/catalogue-pipeline/blob/74b0d98fb40492a7397ec2790f23b32525124f58/pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/rules/ImageDataRule.scala#L19-L37

alexwlchan: I wonder if there should be. I'm going to open a separate ticket for it.

jamieparkinson: It is sort of interesting to consider merging recursively until you reach the fixed point of the merging rule

alexwlchan: Yeah, I wonder if that's the ultimate end-state of the merger. You define a set of merging steps, and apply them all repeatedly until the result stops changing.

alexwlchan: Ticket here, because I don't think this is something we can solve in a pinch: https://github.com/wellcomecollection/platform/issues/5207

PR #1710, opened 2021-06-25
Tei transformer

alicefuzier:
Main + config + terraform will come in a separate PR
Part of #1725


common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
```diff
@@ -26,6 +26,11 @@ object IdentifierType extends Enum[IdentifierType] {
   def apply(id: String): IdentifierType =
     IdentifierType.withName(id)
 
+  case object Tei extends IdentifierType {
+    val id = "tei"
```
alexwlchan, line 30: I wonder if this is the best ID here – TEI describes an approach to manuscript cataloguing, these identifiers are IDs within that system. `tei-manuscript-number` or something?

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -0,0 +1,47 @@
...
+    val identifiersNodes = xml \\ "msDesc" \ "msContents" \ "msIdentifier" \ "altIdentifier"
+    val seq = (identifiersNodes.filter(
+      n => (n \@ "type").toLowerCase == "sierra"
+    ) \ "idno").toList
+    seq match {
+      case List(node) => Right(Some(node.text.trim))
+      case Nil        => Right(None)
+      case _          => Left(new RuntimeException("More than one sierra bnumber node!"))
+    }
+  }
```
alexwlchan, line 23: Could you provide an example of the XML you're trying to parse here? Kinda tricky to know if this code is correct without seeing it. Look at the METS transformer for examples.

Is it worth logging the nodes we got if there was more than one?

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -0,0 +1,47 @@
...
+    for {
+      xml <- Try(XML.loadString(xmlString)).toEither
+      teiXml = new TeiXml(xml)
+      _ <- Try(
+        require(
+          teiXml.id == id,
+          s"Supplied id $id did not match id in XML ${teiXml.id}"
+        )
+      ).toEither
+    } yield teiXml
```
alexwlchan, line 46: Is there a reason the return type is an Either? Seems like a Try might be better, given the internals.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiDataParser.scala
```diff
@@ -0,0 +1,46 @@
...
+  description: Option[String] = None,
+  bNumber: Option[String]
+) {
+  def toWork(time: Instant, version: Int): Work[Source] = {
+    val maybeBnumber = bNumber
+      .map(
+        b =>
+          MergeCandidate(
+            SourceIdentifier(IdentifierType.SierraSystemNumber, "Work", b),
+            "Bnumber present in TEI file"
```
alexwlchan, line 29: nit: keyword arguments please

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiTransformer.scala
```diff
@@ -0,0 +1,22 @@
...
+
+class TeiTransformer(store: Store[S3ObjectLocation, String]) extends Transformer[TeiMetadata]{
+  override def apply(id: String, sourceData: TeiMetadata, version: Int): Result[Work[WorkState.Source]] = sourceData match {
+    case TeiChangedMetadata(s3Location, time) =>
+      for {
+        xmlString <- store.get(s3Location).left.map(_.e)
+        teiXml<-TeiXml(id, xmlString.identifiedT)
+        teiData <- TeiDataParser.parse(teiXml)
+      } yield teiData.toWork(time, version)
+    case TeiDeletedMetadata(time) => Right(Work.Deleted[Source](version, WorkData(), Source(SourceIdentifier(IdentifierType.Tei, "Work", id), time), DeletedReason.DeletedFromSource("Deleted by TEI source")))
```
alexwlchan, line 20: Keyword arguments please!

pipeline/transformer/transformer_tei/src/test/scala/weco/pipeline/transformer/tei/TeiDataParserTest.scala
```diff
@@ -0,0 +1,60 @@
...
+
+class TeiDataParserTest extends AnyFunSpec with Matchers with TeiGenerators{
+  val id = "manuscript_15651"
+
+  it("parses a tei xml and returns TeiData"){
+    val description = "a manuscript about stuff"
+    TeiDataParser.parse(TeiXml(id, teiXml(id = id, summary = Some(summary(description))).toString()).right.get) shouldBe Right(TeiData(id, Some(description), None))
+  }
+  it("parses a tei xml and returns TeiData with bNumber"){
+  val bnumber = "b1234567"
```
alexwlchan, line 15: Suggestion: there's something like SierraDataGenerators that will create Sierra b-numbers for you.

pipeline/transformer/transformer_tei/src/test/scala/weco/pipeline/transformer/tei/TeiTransformerTest.scala
```diff
@@ -0,0 +1,67 @@
...
+import java.nio.charset.StandardCharsets
+
+class TeiTransformerTest
+    extends AnyFunSpec
+    with Matchers
+    with InstantGenerators {
+  it("transforms into a Work") {
+    val teiXml =
+      IOUtils.resourceToString("/WMS_Arabic_1.xml", StandardCharsets.UTF_8)
+    val location = S3ObjectLocation("bucket", "key.xml")
```
alexwlchan, line 24: Suggestion: you can use S3ObjectLocationGenerators to create this.

pipeline/transformer/transformer_tei/src/test/scala/weco/pipeline/transformer/tei/TeiXmlTest.scala
```diff
@@ -0,0 +1,18 @@
...
+
+class TeiXmlTest extends AnyFunSpec with Matchers with TeiGenerators {
+  val id = "manuscript_15651"
+
+  it("fails parsing a tei XML if the supplied id is different from the id in the XML"){
+    val suppliedId = "another_id"
+    val result = TeiXml(suppliedId, teiXml(id = id).toString())
+    result shouldBe a [Left[_,_]]
+    result.left.get.getMessage should include (suppliedId)
+    result.left.get.getMessage should include (id)
```
alexwlchan, line 15: Can we be a bit more explicit about the message here?

kenoir:
This PR is a bit chonky!

alexwlchan (commented):
Mostly looks good, just a couple of suggestions about adding more examples inline. 👍

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiTransformer.scala
kenoir, line 20: Break this up a bit so it's a bit more readable?

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
alicefuzier: Mostly because it plays well when using it in TeiTransformer. I don't really see a big difference between an `Either[Throwable, _]` and a `Try[_]` semantically. Both can do the same thing and carry the same amount of information. So I prefer to have one of the two types everywhere rather than mixing them and converting as necessary

pipeline/transformer/transformer_tei/src/test/scala/weco/pipeline/transformer/tei/TeiXmlTest.scala
alicefuzier: I usually prefer not to be too specific about the error message in tests because effectively it's not testing a functionality but testing logging and logs should be easy to change without having to also change the corresponding test IMO. The main aim was to test that whatever the message is, it mentions the two ids because that's probably what we are going to need if we ever have to debug this

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
jamesgorrie: Don't we use `Result[TeiXml]` in places? I can't find the source type for `Result` though.

alicefuzier: Yes, but it's just an alias for `Either` usually. I think it's defined in the `Store` trait and in a bunch of other places

alicefuzier:
I've addressed your comments @alexwlchan. Is it ok for me to merge?

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
```diff
@@ -26,6 +26,11 @@ object IdentifierType extends Enum[IdentifierType] {
   def apply(id: String): IdentifierType =
     IdentifierType.withName(id)
 
+  case object Tei extends IdentifierType {
+    val id = "tei-manuscript-id"
+    val label = "Tei manuscript id"
```
alexwlchan, line 31: 👍 

Do we need another type of identifier for the multiple works inside a TEI file, as you said at standup?

alexwlchan (commented):
Changes look good. 👍

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
alicefuzier: Not for now, but possibly once I implement Adrian's suggestion for representing those multiple works. But all of that is still in modelling so I wouldn't worry for now

PR #1713, opened 2021-06-28
Improve how we handle the note on Sierra items

alexwlchan:
Improvements for https://github.com/wellcomecollection/platform/issues/5210

* We become a lot pickier about what we call terms and a note.  The terms describe what you can do with the thing if you get hold of it (e.g. copyright restrictions), the note just generally tells you about the thing.
* We add support for the item-level note, when the note is for helping readers distinguish between items rather than telling them how they access it (e.g. "uncoloured impression").


common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
```diff
@@ -40,6 +42,41 @@ object SierraItemAccess extends SierraQueryOps with Logging {
     bibStatus: Option[AccessStatus],
     location: Option[PhysicalLocationType],
     itemData: SierraItemData
+  ): (Option[AccessCondition], Option[String], ItemStatus) =
```
kenoir, line 45: This return type is complex enough to warrant its own type (rather than being a tuple). I don't think that needs to be done in this PR, but I may pick that up when I come back and remove `ItemStatus`.

kenoir (approved):

PR #1716, opened 2021-06-28
Create the 2021-06-27 pipeline, change how we do indexing

alexwlchan:
This sets up the 2021-06-27 pipeline, and starts making changes towards https://github.com/wellcomecollection/platform/issues/5215

Here's how things worked previously:

Screenshot 2021-06-29 at 11 58 26

* We have a pipeline-storage cluster which is created per-pipeline
* The ingestor reads data from the pipeline-storage cluster, and writes it to a `works-{date}` index in the catalogue cluster
* We have auto-following set up from catalogue to catalogue-api, and a cron job that would set up a follow from the catalogue cluster to rank

This had a number of issues:

- We're paying for the intermediate "catalogue" cluster
- We can't easily scale the catalogue cluster up/down when we're doing reindexes
- The reindex load caused issues on the catalogue-api cluster – trying to replicate an index that was being reindexed would cause slowness in the catalogue API

Here's how things work in this patch:

Screenshot 2021-06-29 at 12 02 34

* We have a pipeline-storage cluster which is created per-pipeline, and which scales up for reindexes
* The ingestor writes to the `works-indexed-{date}` index in the pipeline-storage cluster
* When the reindex is done, we use the included script to set up a CCR job into the catalogue-api cluster


alexwlchan:
Currently the rank following is set up by a cron job that runs once a day (see [rank/pages/api/ccr.tx](https://github.com/wellcomecollection/catalogue-api/blob/main/rank/pages/api/ccr.ts), [.github/workflows/crr.yml](https://github.com/wellcomecollection/catalogue-api/blob/main/.github/workflows/ccr.yml)).

I think the rank following should be folded into the script I've added to this repo, rather than duplicated. Reasons:

- My new script is already set up to create multiple CCR replicas of the indexes:

    ```python
    for cluster in (catalogue_api, rank):
        add_pipeline_as_remote_cluster(cluster)
        create_index_follow(cluster, pipeline_date)
    ```

    The existing rank scripts don't add the leader cluster as a remote cluster, so they won't work with the pipeline-storage cluster. We could enhance those scripts, but that seems like redundant work.

- The rank cluster would pick up images automatically.
- The rank cluster would always be up-to-date with catalogue, rather than a day or so behind changes.

kenoir (approved):

kenoir:
We should think about how we deal with deleting old remote servers / follower indices, but we can do that later.

alexwlchan:
> We should think about how we deal with deleting old remote servers / follower indices, but we can do that later.

One approach might be to create a "catalogue-api" cluster per pipeline.

jamesgorrie:
Sorry @alexwlchan - but I am a little lost as to what triggers the `create_index_follow` method to run?

alexwlchan:
> I am a little lost as to what triggers the create_index_follow method to run?

Ah, you mean the script?

Somebody runs that manually after the reindex is finished – it's not done automatically or before the reindex starts, because that would cause performance issues in the catalogue-api cluster.

(I guess I could make it automatic and have it pause/unpause the follows based on whether we're reindexing.)

alexwlchan:
Ticket here to track running the script automatically: https://github.com/wellcomecollection/platform/issues/5223

PR #1718, opened 2021-06-29
Remove unused ItemStatus

kenoir:
ItemStatus was intended to be used in the Items API, but we've changed our plans and now it is unnecessary.

Part of https://github.com/wellcomecollection/catalogue-api/issues/35


alexwlchan (approved):
🎉

PR #1724, opened 2021-07-01
Change how we map item access data (again)

alexwlchan:
Previously: we put access conditions that apply to every item on a work in the top-level "TermsOfUse" note, to avoid repeating them throughout the items table.

The new design for item data means the repetition is less obvious, and it's more useful to copy the conditions down onto individual items – so that’s what we’re doing.

This patch implements three changes:

* Access conditions in the Calm transformer are now copied onto the item
* Access conditions in the Sierra transformer include terms from bib field 506 $a
* We're dropping the "to" field, and putting that information in the access terms. Often the date is repeated there anyway, so having it in a structured field is somewhat tricky.

This has implications for merging that I'm mostly ignoring, because I'm hoping changes to the Calm/Sierra harvest will make it easier without us having to do anything. See comment on the merger.

Closes https://github.com/wellcomecollection/platform/issues/5228


common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
```diff
@@ -388,6 +396,27 @@ object SierraItemAccess extends SierraQueryOps with Logging {
...
     }
+  } match {
+    case Some(accessCondition) =>
+      // We set the access terms on an item from the access terms on the bib.
+      // Note that we need to be careful about merging terms from Calm/Sierra harvested
+      // items, which is why we assert that we haven't set terms based on anything
+      // *except* 506 subfield ǂa.
+      require(accessCondition.terms.isEmpty)
+
+      val bibTerms = {
```
kenoir, line 407: So this means I will definitely need to get the bib data to construct `AccessCondition` correctly?

kenoir (approved):

PR #1729, opened 2021-07-02
remove catalogue_pipeline_2021-06-08 pipeline

jamesgorrie:
Not applied - will after I get the green light.
We've gone far away enough from this now for it to feel safe to remove.


alexwlchan (approved):
Note that you can also bin the `catalogue` cluster: https://github.com/wellcomecollection/catalogue-pipeline/blob/main/infrastructure/critical/elastic_catalogue.tf

PR #1731, opened 2021-07-02
remove catalogue ec outputs

jamesgorrie:
Missed these.


jamesgorrie:
Merged as it was blocking `apply`s

PR #1733, opened 2021-07-02
Miro removal script changes

jamieparkinson:
This adds a simple CLI for suppressing a Miro work (which can be given either a Miro ID or a catalogue ID). It also uses the pipeline cluster rather than the API cluster, which require the ES clients to be quite granular in terms of permissions.


scripts/suppress_miro.py
```diff
@@ -0,0 +1,50 @@
...
+            identifiers = response_data["identifiers"]
+            miro_id = next(
+                i["value"]
+                for i in identifiers
+                if i["identifierType"]["id"] == "miro-image-number"
+            )
+            print(f"Miro identifier: {miro_id}")
+        except Exception:
+            raise click.ClickException(
+                "It looks like that isn't an identifier for a Miro image!"
```
alexwlchan, line 43: minor: log the thing you don't think looks like a Miro ID.

Also, is this telling us it's not a Miro ID, or that it's not present in the catalogue?

scripts/_common.py
```diff
@@ -26,14 +27,33 @@ def get_secret_string(session, *, secret_id):
...
-    password = get_secret_string(session, secret_id="catalogue/ingestor/es_password")
+    secret_path = f"elasticsearch/pipeline_storage_{pipeline_date}"
+
+    if doc_type not in ["work", "image"]:
+        raise Exception("doc_type must be 'work' or 'image'")
+
+    if action == "read":
+        user = "read_only"
+    elif action == "write":
+        user = f"{doc_type}_ingestor"
```
jamesgorrie, line 42: Should we create a new `scripts` user to avoid having to recycle more that 1 service if a key leaks?

jamesgorrie: They're quite easy to add with the [script we use for this]https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/terraform/scripts/create_pipeline_storage_users.py).

jamesgorrie (approved):
nice

scripts/_common.py
```diff
@@ -26,28 +27,57 @@ def get_secret_string(session, *, secret_id):
...
+        session,
+        secret_id=f"elasticsearch/pipeline_storage_{date}/catalogue_api/es_password",
     )
+
+    return Elasticsearch(f"{protocol}://{username}:{password}@{host}:{port}")
+
+
+def get_ingestor_es_client(date, doc_type):
+    """
+    Returns an Elasticsearch client for the catalogue cluster.
```
alexwlchan, line 63: You can use these docstrings to explain why these clients are different (one has write privileges, the other read, by the looks of things?).

scripts/miro_updates.py
```diff
@@ -155,51 +170,42 @@ def _remove_image_from_elasticsearch(*, miro_id):
...
             for hit in works_resp["hits"]["hits"]
-            if hit["_source"]["state"]["sourceIdentifier"]["identifierType"]["id"]
+            if hit["_source"]["debug"]["source"]["identifier"]["identifierType"]["id"]
             == "miro-image-number"
         )
     except StopIteration:
         print(f"Could not find a work for {miro_id} in {works_index}", file=sys.stderr)
+        print(
+            "It could be that the canonical work for this Miro ID is a Sierra work - that should be suppressed by collections information first"
+        )
```
alexwlchan, line 191: print to stderr here, like the previous line

alexwlchan (approved):

PR #1737, opened 2021-07-06
Tei merger rules

alicefuzier:


pipeline/merger/src/test/scala/weco/pipeline/merger/MergerScenarioTest.scala
```diff
@@ -304,5 +304,42 @@ class MergerScenarioTest
...
+        .getMerged(teiWork)
+        .data
+        .otherIdentifiers should contain allElementsOf physicalSierra.data.otherIdentifiers
+        .filter(_.identifierType == IdentifierType.SierraIdentifier)
+      outcome
+        .getMerged(teiWork)
+        .data
+        .otherIdentifiers should contain allElementsOf digitalSierra.data.otherIdentifiers
+        .filter(_.identifierType == IdentifierType.SierraIdentifier)
+    }
```
jamesgorrie, line 343: Really  clear!

pipeline/merger/src/test/scala/weco/pipeline/merger/rules/OtherIdentifiersRuleTest.scala
```diff
@@ -82,6 +84,23 @@ class OtherIdentifiersRuleTest
       )
     )
 
+  it("merges METS, Miro, Calm and Sierra source IDs into Tei target") {
```
jamesgorrie, line 87: Are these tests replications of the scenarios above?

jamesgorrie (approved):

pipeline/merger/src/test/scala/weco/pipeline/merger/rules/OtherIdentifiersRuleTest.scala
jamieparkinson: fwiw - we do err on the side of testing behaviour both in the rule tests and in the scenario tests, because historically stuff has gone wrong "in between" and it's been hard to tell which part has broken

alicefuzier:
On second thought, I'm going to merge this. The transformer is not deployed and doesn't even have a main class so won't do anything even if accidentally deployed, so this should be a no-op until the transformer is in place.

PR #1742, opened 2021-07-07
Extract Miro merge candidates from CALM records

jamieparkinson:
We are now cataloguing M-series Miro records, which can be linked to directly from a CALM record rather than via an intermediary Sierra bib. These links live in the MiroID/Wheels field, and are extracted by this PR.


pipeline/transformer/transformer_calm/src/main/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidates.scala
```diff
@@ -0,0 +1,42 @@
...
+    record.getList("Wheels").map { miroIdentifier =>
+      MergeCandidate(
+        IdState.Identifiable(
+          SourceIdentifier(
+            identifierType = IdentifierType.MiroImageNumber,
+            ontologyType = "Work",
+            value = miroIdentifier
+          )
+        )
+      )
```
alexwlchan, line 27: Can we put a reason on these MergeCandidates?

(Currently the "reason" is an `Option[String]`, but this is currently the only transformer where we don't set one. It feels like the reason is more and more useful as merging gets more complicated, and requiring it in the type system would be a way to keep us honest.

Changing the MergeCandidate model would be a breaking change to internal model – anything that came through the old Calm transformer would have a `null` – but it opens the door to it in a future update.)

pipeline/transformer/transformer_calm/src/main/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidates.scala
```diff
@@ -0,0 +1,42 @@
...
+      MergeCandidate(
+        IdState.Identifiable(
+          SourceIdentifier(
+            identifierType = IdentifierType.SierraSystemNumber,
+            ontologyType = "Work",
+            value = bNumber
+          )
+        )
+      )
+    }
```
alexwlchan, line 41: I think, both here and for the Miro ID, that we should so some validation on what's in this field, and skip values that don't look exactly like a Miro ID or a Sierra ID – possibly dropping a warning log if they're not right.

e.g. [6b9200c4-91e8-45dd-9085-1b0834ff1bdd](https://reporting.wellcomecollection.org/s/calm/app/discover#/doc/90cfbc80-d19f-11ea-8175-efb55f54ba0b/calm_catalog?id=6b9200c4-91e8-45dd-9085-1b0834ff1bdd)

```
Wheels
<a href="http://archives.wellcomelibrary.org/DServe/dserve.exe?&dsqIni=Dserve.ini&dsqApp=Archive&dsqCmd=show.tcl&dsqDb=Catalog&dsqPos=0&dsqSearch=(AltRefNo='WA/HSW/PH/A.3')" target="_blank">Click here to see the description of this item in the full archive catalogue</a>
```

which is pretty clearly not a Miro ID.

alexwlchan: Also, be aware that SourceIdentifier will throw an IllegalArgumentException if you pass it certain invalid values, e.g. with leading/trailing whitespace. In that case we should ignore the merge candidate and transform the Work anyhow – which you could either fix by wrapping this in a `Try` block, or by filtering out illegal values before you pass them into SourceIdentifier.

pipeline/transformer/transformer_calm/src/test/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidatesTest.scala
```diff
@@ -0,0 +1,90 @@
...
+import weco.catalogue.internal_model.work.MergeCandidate
+import weco.catalogue.source_model.generators.CalmRecordGenerators
+
+class CalmMergeCandidatesTest
+    extends AnyFunSpec
+    with Matchers
+    with CalmRecordGenerators {
+
+  it("creates a Sierra mergeCandidate from the BNumber field") {
+    val bnumber = "b1234567"
```
alexwlchan, line 19: This is an interesting example, because if this was the value we wouldn't be able to merge the b-number.

Sierra b-numbers are presented in one of two flavours:

* Sierra identifier: seven digits (e.g. `1234567`), or
* Sierra system number: eight digits with a prefix (e.g. `b12345672`)

but never seven digits and a prefix.

A quick skim suggests that the Calm catalogue mostly uses the eight-digit-with-prefix variant. This is another reason we might want to be stricter about what we use as merge candidates.

(I realise this is an existing problem, but since you're here I'm going to mention it.)

pipeline/transformer/transformer_calm/src/test/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidatesTest.scala
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  it("creates Miro mergeCandidates from Wheels fields") {
+    val miroIds = List("M0000001", "M0000002", "M0000003")
+    val record = createCalmRecordWith(
+      miroIds.map(id => "Wheels" -> id): _*
+    )
+    val mergeCandidates = CalmMergeCandidates(record)
+
+    mergeCandidates should contain allElementsOf miroIds.map(
```
alexwlchan, line 43: I would expect:

```suggestion
    mergeCandidates shouldBe miroIds.map(
```

to work. What am I missing?

jamieparkinson: Strictly, it's that I don't mind whether ordering is preserved

pipeline/transformer/transformer_calm/src/main/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidates.scala
jamieparkinson: Forgot that there was a `reason` field, will add it

pipeline/transformer/transformer_calm/src/main/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidates.scala
jamieparkinson: Strangely, the example you link to is not actually how that record appears in the CALM API (the live version doesn't have a `Wheels` field, and its `Modified` date is `19/03/2021` rather than `26/03/2020`) - is there something up with the way CALM data gets into the reporting cluster?

alexwlchan: That seems likely – I've had at least one other case where the reporting data is wrong.

Filed a ticket for it: https://github.com/wellcomecollection/platform/issues/5240

pipeline/transformer/transformer_calm/src/test/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidatesTest.scala
jamieparkinson: Going to address this and the other validation comment after https://github.com/wellcomecollection/catalogue-pipeline/pull/1746

alexwlchan (approved):

PR #1746, opened 2021-07-08
Consistent source identifier validation in transformers

jamieparkinson:
@alexwlchan's comments on https://github.com/wellcomecollection/catalogue-pipeline/pull/1742 made me think that it would be good to consolidate some of our ID-validating logic in the transformers, rather than maintaining slightly different regexes across all of them. This achieves that.


pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/IdentifierRegexes.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.util.matching.Regex
+
+object IdentifierRegexes {
+  /*
+   * A Sierra system number (aka a bnumber) is composed of:
+   * - A prefix 'b'
+   * - A 7 digit identifier
+   * - A check character, which can be any digit or an 'x'
+   */
+  lazy val sierraSystemNumber: Regex = "^b[0-9]{7}[0-9x]$".r
```
alexwlchan, line 12: Depending on how strict you want to be, you could also bring in the SierraRecordNumber class, which validates a correct check digit – but that's probably unnecessary.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/IdentifierRegexes.scala
```diff
@@ -0,0 +1,63 @@
...
+   *   B=Bottom
+   *   TR =Top Right
+   *   etc.
+   *
+   * To add to that, there are also M series images with suffix EA, EB, EC, ED, EE, and EF.
+   * They don't seem to mean anything (like L = left), they seem to just be running letters
+   * to slip the image in line.
+   */
+  lazy val miroImageNumber: Regex =
+    "^[A-Z][0-9]{7}[A-Z]{0,3}[0-9]?$".r
```
alexwlchan, line 44: I would double-check this – things like `AS0000081FH23` are real Miro IDs, just very weird ones.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/IdentifierRegexes.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  /*
+   * Iconographic numbers are a series of digits followed by an 'i'
+   * TODO: Are there any rules about the number of digits?
```
alexwlchan, line 48: I think there's a non-zero number of them, but that's it.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
```diff
@@ -0,0 +1,65 @@
...
+        case IdentifierType.CalmAltRefNo         => calmRefNo.toPredicate
+        case IdentifierType.CalmRecordIdentifier =>
+          /*
+           * Calm IDs follow the UUID spec
+           *
+           * e.g: f5217b45-b742-472b-95c3-f136d5de1104
+           * see: `https://search.wellcomelibrary.org/iii/encore/record/C__Rb1971204?marcData=Y`
+           */
+          str =>
+            Try(UUID.fromString(str)).isSuccess
```
alexwlchan, line 50: Could we get a `uuid.toPredicate` or something? Having the one comment among a bunch of case statements is a bit jarring.

alexwlchan (commented):
Quick comments based on an initial skim; I'll read more thoroughly tomorrow.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/IdentifierRegexes.scala
jamieparkinson: Yeah, think I'll leave it here

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/IdentifierRegexes.scala
jamieparkinson: Ha, in that case I don't think we'd ever have been able to merge them as this came from the Sierra transformer. I'll make it so that the suffix-of-the-suffix can be any number of digits?

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
jamieparkinson: Yeah I'll put it in a method. The correct regex for UUIDs is a bit scary.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/IdentifierRegexes.scala
alexwlchan: A good compromise here might be to look at Miro IDs that are currently cleared for the catalogue API, because those are the things we can care about matching.

(I realise I said "just add validation" and have then left you the fun of actually dealing with the hard parts of it.)

jamieparkinson: Of the Miro IDs that are currently cleared for the catalogue API, these 5 do not match the regex:

```
V0009118ETLL
L0025088F01
V0009118EBLL
V0009118ETRR
V0009118EBRR
```

So I'm not going to overthink it and will change the regex to:

```
^[A-Z][0-9]{7}[A-Z]{0,4}[0-9]{0,2}$
```

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+    def validatedWithWarning: Option[SourceIdentifier] = validated match {
+      case None =>
+        warn(
+          s"Invalid ${sourceIdentifier.identifierType.id}: ${sourceIdentifier.value}"
+        )
+        None
+      case some => some
+    }
```
alexwlchan, line 33: Is there ever a reason I wouldn't want a warning of an invalid identifier? Why does this exist separate from `.validated`?

alexwlchan (approved):
Do you know what's up with the massive unrelated formatting changes on your PRs? It makes it harder to review.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
```diff
@@ -0,0 +1,65 @@
...
+import scala.Function.const
+import scala.util.Try
+import scala.util.matching.Regex
+
+object SourceIdentifierValidation {
+  implicit class SourceIdentifierOps(sourceIdentifier: SourceIdentifier)
+      extends Logging {
+    import IdentifierRegexes._
+
+    def validated: Option[SourceIdentifier] =
```
kenoir, line 19: Do we ever use this outside of tests?

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
jamieparkinson: Yes - sometimes it's fine/expected for a particular type to be invalid, as we can fall back (ie for Miro library references)

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
jamieparkinson: Yes - https://github.com/wellcomecollection/catalogue-pipeline/blob/14674e27fd87910a1eb40658b3e937cd9a79c708/pipeline/transformer/transformer_miro/src/main/scala/weco/pipeline/transformer/miro/transformers/MiroIdentifiers.scala#L89

jamieparkinson:
I do not know what's up with the formatting changes and it's really irritating - my local scalafmt and the autoformatter seem to disagree - but the config/version is the same! I've had a couple of attempts at sorting it and will give it another...

kenoir (approved):

PR #1749, opened 2021-07-09
Remove the inferred "Copy N" title if it's the same on all items

alexwlchan:
It's not helpful to have a work where every item is titled "Copy 1", say. This affects ~38k items.

Closes https://github.com/wellcomecollection/platform/issues/5237

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraItems.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraItems.scala
@@ -30,8 +30,6 @@ object SierraItems extends Logging with SierraLocation with SierraQueryOps {
 
   type Output = List[Item[IdState.Unminted]]
 
-  type HasInferredTitle = Boolean
-
   /** We don't get the digital items from Sierra.
     * The `dlnk` was previously used, but we now use the METS source.
     *
@@ -112,8 +110,12 @@ object SierraItems extends Logging with SierraLocation with SierraQueryOps {
         )
     }.toList
 
-    tidyInferredTitles(items)
+    tidyTitles(items)
   }
+
+  // A "automated" title is a title that we created automatically, as opposed
+  // to one that was written by a human cataloguer.
+  private type HasAutomatedTitle = Boolean
 
   private def transformItemData(
     bibId: SierraBibNumber,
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraItems.scala
```diff
@@ -193,6 +199,39 @@ object SierraItems extends Logging with SierraLocation with SierraQueryOps {
...
 
+  /** Tidy up the inferred titles (Copy 1, Copy 2, Copy 3, etc.)
+    *
+    * The purpose of a title is to help users distinguish between multiple items.
+    * We remove the inferred title if they aren't doing that, in particular if:
+    *
+    *   1) There's only one item, and it has an inferred title
+    *   2) Every item has the same inferred title
+    *
+    * Note: (1) is really a special case of (1) for the case when there's a single item.
```
kenoir, line 210:
```suggestion
    * Note: (1) is really a special case of (2) for the case when there's a single item.
```

kenoir (approved):

kenoir:
I'm not really sure how we end up in the state where there are multiple items with the same inferred title? Don't we count them and add differing names? Or is this coming from Sierra and this is the case where it's not inferred?

alexwlchan:
> I'm not really sure how we end up in the state where there are multiple items with the same inferred title? Don't we count them and add differing names? Or is this coming from Sierra and this is the case where it's not inferred?

Right, I can see this is confusing, and the term "inferred" isn't helping.

So we can get a title in two ways:

- A cataloguer explicitly writes a title in field tag `v`
- We look at the `copyNo` field in the Sierra API response, and use that to create a title `Copy N`. e.g. if the copyNo is "1" we'd create the title "Copy 1". These are sometimes referred to explicitly in the notes on the bib (_"Copy 1 was given to us by XYZ, Copy 2 was purchased at auction, Copy 3 is signed"_).

The latter is what I meant by "inferred title", but that's not very clear – you could argue the field tag `v` title is also inferred. To make this distinction clearer, I've renamed this to an "automatic title", to distinguish it from a "manual title" written by a human.

Problem is, that `copyNo` field isn't always populated very well – and on ~38k works, there are multiple items where the underlying Sierra record has `copyNo = 1`. And in that case we'll create a whole bunch of items with the title "Copy 1", which is completely unhelpful for actually distinguishing the items, so we remove it.

PR #1751, opened 2021-07-09
Update run_terraform.sh to generate own profile

kenoir:
[Follow the pattern](https://github.com/wellcomecollection/catalogue-api/blob/main/terraform/shared/run_terraform.sh#L3) of creating an AWS profile for users, rather than expecting particular profile/role combinations to exist.


jamesgorrie (approved):

jamieparkinson (approved):

kenoir:
Spun off from https://github.com/wellcomecollection/catalogue-pipeline/pull/1747

PR #1760, opened 2021-07-13
More tei data from the transformer

alicefuzier:


pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiDataParser.scala
```diff
@@ -38,14 +46,17 @@ case class TeiData(
...
-        mergeCandidates = maybeBnumber.toList
+        mergeCandidates = maybeBnumber.toList,
+        languages = languages,
+        format = Some(Format.ArchivesAndManuscripts)
       )
     Work.Visible[Source](
-      version,
-      value,
+      version = version,
+      data = value,
```
alexwlchan, line 57: nit: should this variable be renamed to `value`?

pipeline/transformer/transformer_tei/src/test/scala/weco/pipeline/transformer/tei/TeiTransformerTest.scala
```diff
@@ -41,13 +46,29 @@ class TeiTransformerTest
...
+  it("extracts languages") {
+    val teiXml =
+      IOUtils.resourceToString("/Javanese_4.xml", StandardCharsets.UTF_8)
+    val location = createS3ObjectLocation
+    val store =
+      new MemoryStore[S3ObjectLocation, String](Map(location -> teiXml))
+    val transformer = new TeiTransformer(store)
+    val timeModified = instantInLast30Days
+    val id = "Wellcome_Javanese_4"
+    transformer(id, TeiChangedMetadata(location, timeModified), 1).right.get.data.title shouldBe Some(
```
alexwlchan, line 68: Should this be a language-based assertion?

alexwlchan (approved):

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -68,6 +72,120 @@ class TeiXml(xml: Elem) {
...
+        case (id1, id2) if id2.isEmpty && id1.isEmpty =>
+          Left(
+            new RuntimeException(
+              s"Cannot find a language id in ${n.toString()}"
+            )
+          )
+        case _ =>
+          Left(
+            new RuntimeException(s"Multiple language ids in ${n.toString()}")
+          )
```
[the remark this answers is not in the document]
alicefuzier: yes, but this is the languade id in a single textLang node, so the case where you have `<textLang  mainLang="sa" otherLang="es">Sanskrit</textLang>` In this case we don't know which one is the id. In practice it should not happen. If you have multiple languages, you have multiple `textLang` nodes (and I think there is at least one test with more than one language)

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -68,6 +72,120 @@ class TeiXml(xml: Elem) {
...
+      }
+      langId.map(id => Language(id, langText))
+    }
+    eitherLanguages.sequence
+  }
+
+  /**
+    * If the Tei has a single msItem node under msContents, we get the title from that
+    * Otherwise we get the title from the titleStmt
+    */
```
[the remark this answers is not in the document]
alicefuzier: The XML example is down in the comments to the individual parsing functions. I don't think it would be particularly helpful here

pipeline/transformer/transformer_tei/src/test/scala/weco/pipeline/transformer/tei/TeiTransformerTest.scala
alicefuzier: err right I think the test name is wrong, will fix

PR #1763, opened 2021-07-14
Don't require empty WorkData on deleted works

alexwlchan:
We can hard-code this as empty data, because deleted works will never be merged or combined with anything else -- this is what we already do on redirected works.

This is done with an eye towards more radical changes to WorkData, in particular making "title" compulsory in some future revision.

Because Circe ignores fields in the JSON that it doesn't care about, this should be fine to drop into an existing pipeline.  Additionally, this might make reindexes a bit cheaper!  We have a lot of deleted works, and not having to send an empty `WorkData` JSON blob back and forth will reduce the amount of traffic we have to send over the VPC endpoint.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
@@ -43,8 +43,8 @@ sealed trait Work[State <: WorkState] {
         })
       case Work.Invisible(version, _, _, invisibilityReasons) =>
         Work.Invisible(version, outData, outState, invisibilityReasons)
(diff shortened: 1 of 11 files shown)


jamesgorrie (approved):
Is the reason we need to hardcode it because we require in the trait further up the inheritance chain?

Why not make WorkData a property of the Works where we care about it?

alexwlchan:
> Is the reason we need to hardcode it because we require in the trait further up the inheritance chain?

Yes.

> Why not make WorkData a property of the Works where we care about it?

We probably could, that's just a more invasive change that I think might require another reindex, and I didn't want to tackle right now.

PR #1765, opened 2021-07-14
Add id field to SierraItemData

kenoir:
For https://github.com/wellcomecollection/catalogue-api/pull/191

If we request items from the Sierra API directly, we can't serialise them into a shape with an ID which makes it quite confusing if you have a list of them! This change adds the ID to `SierraItemData` which is present in the response from Sierra source data, but up till now we haven't needed it.

This allows us to simplify some things in the `SierraTransformer` also.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraItems.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraItems.scala
@@ -125,13 +125,23 @@ object SierraItems extends Logging with SierraLocation with SierraQueryOps {
     debug(s"Attempting to transform ${itemData.id}")
 
     val location =
-      getPhysicalLocation(bibId, itemData.id, itemData, bibData, fallbackLocation)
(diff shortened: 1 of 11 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraItems.scala
```diff
@@ -85,29 +87,27 @@ object SierraItems extends Logging with SierraLocation with SierraQueryOps {
...
-          itemId = itemId,
-          itemData = itemData,
-          bibData = bibData,
-          fallbackLocation = fallbackLocation
-        )
+    val items = itemDataEntries.map { itemData =>
+      transformItemData(
+        bibId = bibId,
+        itemId = itemData.id,
+        itemData = itemData,
```
alexwlchan, line 107: Minor: let's simplify this function so it only takes `itemData`, and then we pick up the `.id` inside the function itself.

alexwlchan (commented):
Looks good so far. 👍

alicefuzier (approved):

PR #1769, opened 2021-07-16
A few tidy-ups in the TEI extractor tests

alexwlchan:
I was reading some of the TEI code and spotted a few code tidies.


alicefuzier (approved):
Thank you

PR #1772, opened 2021-07-19
Fix the ordering of TEI languages

alexwlchan:
This swaps the id/label on languages from the TEI transformer, and also in the case class to make this mistake less likely to occur again (I checked the other uses of `Language.apply()`, they're all keyword-argumented-up).

I suspect we're about to pile in more logic here, so I pulled it out into a separate trait to match the other transformers – and we can exhaustively test the language-related code more thoroughly.

For https://github.com/wellcomecollection/platform/issues/5243


pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -1,13 +1,10 @@
...
 
-import weco.catalogue.internal_model.languages.Language
-
 import scala.util.Try
 import scala.xml.{Elem, Node, XML}
-import cats.implicits._
 import grizzled.slf4j.Logging
 
-class TeiXml(xml: Elem) extends Logging {
+class TeiXml(val xml: Elem) extends Logging {
```
alicefuzier, line 7: Why val?

alicefuzier (approved):

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
alexwlchan: So we can access it elsewhere – if I have an instance of `val t = TeiXml`, I can access the elem with `t.xml` – see what I'm doing in TeiDataParser.

common/internal_model/src/main/scala/weco/catalogue/internal_model/languages/Language.scala
```diff
@@ -1,3 +1,3 @@
 package weco.catalogue.internal_model.languages
 
-case class Language(label: String, id: String)
+case class Language(id: String, label: String)
```
alicefuzier, line 3: I wonder if doing this causes issues in the other transformers? It will still compile cause they are both strings. Did you check if we use Language without named arguments elsewhere?

alexwlchan: Yup!

PR #1775, opened 2021-07-20
Map TEI languages to the same MARC language codes as the rest of the pipeline

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5243

TEI is using IANA language codes, which are a mixture of 2 and 3 characters, and not consistent with the Sierra/Calm transformer. This PR modifies the TEI transformer to map these codes to MARC language codes, so we can filter/aggregate TEI works by language alongside other works.

The key object is `TeiLanguageData`, which contains the language mapping.

I wrote it iteratively, and I've included the core code in this PR. I started by writing a test that walks every file in the TEI repo, extracts the `<textLang>` elements, and tries to map them to the MARC codes with the new object. If an element can't be mapped, it prints a warning and carries on. I kept adding cases or fixing the data (see https://github.com/wellcomecollection/wellcome-collection-tei/pull/33) until I got down to two warnings:

```
./Indic/Indic_Alpha_978.xml: Unable to map TEI <textLang> node id=sa label=Sanskrit, with additional title entries in Persian script.
./Indic/Tamil/Tamil_3.xml: Unable to map TEI <textLang> node id=ta label=TamilOne side gives the Tamil version
```

The test is disabled by default because it requires a local checkout of the TEI repo, but I include it for reference and in case it's useful for other TEI work.

This patch also changes the behaviour of the TEI transformer to match our other transformers: if it can't map a `<textLang>` element, it logs a warning and skips the element rather than failing the entire work. This is based on the idea that it's better to push incomplete data through the pipeline than have works vanish post-reindex because of a data issue.

### DIFF
--- a/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/data/TeiLanguageData.scala
+++ b/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/data/TeiLanguageData.scala
@@ -84,7 +84,7 @@ object TeiLanguageData extends Logging {
       case ("bbc", "Toba-Batak")           => Some(Language(id = "btk", label = "Toba-Batak"))
       case ("btk", "Toba-Batak")           => Some(Language(id = "btk", label = "Toba-Batak"))
       case ("gu", "(Old) Gujarati")        => Some(Language(id = "guj", label = "(Old) Gujarati"))
-      case ("btd", "Batak Dairi")          => Some(Language(id = "btd", label = "Batak Dairi"))
+      case ("btd", "Batak Dairi")          => Some(Language(id = "btk", label = "Batak Dairi"))
       case ("ms", "Middle Malay")          => Some(Language(id = "may", label = "Middle Malay"))
       case ("pka", "Ardhamāgadhi Prakrit") => Some(Language(id = "pra", label = "Ardhamāgadhi Prakrit"))
       case ("pka", "Ardhamāgadhī Prākrit") => Some(Language(id = "pra", label = "Ardhamāgadhī Prākrit"))
--- a/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/data/TeiLanguageData.scala
+++ b/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/data/TeiLanguageData.scala
@@ -26,11 +26,9 @@ object TeiLanguageData extends Logging {
     *
     * Create a Language based on the MARC language code list.
     */
-  def apply(id: String, label: Option[String]): Option[Language] =
+  def apply(id: String, label: String): Option[Language] =
     (id, label) match {
 
-      case ("", None) => None
-      case ("", Some(s)) if s.trim.isEmpty => None
 
       // Map languages where there's a 1:1 correspondence between the IANA language
       // and the MARC language codes.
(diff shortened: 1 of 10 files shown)


alicefuzier:
> This patch also changes the behaviour of the TEI transformer to match our other transformers: if it can't map a <textLang> element, it logs a warning and skips the element rather than failing the entire work. This is based on the idea that it's better to push incomplete data through the pipeline than have works vanish post-reindex because of a data issue.

Given that there aren't many tei files, finding issues and fixing them is relatively easy. I think it makes sense to keep the behaviour of the tei transformer different and make it more strict. We've already found and fixed many issues by doing this and the amount of files means that issues are generally fixed almost immediately, so there's no data lost. We can revisit when size of tei data makes this impractical, but for now I would keep it strict. Also, sierra/calm/mets data is or was visible elsewhere other than weco.org, so it had more chance of having errors found and fixed before it even hit the pipeline. I don't think that's true for many tei files, so keeping the transformer strict makes sense to me.

alexwlchan:
> I think it makes sense to keep the behaviour of the tei transformer different and make it more strict.

I'll revert it for now, but I'm not sure I agree with the general rule – once our attention moves away from a transformer, we tend to be quite lax about addressing issues in the data or messages on the DLQ (witness Sierra transformer issues that persisted across multiple reindexes).

We'll be hot on these issues while TEI is front of mind, but once attention moves elsewhere I think the best-effort-transformation is more appropriate.

alicefuzier:
> We'll be hot on these issues while TEI is front of mind, but once attention moves elsewhere I think the best-effort-transformation is more appropriate.


That's possible, and if we can't keep up with things in dlqs, we can change then. But I'd say let's wait for that moment to come to do that rather than anticipate it. For now, the tei transformer strictness has been useful and with the parts work coming, I'm sure it will come up useful again.

alexwlchan:
Yeah, I agree it makes sense right now, particularly as the TEI works aren't publicly visible.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/data/TeiLanguageData.scala
```diff
@@ -0,0 +1,130 @@
...
+      case ("itk", "Judeo-Italian") =>
+        customLanguage("Italian", overrideLabel = "Judeo-Italian")
+      case ("jv", "Java") => customLanguage("Javanese", overrideLabel = "Java")
+
+      // If we're not sure what to do, don't map any language for now.  Drop a warning in
+      // the logs for us to come back and investigate further.
+      case (id, label) =>
+        warn(
+          s"Unable to map TEI language to catalogue language: id=$id, label=$label")
+        None
```
alicefuzier, line 120: Should we hard fail here as well?

alexwlchan: Ah yeah, good suggestion. Done.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiLanguages.scala
```diff
@@ -22,30 +20,49 @@ object TeiLanguages {
...
+
+        val mainLangId = (n \@ "mainLang").toLowerCase
+        val otherLangId = (n \@ "otherLangs").toLowerCase
+
+        val langId = (mainLangId, otherLangId) match {
+          case (id1, id2) if id2.isEmpty && id1.nonEmpty => Some(id1)
+          case (id1, id2) if id1.isEmpty && id2.nonEmpty => Some(id2)
+          case (id1, id2) if id2.isEmpty && id1.isEmpty =>
+            throw new RuntimeException(s"Cannot find a language ID in $n")
+            None
```
alicefuzier, line 39: This is unreachable code. It's no longer possible for this block to return None, either here or below, I think? I think you don't need options at all here

alexwlchan:
@alicefuzier Any more comments on this, or is it good to go?

alicefuzier (approved):

alicefuzier:
Nope, looks good, very meticulous!

PR #1776, opened 2021-07-20
2021-07-19 pipeline

alicefuzier:


pipeline/terraform/stack/provider.tf
```diff
@@ -3,15 +3,13 @@
...
-      configuration_aliases = [aws.catalogue]
-    }
-
     ec = {
       source  = "elastic/ec"
       version = "0.2.1"
     }
   }
 }
+
```
alicefuzier, line 12: I reverted this change cause it doesn't allow me to target in the new version, now sure why 🤷‍♀️

pipeline/terraform/main.tf
```diff
@@ -163,3 +163,90 @@ module "catalogue_pipeline_2021-07-13" {
...
   logging_cluster_id = local.logging_cluster_id
 }
+
+module "catalogue_pipeline_2021-07-19" {
+  source = "./stack"
+
+  pipeline_date = "2021-07-19"
+  release_label = "2021-07-19"
+
+  is_reindexing = true
```
alexwlchan, line 173: Can we flip this to `false` now? It's gonna run us up a bill if left on.

pipeline/terraform/stack/locals.tf
```diff
@@ -74,7 +74,7 @@ locals {
   sierra_adapter_topic_arns = var.is_reindexing ? concat(var.adapters["sierra"].topics, [var.adapters["sierra"].reindex_topic]) : var.adapters["sierra"].topics
   miro_adapter_topic_arns   = var.is_reindexing ? concat(var.adapters["miro"].topics, [var.adapters["miro"].reindex_topic]) : var.adapters["miro"].topics
   mets_adapter_topic_arns   = var.is_reindexing ? concat(var.adapters["mets"].topics, [var.adapters["mets"].reindex_topic]) : var.adapters["mets"].topics
-  tei_adapter_topic_arns    = var.is_reindexing ? concat(var.adapters["tei"].topics, [var.adapters["tei"].reindex_topic]) : var.adapters["tei"].topics
+  tei_adapter_topic_arns    = []
```
alexwlchan, line 77: I wonder if doing this here is the right approach? This means you can never create a pipeline that transforms TEI works. Would it be better to have the topics commented out at the top level, so we can create a one-off pipeline to do TEI if we want to?

Either way, let's get a comment in somewhere explaining why we've removed this.

alexwlchan (approved):

PR #1777, opened 2021-07-20
A couple of transformer improvements

alexwlchan:
* Ashley spotted an encoding issue with the Calm data; I've applied some blunt "find-and-replace" fixes. I could track down the issue properly, but it's more effort than I want to spend on this right now. For https://github.com/wellcomecollection/platform/issues/5244
* The Sierra rules for requesting now include a bit more information to help devs diagnose issues with the item access data – particularly important as we start making it more visible.

### DIFF
--- a/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
+++ b/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
@@ -30,5 +30,11 @@ object NotRequestable {
 
   case class BelongsInStrongroom(message: String) extends NotRequestable
 
-  case class NoPublicReason(message: String) extends NotRequestable
+  // This is used for items that are prevented from requesting, and the rule
+  // doesn't include a message to display to users.
+  //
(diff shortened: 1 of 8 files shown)


common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
```diff
@@ -30,5 +30,5 @@ object NotRequestable {
 
   case class BelongsInStrongroom(message: String) extends NotRequestable
 
-  case object NoReason extends NotRequestable
+  case class NoPublicReason(message: String) extends NotRequestable
```
kenoir, line 33: Might benefit from a comment explaining what "NoPublicReason" means.

kenoir (approved):

common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
alexwlchan: Good suggestion, added.

PR #1779, opened 2021-07-21
Set the match time when still holding the locks

alicefuzier:
This is implementing what is discussed here https://wellcome.slack.com/archives/C3TQSF63C/p1626278361497300?thread_ts=1626276850.494900&cid=C3TQSF63C
Basically, there was an issue with merging where something wasn't merged at all when it should have. Resending the same work unchanged to the matcher solved the issue. I'm not 100% sure what caused it but a race condition due to when we create the timestamp is the only thing I could think of.
Even if it isn't the cause of the problem, creating the timestamp while holding the locks seems more robust in general.

### DIFF
--- a/pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
+++ b/pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
@@ -24,21 +24,16 @@ import java.time.Instant
 import java.util.UUID
 
 class WorkMatcher(workGraphStore: WorkGraphStore,
-                  lockingService: LockingService[Set[MatchedIdentifiers],
+                  lockingService: LockingService[MatcherResult,
                                                  Future,
                                                  LockDao[String, UUID]])(
   implicit ec: ExecutionContext)
     extends Logging {
 
-  type Out = Set[MatchedIdentifiers]
-
   def matchWork(links: WorkLinks): Future[MatcherResult] =
     doMatch(links)
-      .map { works =>
-        MatcherResult(works = works, createdTime = Instant.now())
-      }
 
-  private def doMatch(links: WorkLinks): Future[Out] =
+  private def doMatch(links: WorkLinks): Future[MatcherResult] =
     withLocks(links, links.ids.map(_.toString)) {
       for {
         beforeGraph <- workGraphStore.findAffectedWorks(links)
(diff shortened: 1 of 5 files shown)


pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
```diff
@@ -56,27 +50,28 @@ class WorkMatcher(workGraphStore: WorkGraphStore,
         // but I haven't thought hard enough about whether it might introduce a
         // hard-to-debug consistency error if another process updates the graph
         // between us reading it and writing it.
-        _ <- if (updatedNodes.isEmpty) {
-          Future.successful(())
+        res <- if (updatedNodes.isEmpty) {
+          Future.successful(
+            MatcherResult(toMatchedIdentifiers(afterGraph), Instant.now()))
```
alexwlchan, line 55: nit:

```suggestion
            MatcherResult(toMatchedIdentifiers(afterGraph), createdTime = Instant.now()))
```

pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
```diff
@@ -56,27 +50,28 @@ class WorkMatcher(workGraphStore: WorkGraphStore,
...
 
           withLocks(links, ids = affectedComponentIds) {
-            // We are returning empty set here, as LockingService is tied to a
-            // single `Out` type, here set to `Set[MatchedIdentifiers]`.
-            // See issue here: https://github.com/wellcometrust/platform/issues/3873
-            workGraphStore.put(afterGraph).map(_ => Set.empty)
+            workGraphStore
+              .put(afterGraph)
+              .map(_ =>
+                MatcherResult(toMatchedIdentifiers(afterGraph), Instant.now()))
```
alexwlchan, line 65: nit:

```suggestion
                MatcherResult(toMatchedIdentifiers(afterGraph), matcherResult = Instant.now()))
```

pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
```diff
@@ -56,27 +50,28 @@ class WorkMatcher(workGraphStore: WorkGraphStore,
         // but I haven't thought hard enough about whether it might introduce a
         // hard-to-debug consistency error if another process updates the graph
         // between us reading it and writing it.
-        _ <- if (updatedNodes.isEmpty) {
-          Future.successful(())
+        res <- if (updatedNodes.isEmpty) {
```
alexwlchan, line 53: nit:

```suggestion
        result <- if (updatedNodes.isEmpty) {
```

alexwlchan (approved):

jamieparkinson (approved):

jamesgorrie (approved):

PR #1781, opened 2021-07-23
Add config flag to merger to activate "tei mode"

alicefuzier:


pipeline/merger/src/main/scala/weco/pipeline/merger/Main.scala
```diff
@@ -43,9 +45,12 @@ object Main extends WellcomeTypesafeApp {
         namespace = "identified-works"
       )
     )
-
+    val mergerMode =
+      config.getStringOption("merger.rules.mode").getOrElse("default")
+    val mergerRules =
+      if (mergerMode == "tei") TeiPlatformMerger else DefaultPlatformMerger
```
alexwlchan, line 51:
```suggestion
    val mergerRules = mergerMode match {
      case "tei" => TeiPlatformMerger
      case _ => DefaultPlatformMerger
    }
```

Using pattern-matching would make it slightly easier to add/remove merger rules in future.

pipeline/merger/src/main/scala/weco/pipeline/merger/rules/TargetPrecedence.scala
```diff
@@ -2,18 +2,36 @@ package weco.pipeline.merger.rules
...
+  WorkPredicate
+}
+object TeiTargetPrecedence extends BaseTargetPrecedence {
+  override val targetPrecedence: Seq[WorkPredicate] = Seq(
     teiWork,
     singlePhysicalItemCalmWork,
     sierraElectronicVideo,
     physicalSierra,
     sierraWork
   )
```
alexwlchan, line 20:
```suggestion
  override val targetPrecedence: Seq[WorkPredicate] = Seq(teiWork) ++ DefaultTargetPrecedence.targetPrecedence
```

This means the TEI merger will stay in sync with the rest of the pipeline, and emphasises the difference between them.

pipeline/merger/src/main/scala/weco/pipeline/merger/rules/TargetPrecedence.scala
```diff
@@ -2,18 +2,36 @@ package weco.pipeline.merger.rules
...
+    sierraElectronicVideo,
+    physicalSierra,
+    sierraWork
+  )
+}
+trait BaseTargetPrecedence {
+  import WorkPredicates._
+
+  // This is the canonical list of the order in which we try to select target works
+  val targetPrecedence: Seq[WorkPredicate]
```
alexwlchan, line 34: Minor: the order of definitions in this file seems backwards.

I'd expect it to go:

* Base trait
* Default implementation
* Specialised implementations

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/InvisibilityReason.scala
```diff
@@ -9,4 +9,5 @@ object InvisibilityReason {
   case object UnlinkedHistoricalLibraryMiro extends InvisibilityReason
   case class UnableToTransform(message: String) extends InvisibilityReason
   case object MetsWorksAreNotVisible extends InvisibilityReason
+  case object TeiWorksAreNotVisible extends InvisibilityReason
```
alexwlchan, line 12: nit: possibly rename this to emphasise that it's a temporary condition?

e.g. `TeiWorksAreInDevelopment`

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
```diff
@@ -115,13 +116,13 @@ trait Merger extends MergerLogging {
               state = result.mergedTarget.state,
               redirectSources = result.mergedTarget.redirectSources ++ redirectedIdentifiers
             )
-
           MergerOutcome(
-            resultWorks = redirects.toList ++ remaining ++ deleted :+ targetWork,
+            resultWorks = redirects.toList ++ passThroughTransform(
+              (remaining ++ deleted).toSeq :+ targetWork),
```
alexwlchan, line 121: Is this the best place to be running the passThroughTransform? Wouldn't it make more sense to run this *before* we put the works into the merger rules? Here you could end up mapping away the work that the others redirect to, no?

e.g. you have `redirects = Seq(A)` and `targetWork = B`, so `A` redirects to `B`. If you mess this up, you'll get `A` redirecting to `B` and `B` invisible, which is wrong.

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
```diff
@@ -115,13 +116,13 @@ trait Merger extends MergerLogging {
...
             imagesWithSources = result.imageDataWithSources
           )
       }
-      .getOrElse(MergerOutcome.passThrough(works))
+      .getOrElse(MergerOutcome(passThroughTransform(works), Nil))
```
alexwlchan, line 125: Keyword arguments!

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
```diff
@@ -176,14 +177,26 @@ object Merger {
...
+  override val targetPrecedence: BaseTargetPrecedence = DefaultTargetPrecedence
+  override val passThroughTransform: PassThroughTransform =
+    DefaultPassThroughTransform$
+}
+
+object TeiPlatformMerger extends BasePlatformMerger {
+  override val targetPrecedence: BaseTargetPrecedence = TeiTargetPrecedence
+  override val passThroughTransform: PassThroughTransform =
+    TeiPassThroughTransform$
+}
```
alexwlchan, line 190: Again, the approach of putting specialised implementations ahead of the base trait feels weird to me when reading this file.

pipeline/merger/src/main/scala/weco/pipeline/merger/rules/ImageDataRule.scala
```diff
@@ -19,9 +19,9 @@ object ImageDataRule extends FieldMergeRule {
     // We first try to merge images into Sierra targets, regardless of whether this is the principal
     // target of the graph we're currently merging (ie if there's a Calm target, it's ignored).
     // If this fails, we try to merge into a Calm target
-    TargetPrecedence
+    DefaultTargetPrecedence
```
alexwlchan, line 22: Should this be parameterised?

pipeline/merger/src/main/scala/weco/pipeline/merger/rules/TargetPrecedence.scala
jamesgorrie: Tests should catch this, and the coupling makes me feel a bit weird, but makes sense in this instance.

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
```diff
@@ -176,14 +177,26 @@ object Merger {
...
+trait BasePlatformMerger extends Merger {
```
kenoir, line 191: What's the difference between `BasePlatformMerger` and `Merger`?

pipeline/merger/src/test/scala/weco/pipeline/merger/PlatformMergerScenarioTest.scala
```diff
@@ -0,0 +1,37 @@
+package weco.pipeline.merger
+
+import weco.catalogue.internal_model.work.Work
+import weco.pipeline.merger.services.{DefaultPlatformMerger, Merger}
+
+class PlatformMergerScenarioTest extends MergerScenarioTest {
```
kenoir, line 6: Again marking out `PlatformMergerScenarioTest` and `MergerScenarioTest` - why is this one "Platform" - perhaps it should be `TEIMergerScenarioTest`?

pipeline/merger/src/main/scala/weco/pipeline/merger/rules/TargetPrecedence.scala
alexwlchan: Is the coupling wrong? To me it feels like the TEI target precedence is a specialisation of the default target precedence.

kenoir: > and the coupling makes me feel a bit weird
Coupling which things?

jamesgorrie: Coupling the default and tei... which I think makes sense.

pipeline/merger/src/test/scala/weco/pipeline/merger/PlatformMergerScenarioTest.scala
kenoir: Hmm, there already is a `TEIMergerScenarioTest` ... I am finding it confusing because the scenarios in `PlatformMergerScenarioTest` seem to concern TEI merging exclusively.

alexwlchan:
For what it's worth, this is a bit different to what I envisioned when I suggested it (which is not to say that it's wrong, just different).

The merger broadly breaks into three steps:

Screenshot 2021-07-26 at 15 36 49

This patch adds divergence at the "apply merging rules" stage – we have two different sets of rules, depending on whether or not TEI is enabled.

In my head, I was imagining we'd diverge at the "fetch works from pipeline storage" stage – with TEI works being marked as deleted/invisible at that stage. Add the right condition to the TEI work predicate (isVisible), and I think the merger rules would automatically ignore them.

TBH, I find the merger rules a bit scary and complicated and try to avoid touching them as much as possible – which might be why I thought of an approach that doesn't have to.

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
jamesgorrie: In this instance, I imagine it's just about editing the file for this work, it's easy to have the TEI (what's in focus) at the top of the file. Agree generally I'd go with the trait => trait implementations

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
jamesgorrie: I think in this instance, it is just applying the same split in inheritance as everywhere else.

I think in this case Merger is actually just a set type class ops - I'll see if I can make this more clear.

PR #1784, opened 2021-07-26
Fix multi-line physical description on Sierra records

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4201

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraPhysicalDescription.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraPhysicalDescription.scala
@@ -17,15 +17,10 @@ import weco.catalogue.source_model.sierra.source.SierraQueryOps
 // Notes:
 //
 //  - MARC field 300 and subfield ǂb are both labelled "R" (repeatable).
-//    According to Branwen, this field does appear multiple times on some
-//    of our records -- not usually on books, but on some of the moving image
-//    & sound records.
+//    This field does appear multiple times on some of our records, in particular
+//    AV records.  See the tests for an example of this.
 //
-//  - So far we don't do any stripping of punctuation, and if multiple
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraPhysicalDescription.scala
```diff
@@ -35,13 +30,17 @@ object SierraPhysicalDescription
...
-      .contentString(" ")
+  def apply(bibData: SierraBibData): Option[String] = {
+    val lines =
+      bibData
+        .varfieldsWithTag("300")
+        .flatMap { vf =>
+          vf.subfieldsWithTags("a", "b", "c", "e").contentString(" ")
+        }
+
+    lines.mkString("<br/>") match {
```
jamesgorrie, line 41: Could you give some examples of what content we have the we'll split across multiple lines? I ask as I wonder if this is the correct semantics for it?

Ping @gestchild

jamesgorrie (commented):
Looks good, I just wonder about the semantics of the `<br />` - but it's hard to tell without knowing how we use this in the source data.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraPhysicalDescription.scala
alexwlchan: > 1 videocassette (VHS) (1 min.) : sound, color, PAL.
> 1 DVD (1 min.) : sound, color

is a typical example.

jamesgorrie: Yea, strange that's probably more of a list, which it would be good if we could return that, _but_ that would be a massive change.

kenoir (approved):

jamesgorrie (approved):

PR #1786, opened 2021-07-27
Add some scalacheck implicits to use upstream

kenoir:
Following https://github.com/wellcomecollection/catalogue-api/pull/202#pullrequestreview-714796274

Moving some implicits for scalacheck into generators in the internal model. The internal model is defined in this repo so it makes sense for any generators to be provided here too.

### DIFF
--- a/common/internal_model/src/test/scala/weco/catalogue/internal_model/generators/IdentifiersGenerators.scala
+++ b/common/internal_model/src/test/scala/weco/catalogue/internal_model/generators/IdentifiersGenerators.scala
@@ -1,17 +1,27 @@
 package weco.catalogue.internal_model.generators
 
+import org.scalacheck.Arbitrary
 import weco.fixtures.RandomGenerators
-import weco.catalogue.internal_model.identifiers.{
-  CanonicalId,
-  IdentifierType,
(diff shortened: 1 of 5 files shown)


common/internal_model/src/test/scala/weco/catalogue/internal_model/generators/IdentifiersGenerators.scala
```diff
@@ -1,17 +1,27 @@
...
-    CanonicalId(randomAlphanumeric(length = 8).toLowerCase())
+  implicit val arbitraryCanonicalId: Arbitrary[CanonicalId] =
+    Arbitrary {
+      createCanonicalId
+    }
+
+  // We have a rule that says SourceIdentifier isn't allowed to contain whitespace,
+  // but sometimes scalacheck will happen to generate such a string, which breaks
+  // tests in CI.  This generator is meant to create SourceIdentifiers that
+  // don't contain whitespace.
```
jamesgorrie, line 22: I get this - but it did confuse me for a second as I thought that this related to the `createCanonicalId` below. Probably just the fact they're in the same PR though.

jamesgorrie (approved):

PR #1787, opened 2021-07-28
Filter out TEI works when using the TeiOffMerger

jamesgorrie:
Based on the good work done [here](https://github.com/wellcomecollection/catalogue-pipeline/pull/1781) just moving where we exlclude works closer to the edge.

This is Mod2.0, current is #1781 
merger_tei(1)

`TeiOnMerger` continues as usual and is where we want to be. 

`TeiOffMerger` is where we are at and filters out any TEI works at the very start of the merger step.

I've moved the TEI scnario tests into it's own suite which we will fold into the main suite once we're happy.

I played with the idea of making the TEI works `Deleted` or `Invisible` in the `TeiOffMerger` but landed up with too many questions around what if you merge into an Invisible work etc, so as with most feature switching work, I just went with what will change the least code.

I'll add in the feature switch at application level as a separate PR.


pipeline/merger/src/test/scala/weco/pipeline/merger/MergerScenarioTest.scala
```diff
@@ -332,42 +332,5 @@ class MergerScenarioTest
       And("the Sierra physical work is unaffected")
       outcome.getMerged(workWithPhysicalVideoFormats) shouldBe workWithPhysicalVideoFormats
     }
-
-    Scenario("A Tei and a Sierra digital and a sierra physical work are merged") {
```
kenoir, line 336: Can we keep this test, but use a `TeiOnMerger`?

kenoir: Seen, you've done this already 👍

kenoir (approved):

PR #1797, opened 2021-08-05
Map another bib/item status combination within the catalogue pipeline

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5253

If the bib has status "Temporarily unavailable" and the item has status "Unavailable", we now mark the whole item as "Unavailable" and don't include an outgoing link to wl.org.

This patch also adds the test harness I use for running the local Sierra transformer code against live records. You supply a b number, it fetches the records and transforms it. Quite handy for working out why something doesn't have access conditions.


jamieparkinson (commented):
As useful as this undoubtedly is, it feels uncomfortable to bring this coupling/dependency into the transformer tests proper - I know there's a comment, but I think that if there's local-only behaviour then that separation and intent should be enforced/expressed more clearly. As it stands, this means that (eg) a change in the account/role config would cause a change in the application image (since we build everything into one).

Could we put this in a different source root like `/src/test-local`? Could we make it a standalone script? Or maybe we could have it as an automated test that looks at a local file that's checked into version control, alongside a script that can fetch other instances of said file?

jamieparkinson:
Don't know why that comment didn't show up at the place that I wrote it - as you might have guessed it's about the STS stuff

alexwlchan:
> As useful as this undoubtedly is, it feels uncomfortable to bring this coupling/dependency into the transformer tests proper - I know there's a comment, but I think that if there's local-only behaviour then that separation and intent should be enforced/expressed more clearly. As it stands, this means that (eg) a change in the account/role config would cause a change in the application image (since we build everything into one).
>
> Could we put this in a different source root like /src/test-local? Could we make it a standalone script? Or maybe we could have it as an automated test that looks at a local file that's checked into version control, alongside a script that can fetch other instances of said file?

How would you feel about putting it as a code snippet in a README that you copy/paste into a local checkout?

Because you're right, I don't think we care if the account/role config drifts unless somebody is actively trying to use it – and so the snippet is provided as a starting point, and we defer the work of keeping it up-to-date until somebody wants to use it.

jamieparkinson:
Yep that makes sense too (plus a note about the dependency) - just something to make the separation really intentional.

alexwlchan:
How about now?

project/Dependencies.scala
```diff
@@ -103,7 +103,7 @@ object ExternalDependencies {
     val enumeratumScalacheck = "1.6.1"
     val jsoup = "1.13.1"
     val logback = "1.1.8"
-  }
+å  }
```
jamieparkinson, line 106: å no

jamieparkinson (approved):

project/Dependencies.scala
alexwlchan, line 106:
```suggestion
  }
```

årgh

PR #1804, opened 2021-08-10
Reduce repetition between the ingestor apps

alexwlchan:
I was reading the ingestor code and spotted some low-hanging fruit where we can tidy up the code and reduce repetition. And bin several unnecessary classes! ✨


kenoir (approved):
Very nice.

PR #1807, opened 2021-08-11
Use DynamoDB autoscaling when we're reindexing in a pipeline

alexwlchan:
So we'll get a big pile of provisioned capacity when everything goes through the matcher, but when we're waiting for the ingestor it'll scale away. 💸 💸 💸

### DIFF
--- a/pipeline/terraform/main.tf
+++ b/pipeline/terraform/main.tf
@@ -93,7 +93,7 @@ module "catalogue_pipeline_2021-08-09" {
   pipeline_date = "2021-08-09"
   release_label = "2021-08-09"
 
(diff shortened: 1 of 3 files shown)


kenoir (approved):

jamieparkinson (approved):
Maybe add a comment somewhere obvious about how often this can (or rather, can't) be changed?

alexwlchan:
> Maybe add a comment somewhere obvious about how often this can (or rather, can't) be changed?

There's a comment a bit further up the file, is that good enough?

jamieparkinson:
Ah yes, so there is 👍

PR #1813, opened 2021-08-12
Tei on pipeline

alicefuzier:


pipeline/terraform/main.tf
```diff
@@ -91,6 +91,99 @@ module "catalogue_pipeline_2021-08-09" {
...
+    # we want to keep tei out of the prod pipeline for now,
+    # while me make sure that the data is good enough to present
+    # and while we extend the transformer to extract more of it.
+    #
+    # Note: we have to supply a real topic ARN here, not just 'null',
+    # or the module gets confused.  For now we just create a dummy topic
+    # that will never receive messages.
+    tei = {
+      topics = [
+        # local.tei_adapter_topic_arn,
```
jamesgorrie, line 144: Do we want this back on?

pipeline/terraform/stack/elastic_pipeline.tf
```diff
@@ -180,8 +180,8 @@ locals {
     "relation_embedder",
     "router",
     "inferrer",
-    "ingestor_works",
-    "ingestor_images",
+    "work_ingestor",
```
jamesgorrie, line 183: Why'd we change this? Would it make more sense to pluralise it?

pipeline/terraform/stack/tei_off_branch.tf
```diff
@@ -0,0 +1,55 @@
+module "tei_off_branch" {
+  source = "./branch"
```
jamesgorrie, line 2: Is it worth naming this branch? Just in case we want to keep it later, it'll will be a bit harder to rename later (althought we've already applied this it seems).

pipeline/terraform/stack/branch/locals.tf
```diff
@@ -0,0 +1,10 @@
+locals {
+  tei                         = var.toggle_tei_on ? "tei-on" : "tei-off"
```
jamesgorrie, line 2:
```suggestion
  tei_suffix                         = var.toggle_tei_on ? "-tei-on" : ""
```

Then you can drop all the conditionals below.

jamesgorrie (approved):

pipeline/terraform/main.tf
alicefuzier: ah yes, thanks

pipeline/terraform/stack/elastic_pipeline.tf
alicefuzier: because it was not in sync with what is created in the create_elasticsearch_users script. We could pluralize, but we have to do it there too. I just opted for the simplest change which was this

pipeline/terraform/stack/tei_off_branch.tf
alicefuzier: yeah, I don't love the name, but I'm not sure what a good name would be. It os true that is already applied but it's not a particularly massive effort to rename this

pipeline/terraform/stack/branch/locals.tf
alicefuzier: hmm it doesn't quite work cause that tei flag is used to name the services with a "tei-on" or "tei-off" suffix, but for the indices we don't want a tei-off suffix, we just want no suffix. So I think the conditionals will have to stay. I think tei_suffix is a better name though

pipeline/terraform/stack/tei_off_branch.tf
jamesgorrie: No need, we can do it if we find it useful when spinning up new pipelines.

jamesgorrie (approved):

PR #1818, opened 2021-08-16
Support internal works in merger and router

alicefuzier:
Part of #1806

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
@@ -154,9 +154,9 @@ sealed trait WorkState {
 object WorkState {
 
   case class Source(
-                     sourceIdentifier: SourceIdentifier,
-                     sourceModifiedTime: Instant,
-                     internalWorks: List[Work[Source]]
+    sourceIdentifier: SourceIdentifier,
+    sourceModifiedTime: Instant,
+    internalWorks: List[Work[Source]]
   ) extends WorkState {
 
     type WorkDataState = DataState.Unidentified
@@ -169,10 +169,10 @@ object WorkState {
   }
 
   case class Identified(
-                         sourceIdentifier: SourceIdentifier,
-                         canonicalId: CanonicalId,
-                         sourceModifiedTime: Instant,
-                         internalWorks: List[Work.Visible[Identified]]
+    sourceIdentifier: SourceIdentifier,
+    canonicalId: CanonicalId,
+    sourceModifiedTime: Instant,
+    internalWorks: List[Work.Visible[Identified]]
   ) extends WorkState {
 
     type WorkDataState = DataState.Identified
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
@@ -154,8 +154,9 @@ sealed trait WorkState {
 object WorkState {
 
   case class Source(
-    sourceIdentifier: SourceIdentifier,
-    sourceModifiedTime: Instant
+                     sourceIdentifier: SourceIdentifier,
+                     sourceModifiedTime: Instant,
+                     internalWorks: List[Work[Source]]
   ) extends WorkState {
 
     type WorkDataState = DataState.Unidentified
@@ -168,9 +169,10 @@ object WorkState {
   }
 
   case class Identified(
-    sourceIdentifier: SourceIdentifier,
-    canonicalId: CanonicalId,
-    sourceModifiedTime: Instant,
+                         sourceIdentifier: SourceIdentifier,
+                         canonicalId: CanonicalId,
+                         sourceModifiedTime: Instant,
+                         internalWorks: List[Work[Identified]]
   ) extends WorkState {
 
     type WorkDataState = DataState.Identified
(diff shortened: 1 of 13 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
```diff
@@ -155,7 +155,8 @@ object WorkState {
 
   case class Source(
     sourceIdentifier: SourceIdentifier,
-    sourceModifiedTime: Instant
+    sourceModifiedTime: Instant,
+    internalWorks: List[Work.Visible[Source]] = Nil
```
alexwlchan, line 159: As a minor note: I think this probably means we can deploy directly into the current prod pipeline, because existing Source works will still deserialise correctly?

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
```diff
@@ -235,6 +242,21 @@ object PlatformMerger extends Merger {
           }
         )
 
+  private def addItemsToInternalWorks(items: List[Item[IdState.Minted]],
+                                      work: Work.Visible[Identified]) = {
+    // Inteernal works are in TEI works. If they are merged with Sierra, we want the Sierra
+    // items to be added to TEI internal works so that the user can request the item
+    // containing that work without having to find the wrapping work.
```
alexwlchan, line 249: 👍 Useful comment.

pipeline/merger/src/main/scala/weco/pipeline/merger/services/MergerManager.scala
```diff
@@ -19,7 +19,7 @@ trait MergerManager {
     val modifiedWorks = preMergeModify(works)
     if (works.size == maybeWorks.size) {
       val result = mergerRules.merge(modifiedWorks)
-      assert(result.resultWorks.size == modifiedWorks.size)
+      assert(result.resultWorks.size >= modifiedWorks.size)
```
alexwlchan, line 22: Possibly worth a comment explaining why this is the case? The "greater than" case isn't obvious unless you know about internal works.

pipeline/relation_embedder/router/src/main/scala/weco/pipeline/router/RouterWorkerService.scala
```diff
@@ -33,16 +33,24 @@ class RouterWorkerService[MsgDestination](
...
+    work: Work[Merged]): Try[List[Work[Denormalised]]] = {
+    (work.data.collectionPath, work.state.relations) match {
+      // For TEI works relations are already populated based on
+      // inner works extracted by the TEI transformer. We don't need
+      // to repopulate them in the relation embedder.
+      // We don't expect TEI works to have a collectionPath field populated.
+      case (None, relations) =>
+        Success(List(work.transition[Denormalised]((relations, Set.empty))))
+      case (Some(CollectionPath(path, _)), relations)
+          if relations == Relations.none =>
```
alexwlchan, line 48: Does this work? I'm guessing there's a reason you didn't do it this way, but just checking.

```suggestion
      case (Some(CollectionPath(path, _)), Relations.none) =>
```

pipeline/relation_embedder/router/src/test/scala/weco/pipeline/router/RouterWorkerServiceTest.scala
```diff
@@ -123,7 +146,7 @@ class RouterWorkerServiceTest
                            retriever: Retriever[Work[Merged]])(
     testWith: TestWith[(QueuePair, MemoryMessageSender, MemoryMessageSender),
                        R]): R =
-    withLocalSqsQueuePair() {
+    withLocalSqsQueuePair(1 second) {
```
alexwlchan, line 149:
```suggestion
    withLocalSqsQueuePair(visibilityTimeout = 1 second) {
```

pipeline/transformer/transformer_common/src/test/scala/weco/pipeline/transformer/example/ExampleTransformer.scala
```diff
@@ -28,7 +28,7 @@ object ExampleTransformer extends Transformer[ExampleData] with WorkGenerators {
       case ValidExampleData(id, title) =>
         Right(
           Work.Visible[Source](
-            state = Source(id, modifiedTime),
+            state = Source(id, modifiedTime, List()),
```
alexwlchan, line 31:
```suggestion
            state = Source(id, modifiedTime),
```

alexwlchan (approved):
Looks good! A few small suggestions but nothing major.

pipeline/relation_embedder/router/src/main/scala/weco/pipeline/router/RouterWorkerService.scala
alicefuzier: This doesn't compile. `Relations.none` is not a type so you can't pattern match. It's not particularly nice, I agree. I tried making `Relations` an `Option` but I suspected it wouldn't work otherwise Nick would have done it. It ended up being very convoluted and I backtracked

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
alicefuzier: I think so? I can't think of a reason why they would break

alexwlchan (approved):

PR #1825, opened 2021-08-18
Add comments and tests to CollectionPathSorter

alexwlchan:
While trying to read the relation embedder code for wellcomecollection/platform#5270, I wasn't entirely clear what was going on in CollectionPathSorter – and it's only tested through its side effects.

This is less of a radical refactor than #1824; it's mostly just adding some comments and tests, tightening up the variable names, and rearranging the order of some functions to make similarities/distinctions more obvious.

### DIFF
--- a/pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/ArchiveRelationsCache.scala
+++ b/pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/ArchiveRelationsCache.scala
@@ -3,7 +3,7 @@ package weco.pipeline.relation_embedder
 import grizzled.slf4j.Logging
 import weco.catalogue.internal_model.work.WorkState.Merged
 import weco.catalogue.internal_model.work._
-import weco.pipeline.relation_embedder.models.{CollectionPathSorter, PathCollection}
+import weco.pipeline.relation_embedder.models.PathCollection
 
 class ArchiveRelationsCache(works: Map[String, RelationWork]) extends Logging {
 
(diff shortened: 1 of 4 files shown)


pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/CollectionPathSorter.scala
```diff
@@ -1,80 +0,0 @@
...
-    }
-
-  implicit val pathTokenOrdering: Ordering[PathToken] =
-    new Ordering[PathToken] {
-      @tailrec
-      override def compare(a: PathToken, b: PathToken): Int =
-        (a, b) match {
-          case (Nil, Nil) => 0
-          case (Nil, _)   => 1
-          case (_, Nil)   => -1
```
alexwlchan, line 60: These two lines are the cause of the behaviour Victoria is seeing – they used to cause longer parts to sort higher. Notice the difference between this and tokenizedPathOrdering:

```scala
          case (Nil, _)   => -1     // pathTokenOrdering
          case (_, Nil)   => 1

          case (Nil, _)   => 1      // tokenizedPathOrdering
          case (_, Nil)   => -1
```

e.g. consider sorting `A/B/1` and `A/B/1b`.

The old code would sort `1b` above `1` because it was longer. When I created the common `shorterPathsWinOrdering` behaviour, I flipped this behaviour – so we now get the desired behaviour "for free".

(Also this is why you need tests! This is a subtle behaviour change that I wouldn't have noticed if I hadn't been looking for it, and wasn't caught by the tests.)

kenoir (approved):

PR #1828, opened 2021-08-19
Allow the Calm indexer to turn off when it's not in use

alexwlchan:
Spotted while debugging the Calm adapter auto-deployment.

I thought the default min capacity of the `worker` module was 0, but it was actually 1, so it would never scale all the way to zero. This patch allows the Calm indexer to scale down to 0, and removes the default value so we can't make the same mistake again – I've checked all the uses of the module, and everywhere else the min capacity is already explicitly set.

💸

### DIFF
--- a/calm_adapter/terraform/provider.tf
+++ b/calm_adapter/terraform/provider.tf
@@ -5,6 +5,10 @@ provider "aws" {
     role_arn = "arn:aws:iam::760097843905:role/platform-developer"
   }
 
+  ignore_tags {
+    keys = ["deployment:label"]
+  }
+
   default_tags {
     tags = {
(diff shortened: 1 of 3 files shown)


jamesgorrie (approved):

calm_adapter/terraform/provider.tf
```diff
@@ -5,6 +5,10 @@ provider "aws" {
     role_arn = "arn:aws:iam::760097843905:role/platform-developer"
   }
 
+  ignore_tags {
```
jamesgorrie, line 8: Worth explaining what this is doing as it's related to weco-deploy?

PR #1829, opened 2021-08-19
Automatically deploy the catalogue pipeline if there are no internal model changes

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5271

How it works:

* At the end of a build on main, Buildkite triggers a "catalogue pipeline deploy" pipeline: https://buildkite.com/wellcomecollection/catalogue-pipeline-deploy-prod
* The script that pipeline runs gathers a bunch of info:

    - What's the live pipeline/API?
    - What versions of internal model does the associated index know about?
    - What changed between the current commit and the latest internal model in the index?

* If nothing in `internal_model/src/main` has changed in the interim, it kicks off an automatic deploy into the prod pipeline. Otherwise, it fails the build and waits for a human to deploy locally.

As a future enhancement, I wouldn't mind making that a button you can click in Buildkite, but this is enough for a first pass. I've tested it in the branch and seen it get as far as running weco-deploy, but because I've not deployed any images from this branch, it's hard to know it's working – I'll test it properly when I merge https://github.com/wellcomecollection/catalogue-pipeline/pull/1825

I've also pulled all the deployment-y bits out into this new pipeline, so they can succeed/fail together.


jamesgorrie (commented):
Is there anyway to have the python script be a condition to a deploy step in the buildkite config to keep all deployment stuff there? This would then have the same visibility as the rest of the deployment (fails etc).

I suppose you could have it `exit(1)` as a step, and then have the prod deployment reliant on that not failing?

alexwlchan:
> Is there anyway to have the python script be a condition to a deploy step in the buildkite config to keep all deployment stuff there?

I'm not sure I understand what you're suggesting? Do you mean using the Python script to gate whether we run the weco-deploy command, but having that command declared in pipeline.yml? 

Where this is a little tricky is that unlike the other services, the pipeline doesn't have a static environment ID – we have to look up the pipeline date, then pass that to the weco-deployment command. The pipeline doesn't deploy like the other services. I'd need to come up a way to pass state from the script to the weco-deploy command. That's probably possible, but it's a bit of work and I want to check that's what you're suggesting.

> This would then have the same visibility as the rest of the deployment (fails etc).

The parenthetical is a bit confusing – we still have the same failure behaviours as before. If the weco-deploy command fails, the Python script will exit(1) and bubble up a failure to weco-deploy

jamesgorrie:
That's what I meant. I see the issue. Why are we deploying to `prod` in previous steps if `prod` doesn't exists?

alexwlchan:
> Why are we deploying to prod in previous steps if prod doesn't exists?

We're deploying to prod for the adapter services (CALM adapter, Sierra adapter, etc.), not the pipeline services (transformers, ID minter, ingestor, etc.).

jamesgorrie (approved):
Thanks for clearning that up!

alexwlchan:
Right, let's get this into main and see if it works!

PR #1833, opened 2021-08-21
Fix a flaky test in the batcher

alexwlchan:
Sometimes in CI we see one of the batcher tests fail because the service sends two batches, instead of one.  You can reproduce this locally by cranking down the flushInterval until the batcher flushes the first message before the second is received.

Conversely, it seems like cranking the interval up will probably stop this test being flaky in CI.


alexwlchan:
I think I merged this in a different PR.

PR #1835, opened 2021-08-23
Add an option to display the Miro image in the visual graph

alexwlchan:
This will make it easier to reason about what's going on in the graph, e.g. if a Miro image has been incorrectly linked to a Sierra record. An example:

Screenshot 2021-08-23 at 12 21 31

I've done this by hand a few times, but this is what computers are good for!

This is enabled by default, but there's a flag to disable it if you just want the pure-text graph.


jamieparkinson (approved):
nice ✨

PR #1837, opened 2021-08-23
Add a title to works from the METS transformer

alexwlchan:
Follows https://github.com/wellcomecollection/catalogue-pipeline/pull/1836, for https://github.com/wellcomecollection/platform/issues/5242

We assume the presence of a title in a bunch of places, including the design of the works pages – and indeed, the title will always be set on the WorkData for a visible Work – but this isn't enforced by the type system. We have `WorkData { title: Option[String] }` is because we don't set a title in the METS transformer. The reason we don't set a title in the METS transformer is that it'll be blatted in the merger – but there is a title in the METS files (and I checked every METS file to verify – yes, it really is there).

This PR updates the transformer to set a title on every METS work.

It **doesn't** update the internal model to assume the use of title – if we do that, we can't deploy it into the current pipeline, because the existing METS works without a title will break. My plan is:

* Get this patch reviewed, merged and deployed into the current pipeline
* Reindex just the METS works, so every METS work in the pipeline has a title
* Start enforcing a required title in internal model, and deploy that into the current pipeline

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/weco/pipeline/transformer/mets/transformer/MetsData.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/weco/pipeline/transformer/mets/transformer/MetsData.scala
@@ -50,6 +50,7 @@ case class DeletedMetsData(recordIdentifier: String) extends NewMetsData {
 
 case class InvisibleMetsData(
   recordIdentifier: String,
+  title: String,
   accessConditionDz: Option[String] = None,
   accessConditionStatus: Option[String] = None,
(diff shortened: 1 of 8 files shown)


jamesgorrie (approved):
Maybe an extra verification step of "all works at all stages have titles set in the elatic index"

alexwlchan:
> Maybe an extra verification step of "all works at all stages have titles set in the elatic index"

Yeah, good suggestion.

alexwlchan:
Although TBH this is looking like it's going to be a sufficiently invasive change that I might just reindex instead.

PR #1840, opened 2021-08-24
Make the invisibilityReason a required value, rather than a list

alexwlchan:
Follows https://github.com/wellcomecollection/catalogue-pipeline/pull/1839, spotted while looking at https://github.com/wellcomecollection/platform/issues/5242

We always set exactly one invisibilityReason, let's reflect that in the model.

This is a backwards-incompatible change to internal model, so I'll hold off merging this until I'm ready to merge and deploy all the changes I want to make.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
@@ -174,7 +174,7 @@ object WorksIndexConfig extends IndexConfigFields {
         search,
         keywordField("type"),
         data.withDynamic("false"),
-        objectField("invisibilityReasons")
+        objectField("invisibilityReason")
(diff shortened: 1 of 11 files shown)


jamesgorrie (commented):
The main reason for this as the intended use for this was to report back to collections staff why something might not be turning up e.g. an interface or report with `Invisible` works and why they're invisible.

The reason this was a list as a work might become invisible for multiple reasons, and a person would want to address all of them rather than address 1, and then have another pop up.

e.g.
* We're missing a field for the transformer
* There's an invalid field in the transformer

As a person correcting this data, you'd want both reasons at once, rather than sequentially.

alexwlchan:
> As a person correcting this data, you'd want both reasons at once, rather than sequentially.

\*nod*

In which case I'll roll this back and put a comment to that effect, but it's completely non-obvious from the code (and a bit of a faff to do with our current transformer architecture, but I see the possible rationale).

PR #1844, opened 2021-08-24
Fix a bug in merging Sierra works when the 776 $w is on the e-bib

alexwlchan:
## The problem

Some works are missing their identifiers. e.g. https://api.wellcomecollection.org/catalogue/v2/works/ym7pjfz2?include=identifiers:

```json
  "identifiers": [
    {
      "identifierType": {
        "id": "sierra-system-number",
        "label": "Sierra system number",
        "type": "IdentifierType"
      },
      "value": "b13456143",
      "type": "Identifier"
    }
  ],
```

This is a combination of a Sierra bib, e-bib and METS work – but it only has a single identifier. What's up with that?

## The cause

We use the 776 $w linking field to merge physical/digitised Sierra pairs, and we expect to see a 776 $w on the physical bib – but sometimes, as is the case here, that linking record is actually on the e-bib only. In this case, we'd drop all the identifiers on the work (and maybe some other stuff)

See
https://search.wellcomelibrary.org/iii/encore/record/C__Rb1345614?lang=eng
https://search.wellcomelibrary.org/iii/encore/record/C__Rb2044212?lang=eng

## The solution

Update the code that looks for the matching work in `Sources`, so if there are no merge candidates on the physical work, we look through the electronic works instead for the first match.

### DIFF
--- a/pipeline/merger/src/main/scala/weco/pipeline/merger/models/Sources.scala
+++ b/pipeline/merger/src/main/scala/weco/pipeline/merger/models/Sources.scala
@@ -10,26 +10,48 @@ object Sources {
   def findFirstLinkedDigitisedSierraWorkFor(
     target: Work.Visible[Identified],
     sources: Seq[Work[Identified]]): Option[Work[Identified]] =
-    if (physicalSierra(target)
+    target match {
+      // Audiovisual works are catalogued as multiple bib records, one for the physical
+      // format and another for the digitised version, where available.
+      //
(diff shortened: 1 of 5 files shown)


alexwlchan:
Once this is merged and deployed, I'll look for Sierra works with a single identifier and reindex just those through the merger. Yay continuous deployment!

jamieparkinson (approved):
The e-bib boulder rolls once again, Sisyphus

PR #1846, opened 2021-08-27
Don't use the bib data to create the item status

alexwlchan:
Currently we get the item status in two ways:

* In the Catalogue API, we look at both the bib data and the item data.
* In the Items API, we just look at the item data. We skip fetching the bib data to avoid an extra request to Sierra.

This is because I had the “clever idea” of looking at them both, so we could spot the inconsistencies. Smart!

Unfortunately, this means we have two different code paths in play, which behave inconsistently, and the code wasn’t written to handle getting no bib data – so when you use the items API, a bunch of items get blatted with the _"I don't know what to do"_ generic message.

With agreement from Victoria, this removes all the bib-related code for getting the item status. We just use the item data, which means both APIs are using exactly the same code – so we shouldn't get any more inconsistencies.

### DIFF
--- a/pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
+++ b/pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
@@ -32,8 +32,6 @@ case class TransformerError[SourceData, Key](t: Throwable,
   * - Runs it through a transformer and transforms the `SourceData` to `Work[Source]`
   * - Emits the message via `MessageSender` to SNS
   */
-class NewTransformerWorker
-
 trait TransformerWorker[Payload <: SourcePayload, SourceData, SenderDest]
     extends Logging {
   type Result[T] = Either[TransformerWorkerError, T]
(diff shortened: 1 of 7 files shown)


pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
```diff
@@ -32,6 +32,8 @@ case class TransformerError[SourceData, Key](t: Throwable,
   * - Runs it through a transformer and transforms the `SourceData` to `Work[Source]`
   * - Emits the message via `MessageSender` to SNS
   */
+class NewTransformerWorker
```
kenoir, line 35: 😕

kenoir (approved):
👏 Presumably this needs a corresponding PR in catalogue-api to copy these changes there?

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
alexwlchan: Oops.

alexwlchan:
> Presumably this needs a corresponding PR in catalogue-api to copy these changes there? 

Yup, I'll do it once this is merged.

PR #1848, opened 2021-09-06
Speed up the WorkIndexConfig tests, and internal model tidying

alexwlchan:
The WorkIndexConfig tests work by creating dozens of different instances of Work and trying to store them in the index.  The way we're doing this is very inefficient -- we were creating a new ES index per Work, and only looking to see if the index contained *a* Work.

This new approach shares the index, and indexes Works under a known ID, so we only have to create one index per test.

Where previously an individual test took ~13s locally, now the entire suite passes in the same time. Much faster!

---

I'm also stripping out the ElasticsearchFixtures helpers `assertObjectIndexed`, `indexObject` and `indexObjectCompressed`, which index a single document without an ID – if you look at the implementations, their behaviour isn't what you'd expect. At some point I'd like to remove them entirely, but we need to bin all uses of them first.

This is some preliminary refactoring I spotted while working on https://github.com/wellcomecollection/platform/issues/5242

### DIFF
--- a/common/internal_model/src/test/scala/weco/catalogue/internal_model/generators/ImageGenerators.scala
+++ b/common/internal_model/src/test/scala/weco/catalogue/internal_model/generators/ImageGenerators.scala
@@ -1,6 +1,5 @@
 package weco.catalogue.internal_model.generators
 
-import weco.catalogue.internal_model.work.generators._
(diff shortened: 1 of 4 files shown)


kenoir (approved):
👏

PR #1858, opened 2021-09-07
Disable the image inferrer in the "tei-on" pipeline

alexwlchan:
Something weird is happening here (see comment in the Terraform) that I don't care to debug properly right now.

### DIFF
--- a/pipeline/terraform/stack/branch/service_image_inferrer.tf
+++ b/pipeline/terraform/stack/branch/service_image_inferrer.tf
@@ -170,6 +170,9 @@ module "image_inferrer" {
   #
   # Disabling the image inferrer in the "tei-on" pipeline is a hack
   # to fix the image inferrer in the publicly visible pipeline.
+  # The "tei-off" inferrer gets exclusive use of the EC2 instance and is
+  # able to start correctly.
+  #
   # At some point it'd be nice to come back and sort this out properly,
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
Good catch a quick-fix, but we should definitely fix this as I am assuming this means images are coming through on the `tei-on` pipeline?

alexwlchan:
> I am assuming this means images are coming through on the tei-on pipeline

Yes.

AIUI, the "tei-on" and "tei-off" pipelines both get the same updates from the source records, but they split post-merger. Image updates will be coming through the "tei-on" pipeline, they're just being ignored. So you won't be able to search for images in "tei-on" – @alicefuzier will know if the TEI work is going to affect images and if that's an issue we strongly care about.

This is mostly about codifying the fix I made manually to unbreak the pipeline, but if people can work out how to fix it properly I'm all ears.

jamesgorrie:
> @alicefuzier will know if the TEI work is going to affect images and if that's an issue we strongly care about.

We have spoken about this previously, and it was definitely something we'd want to be testing, so maybe we'll have to prioritise a fix to make sure we can continue TEI work.

> This is mostly about codifying the fix I made manually to unbreak the pipeline

This is more important so we should definitely get this in soonish

alexwlchan:
Okay, I've merged this for now, and created an issue to track it: https://github.com/wellcomecollection/platform/issues/5292

PR #1861, opened 2021-09-09
Bump the flushInterval to make this test less flaky in CI

alexwlchan:
Sometimes it sends two batches instead of one; I suspect because it can't quite process both messages in time.  Maybe bumping the interval will help?


jamieparkinson (approved):
🚽

PR #1863, opened 2021-09-09
Various pipeline fixes for CCR reliability

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5295, closes https://github.com/wellcomecollection/catalogue-pipeline/pull/1862

- Image inferrers now get dedicated capacity providers for TEI on/off (h/t @alicefuzier)
- We now get the version of Elasticsearch that the API is using, so we never step ahead of it. This means CCR can't get its versioning in a twist.
- We now run two, smaller Elasticsearch nodes in the pipeline cluster when we're not reindexing – for high availability and offset costs
- We no longer depend on the Elasticsearch cluster for all the pipeline resources. This is a "clever idea" I had when setting this up that means services won't be created before the relevant secrets are ready, but:
     1. It creates unreadable Terraform plans.
     2. If those secrets aren't ready, the services will just restart until they are. Not a big issue.

### DIFF
--- a/pipeline/terraform/stack/cluster.tf
+++ b/pipeline/terraform/stack/cluster.tf
@@ -1,12 +1,33 @@
 resource "aws_ecs_cluster" "cluster" {
   name               = local.namespace_hyphen
-  capacity_providers = [module.inference_capacity_provider.name]
+  capacity_providers = [module.inference_capacity_provider_tei_off.name,module.inference_capacity_provider_tei_on.name]
 }
 
-module "inference_capacity_provider" {
+module "inference_capacity_provider_tei_on" {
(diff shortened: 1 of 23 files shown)


pipeline/terraform/stack/cluster.tf
```diff
@@ -1,12 +1,33 @@
 resource "aws_ecs_cluster" "cluster" {
   name               = local.namespace_hyphen
-  capacity_providers = [module.inference_capacity_provider.name]
+  capacity_providers = [module.inference_capacity_provider_tei_off.name, module.inference_capacity_provider_tei_on.name]
```
kenoir, line 3: Nice.

kenoir (approved):

pipeline/terraform/stack/cluster.tf
alexwlchan: All @alicefuzier’s work :D

PR #1871, opened 2021-09-13
Add more implicit encoders/decoders to speed up the compilation of internal model

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5298 I followed the instructions in the blog post linked at the top of `Implicits.scala`, identified the slowest parts, and added implicits for them. Unsurprisingly, implicit matching is still ~99% of our compilation time. It has a roughly 50% improvement:

Before:

```
Executed in  104.56 secs   fish           external
   usr time   72.90 millis   90.00 micros   72.81 millis
   sys time   87.06 millis  426.00 micros   86.63 millis
```

After:

```
Executed in   47.56 secs   fish           external
   usr time   63.90 millis   87.00 micros   63.81 millis
   sys time   78.13 millis  353.00 micros   77.77 millis
```

I'm at a point of diminishing returns – the last few changes made barely any dent in the compilation time.

I've uploaded the current flame graph here, so if anyone can see where I should make the next intervention, I can see if they make a difference: https://wellcome.slack.com/archives/C3TQSF63C/p1631533771051400

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/Implicits.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/Implicits.scala
@@ -4,6 +4,7 @@ import io.circe.generic.extras.semiauto._
 import io.circe._
 import weco.catalogue.internal_model.identifiers._
 import weco.catalogue.internal_model.image._
+import weco.catalogue.internal_model.languages.Language
 import weco.catalogue.internal_model.locations.{
   AccessCondition,
   DigitalLocation,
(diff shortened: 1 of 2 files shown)


alexwlchan:
New record:

```
Executed in   41.39 secs   fish           external
   usr time   58.91 millis   91.00 micros   58.82 millis
   sys time   70.29 millis  363.00 micros   69.93 millis
```

alexwlchan:
And a little bit more:

```
Executed in   38.56 secs   fish           external
   usr time   53.41 millis  101.00 micros   53.31 millis
   sys time   66.16 millis  421.00 micros   65.73 millis
```

jamieparkinson (dismissed):
v nice

alicefuzier (approved):

jamesgorrie:
@alexwlchan do you think it would be worth removing any implicit codecs to ensure we always add them?

alexwlchan:
> do you think it would be worth removing any implicit codecs to ensure we always add them?

Do you mean removing the autoderived decoders/encoders? Possibly, but not right now – it seems like a moderate amount of work to remove them just for internal model and not break every other bit of code that relies on JSON parsing.

The point of this change was to get internal model compiling in a sensible time, because that was blocking the TEI work (it was so slow that neither of us could run builds in IntelliJ). Now the TEI work is unblocked, I don't want to spend more time on this.

If we really want to fix this properly, I think simplifying the type hierarchy is a better fix than piling more stuff into the JSON handling, but that's a bigger change that needs more discussion.

PR #1878, opened 2021-09-14
Add an InternalWork type to avoid a circular model + internal model slowness

alexwlchan:
An attempt to make progress on https://github.com/wellcomecollection/platform/issues/5298


alicefuzier (dismissed):

.buildkite/scripts/publish.py
```diff
@@ -47,5 +43,7 @@ def publish(project_name):
         print(f"Nothing in this patch affects {projects}, so stopping.")
         sys.exit(0)
 
-    for p in projects:
-        publish(p)
+    for project_name in projects:
+        run_build_script(
```
alicefuzier, line 47: Does this propagate the environment variables to the task in docker? Causeto publish it needs the BUILDKITE_BUILD_NUMBER otherwise it will put a build number 0, which is only supposed to happen locally. The previous YAML way had `propagate-environment: true` for this reason

alexwlchan: Yes: https://github.com/wellcomecollection/catalogue-pipeline/blob/main/builds/run_sbt_task_in_docker.sh#L32

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
```diff
@@ -151,12 +151,26 @@ sealed trait WorkState {
   def id: String
 }
 
+object InternalWork {
+  // todo: explain why we don't just use Work[Source] because BOOM RECURSION
```
alicefuzier, line 155: Is this still work in progress? Do we want to add a comment now here?

alicefuzier (approved):

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
alexwlchan: How about the comment now?

alexwlchan:
Note that internal model is still noticeably slower than it was – about 11 minutes in CI (https://buildkite.com/wellcomecollection/catalogue-pipeline/builds/5496#ccdf59fd-6505-40c3-a3d5-1c12de9c8c0c) vs 7 minutes before the changes to add internal model (https://buildkite.com/wellcomecollection/catalogue-pipeline/builds/5437)

alicefuzier (approved):

alexwlchan:
This also has a fix for https://github.com/wellcomecollection/platform/issues/5303

At a minimum, I'm hoping it unblocks the publishing of internal model in CI. We still need to get the internal model build back under control, but this takes us in the right direction.

PR #1883, opened 2021-09-15
What if we don't use the automatically derived encoders/decoders?

alexwlchan:
Experimenting with https://github.com/wellcomecollection/platform/issues/5298

We spend a lot of time coming up with implicit encoders/decoders. What if we cut Shapeless out of the equation (somewhat) and tell Circe exactly how do decode/encode some of the more expensive classes?

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/Implicits.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/Implicits.scala
@@ -32,21 +32,21 @@ object Implicits {
   // dependencies (thus preventing duplicate work)
 
   implicit val _decAccessCondition: Decoder[AccessCondition] =
-    deriveConfiguredDecoder
+    AccessCondition.decoder
   implicit val _decNote: Decoder[Note] = deriveConfiguredDecoder
 
   implicit val _decIdentifierType: Decoder[IdentifierType] =
     IdentifierType.identifierTypeDecoder
   implicit val _decSourceIdentifier: Decoder[SourceIdentifier] =
-    deriveConfiguredDecoder
+    SourceIdentifier.decoder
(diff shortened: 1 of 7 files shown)


alicefuzier:
FWIW this is probably an idea that has a great impact on compilation time. But I would be very wary of doing this in general. The time that it will take us to write the decoders, debug which ones we forgot when we add new things and so on would much greater than the time spent waiting for compilation IMO. Not to mention all the inconsistencies/errors  that we risk introducing which also will take a long time to debug. IMO, we should stick to automagically generated JSON. 

Whether Circe & Shapeless are the right tools for this, it's another question and I'm no longer sure of the answer. We might switch to scala libraries that use reflection as opposed to shapeless ifthis continues to be an issue

But, I think the main issue here is the complexity of the internal model that is becoming too entangled, so I'm not sure that I would feel too keen to change JSON library to work around the internal model complexity

alexwlchan:
> FWIW this is probably an idea that has a great impact on compilation time. But I would be very wary of doing this in general. The time that it will take us to write the decoders, debug which ones we forgot when we add new things and so on would much greater than the time spent waiting for compilation IMO. Not to mention all the inconsistencies/errors that we risk introducing which also will take a long time to debug. IMO, we should stick to automagically generated JSON.

Yeah, I think this needs a broader discussion among the team – I wanted to get an idea of whether this was even a viable path. (The compilation time is down to 10 minutes, which is good but still a far cry from where we were a few days ago.)

I do think we're somewhat protected against screwing this up – in particular, the WorksIndexConfigTest that generates random instances of case classes should notice if we're sending objects to Elasticsearch that come back differently.

> But, I think the main issue here is the complexity of the internal model that is becoming too entangled, so I'm not sure that I would feel too keen to change JSON library to work around the internal model complexity

Yeah, that's just a lot harder. 😅

PR #1884, opened 2021-09-15
Don't use the version in the matcher/merger; use the modifiedTime

alexwlchan:
Follows #1882, for wellcomecollection/platform#5302

I've switched to using modifiedTime for works in the matcher, which is more consistent with what we do elsewhere. This paves the way to removing version entirely in a subsequent patch.

I've also tidied up some of the classes in the matcher/merger, in particular:

* WorkLinks is now WorkStub, to reflect the fact that it's really a slimmed-down Work because we don't want to retrieve the whole thing from Elasticsearch
* The MatcherResult changes from:

    ```scala
    MatcherResult {
      works: Set[MatchedIdentifiers] = Set[Set[WorkIdentifier]]
      createdTime: Instant
    }
    ```

    to

    ```scala
    MatcherResult {
      works: Set[MatchedIdentifiers] = Set[Set[WorkStub]]
      createdTime: Instant
    }
    ```

    so we're using the same "stub" to identify works throughout. We also stop passing the merger information about works it cannot possibly merge (previously those were instances of WorkIdentifier with `version = None`), where we know that A links to B, but we haven't seen B yet. It lets us do away with a class and make the namespace less crowded, and means we're sending less data between the two apps.


jamieparkinson (commented):
I know we used `modifiedTime` once before and changed back but can't remember any context; would definitely want @alicefuzier's thoughts on this.

alexwlchan:
Yeah, tbh I should probably split this into "WorkStub" and "version" changes.

PR #1889, opened 2021-09-16
Explain why we have the "version" field on the Work model

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5302

I had a go at removing this, but it is useful for very particular reasons. Ideally it'd be renamed "sourceVersion", but that's a bigger problem than I care to undertake right now. I can at least document why it's here to save the next person scratching their head about it.


alicefuzier (approved):
Thank you for spending time investigating!

PR #1899, opened 2021-09-20
Add the initial RelationPath model

alexwlchan:


common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/RelationPath.scala
```diff
@@ -0,0 +1,31 @@
...
+  *
+  * This allows us to identify that, for example, PP/CRI/B/1 is a direct child
+  * of PP/CRI/B and a descendent of PP/CRI.
+  *
+  */
+class RelationPath(val underlying: String)
+  extends TypedString[ReferenceNumber] {
+
+  def join(other: RelationPath): RelationPath =
+    RelationPath(s"$underlying/${other.underlying}")
```
alexwlchan, line 25: Note: this needs a test

alexwlchan:
Looking at this again, I wonder if this is wrong, and if we should in fact revert https://github.com/wellcomecollection/catalogue-pipeline/pull/1897

Because every collectionPath.label is a referenceNumber, but not vice versa. By smushing the two together, we'll expose the i-numbers to the boosting in search.relations. Which is probably fine, but I wouldn't swear to it.

PR #1900, opened 2021-09-21
What if we do relations on TEI works by just renaming the CollectionPath field?

alexwlchan:


common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
```diff
@@ -37,10 +37,22 @@ object WorksIndexConfig extends IndexConfigFields {
   val merged = WorksIndexConfig(
     Seq(
       keywordField("type"),
+      //
+      // We copy the relationPath field to enable two types of query in the
+      // relation embedder:
```
jamesgorrie, line 42: This is also in the API.

alexwlchan: This is a comment on the config for the merged-works index – there's a separate comment for the API further down.

PR #1903, opened 2021-09-21
Remove logic to check existence of relations from router

alicefuzier:


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
```diff
@@ -233,15 +233,15 @@ object WorkState {
...
-    relations: Relations = Relations.none
+    availabilities: Set[Availability] = Set.empty
   ) extends WorkState {
 
     type WorkDataState = DataState.Identified
     type TransitionArgs = Instant
 
     def id: String = canonicalId.toString
 
+    override val relations: Relations = Relations.none
```
alexwlchan, line 244: Do we need Relations on the core Work model at this point?

pipeline/relation_embedder/router/src/main/scala/weco/pipeline/router/RouterWorkerService.scala
```diff
@@ -36,19 +36,19 @@ class RouterWorkerService[MsgDestination](
...
   private def processMessage(
-    work: Work[Merged]): Try[List[Work[Denormalised]]] =
-    (work.data.collectionPath, work.state.relations) match {
+    work: Work[Merged]): Try[List[Work[Denormalised]]] = {
+    work.data.collectionPath match {
       // For TEI works relations are already populated based on
       // inner works extracted by the TEI transformer. We don't need
       // to repopulate them in the relation embedder.
       // We don't expect TEI works to have a collectionPath field populated.
-      case (None, relations) =>
```
alexwlchan, line 45: Can we remove the comment above this line? It's no longer accurate.

alexwlchan (approved):
Couple of suggestions, but looks good. I don't need to review this again.

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
alicefuzier: Maybe? I feel that it could and probably should exist only in WorkState.Denormalised and WorkState.Indexed, but I'm not sure I would do this change in this PR. This is mostly to just get the router back to what it was doing before

alexwlchan: Sounds good, let's park it as something to address later.

(I hope we can get a TEI pipeline out today; I don't want to throw in more internal model changes that aren't blocking this work.)

alexwlchan (approved):
✨

PR #1905, opened 2021-09-21
Merger copies collectionPath from calm to tei

alicefuzier:


alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5309

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
```diff
@@ -234,16 +239,34 @@ object PlatformMerger extends Merger {
...
+  private def modifyInternalWorks(work: Work.Visible[Identified],
+                                  items: List[Item[IdState.Minted]],
+                                  collectionPath: Option[CollectionPath]) = {
+    // Internal works are in TEI works. If they are merged with Sierra, we want the Sierra
     // items to be added to TEI internal works so that the user can request the item
     // containing that work without having to find the wrapping work.
     work
       .mapState { state =>
         state.copy(internalWorkStubs = state.internalWorkStubs.map { stub =>
+          val updatedCollectionPath =
```
alexwlchan, line 251: Maybe worth a comment about why we're concatenating collection paths?

pipeline/merger/src/test/scala/weco/pipeline/merger/services/TeiOnMergerScenarioTest.scala
```diff
@@ -127,17 +154,72 @@ class TeiOnMergerScenarioTest
       .getMerged(secondInternalWork)
       .data
       .items should contain allElementsOf physicalSierra.data.items
+
+    And("the tei internal works retain their collectionsPath")
```
alexwlchan, line 158: We should assert that the top-level work also retains its collection path, so we know we can build the collection tree.

pipeline/merger/src/test/scala/weco/pipeline/merger/services/TeiOnMergerScenarioTest.scala
```diff
@@ -127,17 +154,72 @@ class TeiOnMergerScenarioTest
...
 
     When("the tei work is merged")
     val outcome = merger.applyMerge(List(Some(teiWork)))
 
     Then("the tei work should be a TEI work")
-    outcome
-      .getMerged(teiWork) shouldBe teiWork
+    outcome.getMerged(teiWork) shouldBe teiWork
+
+    And("the the tei inner works shou;ld be returned")
```
alexwlchan, line 183:
```suggestion
    And("the the tei inner works should be returned")
```

alexwlchan (dismissed):
Couple of small comments, but nothing major.

pipeline/merger/src/test/scala/weco/pipeline/merger/services/TeiOnMergerScenarioTest.scala
alicefuzier: This is a bit of a weird test because it's a case that shouldn't be happening where internal works have collectionspath but the wrapper work doesn't. It should log a warning at least cause it's a data issue and probably a transformer bug if it happens, so I'll add that. But it needed to be tested because it has to be dealt with in the match statement. [this test](https://github.com/wellcomecollection/catalogue-pipeline/pull/1905/files#diff-0b198acf186857fb7019d066fba330d14bab1b0e0b965fd4ad1106b437f0befbR192) tests the case where the tei work is not merged and the collectionpath are correct and [this other test](https://github.com/wellcomecollection/catalogue-pipeline/pull/1905/files#diff-0b198acf186857fb7019d066fba330d14bab1b0e0b965fd4ad1106b437f0befbR58) tests the case where a calm work is redirected to the tei work and so the fake collectionpath is blatted

alexwlchan (approved):

PR #1908, opened 2021-09-22
Add the 2021-09-23 pipeline

alexwlchan:


infrastructure/modules/worker/main.tf
```diff
@@ -1,13 +1,15 @@
 locals {
   # Override the default service name if requested
   deployment_service_name = var.deployment_service_name == "" ? var.name : var.deployment_service_name
+
+  service_name = var.service_name == "" ? var.name : var.service_name
```
alicefuzier, line 5: I'm struggling to understand this, or the other related changes. Would it make sense to add a comment explaining?

alexwlchan: I've added a comment, any better?

alicefuzier: Sort of.. doesn't that make it clash with other pipelines though?

alexwlchan: No:

* In the ECS console, service names only have to be unique within a cluster. You can repeat names across clusters.
* In the logging cluster, you can filter by (cluster, service name) to find just those services – but now you can actually find the service name.

reindexer/get_reindex_status.py
```diff
@@ -242,6 +275,21 @@ def main(reindex_date):
...
+        rows = [
+            [q_url.split("/")[-1].replace(f"catalogue-{reindex_date}_", ""), size]
+            for q_url, size in sorted(non_empty_dlqs.items())
+        ]
+        print(tabulate.tabulate(rows, tablefmt="plain", colalign=("left", "right")))
+    else:
+        print("There are no messages on DLQs")
+
+    print("")
+
```
alexwlchan, line 292: This change is to highlight any DLQs in this pipeline that have failing messages:

Screenshot 2021-09-24 at 10 15 15

alicefuzier (approved):

PR #1909, opened 2021-09-23
Don't declare separate types for each instance of Note – make internal model compile faster?

alexwlchan:
Creating the implicit JSON encoders/decoders for Note is surprisingly expensive (~10% of implicit resolution) – because we have a lot of different subtypes to create encoders/decoders for.

We don't really need the separate types, so I want to try removing them and switching to something closer to what we do in the API, and see if it helps the compilation times of internal model.

This also means we can change the Note types in the pipeline without having to make a corresponding change in the API repo.

For https://github.com/wellcomecollection/platform/issues/5298

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Note.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Note.scala
@@ -1,61 +1,46 @@
 package weco.catalogue.internal_model.work
 
-sealed trait Note {
-  val content: String
+case class NoteType(id: String, label: String)
+
+case object NoteType {
+  val GeneralNote                = NoteType(id = "general-note", label = "Notes")
+  val BibliographicalInformation = NoteType(id = "bibliographic-info", label = "Bibliographic information")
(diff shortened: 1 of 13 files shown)


alicefuzier (dismissed):

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Note.scala
```diff
@@ -1,61 +1,61 @@
 package weco.catalogue.internal_model.work
 
-sealed trait Note {
-  val content: String
+case class NoteType(id: String, label: String)
+
+case object NoteType {
```
jamesgorrie, line 5: I wonder if it's worth leaving a note here as to why we're changing this to avoid a future person thinking they might make the code neater by using the ADT.

jamesgorrie: We have a few patterns like this used across different places, are they worth documenting globally?

alexwlchan:
I'm going to merge this as it's approved and we know the code it's forked from is stable enough for a reindex, it's just wrong.

PR #1912, opened 2021-09-23
Remove the normalised concept constructors from internal model

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5298

## Before

We have a bunch of methods in the internal model library that:

- allow us to construct "normalised" concepts, e.g.

    ```scala
    object Concept {
      def normalised(label: String): Concept(label = label.trimWhitespace)
    }
    ```

- allow us to parse values from source data for time periods, e.g.

    ```scala
    object Period {
      def parse(label: String): Period(label = label.trimWhitespace, range = InstantRange.parse(label))
    }
    ```

These are in the internal model library so they're made available to every application, but we only ever use them in the transformers. Stronger – we should never use them outside the transformers, because they're meant for modifying the source data.

## After

This patch moves all this parsing-esque logic into the transformer_common library in the new `ConceptsTransformer` trait and `ParsedPeriod` model. Implicit resolution continues to dominate the compilation time of internal model, so hopefully moving a bunch of those implicits into a less-widely used library will make builds faster.

### DIFF
--- a/pipeline/transformer/transformer_miro/src/main/scala/weco/pipeline/transformer/miro/transformers/MiroGenres.scala
+++ b/pipeline/transformer/transformer_miro/src/main/scala/weco/pipeline/transformer/miro/transformers/MiroGenres.scala
@@ -1,9 +1,9 @@
 package weco.pipeline.transformer.miro.transformers
 
-import weco.catalogue.internal_model.text.TextNormalisation._
 import weco.catalogue.internal_model.identifiers.IdState
 import weco.catalogue.internal_model.work.{Concept, Genre}
 import weco.pipeline.transformer.miro.source.MiroRecord
+import weco.pipeline.transformer.text.TextNormalisation._
 import weco.pipeline.transformer.transformers.ConceptsTransformer
 
 trait MiroGenres extends ConceptsTransformer {
--- a/pipeline/transformer/transformer_miro/src/main/scala/weco/pipeline/transformer/miro/transformers/MiroGenres.scala
+++ b/pipeline/transformer/transformer_miro/src/main/scala/weco/pipeline/transformer/miro/transformers/MiroGenres.scala
@@ -13,9 +13,11 @@ trait MiroGenres extends ConceptsTransformer {
     (miroRecord.physFormat.toList ++ miroRecord.lcGenre.toList).map { label =>
       val normalisedLabel = label.sentenceCase
 
+      val concept = Concept(label = normalisedLabel).normalised
+
       val genre = Genre(
         label = normalisedLabel,
-        concepts = List(Concept.normalised(label = normalisedLabel))
+        concepts = List(concept)
       )
 
       genre.normalised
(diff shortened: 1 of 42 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraGenres.scala
```diff
@@ -66,7 +68,12 @@ object SierraGenres
...
     primarySubfields.map { subfield =>
-      val concept = Concept.normalised(label = subfield.content)
-      concept.copy(id = identifyConcept(concept, varField))
+      Concept(
+        id = identifyConcept(
+          ontologyType = "Concept",
+          varField = varField
+        ),
+        label = subfield.content
+      ).normalised
```
alexwlchan, line 77: This illustrates one benefit of moving these to methods on an existing instance of Concept/Period/et al – they'll drop in just fine for identified concepts when that work comes along.

alexwlchan:
This doesn't seem to make the dent in compile times I was hoping for, but it just get some more code out of internal model and reduce the surface area of stuff we have to consider for making it compile faster.

pipeline/transformer/transformer_miro/src/main/scala/weco/pipeline/transformer/miro/transformers/MiroGenres.scala
```diff
@@ -1,19 +1,25 @@
...
+import weco.pipeline.transformer.text.TextNormalisation._
+import weco.pipeline.transformer.transformers.ConceptsTransformer
 
-trait MiroGenres {
+trait MiroGenres extends ConceptsTransformer {
   def getGenres(miroRecord: MiroRecord): List[Genre[IdState.Unminted]] =
     // Populate the genres field.  This is based on two fields in the XML,
     // <image_phys_format> and <image_lc_genre>.
     (miroRecord.physFormat.toList ++ miroRecord.lcGenre.toList).map { label =>
       val normalisedLabel = label.sentenceCase
```
jamesgorrie, line 14: Should this be moved into the `ConceptsTransformer`?

jamesgorrie (commented):
Am I right in saying that the only time we ever use the `normalised` method is now the transformers?

If that's the case - why not remove the method from the `Concept` & `Agent` types, and just have the transformer deal with it, further indicating it's not actually part of the type, but rather a transformer method.

alexwlchan:
> Am I right in saying that the only time we ever use the normalised method is now the transformers?
>
> If that's the case - why not remove the method from the Concept & Agent types, and just have the transformer deal with it, further indicating it's not actually part of the type, but rather a transformer method.

Yes, that's what this patch is doing – have I explained it badly?

jamesgorrie:
Nope, I think I got that from the description / code - and wondered if we could push it further and not have the method on `ConceptOps` - as it's not something we want to expose evreywhere the Concepts are available, but rather it's just a transformer method?

alexwlchan:
> wondered if we could push it further and not have the method on `ConceptOps` - as it's not something we want to expose evreywhere the Concepts are available, but rather it's just a transformer method?

ConceptOps is only available in the transformer – it's defined as part of the ConceptsTransformer trait, so you can only use them if you extend that trait. That trait is only in the transformer common lib.

alicefuzier (approved):

PR #1913, opened 2021-09-24
Add a timeout to internal model builds

alexwlchan:


.buildkite/pipeline.yml
```diff
@@ -4,6 +4,18 @@ steps:
...
+    # This will fail the build if internal model takes more than
+    # 13 minutes to build and test.  This is to make sure we know if
+    # the internal model compilation starts creeping again.
+    #
+    # We may increase this timeout, but we should do so intentionally.
+    # We should only increase it when we know what's causing the slowness,
+    # and we think it's an acceptable tradeoff.
+    #
+    # See https://buildkite.com/docs/pipelines/command-step
+    timeout_in_minutes: 13
```
jamesgorrie, line 17: Assuming this is about the average of where we are now?

jamesgorrie (approved):
Nice line in the sand!

.buildkite/pipeline.yml
alexwlchan: Yeah, we're averaging 10–12m depending on whether Buildkite has a warm cache. The last big change doubled that to 20 minutes.

PR #1917, opened 2021-09-28
Extract msPart inner works

alicefuzier:


jamesgorrie (approved):

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -70,7 +71,50 @@ class TeiXml(val xml: Elem) extends Logging {
...
+            id = id,
+            title = s"$wrapperTitle part $partNumber",
+            languages = languages,
+            description = description)
+      }
+      .toList
+      .sequence
+
+  /**
+    * Extract information about inner works for single part manuscripts.
```
alexwlchan, line 114: Can you explain what a single part manuscript is? Or perhaps point to an example of one?

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -70,7 +71,50 @@ class TeiXml(val xml: Elem) extends Logging {
...
+    * check https://github.com/wellcomecollection/wellcome-collection-tei/blob/main/docs/TEI_Manual_2020_V1.pdf
+    * for more info.
+    */
+  def nestedTeiData = nestedTeiDataFromItems.flatMap {
+    case Nil      => title.flatMap(nestedTeiDataFromParts)
+    case teiDatas => Right(teiDatas)
+  }
+
+  /**
+    * Extract information about inner works for multi part manuscripts.
```
alexwlchan, line 89: Can you explain what a multi-part manuscript is? Or point to an example of one?

alexwlchan:
Couple of suggestions, but nothing major.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
alicefuzier: In the comment on the public function for nested data there's a link to the tei manual. I think that's the best source of information rather than me trying to rephrase it here

alexwlchan (approved):

PR #1922, opened 2021-10-01
Split deployment of the pipeline/adapters

alexwlchan:
We expect the pipeline deployment might fail sometimes if the internal model version has changed, but we never expect the adapter deployment to fail.

Closes https://github.com/wellcomecollection/platform/issues/5315


jamesgorrie (approved):
Nice

PR #1947, opened 2021-10-28
Add scribe contributor from tei

alicefuzier:
This turned out to be very fiddly with a million corner cases. 
I've tried to comment and break down into smaller functions as much as possible to make it as readable as possible


alexwlchan:
👀

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
```diff
@@ -2,13 +2,41 @@ package weco.pipeline.transformer.tei
...
+  lazy val scribesMap = TeiContributors.scribes(xml, id)
+
+  def parse: Result[TeiData] =
+    for {
+      summary <- summary
+      bNumber <- bNumber
+      title <- title
+      languageData <- TeiLanguages(xml)
+      (languages, languageNotes) = languageData
+      scribes <- scribesMap
```
alexwlchan, line 26: Two notes:

* For future!me/other reviewers: this is almost the same as the `parse()` function removed form TeiTransformer; the `scribes` part is the only change.
* Could we rearrange the order of the `for` comprehension so the order matches the order in which they're consumed by `TeiData`? Just to make the different parts easier to compare.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+          res <- createContributor(
+            label = label,
+            id = id,
+            isFihrist = isFihrist,
+            ContributionRole("author"))
+        } yield res
+
+      }
+      .toList
+      .sequence
```
alexwlchan, line 47: Copied unmodified from TeiAuthors.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+    * Scribes appear in the physical description section of the manuscript and can appear
+    * within a handNote tag with attribute "scribe" as in this example
+    * <physDesc>
+    *     <handDesc>
+    *         <handNote scope="minor" scribe="Scribe_A"> <locus target="#Wellcome_Batak_36801_1 #Wellcome_Batak_36801_2" >a 2-62, b 2-7; b 8-22</locus> Gorak-gorahan, etc. Southern form of ta.</handNote>
+    *         <handNote scope="minor" scribe="Scribe_B"> <locus target="#Wellcome_Batak_36801_3">b 23-35</locus>Another hand, which uses the northern ta.</handNote>
+    *         <handNote scope="minor" scribe="Scribe_C"> <locus target="#Wellcome_Batak_36801_9">b 44-45</locus>Another hand; careless writing, at first with northern ta, later southern ta.</handNote>
+    *         <handNote scope="minor" scribe="Scribe_D"> <locus target="#Wellcome_Batak_36801_10">b 45-56</locus>A different handwriting, using the southern form of ta but not the peculiar form of ya which is found in the text written by <persName>Datu Poduwon</persName>.</handNote>
+    *     </handDesc>
+    * </physDesc>
```
alexwlchan, line 59:
```suggestion
    * <physDesc>
    *     <handDesc>
    *         <handNote scope="minor" scribe="Scribe_A">
    *           <locus target="#Wellcome_Batak_36801_1 #Wellcome_Batak_36801_2" >a 2-62, b 2-7; b 8-22</locus> Gorak-gorahan, etc. Southern form of ta.
    *         </handNote>
    *         <handNote scope="minor" scribe="Scribe_B">
    *           <locus target="#Wellcome_Batak_36801_3">b 23-35</locus>Another hand, which uses the northern ta
    *         </handNote>
    *         <handNote scope="minor" scribe="Scribe_C">
    *           <locus target="#Wellcome_Batak_36801_9">b 44-45</locus>Another hand; careless writing, at first with northern ta, later southern ta.
    *         </handNote>
    *     </handDesc>
    * </physDesc>
```

This may not be strictly accurate to the original TEI, but the extra newlines just make the structure a little easier to follow on a narrow screen.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+            scribesMap ++ addIdsToMap(scribesMap, contributor, ids)
+
+        }
+
+    }
+  }
+
+  /**
+    * The scribe name can be within the handNote tag directly or within a
+    * persName node with role="src" inside the handNote tag. If there is a persName tag,
```
alexwlchan, line 143: `role="src"` doesn't match the code; which is correct?

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+    *            <persName role="scr">Mahādeva Pāṇḍe</persName>
+    *        </handNote>
+    *    </handDesc>
+    *</physDesc>
+    * If the scribe refers to a part or item of the manuscript, the handNote tag will have a locus node with the item or part ids that it refers to
+    *
+    */
+  def scribes(
+    xml: Elem,
+    workId: String): Result[Map[String, List[Contributor[Unminted]]]] =
```
alexwlchan, line 74: Can you explain what the keys of this map are? What are these string values?

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+
+    }
+  }
+
+  /**
+    * The scribe name can be within the handNote tag directly or within a
+    * persName node with role="src" inside the handNote tag. If there is a persName tag,
+    * we pick that to construct the scribe name. Otherwise we use the text directly inside the handNote tag
+    */
+  private def parseScribeLabel(n: Node) = {
```
alexwlchan, line 146: To make this easier to follow/understand:

* Let's change the variable name to `handNote`, as that's the only type of node we're going to pass in here. This would be good to do throughout the helper functions – it'd be easier to map these helper functions back to the TEI XML.
* Pull out `n.attribute("scribe")` into a variable like `handNoteText`, so it more clearly maps to your comment.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+            s"Multiple persName nodes with type=original in author $n"
+          )
+        )
+    }
+
+  private def getFromPersNode(n: Node, persNode: Node) =
+    Right((persNode.text.trim, getId(n, persNode)))
+
+  private def getFromAuthorNode(n: Node) =
+    Right((n.text.trim, (n \@ "key").trim))
```
alexwlchan, line 261: These are copied directly from TeiAuthors, but for `getFromOriginalPersNode`, could we choose different variable names to make the intent more obvious? e.g. `persNode` on `getFromPersNode` is helpful.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+    * and sometimes it is on the wrapping author node and we must deal with both cases.
+    */
+  private def getId(n: Node, persNode: Node) = {
+    val persNodeId = (persNode \@ "key").trim
+    if (persNodeId.isEmpty) {
+      (n \@ "key").trim
+    } else {
+      persNodeId
+    }
+  }
```
alexwlchan, line 274: Also copied from TeiAuthors, is it a `<pers>` per the variable or a `<persName>` per the comment? Could we make those names consistent?

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+        workIds match {
+          case Nil =>
+            scribesMap + addIdToMap(workId, scribesMap, contributor)
+          case ids =>
+            scribesMap ++ addIdsToMap(scribesMap, contributor, ids)
+
+        }
+
+    }
+  }
```
alexwlchan, line 139: nit: bin the outermost parentheses, they're just visual noise in a busy file.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+
+  private def addIdToMap(workId: String,
+                         scribesMap: Map[String, List[Contributor[Unminted]]],
+                         c: Contributor[Unminted]) = {
+    scribesMap.get(workId) match {
+      case None =>
+        (workId, List(c))
+      case Some(list) => (workId, list :+ c)
+    }
+  }
```
alexwlchan, line 197: What do you think of something like:

```suggestion
    (workId, scribesMap.getOrElse(workId, List()) :+ c)
```

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+    case Nil            => getFromAuthorNode(n)
+    case List(persNode) => getFromPersNode(n, persNode)
+    case list           => getFromOriginalPersNode(n, list)
+  }
+
+  private def getScribeContributor(n: Node) =
+    for {
+      maybeLabel <- parseScribeLabel(n)
+      contributor <- maybeLabel.map { label =>
+        createContributor(label, ContributionRole("scribe"))
```
alexwlchan, line 110: Because we're not doing anything fancy with IDs or `isFihrist` here, let's inline the contributor to make this a tad easier to follow:

```suggestion
        Contributor(Person(label), roles = List(ContributionRole("scribe")))
```

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
```diff
@@ -0,0 +1,276 @@
...
+        (locus \@ "target").trim.replaceAll("#", "").split(" ")
+      }
+      .toList
+      .flatten
+  }
+
+  private def createContributor(
+    label: String,
+    role: ContributionRole): Result[Contributor[Unminted]] =
+    createContributor(label = label, id = "", isFihrist = false, role = role)
```
alexwlchan, line 211: Possibly daft question: could we just set default arguments for `id` and `isFihrist` on the other `createContributor` method?

alexwlchan (approved):
I've tried to review this as best I can, but a diff this big is a bit large for me to review thoroughly. I've focused on TeiContributors as the most important bit of the code; I've only skimmed the tests.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
alicefuzier: TeiAuthors was renamed to TeiContributors but the changes were so many that gihub didn't figure it out

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
alicefuzier: I think this comment is correct but not clear enough, will change

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
alicefuzier: They are the work ids of the works they belong to. Will add in the comment

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
alexwlchan: Ah, sorry, half of these comments are for my benefit as I was trying to read the diff.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
alicefuzier: Not a daft question. I think I just felt icky about putting an empty string as a default value for an id and instead of using options, which would probably be the best option. So I decided to hide it into an overloaded method which is even uglier

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiContributors.scala
alicefuzier: Nicer, thanks!

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiXml.scala
alicefuzier: > For future!me/other reviewers: this is almost the same as the parse() function removed form TeiTransformer; the scribes part is the only change



Just to add the reasoning for moving it here: the `scribes` function is a bit weird in that it returns a map and then each usage needs to know what to pick out of it. Also, because all scribes for all works in the tei are in the same place in the xml, it makes sense to parse it once and then pipe that map down when we're parsing nested data. But putting all of that in a `parse` function in `TeiTransformer` suddenly seemed to put too much knowledge about what needs to know what when parsing the Xml, and `TeiTransformer` didn't seem the right place anymore

PR #1952, opened 2021-11-04
Add the implicit Work encoders in the transformers

alexwlchan:
These are among the slowest apps to build, and they all need a full copy of the Work encoder.  Why are we creating it from scratch, when we could be importing a prebuilt version?

🤞 targeting the slowest bit of the current pipeline build


jamesgorrie (approved):
Simple, yet (very) effective!

jamesgorrie:
(asks again...) Do you think it's worth turning implicit derivation off?

alexwlchan:
> Do you think it's worth turning implicit derivation off?

I'm certainly inching towards it. I was surprised by how little broke when I turned off implicit derivation in the transformers, certainly.

(The next bit to tackle would be source_model.)

PR #1953, opened 2021-11-05
Add the ESTC identifier type to internal model

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5358

This should be merged before we make any transformer changes, so we can get it into the catalogue API – otherwise the catalogue API will be unable to decode works with ESTC identifiers, and will 500 whenever they appear in search results.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
@@ -126,7 +126,7 @@ object IdentifierType extends Enum[IdentifierType] {
     val label = "Fihrist Authority"
   }
 
-  // http://estc.bl.uk/
+  // http://estc.bl.uk/ / https://www.wikidata.org/wiki/Property:P3939
   case object ESTC extends IdentifierType {
     val id = "bl-estc-citation-no"
     val label = "British Library English Short Title Catalogue"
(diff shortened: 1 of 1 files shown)


jamesgorrie (commented):
Naming question is all

alexwlchan:
I don't see a naming question…?

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
```diff
@@ -125,4 +125,10 @@ object IdentifierType extends Enum[IdentifierType] {
     val id = "fihrist"
     val label = "Fihrist Authority"
   }
+
+  // http://estc.bl.uk/
+  case object ESTC extends IdentifierType {
+    val id = "bl-estc-citation-no"
```
jamesgorrie, line 131: Why `citation`? I don't see it mentioned in the spec, they seem to call them [ESTC IDs, or estc identification number](https://estc.ucr.edu/starhelp/basic_search.html)? `bl-estc`? Or is this something more specific in the spec?

jamesgorrie:
@alexwlchan not sure why it didn't submit with the review...

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
alexwlchan: This comes from Jonathan's initial issue; it does seem to used elsewhere: https://www.wikidata.org/wiki/Property:P3939

jamesgorrie: Lovely, I was just wondering if it was referring to something specific. The WikiData article is more descriptive, and provbably useful to us, might be worth linking through to that.

jamesgorrie (approved):

common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/IdentifierType.scala
alexwlchan: Added it to the comment.

PR #1955, opened 2021-11-07
Consolidate and tidy up our typesafe builders

alexwlchan:
* Inline all uses of ElasticIndexerBuilder and ElasticSourceRetrieverBuilder, which weren't adding anything
* Create a TransformerMain trait to make the Main objects in the transformers consistent
* Create an IngestorMain trait to make the Main objects in the ingestors consistent
* Push the creation of the SQSStream and ActorSystem into the pipeline-storage typesafe builder, rather than repeat it everywhere


pipeline/ingestor/ingestor_common/src/main/scala/weco/pipeline/ingestor/common/IngestorMain.scala
```diff
@@ -0,0 +1,67 @@
...
+import weco.pipeline_storage.Indexable
+import weco.pipeline_storage.elastic.{ElasticIndexer, ElasticSourceRetriever}
+import weco.pipeline_storage.typesafe.PipelineStorageStreamBuilder
+import weco.typesafe.WellcomeTypesafeApp
+import weco.typesafe.config.builders.AkkaBuilder
+import weco.typesafe.config.builders.EnrichConfig._
+
+import scala.concurrent.ExecutionContext
+
+trait IngestorMain[In, Out] extends WellcomeTypesafeApp {
```
alicefuzier, line 17: why does this need to be a trait and not a class with a contructor? it doesn't look like we override the behaviour in its usages

alexwlchan: I think you want to extend this in the `Main` object, no? I'm not sure what this would look like with a class.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerMain.scala
```diff
@@ -0,0 +1,70 @@
...
+import weco.pipeline_storage.elastic.{ElasticIndexer, ElasticSourceRetriever}
+import weco.pipeline_storage.typesafe.PipelineStorageStreamBuilder
+import weco.storage.typesafe.S3Builder
+import weco.typesafe.WellcomeTypesafeApp
+import weco.typesafe.config.builders.AkkaBuilder
+import weco.typesafe.config.builders.EnrichConfig._
+
+import scala.concurrent.ExecutionContext
+
+trait TransformerMain[Payload <: SourcePayload, SourceData]
```
alicefuzier, line 22: ditto as above. Is there a reson to prefer a trait over a class with injected parameters? Reason for me to prefer a class in this case where we don't override the behavipour is that if it's a final class we can test the class once and trust that its behaviour cannot change where it used. 
In the case of traits it can be changed so tests on the trait are not enough imo, you need to test each implementation too with a pattern like the test cases. 
Also, I think not having to follow overrides makes for much readable code, and ingle test on a final class are also much more redable than TestCases.

pipeline/ingestor/ingestor_common/src/main/scala/weco/pipeline/ingestor/common/IngestorMain.scala
alicefuzier: hmm I think you could avoid extending `WellcomeTypesafeApp` here and keep that duplicated and just have a class for everything else that is duplicated between the 2 ingestors no?
I think there is more to be gained avoiding too many overrides than by removing absolutely all code duplication, especially in a class that is actually never tested. It should be as readable as possible and overrides make everything harder to follow for me.

alicefuzier (approved):

PR #1963, opened 2021-11-11
Extract the colophon note in TEI works

alexwlchan:
For https://github.com/wellcomecollection/catalogue-pipeline/issues/1959 / https://github.com/wellcomecollection/platform/issues/5118


pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiNotes.scala
```diff
@@ -0,0 +1,35 @@
...
+    *       <sourceDesc>
+    *         <msDesc xml:lang="en" xml:id="MS_Arabic_1">
+    *           <msContents>
+    *             <colophon> <locus>F. 9v</locus> </colophon>
+    *
+    */
+  private def getColophon(value: NodeSeq): Seq[Note] =
+    (value \ "colophon")
+      .map { n =>
+        n.text.trim
```
alicefuzier, line 29: I _think_ that if you use `.text` you shouldn't get comments? Have you tried to see if that tests passes even without the strip comments function?

alexwlchan: Oh yeah, it totally does. My bad, fixed, thanks. 🙈

alicefuzier (approved):

PR #1964, opened 2021-11-11
Map reference numbers

harrisonpim:
- This is part of https://github.com/wellcomecollection/platform/issues/5333, but won't close it until the query has also been updated

This PR: 
- adds `data.referenceNumber` to the mapping
- adds a new `search.identifiers` field, and copies relevant identifiers into it, ready for querying

Adding the `search.identifiers` will allow us to simplify [the section of our query which matches identifiers](https://github.com/wellcomecollection/catalogue-api/blob/main/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala#L36-L56), rather than extending it even further


alexwlchan (approved):

harrisonpim:
@jamesgorrie pointed out that setting the `search.identifiers` field type as `keyword` could result in a long single keyword of all ids, rather than a list of separate keywords, which would make matching near impossible. I'm going to run some tests in the rank cluster when it's up and running, but might come back to change that `keyword` to `text` if james's hypothesis is correct.

PR #1965, opened 2021-11-11
Add the incipit/explicit notes on TEI works

alexwlchan:
For https://github.com/wellcomecollection/catalogue-pipeline/issues/1959 / https://github.com/wellcomecollection/platform/issues/5118

I've only been able to find examples of this note in an msItem, not in an msPart, so it just fetches the incipit/explicit from the items for now. (This probably isn't worth worrying about too much – if we find some other notes we want from msPart, we'll start running `TeiNotes` against msParts, and that will pull in these notes for us.)


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Note.scala
```diff
@@ -60,6 +60,9 @@ case object NoteType {
     NoteType(id = "references-note", label = "References note")
   val ColophonNote =
     NoteType(id = "colophon", label = "Colophon note")
+
+  val Incipit: NoteType = NoteType(id = "incipit", label = "Incipit")
```
alicefuzier, line 64: Sorry, that isn't very clear from the modeling document, but the proposal is to rename incipit to begins and explicit to ends. The idea is that those terms are more user friendly

alicefuzier (approved):
Just a comment on the name of the note, but otherwise looks good!

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Note.scala
alexwlchan: Interesting. I've changed it for now, but I'm not sure I agree – it seems like this is a technical term from the world of manuscripts that would be understood here (cf folio) – but we can change it later. I don't think it should hold up this PR.

alicefuzier: For what is worth, this is not my suggestion. According to Adrian (and Lara and Branwen) a lot of these latin terms used in manuscript description are very obscure to them too. Incipit /explicit might be well nown among manuscript researchers, but they are completely obscure to anyone else. Begins/ends has the advantage of being accessible to both manuscript researchers and other people. It doesn't lose anything and adds understanding to a wider range of people.
 Colophon is an exception apparently because it's a widely accepted term in the community and there is no easy way to describe this with another more widely known term.

PR #1971, opened 2021-11-15
Copy thumbnails to inner works in the TEI merger

alexwlchan:
For https://github.com/wellcomecollection/catalogue-pipeline/issues/1970

Plus making `thumbnail` a DigitalLocation rather than a Location – it was already, but now that's enforced by the type system.


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
```diff
@@ -222,20 +216,6 @@ object WorkState {
     val relations = Relations.none
 
     override val modifiedTime: Instant = sourceModifiedTime
-
-    def internalWorksWith(version: Int): List[Work.Visible[Identified]] =
```
alexwlchan, line 226: I moved this into the merger because that's the only place it's used, and the next time we tweak it I don't want to cause a full rebuild of internal model.

pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala
```diff
@@ -102,8 +103,11 @@ trait Merger extends MergerLogging {
           val remaining = sources.toSet -- redirectedSources
           val redirects = redirectedSources.map(redirectSourceToTarget(target))
           logResult(result, redirects.toList, remaining.toList)
-          val internalWorks = result.mergedTarget.state.internalWorksWith(
-            version = result.mergedTarget.version)
+
+          val internalWorks = result.mergedTarget.internalWorksWith(
+            thumbnail = result.mergedTarget.data.thumbnail,
```
alicefuzier, line 108: 🤦‍♀️  sorry, this is probably me forgetting it

alicefuzier (approved):

Issue #1970, opened 2021-11-15
Identify/fix the outstanding TEI issues


alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5118

- [x] Why isn't the merger copying across thumbnails? (https://github.com/wellcomecollection/catalogue-pipeline/pull/1971)
- [x] Front-end shouldn't show "undefined" if there's a missing referenceNumber (https://github.com/wellcomecollection/wellcomecollection.org/pull/7313)
- [x] Add a referenceNumber to the top-level work
- [x] Don't always assign production to the wrapper work?
- [x] Why did this work pass transformation? https://wellcomecollection.org/works/pcgwbkea?toggle=tei (https://github.com/wellcomecollection/catalogue-pipeline/pull/1980)
- [ ] What's up with the requesting of nested items?
- ~Decide what we want to do about works with a single inner work~

alexwlchan:
Closing the last item in favour of https://github.com/wellcomecollection/platform/issues/5377

PR #1973, opened 2021-11-16
Subjects tei

alicefuzier:
For #1970

### DIFF
--- a/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiSubjects.scala
+++ b/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiSubjects.scala
@@ -1,0 +1,33 @@
+package weco.pipeline.transformer.tei.transformers
+
+import weco.catalogue.internal_model.identifiers.IdState.{Identifiable, Unidentifiable}
+import weco.catalogue.internal_model.identifiers.{IdState, IdentifierType, SourceIdentifier}
+import weco.catalogue.internal_model.work.{Concept, Subject}
+import weco.pipeline.transformer.tei.NormaliseText
+
+import scala.xml.{Elem, Node}
+
+object TeiSubjects {
+  def apply(xml: Elem): List[Subject[IdState.Unminted]] = (xml \\ "profileDesc" \\ "keywords").flatMap { keywords =>
+    val identifierType = (keywords \@ "scheme").toLowerCase.trim match {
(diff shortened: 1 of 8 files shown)


pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiSubjects.scala
```diff
@@ -0,0 +1,65 @@
...
+        val label = NormaliseText(term.text)
+        val reference = parseReference(term)
+        val id = (reference, identifierType) match {
+          case (Some(r), Some(identifierType)) =>
+            Identifiable(SourceIdentifier(identifierType, "Subject", r))
+          case _ => Unidentifiable
+        }
+        label.map(l =>
+          Subject(id = id, label = l, concepts = List(Concept(label = l))))
+      }
```
alexwlchan, line 51: Really easy to follow. 👍

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiSubjects.scala
```diff
@@ -0,0 +1,65 @@
...
+    }.toList
+
+  private def parseReference(term: Node) = {
+    val referenceString = term \@ "ref"
+    // arabic manuscripts seem to have the subject id in the
+    // attribute "key" instead of "ref" ¯\_(ツ)_/¯
+    val idValue =
+      if (referenceString.isEmpty) term \@ "key" else referenceString
+    // some of the subject ids are prepended with "subject_sh" for lcsh or " subject" for mesh.
+    // So here we remove that to get the id
```
alexwlchan, line 61: Is the 'sh' part of the LCSH identifier, or do we want to bin that too? It's not obvious to me.

e.g. in your example is the identifier `sh85083116` or `85083116`

alexwlchan (approved):
Super nice. ✨ 

One question about LCSH identifiers but otherwise this can go in.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiSubjects.scala
alicefuzier: Yeah that comment was written poorly, I've updated it.
Yes, for lcsh `sh` is part of the id so we keep it , we only discard the `subject_` bit before

PR #1981, opened 2021-11-22
Update rules for requesting to enable more requests

alexwlchan:
Branwen noticed that some Hebrew manuscripts (somhe) weren't requestable on wc.org/works but they were in Encore, because the Rules for Requesting had changed in Sierra but not in the pipeline. This brings them back up-to-date.


jamieparkinson (approved):

jamieparkinson:
Hang on - is this just changing the comments?

alexwlchan:
> is this just changing the comments?

…

oops

PR #1984, opened 2021-11-25
Highlight nodes in the matcher graph which don't correspond to a Work processed by the matcher

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5375

This changes the matcher graph to expose the "dummy" nodes it creates when Work A points to Work B, but we haven't seen Work B yet. Previously these could be filtered out if (as in this case) the unseen Work had been deleted. The presence of this Work which the matcher hasn't seen causes problems in the merger, because a single Work the matcher hasn't seen will cause the merger to skip merging anything.

Screenshot 2021-11-25 at 16 16 33

For some reason the matcher never saw the Miro work, and so the Calm/Sierra works are being left unmerged.


jamesgorrie (approved):
Makes sense

PR #1986, opened 2021-11-26
Tidy up the matcher/merger tests

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/5375

This is just some code tidying and refactoring to clean things up before I start fixing this patch properly, which is ultimately going to require changing the matcher–merger interface. Although lots of lines, this patch isn't that complicated – most of the diff is folding the TEI-specific tests into the merger tests.

Next step is writing some proper integration tests for the services as a single unit, but I'll do that as a single patch.


pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
```diff
@@ -47,11 +47,12 @@ class WorkMatcher(
         // but I haven't thought hard enough about whether it might introduce a
         // hard-to-debug consistency error if another process updates the graph
         // between us reading it and writing it.
+        result = MatcherResult(
+          works = toMatchedIdentifiers(afterNodes),
+          createdTime = Instant.now())
```
jamesgorrie, line 52: Is there potentially a reason these were created separately? I ask as it looks like createdTime _might_ be different?

alexwlchan: You're right, the createdTime on the latter will be after the successful write to the store.

pipeline/merger/src/test/scala/weco/pipeline/merger/MergerScenarioTest.scala
```diff
@@ -332,5 +313,273 @@ class MergerScenarioTest
...
+
+      teiItems should contain allElementsOf sierraWork.data.items
+      teiItems should contain allElementsOf metsWork.data.items
+      teiItems should contain noElementsOf calmWork.data.items
+
+      And("it gets the METS thumbnail")
+      outcome.getMerged(teiWork).data.thumbnail shouldBe metsWork.data.thumbnail
+    }
   }
+
```
jamesgorrie, line 577: ☝️ really nice.

jamesgorrie (approved):
Nice 🧹

PR #1987, opened 2021-11-26
Add some integration tests for the matcher/merger as a single unit

alexwlchan:
The next step towards wellcomecollection/platform#5375

The actual bug in that ticket is quite subtle and difficult to reproduce without a working matcher+merger setup – you have to send the messages from the matcher in the right order, and then those unusual messages cause the merger to do the wrong thing. It's more a bug in the inter-app messaging than in either app.

This patch converts all the merger scenario tests into full-blown matcher/merger integration tests. It creates an instance of both apps, connects a matcher -> merger queue, and then sends works to the matcher queue. The tests check to see if the merger created the right works at the end. It's testing the matcher/merger as a single unit, not as distinct applications. It also gives me a test harness I can use to reproduce the bug in wellcomecollection/platform#5375.

This patch only converts the existing tests (of which there were many) – I'll do a separate patch which adds a new test to illustrates the current problem.


pipeline/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala
```diff
@@ -299,89 +373,66 @@ class MergerScenarioTest
...
+      context
         .getMerged(teiWork)
         .data
         .otherIdentifiers should contain allElementsOf digitalSierra.data.otherIdentifiers
         .filter(_.identifierType == IdentifierType.SierraIdentifier)
     }
+  }
 
-    Scenario("A Tei with internal works and a Calm are merged") {
-      Given("a Tei and a Calm record")
```
alexwlchan, line 355: There's nothing in the TEI transformer to set up this relationship (instead, it goes via Sierra b numbers), so I removed this test.

alexwlchan:
The heavy lifting happens in `IntegrationTestHelpers`, which contains all the setup logic and syntax sugar for the tests. Each test is wrapped in a `withContext { … }` fixture, so the Given/When/Then/And all line up as they do in the current tests.

A lot of it is ported or adapted from the existing tests; they're pretty readable and I didn't want to make unnecessary changes.

jamieparkinson (approved):
This is fantastic, and it sets us up very nicely for the kind of refactors we've discussed which might shift responsibilities around between the matcher and merger

PR #1988, opened 2021-11-26
Fix an issue where we don't fetch suppressed works from the matcher graph

alexwlchan:
This means we end up blatting suppressed works in the matcher graph, which in turn breaks merging of any works in the same graph.

Follows https://github.com/wellcomecollection/catalogue-pipeline/pull/1987. This patch adds a regression test for the bug described in https://github.com/wellcomecollection/platform/issues/5375#issuecomment-980061590, and then the least invasive/likely-to-break fix I could think of.

### DIFF
--- a/pipeline/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala
+++ b/pipeline/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala
@@ -4,6 +4,7 @@ import org.scalatest.GivenWhenThen
 import org.scalatest.featurespec.AnyFeatureSpec
 import org.scalatest.matchers.should.Matchers
 import weco.catalogue.internal_model.identifiers.IdentifierType
+import weco.catalogue.internal_model.work.DeletedReason.SuppressedFromSource
 import weco.catalogue.internal_model.work.{
   CollectionPath,
   Format,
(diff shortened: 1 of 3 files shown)


pipeline/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala
```diff
@@ -640,4 +641,40 @@ class MergerIntegrationTest
...
+      //
+      //      Sierra --> Calm --> Miro
+      //
+      // where the Miro work was suppressed.
+      //
+      // Because the matcher graph doesn't include links for suppressed works,
+      // this graph would end up with two components (Sierra/Calm) and (Miro).
+      // If you later updated the Sierra work, which had no direct link to the Miro
+      // work, the matcher wouldn't find the Miro work in the matcher database
+      // (because it was in a different component) and the Miro work got blatted.
```
jamieparkinson, line 657: Interesting, I hate it

jamieparkinson (approved):

pipeline/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala
alexwlchan: ikr

PR #1991, opened 2021-11-29
Don't wait after autoformatting

alexwlchan:
This means builds will finish faster!  If we have a pull request build that fails because we pushed an autoformat commit, that kicks off a new build – and then Buildkite cancel the intermediate builds, so we're not running tasks we don't care about.

(See settings in https://buildkite.com/wellcomecollection/catalogue-pipeline/settings/builds)


jamesgorrie (approved):

alexwlchan:
Gonna leave this open until I have time to push it to the other repos.

PR #1992, opened 2021-11-29
Add the 2021-11-29 pipeline

alexwlchan:
This adds a new pipeline… and then some.

Essentially I started tugging on a thread around consistency, and ended up rearranging a bunch of modules. In particular:

* All the pipeline services are consistent about how they name and create their input SQS queues and output SNS topics
* The creation of the queue is now handled by `modules/fargate_worker`, rather than on each service
* I rearranged how the modules are called so all the copy/pasted boilerplate is at the bottom, and the differences in each service are more prominent
* Every service gets permissions to publish CloudWatch metrics, so I pushed that all the way down into `infrastructure/modules/worker`


pipeline/transformer/transformer_common/src/test/scala/weco/pipeline/transformer/transformers/NotesTransformerTest.scala
```diff
@@ -0,0 +1,30 @@
...
+//        "A digitised copy is held by the Wellcome Library as part of Codebreakers: Makers of Modern Genetics",
+//        "A digitised copy is held by the Wellcome Library as part of The Mental Health Archives digitisation project..",
+//      )
+//
+//      forAll(digitisationNotes) {
+//        transformer.createLocationOfDuplicatesNote(_) shouldBe None
+//      }
+//    }
+//  }
+//}
```
jamesgorrie, line 30: Did you mean to comment this all out?

jamesgorrie (approved):
PR diffs are hard tp read, but when you check it out the structure is way neater!

PR #1997, opened 2021-12-20
Fix a bug where METS/Sierra works weren't being merged correctly

alexwlchan:
In particular, if you sent the works in this order:

1. The suppressed Sierra e-bib
2. The METS work
3. The unsuppressed Sierra e-bib

We wouldn't create a Sierra->METS link in step 2, so we wouldn't retrieve the METS work on step 3.

This is a side-effect of us attempting to be clever about how we put suppressed works in the store.

Now we give the same "component ID" (now a misnomer) to all the works that are written together, so if two works were written together they'll always be retrieved together.  We then have to rejig the rest of the matcher a bit so that individual components are written together, but overall it's simpler so hopefully we won't have more bugs like this in future.

---

~Still a WIP because I'm struggling to get tests passing locally – they pass individually, but not if you run the whole suite, which makes me think a test is leaking somewhere.~ Turns out these tests care quite a lot about the version of individual works, and I had to set those explicitly.

I'm looking for alternative names to "component ID". Maybe "group ID"?

Once this is merged and we rename component ID (separate PR), I'd quite like to do a reindex to get this change in, and flush out any other works affected by this issue.

### DIFF
--- a/pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
+++ b/pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
@@ -90,23 +90,23 @@ class WorkMatcher(
 
   import WorkGraphUpdater.GraphComponents
 
-  private def toMatchedIdentifiers(components: GraphComponents): Set[MatchedIdentifiers] =
+  private def toMatchedIdentifiers(
+    components: GraphComponents): Set[MatchedIdentifiers] =
     components
-      .map {
-        workNodes =>
-          // The matcher graph may include nodes for Works it hasn't seen yet, or which
-          // don't exist.  These are placeholders, in case we see the Work later -- but we
-          // shouldn't expose their existence to other services.
-          //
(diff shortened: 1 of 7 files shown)


pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
```diff
@@ -35,7 +35,7 @@ class WorkMatcher(
         beforeNodes <- workGraphStore.findAffectedWorks(work)
         afterNodes = WorkGraphUpdater.update(work, beforeNodes)
 
-        updatedNodes = afterNodes -- beforeNodes
+        updatedNodes = afterNodes.flatten -- beforeNodes
```
jamieparkinson, line 38: Are we `flatten`ing `afterNodes` 3 times? Can we just return it in flattened form?

jamieparkinson (approved):
"Group ID" sounds reasonable to me

alexwlchan:
Closing in favour of #2000.

PR #1999, opened 2021-12-21
Rejig the WorkNode model so we can simplify the way we retrieve changing graphs

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5389

The key is the new WorkNode model.

Before:

```scala
case class WorkNode(
  id: CanonicalId,
  version: Option[Int],
  linkedIds: List[CanonicalId],
  componentId: String,
  suppressed: Boolean = false
)
```

After:

```scala
case class WorkNode(
  id: CanonicalId,
  subgraphId: String,
  componentIds: List[CanonicalId],
  sourceWork: Option[SourceWorkData] = None,
)

case class SourceWorkData(
  id: SourceIdentifier,
  version: Int,
  suppressed: Boolean = false,
  mergeCandidateIds: List[CanonicalId] = List(),
)
```

Changes:

- The `subgraphId` records works that should be matched together; the `componentIds` record works that should be merged together.
- We now record the source identifier alongside the source work (which I've wanted for a while to make matcher analysis easier), and `linkedIds` has been renamed to `mergeCandidateIds`, which is what it actually is. I've also grouped all the source data, to make it more obvious whether a node corresponds to a known source work or not.

We use the new `subgraphId` to look up related works when matching, and it makes handling suppressed/unlinked works much simpler. If A and B are *ever* in the same matcher subgraph, they will *always* have the same subgraphId from that point forward, even if they're not in the same component. (Until the matcher database is reset.) It's a little inefficient, but much simpler and therefore less likely to be buggy.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/SourceIdentifier.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/identifiers/SourceIdentifier.scala
@@ -10,5 +10,5 @@ case class SourceIdentifier(
     value == value.trim,
     s"SourceIdentifier value has trailing/leading whitespace: <$value>")
 
-  override def toString = s"$ontologyType<${identifierType.id}/$value>"
+  override def toString = s"$ontologyType[${identifierType.id}/$value]"
 }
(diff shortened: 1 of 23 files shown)


jamieparkinson (approved):
Seems very straightforward (suspicious)

PR #2003, opened 2021-12-22
Don't use subfield ǂh for title if it's the last subfield

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5362


alexwlchan:
I'm going to merge by fiat because this is only a small change.

PR #2005, opened 2021-12-22
Update the matcher script to use the new matcher table format

alexwlchan:
Now we've changed the structure of the matcher DynamoDB table (for https://github.com/wellcomecollection/platform/issues/5389), we need to update the script that reads the table. I decided to rewrite it in TypeScript rather than update the Python version, because that's the clear direction of the platform.

This new version picks up a couple of nice user-facing changes as well:

* A brief README for the matcher and the script: https://github.com/wellcomecollection/catalogue-pipeline/blob/new-matcher-script/pipeline/matcher/README.md

* We now get thumbnails for METS works, if available:
 
  Screenshot 2021-12-22 at 16 34 34
 
* Suppressed works are highlighted with dashed lines, because the matcher knows about them even if it doesn't use them:

   Screenshot 2021-12-22 at 16 35 01


pipeline/matcher/scripts/getMatcherGraph.ts
```diff
@@ -0,0 +1,191 @@
...
+
+type ElasticConfig = {
+  worksIndex: string;
+  imagesIndex: string;
+}
+
+
+async function getInput(): Promise<UserInput> {
+  let id: string;
+  if (process.argv.length != 3) {
```
jamieparkinson, line 24: Never the double equals!
```suggestion
  if (process.argv.length !== 3) {
```

pipeline/matcher/scripts/getMatcherGraph.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    if (result.Items && result.Items.length > 0) {
+      allItems = [...allItems, ...result.Items]    
+    }
+
+    if (typeof result.LastEvaluatedKey === 'undefined') {
+      break;
+    }
+
+    query.ExclusiveStartKey = result.LastEvaluatedKey;
```
jamieparkinson, line 64: Can we do this without mutating arguments? We could clone the `query` or - my preference - this function could be implemented recursively

pipeline/matcher/scripts/getMatcherGraph.ts
```diff
@@ -0,0 +1,191 @@
...
+  await Promise.all(works.map(async (w: SourceWork) => {
+    let attributes = await getAttributes(w);
+
+    const newNode = g.createNode(w.canonicalId, attributes);
+
+    nodes.set(w.canonicalId, newNode);
+  }));
+
+  // Add all the edges
+  works.map((w: SourceWork) => {
```
jamieparkinson, line 136: Since we're not using the return value, let's be clear that there are side effects
```suggestion
  works.forEach((w: SourceWork) => {
```

pipeline/matcher/scripts/getMatcherGraph.ts
```diff
@@ -0,0 +1,191 @@
...
+      let attributes = {};
+      if (!w.componentIds.includes(target)) {
+        attributes['style'] = 'dotted';
+      }
+
+      g.createEdge([nodes.get(w.canonicalId), nodes.get(target)], attributes);
+    });
+  });
+
+  return g;
```
jamieparkinson, line 157: No need for this if the above is true I think

pipeline/matcher/scripts/graphAttributes.ts
```diff
@@ -0,0 +1,108 @@
...
+    return undefined;
+  }
+
+  let json: CatalogueResults = await resp.json();
+
+  return json.results.map(w => w.thumbnail?.url )?.[0];
+}
+
+async function getLabelAttributes(w: SourceWork): Promise<Record<string, string>> {
+  if (w.sourceIdentifier.identifierType == 'SierraSystemNumber') {
```
jamieparkinson, line 71: switch / case?

pipeline/matcher/scripts/graphAttributes.ts
```diff
@@ -0,0 +1,108 @@
...
+    return {label: `Sierra\n${w.sourceIdentifier.value}`};
+  }
+
+  if (w.sourceIdentifier.identifierType == 'CalmRecordIdentifier') {
+    return {label: `CALM\n${w.sourceIdentifier.value}`}
+  }
+  
+  if (w.sourceIdentifier.identifierType == 'MiroImageNumber') {
+    return await getImageAttributes(
```
jamieparkinson, line 80: There is no* difference between returning a promise and awaiting it before it's returned - this function needn't be `async` and it needn't use `awaits`.

<sup>*this is not technically correct, but the difference [only matters](https://jakearchibald.com/2017/await-vs-return-vs-return-await/#return-awaiting) in a try/catch block</sup>

pipeline/matcher/scripts/graphAttributes.ts
```diff
@@ -0,0 +1,108 @@
...
+      `Miro\n${w.sourceIdentifier.value}`,
+      async () => `https://iiif.wellcomecollection.org/thumbs/${w.sourceIdentifier.value}/full/!100,100/0/default.jpg`
+    );
+  }
+
+  if (w.sourceIdentifier.identifierType == 'METS') {
+    return await getImageAttributes(
+      w.sourceIdentifier,
+      `METS\n${w.sourceIdentifier.value}`,
+      async () => await getMetsThumbnail(w.sourceIdentifier.value),
```
jamieparkinson, line 91: redundant `async`/`await`

pipeline/matcher/scripts/graphAttributes.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+async function getMetsThumbnail(bnumber: string): Promise<string | undefined> {
+  const resp = await fetch(`https://api.wellcomecollection.org/catalogue/v2/works?identifiers=${bnumber}`);
+
+  if (resp.status !== 200) {
+    return undefined;
+  }
+
+  let json: CatalogueResults = await resp.json();
```
jamieparkinson, line 65: I know people on the internet have varying opinions on this, but in wc code I believe we only use `let` when we are going to mutate a variable (ie it is a statement of intent)

jamieparkinson (changes requested):
More than any of my other comments - let's configure prettier & eslint for this repo before adding JS/TS to it!

pipeline/matcher/scripts/getMatcherGraph.ts
alexwlchan: I'm not sure I understand what you're suggesting? The variable is declared within this function, so we have to return it for callers to get to it.

alexwlchan:
> let's configure prettier & eslint for this repo before adding JS/TS to it!

Done! ✨

pipeline/matcher/scripts/getMatcherGraph.ts
jamieparkinson: Misread the bracket matching!

jamieparkinson:
Did you copy the eslint config from the wc.org repo? I don't like that we run prettier with eslint (rather than directly) there, and the prettier team now [recommend against that](https://prettier.io/docs/en/integrating-with-linters.html#notes).

I'd be in favour of starting from scratch (or close to scratch) with the eslint config here, so as not to propagate old decisions. For example - you can remove all of the references to react from here since I doubt we'll be using it in this repo.

alexwlchan:
> Did you copy the eslint config from the wc.org repo?

Yes.

jamieparkinson:
I'm not sure... this PR does so much more than the title suggests - it's a complete rewrite in a language that's new to the repo. I don't want to forget about it, I want us to get the tooling right without propagating tech debt from elsewhere. 

It's now a bit tricky to split up, so actually perhaps my preference now is to remove the tooling (ie prettier and eslint) and come back later to add it properly?

PR #2028, opened 2022-02-02
Add some notes about the reindexer

alexwlchan:
Rendered version: https://github.com/wellcomecollection/catalogue-pipeline/blob/reindexer-docs/reindexer/README.md

Based on a conversation with Paul, this captures the key aspects of reindexing.


jamieparkinson:
Is it worth clarifying the fact that the reindexer doesn't actually send full records, but rather identifiers in the source stores?

alexwlchan:
> Is it worth clarifying the fact that the reindexer doesn't actually send full records, but rather identifiers in the source stores?

Mmm. If you want to be picky, it doesn't send identifiers either – it sends "source payloads".

We've changed the reindexer payload semi-regularly – that's a bit close to the sort of implementation detail I kinda don't want here, because it'll get out of date quickly.

PR #2033, opened 2022-02-09
Remember to lock over the IDs we learn about from the store

alexwlchan:
I had a brainwave about the matcher shenanigans.

If we process two works in a graph simultaneously, and neither work sees the other directly, we don't lock correctly and we can get inconsistent data. The test case explains this in more detail.

I went digging to find out when we changed this (August 2020, https://github.com/wellcomecollection/catalogue-pipeline/pull/787). I think it's only since reindexing has become faster that we've started to notice, because works in the same graph are more likely to be processed simultaneously – when reindexing was slower and throughput less, this failure case was much less likely.

🤞 this stems the matcher weirdness


alexwlchan:
We'll need a full reindex to see the effect of this, I think.

alexwlchan:
> What sort of increase in the number of locks is this / how is this increase distributed over works?

Not sure, and I don't know how I'd tell.

I think the C->B->A case is probably the most common one we're encountering, where to update (C) we've already locked C, B, and a before/after subgraph ID – so four locks. This would be a fifth lock, or a 25% increase.

> I remember we used to lock over all the work IDs individually and it caused major throughput issues

That's probably for the really big work graphs, isn't it?

Now I look at this again – your approach of locking just the subgraph IDs should work, shouldn't it? I wonder why it doesn't. 🤔 

Let me go away and look at this some more.

alexwlchan:
Ah, so I went ahead and checked what locks we're creating in the existing code. Here's what happens when you don't lock individual work IDs and that test fails:

```
1 @processing[A]: locking Set(A)
2 @processing[C]: locking Set(B, C)
3 @processing[A]: expanding lock Set(subgraphId:A+B)
4 @processing[A]: releasing lock Set(subgraphId:A+B)
5 @processing[A]: releasing lock Set(A)
6 @processing[C]: expanding lock Set(subgraphId:A+B, subgraphId:A+B+C)
7 @processing[C]: releasing lock Set(subgraphId:A+B, subgraphId:A+B+C)
8 @processing[C]: releasing lock Set(B, C)
```

And here's where I think the bug is:

```
1 @processing[A]: locking Set(A)
2 @processing[C]: locking Set(B, C)
     // @processing[C] reads the old value of A from the graph
3 @processing[A]: expanding lock Set(subgraphId:A+B)
     // @processing[A] writes a new value of A to the graph
4 @processing[A]: releasing lock Set(subgraphId:A+B)
5 @processing[A]: releasing lock Set(A)
6 @processing[C]: expanding lock Set(subgraphId:A+B, subgraphId:A+B+C)
     // @processing[C] replaces the new value of A with one based on the old value
7 @processing[C]: releasing lock Set(subgraphId:A+B, subgraphId:A+B+C)
8 @processing[C]: releasing lock Set(B, C)
```

This is kinda tricky to reproduce, but you can do it – see the test in the updated commit. The basic problem boils down to this:

```
A: read_graph()
C: read_graph()
A: write_graph()
C: write_graph()
```

and the graph written by C is using stale data.

It's *very* fiddly, but I think I've fixed this particular edge case. Have another look?

jamieparkinson (approved):
This is very bonkers - we really shouldn't be handling concurrency ourselves - but I agree that it fixes a failure mode. Nicely reasoned!

jamieparkinson:
This patch is a very highly concentrated expression of the matcher's general state!

PR #2042, opened 2022-02-24
Optional relation ids

paul-butcher:
This introduces the possibility that a Relation can exist without an id.  It *does not* actually make that happen.

This creates a breaking change in internal_model that is resolved in the API with this pull (once the internal_model hash has been updated) https://github.com/wellcomecollection/catalogue-api/pull/360/files

Because the API loads the data from Elasticsearch, and there are no actual Relations without ids, the change  won't  break the API until we actually start producing Relations without ids.

Part of https://github.com/wellcomecollection/platform/issues/5408

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
@@ -65,13 +65,13 @@ object Relation {
     )
 
   /**
-   * A Relation created with just a String is a Series.  The string is the title.
-   * A Series is not an entity in its own right, so does not have an id of its own.
-   *
-   * Practically, a Series does not exist in a hierarchy, so does not have depth or
-   * children or descendents, even though (obviously) there are objects that are "partOf"
-   * the series.
-   */
+    * A Relation created with just a String is a Series.  The string is the title.
+    * A Series is not an entity in its own right, so does not have an id of its own.
+    *
+    * Practically, a Series does not exist in a hierarchy, so does not have depth or
+    * children or descendents, even though (obviously) there are objects that are "partOf"
+    * the series.
+    */
   def apply(series: String): Relation =
     new Relation(None, Some(series), None, WorkType.Series, 0, 0, 0)
 
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
@@ -55,7 +55,7 @@ object Relation {
     numDescendents: Int
   ): Relation =
     Relation(
-      id = Option(id),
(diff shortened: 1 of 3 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
```diff
@@ -64,6 +64,17 @@ object Relation {
...
+  /**
+    * A Relation created with just a String is a Series.  The string is the title.
+    * A Series is not an entity in its own right, so does not have an id of its own.
+    *
+    * Practically, a Series does not exist in a hierarchy, so does not have depth or
+    * children or descendants, even though (obviously) there are objects that are "partOf"
+    * the series.
+    */
+  def apply(series: String): Relation =
+    Relation(None, Some(series), None, WorkType.Series, 0, 0, 0)
```
alexwlchan, line 76: Couple of notes:

1. Could we pull this into a dedicated object so it's clear what this is doing elsewhere in the codebase? e.g.

    ```scala
    object SeriesRelation {
      def apply(name: String): Relation = ...
    }

    // elsewhere
    val series = SeriesRelation(name = "Early Printed Books")
    ```

2. Let's add some keyword arguments so we know what these `None` and `0` values mean.

common/internal_model/src/test/scala/weco/catalogue/internal_model/work/RelationsTest.scala
```diff
@@ -0,0 +1,57 @@
...
+  it("Has zero size when empty") {
+    Relations.none.size shouldBe 0
+  }
+
+  it("Shows the total number of known relations when populated") {
+    Relations(
+      List(Relation("Granny"), Relation("Grandpa")),
+      List(Relation("Baby")),
+      List(Relation("Big Sister")),
+      List(Relation("Little Brother"))
```
alexwlchan, line 17: Keyword arguments for readability, please.

common/internal_model/src/test/scala/weco/catalogue/internal_model/work/RelationsTest.scala
```diff
@@ -0,0 +1,57 @@
+package weco.catalogue.internal_model.work
+
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+import weco.catalogue.internal_model.work.generators.WorkGenerators
+
+class RelationsTest extends AnyFunSpec with Matchers {
+  it("Has zero size when empty") {
```
alexwlchan, line 8: nit: codebase convention is that test descriptions start with a lowercase letter, e.g.

```suggestion
  it("has zero size when empty") {
```

I don't have strong feelings either way, but let's be consistent.

alexwlchan (approved):
Couple of suggestions for improving readability, but overall looks good. 👍

PR #2044, opened 2022-03-03
Series links

paul-butcher:
This change presents the four different child -> parent MARC fields as Series relationships.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
@@ -198,7 +198,7 @@ object WorkState {
 
     def id = sourceIdentifier.toString
 
-    override val modifiedTime: Instant = sourceModifiedTime
+    val modifiedTime: Instant = sourceModifiedTime
   }
 
   case class Identified(
@@ -215,7 +215,7 @@ object WorkState {
 
     def id = canonicalId.toString
 
-    override val modifiedTime: Instant = sourceModifiedTime
+    val modifiedTime: Instant = sourceModifiedTime
   }
 
   case class Merged(
@@ -234,7 +234,7 @@ object WorkState {
 
     // This is used to order updates in pipeline-storage.
     // See https://github.com/wellcomecollection/docs/tree/main/rfcs/038-matcher-versioning
-    override val modifiedTime: Instant = mergedTime
+    val modifiedTime: Instant = mergedTime
   }
 
   case class Denormalised(
@@ -253,7 +253,7 @@ object WorkState {
 
     // This is used to order updates in pipeline-storage.
     // See https://github.com/wellcomecollection/docs/tree/main/rfcs/038-matcher-versioning
-    override val modifiedTime: Instant = mergedTime
+    val modifiedTime: Instant = mergedTime
   }
 
   /** Why are there three *Time parameters?
@@ -285,7 +285,7 @@ object WorkState {
 
     def id = canonicalId.toString
 
-    override val modifiedTime: Instant = mergedTime
+    val modifiedTime: Instant = mergedTime
   }
 }
 
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
@@ -206,14 +206,14 @@ object WorkState {
     canonicalId: CanonicalId,
     sourceModifiedTime: Instant,
     mergeCandidates: List[MergeCandidate[IdState.Identified]] = Nil,
-    internalWorkStubs: List[InternalWork.Identified] = Nil
+    internalWorkStubs: List[InternalWork.Identified] = Nil,
+    relations: Relations = Relations.none
   ) extends WorkState {
 
     type WorkDataState = DataState.Identified
     type TransitionArgs = Unit
 
     def id = canonicalId.toString
-    val relations = Relations.none
 
     override val modifiedTime: Instant = sourceModifiedTime
   }
@@ -223,7 +223,8 @@ object WorkState {
     canonicalId: CanonicalId,
     mergedTime: Instant,
     sourceModifiedTime: Instant,
-    availabilities: Set[Availability] = Set.empty
+    availabilities: Set[Availability] = Set.empty,
+    relations: Relations = Relations.none
   ) extends WorkState {
 
     type WorkDataState = DataState.Identified
@@ -231,7 +232,6 @@ object WorkState {
 
     def id: String = canonicalId.toString
 
-    override val relations: Relations = Relations.none
     // This is used to order updates in pipeline-storage.
     // See https://github.com/wellcomecollection/docs/tree/main/rfcs/038-matcher-versioning
     override val modifiedTime: Instant = mergedTime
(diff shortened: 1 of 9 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
```diff
@@ -189,14 +189,14 @@ object WorkState {
     // work to be performed by the pipeline
     sourceModifiedTime: Instant,
     mergeCandidates: List[MergeCandidate[IdState.Identifiable]] = Nil,
-    internalWorkStubs: List[InternalWork.Source] = Nil
+    internalWorkStubs: List[InternalWork.Source] = Nil,
+    relations: Relations = Relations.none
```
alexwlchan, line 193: How much do you understand the different states here (Source, Identified, Merged, etc.)?

This is definitely the right place to put this value, but I think it's going to be dropped at the next stage of the pipeline – the Identified/Merged states always assume that `Relations` is going to be `none` (and the merger/relation embedder may ignore it too, although that's a bit more complicated).

I think you need to extend the Identified and Merged states additionally, then update the WorkFsm described further down in this file.

(This is another case of where the internal model is quite brittle and unwieldy, but it's what we have for now.)

alexwlchan (changes requested):
The transformer stuff is lovely, but you need to do a bit more on the internal Work state.

(There may well be other issues after you fix that, but that one at least stands out to me as something we can fix before we deploy.)

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraLinksTest.scala
```diff
@@ -0,0 +1,187 @@
...
+        content = Some("A Series; but with a space after the separator ; ")
+      )
+    )
+
+    getLinks(varFields) shouldBe List(
+      SeriesRelation(
+        "Published papers (Wellcome Chemical Research Laboratories)"),
+      SeriesRelation("A Series; but with a space after the separator")
+    )
+  }
```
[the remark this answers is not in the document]
paul-butcher: Ah good.  I was looking for that.  I want it somewhere else as well.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraParents.scala
```diff
@@ -0,0 +1,54 @@
...
+    *  All Series Relations resulting from this:
+    *  - Have a non-empty title
+    *  - Are stripped of subfield separators
+    *  - Are unique
+    */
+  def apply(bibData: SierraBibData): List[Relation] =
+    bibData
+      .varfieldsWithTags("440", "490", "773", "830")
+      .flatMap(_.content)
+      .map(title => TailSeparator.replaceFirstIn(title, "").trim())
```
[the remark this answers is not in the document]
paul-butcher: The reason I chose this approach was because I envisaged simply adding a few more characters to the expression as we discover them, but now, having looked at a lot more examples, ' ;' seems to be the suffix of choice. If any others appear, an extra stripSuffix would be clearer.

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Work.scala
paul-butcher: Yes. I think I forgot that this is used to communicate between applications by DB & queue, rather than by in-memory copies in a single application.

paul-butcher: There is something I find a bit odd here - Most of the states have some `override val` lines, which seem unnecessary, as the val in the trait is only declared and not defined.  Is there some magic I'm missing?

alexwlchan: I think the compiler complains if the values aren't defined in the case classes? But I don't recall there being any special magic here; it's possible they're unnecessary. Try removing them and let me know!

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraTransformerTest.scala
```diff
@@ -1254,6 +1254,43 @@ class SierraTransformerTest
...
+         |    {
+         |      "marcTag": "773",
+         |      "content": "Series via 773"
+         |    },
+         |    {
+         |      "marcTag": "830",
+         |      "content": "Series via 830"
+         |    }
+         |  ]
+         |}""".stripMargin
```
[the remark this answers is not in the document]
paul-butcher: Absolutely.  It wouldn't be hard to add JSON output somewhere in the scala-libs SierraDataGenerators

common/internal_model/src/test/scala/weco/catalogue/internal_model/work/StateTest.scala
```diff
@@ -0,0 +1,97 @@
...
+import org.scalatest.prop.TableDrivenPropertyChecks
+import weco.catalogue.internal_model.identifiers.IdentifierType.SierraIdentifier
+import weco.catalogue.internal_model.identifiers.{CanonicalId, SourceIdentifier}
+
+import java.time.Instant
+
+class StateTest
+    extends AnyFunSpec
+    with Matchers
+    with TableDrivenPropertyChecks {
```
alexwlchan, line 18: Couple of notes: we have a file WorkTest which has similar behaviour, and these tests probably belong in there. Good idea to add some tests here 👍

Also, there's a whole raft of traits called `{Something}Generators` that are quite useful for generating values in tests, usually when you don't care about the exact value.

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraLinksTest.scala
```diff
@@ -0,0 +1,204 @@
...
+      VarField(
+        marcTag = "773",
+        subfields = List(Subfield(tag = "t", content = "A Series"))
+      )
+    )
+    getLinks(varFields) shouldBe List(SeriesRelation("A Series"))
+  }
+
+  it("Extracts the title from the body of a 773 field, if title is absent") {
+    // This is not a scenario we expect to encounter, but
```
alexwlchan, line 60: …but? did you mean to write more here?

alexwlchan (approved):

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraLinksTest.scala
```diff
@@ -57,7 +57,8 @@ class SierraLinksTest
   }
 
   it("Extracts the title from the body of a 773 field, if title is absent") {
-    // This is not a scenario we expect to encounter, but
+    // This is not a scenario we expect to encounter, but applying
+    // Postel's Law and logging a warning is better than discarding it
```
alexwlchan, line 61: 👍

PR #2048, opened 2022-03-10
Add 2022-03-10 pipeline

paul-butcher:


pipeline/terraform/stack/dynamo.tf
```diff
@@ -40,16 +40,16 @@ resource "aws_dynamodb_table" "matcher_graph_table" {
 
   # These numbers were chosen by running a reindex and seeing when the
   # matcher started throttling.
-  read_capacity  = local.graph_table_billing_mode == "PROVISIONED" ? 600 : 1
-  write_capacity = local.graph_table_billing_mode == "PROVISIONED" ? 1000 : 1
+  read_capacity  = local.graph_table_billing_mode == "PROVISIONED" ? 600 : 0
+  write_capacity = local.graph_table_billing_mode == "PROVISIONED" ? 1000 : 0
```
jamieparkinson, line 44: This was a bit confusing until I realised that `graph_table_billing_mode` becomes `PAY_PER_REQUEST` when `scale_up_matcher_db` is `false` - it might be clearer if we made the ternary use that boolean directly? I know this code was here before but just a thought...

jamieparkinson (approved):

pipeline/terraform/stack/dynamo.tf
paul-butcher: Possibly. 

The reason I changed it to 0 was because that is the value terraform sees when it should be a NOOP.

PR #2053, opened 2022-03-11
Provide a better message for an "impossible" scenario in the ID minter

alexwlchan:
The ID minter is the app that creates the canonical IDs we use in works URLs, e.g. in /works/k6sy9zz3, the canonical ID is `k6sy9zz3`. Each canonical ID has a 1:1 correspondence with an ID from a source system. The canonical–source ID mappings are stored in an RDS database.

When it runs, the ID minter asks the database _"please give me any canonical IDs you know about for X, Y, Z"_. It uses any that already exist, then creates new canonical IDs for any source IDs that haven't been seen before. So far, so simple.

What happens when the database hands back canonical IDs for P, Q, and R? This should be impossible in practice (and we wrote a comment to that effect), but it is in fact possible – because the database queries are case insensitive. Consider:

> ID minter: Please give me the canonical ID for "METS b12345678"
> Database: Here's the canonical ID for "METS B12345678"
> ID minter: This isn't what I asked for 😱 💥 

Although in general we use lowercase b-identifiers for METS files, there are a handful where we've used uppercase B in the past. When these get corrected, the ID minter is unhappy.

It's rare enough that I don't (yet) want to put in proper logic to fix this automatically, but I can put in a better error message so it's easier for the next person to debug.

I've also put in some notes about how to connect to the ID minter database, which is a thing I do relatively infrequently and have to re-derive from scratch every time. These instructions probably have some gaps, but they're a starting point.

This fixes the issue Ashley was having in Slack: https://wellcome.slack.com/archives/C8X9YKM5X/p1646995044309189

### DIFF
--- a/builds/publish_image_with_weco_deploy.sh
+++ b/builds/publish_image_with_weco_deploy.sh
@@ -43,7 +43,7 @@ docker run --tty --rm \
   --env AWS_PROFILE \
   --volume ~/.aws:/root/.aws \
   --volume /var/run/docker.sock:/var/run/docker.sock \
-  --volume "${DOCKER_CONFIG:-~/.docker}:/root/.docker" \
+  --volume "${DOCKER_CONFIG:-$HOME/.docker}:/root/.docker" \
   --volume "$ROOT:$ROOT" \
   --workdir "$ROOT" \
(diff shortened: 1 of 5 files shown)


jamieparkinson (approved):
Nice error message

PR #2054, opened 2022-03-17
Extract hierarchies from Sierra data

paul-butcher:
This change sets a CollectionPath to make use of the relation embedder, in order to turn 773 and 774 fields into hierarchical partOf relationships.


jamieparkinson (approved):
As a general implementation/style note I might be inclined to "flip around" a lot of the logic present in `SierraCollectionPath` so that it didn't do so much repeated checking for the presence of data and rather relied more on the monadic wizardry of an `Option`'s `map`/`flatMap`/etc, less on `if`/`else`s. 

Also, the comment is very very good

PR #2056, opened 2022-03-23
Fix: Router seems unreliable

paul-butcher:
See: https://github.com/wellcomecollection/platform/issues/5463

The manifestation of this problem is that the router_input_dlq grows.

The cause is that the queue message timeout and the batch flush interval are the same, so it only succeeds when it's busy, and fails when it's quiet.

### DIFF
--- a/common/pipeline_storage/src/main/scala/weco/pipeline_storage/PipelineStorageStream.scala
+++ b/common/pipeline_storage/src/main/scala/weco/pipeline_storage/PipelineStorageStream.scala
@@ -14,6 +14,39 @@ import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future, TimeoutException}
 import scala.util.Try
 
+/**
+ * Configuration for the processing of messages in batches.
+ *
+ * Messages will be processed when either batchSize or flushInterval are reached.
+ *
+ * The intended use for this is where it can be significantly more efficient to process
+ * multiple messages together, rather than one at a time.  For example where
+ * downloading all the data to service a bundle of messages incurs less overhead than
+ * making individual requests for the data to service each message.
+ *
+ * ==Setting batchSize and flushInterval==
+ *
+ * The numbers for batchSize and flushInterval should be chosen based on how responsive
+ * you want the service to be and how many message the service can realistically handle
+ * in one go, as well as the expected frequency with which messages are likely to be received.
+ *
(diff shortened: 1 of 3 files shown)


pipeline/relation_embedder/router/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+```
+
+## Why?
+
+Unlike other pipeline stages, the relation embedder does not operate on the document currently 
+passing through the pipeline, but on *all* documents with which that document has a relationship.
+
+Because of this, the message sent to the embedder is the *content of the collectionPath*, rather 
+than a way to identify the document in question.
```
alexwlchan, line 41: Possibly worth noting that the relation embedder is moderately expensive (it's an ES scroll) and will pass through any works without a collectionPath unmodified, so we want to skip it rather than having it get a work and work on its collectionPath.

pipeline/terraform/stack/service_work_router.tf
```diff
@@ -31,7 +31,7 @@ module "router" {
     es_merged_index        = local.es_works_merged_index
     es_denormalised_index  = local.es_works_denormalised_index
     batch_size             = 100
-    flush_interval_seconds = 30
+    flush_interval_seconds = 20
```
alexwlchan, line 34: Let's link to the ticket/your debugging, so we don't introduce this error again.

common/pipeline_storage/src/main/scala/weco/pipeline_storage/PipelineStorageStream.scala
```diff
@@ -14,6 +14,39 @@ import scala.concurrent.duration._
...
+  *
+  * The intended use for this is where it can be significantly more efficient to process
+  * multiple messages together, rather than one at a time.  For example where
+  * downloading all the data to service a bundle of messages incurs less overhead than
+  * making individual requests for the data to service each message.
+  *
+  * ==Setting batchSize and flushInterval==
+  *
+  * The numbers for batchSize and flushInterval should be chosen based on how responsive
+  * you want the service to be and how many message the service can realistically handle
```
alexwlchan, line 30:
```suggestion
  * you want the service to be and how many messages the service can realistically handle
```

alexwlchan (approved):
Really useful comments, thanks. 👍

PR #2057, opened 2022-03-24
Sierra hierarchy snags

paul-butcher:
This fixes the problem where a child in the hierarchy also gets a series link tag in the UI.

(requires a reindex to actually fix)


alexwlchan (approved):
🧧

alexwlchan:
> requires a reindex to actually fix

Once this is deployed, it's possible to do an in-pipeline reindex, by connecting the transformers to the reindexer topic and then running the "start reindex" script. It only trickles through because it can't scale the tasks to more than 1 at a time, but it does get the change applied.

(Although often we will just queue up changes for the next reindex if they're not urgent.)

PR #2058, opened 2022-03-30
Fix same title hierarchy

paul-butcher:
This is the fix for https://github.com/wellcomecollection/platform/issues/5472

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
@@ -26,19 +26,20 @@ case class Relations(
 
   def +(that: Relations): Relations = {
     Relations(
-      ancestors = RelationSet(this.ancestors ::: that.ancestors),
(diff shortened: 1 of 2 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Relations.scala
```diff
@@ -59,18 +60,11 @@ case class Relations(
...
-    val indexMap = relations.map(_.title).distinct.zipWithIndex.toMap
-    relations
-      .groupBy(_.title)
-      .values
-      .map(_.maxBy(_.id))
-      .toList
-      .sortBy(relation => indexMap(relation.title))
+  def apply(existingRelations: List[Relation],
+            newRelations: List[Relation]): List[Relation] = {
+    val newTitles = newRelations.map(_.title).distinct
```
jamieparkinson, line 65: nit - not sure you need the `distinct`? Would that we had `distinctBy` in 2.12 for this whole block

jamieparkinson (approved):

PR #2071, opened 2022-04-21
Add some models to use in the works ingestor

alexwlchan:

This is a suggestion for models for the first pass of this work. It's a forked version of `Work[WorkState.Indexed]`, starting to rearrange the fields towards something closer to what we want.

* The new `DebugInformation` case class contains information meant for devs to read, but which is of no value for displaying/searching works (e.g. the source identifier/version, or why a record was deleted). This will be stored in the `"debug"` field.

   Currently the API is serialising these fields when it gets a document from Elasticsearch (and I think we index them), but I think it should completely ignored by applications.

   For now I'm going to define/construct it in the ingestor to keep this patch manageable, but there's no reason we couldn't move it forward in future (because everything it contains will be written by the transformer, and doesn't change after that except maybe when merging).

* The `display` field points to a `Json` model; this is an exact representation of how it should be emitted by the API, e.g. no nulls.

* The `IndexedWork` type mirrors the `Work[WorkState.Indexed]` type, but I've moved some of the top-level fields into the `debug` field. The `data` and `state` fields are left as-is, so the meat of the existing API should continue to work.

* The `IndexedImage` type mirrors the `Image[ImageState.Indexed]` type, with the new `display` field. I haven't added a `debug` field yet because I'm not as savvy about what goes into images and what we'd need there.

   (We probably want to review how we do images more thoroughly; the only reason I've included them is so I get the DisplayImage logic in at the same time – it behaves a little differently with some of the includes, and I want to make sure I don't over-delete display code.)

### DIFF
--- a/pipeline/ingestor/ingestor_works/src/main/scala/weco/pipeline/ingestor/works/models/IndexedWork.scala
+++ b/pipeline/ingestor/ingestor_works/src/main/scala/weco/pipeline/ingestor/works/models/IndexedWork.scala
@@ -1,0 +1,36 @@
+package weco.pipeline.ingestor.works.models
+
+import weco.catalogue.internal_model.identifiers.{DataState, IdState}
+import weco.catalogue.internal_model.work.{WorkData, WorkState}
+
+sealed trait IndexedWork
+
+object IndexedWork {
+  case class Visible(
+    data: WorkData[DataState.Identified],
+    state: WorkState.Indexed,
+
+    debug: DebugInformation.Visible,
+    display: DisplayWork
+  ) extends IndexedWork
+
+  case class Redirected(
+    redirectTarget: IdState.Identified,
+    state: WorkState.Indexed,
+
+    debug: DebugInformation.Redirected
+  ) extends IndexedWork
+
+  case class Invisible(
+    data: WorkData[DataState.Identified],
+    state: WorkState.Indexed,
+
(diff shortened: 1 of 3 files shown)


pipeline/ingestor/ingestor_works/src/main/scala/weco/pipeline/ingestor/works/models/IndexedWork.scala
```diff
@@ -0,0 +1,32 @@
...
+object IndexedWork {
+  case class Visible(
+    data: WorkData[DataState.Identified],
+    state: WorkState.Indexed,
+    debug: DebugInformation.Visible,
+    display: DisplayWork
+  ) extends IndexedWork
+
+  case class Redirected(
+    redirectTarget: IdState.Identified,
```
paul-butcher, line 17: All of these have state and debug, then some other parameters.  I think it would be nice to have those two first, then the other parameters after, for consistency.

alexwlchan: 👍 I wrote it this way because originally I had newlines to mark where I'd forked form the Work[WorkState.Indexed] model, but apparently autoformatting has removed those. I'll rearrange them.

alexwlchan:
Ack, what did I do

PR #2074, opened 2022-04-21
Store the new IndexedWork/Image models in the ingestor

alexwlchan:

This realises the models defined in #2071. I've copied across the display models from the catalogue-api repo, and removed the bits we don't need here (e.g. items will always get identifiers if available, whereas the catalogue-api repo adds/removes them based on the user-defined includes).

Then I've updated the works and images ingestor to use the new models, updated the index mappings to support them, and updated the tests to get them passing.


alexwlchan:
Closing as superseded.

PR #2076, opened 2022-04-25
Store the new models in the ingestor

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5449; follows #2071

This patch wires up the ingestors to store the new display model. This will immediately break the API if deployed, so we should only merge it when we're ready to reindex it.


alexwlchan:
Go go go!

PR #2082, opened 2022-04-28
Create JSON examples of what's in the API index that can be consumed by the API repo

alexwlchan:

We've drastically changed the structure of the API index, and the way the API should use it – so rather than continuing to replicate internal model between the pipeline/API repos, I want to investigate breaking that link.

This PR introduces a new trait `ApiFixtures`, with a model that defines what a JSON fixture looks like:

```scala
case class Fixture(
  description: String,
  createdAt: Instant = Instant.now(),
  id: String,
  document: Json
)
```

These fixtures get serialised as JSON in the `fixtures` folder (name tbc), and then in the API tests there are helpers that will index these JSON files into Elasticsearch (specifically, the `document` property).

We'll need way more fixtures than this to get the API tests passing, but hopefully this is enough to illustrate the idea. I want to get a sense of whether others think this is a sensible approach before digging into it.

### DIFF
--- a/fixtures/list-of-works.0.json
+++ b/fixtures/list-of-works.0.json
@@ -1,0 +1,136 @@
+{
+  "description" : "one of a list of works",
+  "createdAt" : "2022-04-28T13:28:29.048Z",
+  "id" : "equ81tpb",
+  "document" : {
+    "debug" : {
+      "source" : {
+        "identifier" : {
+          "identifierType" : {
+            "id" : "miro-image-number"
+          },
+          "ontologyType" : "Work",
+          "value" : "dgZfc8BAUa"
+        },
+        "version" : 46,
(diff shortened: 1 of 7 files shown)


alexwlchan:
> Bit of a bikeshed but I would be in favour of the naming `snapshots` over `fixtures`

I don't love the name "snapshots", because it already has a meaning in the context of our Elasticsearch indexes, but agree `fixtures` is not good either.

jamieparkinson:
Is it not pretty similar to that use of the term?

alexwlchan:
Or maybe `mock_documents`?

jamieparkinson:
+1 for `test_documents`

alexwlchan:
Superseded by other PRs.

PR #2085, opened 2022-04-29
Create test documents that can be used in the API tests

alexwlchan:

This adds a new test class to the works ingestor, which creates "test documents" as JSON files. These files can then be copied into the API repo to use in tests.

This may not be the final set of examples (in fact I'm sure it won't be), but it illustrates the general approach and I think this is suitable for merging on its own. The JSON files are so massive they cause issues in GitHub's diff viewer, so I'll do a separate PR that's just adding those files.


pipeline/ingestor/ingestor_works/src/test/scala/weco/pipeline/ingestor/works/CreateTestWorkDocuments.scala
```diff
@@ -0,0 +1,379 @@
...
+    saveDocuments(documents)
+  }
+
+  implicit class WorkOps(work: Work[WorkState.Denormalised]) {
+    def toDocument: Json = {
+      // This is a fixed date so we get consistent values in the indexedTime
+      // field in the generated documents.
+      val transformer = new WorkTransformer {
+        override protected def getIndexedTime: Instant =
+          Instant.parse("2020-10-15T15:51:00.00Z")
```
jamieparkinson, line 373: Why this date?

pipeline/ingestor/ingestor_works/src/test/scala/weco/pipeline/ingestor/works/CreateTestWorkDocuments.scala
```diff
@@ -0,0 +1,379 @@
...
+    id: String
+  ): Unit =
+    saveWorks(works = List(work), description, id)
+
+  private def saveWorks(
+    works: Seq[Work[WorkState.Denormalised]],
+    description: String,
+    id: String
+  ): Unit = {
+    val documents = if (works.length == 1) {
```
jamieparkinson, line 342: Nit, don't like ifs

```scala
works match {
  case List(work) => ???
  case works => ???
}
```

jamieparkinson (approved):

pipeline/ingestor/ingestor_works/src/test/scala/weco/pipeline/ingestor/works/CreateTestWorkDocuments.scala
alexwlchan: It's somewhat arbitrary; there's code somewhere else that hard-codes an Instant. I'll change it so it's something obviously not meaningful.

PR #2086, opened 2022-05-03
Path concatenator

paul-butcher:
This introduces the path concatenator.

Previously, a record of Sierra origin would only know about itself and its parent.

So where we had three Sierra records that would form a hierarchy a/b/c, they would have had the paths a, a/b and b/c.  Now that b/c path will be resolved to a/b/c

Done: 

*  the string manipulation bit.  
* The database read bit
* The overall control bit - don't do anything if unnecessary
* The database write bit.
* Wiring in to pipeline
   * New terraform/docker stuff to run it
   * New router option
   * Notify downstream
* Wiring it into buildkite
*  Ensure transformer translates both parent and child relationships when required


pipeline/relation_embedder/path_concatenator/README.md
```diff
@@ -0,0 +1,54 @@
+# Path Concatenator
+
+## What?
+
+Unlike records from other systems each store their full path to the root of their hierarchy,
```
alexwlchan, line 5: Missing word:

```suggestion
Unlike records from other systems which each store their full path to the root of their hierarchy,
```

pipeline/relation_embedder/path_concatenator/README.md
```diff
@@ -0,0 +1,54 @@
...
+    * collectionPath is a path_hierarchy, so in the example above, this will match any records with a path that start with `leaf`
+5. Replace the first segment in those collectionPath, with this record's collectionPath
+    * e.g. a path `leaf/1/2` would become `root/branch/leaf/1/2`
+6. Notify downstream (batcher) of all changed paths.
+7. Notify downstream (batcher) of the current path if it is unchanged.
+
+This is a deliberately ordered list of steps, so that the current record *first* gets its own path embellished, *then*
+goes on to embellish the paths of its children with that new path.
+
+When writing to the database, pay attention to version.  If it is not what is expected,
```
alexwlchan, line 37: What do you mean by `version` here? It's a bit of an overloaded term and I'm not convinced it's as important as we think.

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/ChildWork.scala
```diff
@@ -0,0 +1,67 @@
...
+    * This must be called with two Works with compatible collectionPath values.
+    * The last node of the parent path must match the first node of the child path.
+    *
+    * Because of the iterative nature of this stage, it may be that, when this
+    * is executed, the collectionPath on the child contains more than just parent/self,
+    * and the "parent" is not the direct parent of self, but a further ancestor.
+    * i.e. because it has already been executed once to resolve the grandparent/parent
+    * part (and possibly again to add great-grandparent etc.)
+    *
+    * In that case it will still merge the paths as expected,
```
alexwlchan, line 26: Comma – is something missing here?

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathOps.scala
```diff
@@ -0,0 +1,15 @@
+package weco.pipeline.path_concatenator
+
+object PathOps {
+  implicit class StringOps(path: String) {
```
alexwlchan, line 4: We should combine this with [PathOps](https://github.com/wellcomecollection/catalogue-pipeline/blob/92193049aefe60f4e44b21bf35fcede2ee4291ca/pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/models/PathOps.scala) in the relation embedder, put them both in internal model and share them among applications.

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathsModifier.scala
```diff
@@ -0,0 +1,110 @@
...
+      .getParentPath(path)
+      .runWith(Sink.seq[String])
+      .map { parentPaths =>
+        info(s"Received ${parentPaths.size} parents")
+        // Only return the head if the list has exactly one path in it.
+        // If there are more, then there is a data error.
+        if (parentPaths.size <= 1) parentPaths.headOption
+        else
+          throw new RuntimeException(
+            s"More than one Work exists with the path $path")
```
alexwlchan, line 81: A pattern match would be more idiomatic Scala, something like:

```scala
parentPaths match {
  case Nil => None
  case Seq(singlePath) => Some(singlePath)
  case multiplePaths =>
    throw new RuntimeException(
      "More than one work exists with path $path: $multiplePaths"
    )
}
```

(and let's log what the distinct values are)

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathsModifier.scala
```diff
@@ -0,0 +1,110 @@
...
+    works map { work: Work.Visible[Merged] =>
+      ChildWork(parentPath, work)
+    }
+
+  def getParentPath(path: String): Future[Option[String]] = {
+    pathsService
+      .getParentPath(path)
+      .runWith(Sink.seq[String])
+      .map { parentPaths =>
+        info(s"Received ${parentPaths.size} parents")
```
alexwlchan, line 75:
```suggestion
        info(s"Received ${parentPaths.size} parents of $path")
```

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathsModifier.scala
```diff
@@ -0,0 +1,110 @@
...
+      .runWith(Sink.seq[Work.Visible[Merged]])
+      .map { currentWork =>
+        info(s"Received ${currentWork.size} works")
+        // Only return the head if the list has exactly one path in it.
+        // If there are more, then there is a data error.
+        if (currentWork.size == 1) currentWork.head
+        else
+          throw new RuntimeException(
+            s"More than one Work exists with the path $path")
+      }
```
alexwlchan, line 97: Similar comments to above.

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathsModifier.scala
```diff
@@ -0,0 +1,110 @@
...
+      case Some(parentPath) =>
+        getWorkWithPath(path) flatMap { work =>
+          Future(Some(ChildWork(parentPath, work)))
+        }
+      case _ => Future.successful(None) // This is expected, if parent is root
+
+    }
+
+  def modifyChildPaths(path: String): Future[Seq[Work.Visible[Merged]]] =
+    getWorksUnderPath(path) flatMap { works: Seq[Work.Visible[Merged]] =>
```
alexwlchan, line 60: Why do we need a flatMap here? Wouldn't a vanilla map be sufficient?

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathsService.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+    * Add a scroll to the given request and return an Elasticsearch Publisher
+    *
+    * The requests made by this service are expected to yield very few results
+    * getParentPath and getWorkWithPath should only ever return one result each.
+    *
+    * Setting up scroll context in Elasticsearch would appear to be an unnecessary
+    * overhead for this process, However, the Akka Source/publisher way of loading
+    * things requires a scroll query.
```
alexwlchan, line 36: Do we need to use the Akka Source/publisher approach? Is there any case where a vanilla query with a suitably large `size` parameter wouldn't work?

alexwlchan: (Good comment either way 👍)

pipeline/relation_embedder/path_concatenator/README.md
paul-butcher: I mean version as in here: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#docs-index-api-query-params

I agree that it's unlikely to be a problem.

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathOps.scala
paul-butcher: I agree, and that its why I chose to keep the same name, so that it would be just a matter of changing the import path once they were merged.

I was reluctant to make internal-only changes in internal model while the version check still exists to keep pipeline and API in harmony, and was wondering if there is a better place to put common pipeline-only code like this.

alexwlchan (commented):
I haven't finished reading this, but I'm about to wrap up for the day.

Mostly looks good so far – the logic is sound, the documentation is nice – but some places where the naming or control flow is a little unclear or unidiomatic. Let's make sure we sharpen those up to make it really clear what's happening.

pipeline/relation_embedder/path_concatenator/src/test/scala/weco/pipeline/path_concatenator/PathConcatenatorWorkerServiceTest.scala
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+  * Tests covering the Path Concatenator Worker Service, which
+  * responds to SQS messages to change records in a given Elasticsearch database
+  *
+  * These tests require running instances of
+  * - ElasticSearch
+  * - Localstack
+  *   - docker run --env SERVICES=sqs -p4566:4566 localstack/localstack:0.12.20
+  */
```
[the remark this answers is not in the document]
paul-butcher: I added it because I have found this out after the fact when running the tests in other services, by seeing suites go all red in IntelliJ.  I thought that a reminder at the top of the suite might be handy.

pipeline/terraform/stack/service_work_router.tf
```diff
@@ -5,6 +5,13 @@ module "router_path_output_topic" {
   role_names = [module.router.task_role_name]
 }
 
+module "router_path_concatenator_output_topic" {
+  source = "../modules/topic"
+
+  name       = "${local.namespace}_router_path_concatenator_output"
+  role_names = [module.router.task_role_name]
```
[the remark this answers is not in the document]
paul-butcher: This was essentially what caused that Slack conversation about "pull from upstream" topics.  I've given it a clearer name.

pipeline/relation_embedder/path_concatenator/src/main/scala/weco/pipeline/path_concatenator/PathConcatenatorRequestBuilder.scala
```diff
@@ -0,0 +1,81 @@
...
+    * e.g.
+    * a/b/c is the parent of c/d/e
+    *
+    * There _should_, at most, be one record that matches this query,
+    * and callers should act accordingly.
+    *
+    * However, because there is a possibility that errors in the data might
+    * cause this query to yield more than one result, this request leaves
+    * the size constraint set at the default (10), in order to allow the caller
+    * to examine the number of results returned and take appropriate action.
```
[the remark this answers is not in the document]
paul-butcher: Yes. This is an out of date comment.  Originally, this returned multiple and left it to the caller to work out whether it was too many.  I then realised there was a better way.  It is still valuable to return more than one if available, for logging

pipeline/relation_embedder/router/src/main/scala/weco/pipeline/router/RouterWorkerService.scala
```diff
@@ -43,7 +45,11 @@ class RouterWorkerService[MsgDestination](
       case None =>
         Success(List(work.transition[Denormalised](Relations.none)))
       case Some(CollectionPath(path, _)) =>
-        pathsMsgSender.send(path).map(_ => Nil)
+        work.sourceIdentifier.identifierType match {
+          case SierraSystemNumber =>
+            pathConcatenatorMsgSender.send(path).map(_ => Nil)
+          case _ => pathsMsgSender.send(path).map(_ => Nil)
+        }
```
alexwlchan, line 52: nit: how about introducing an intermediate variable to highlight the common parts:

```suggestion
val sender = work.sourceIdentifier.identifierType match {
  case SierraSystemNumber => pathConcatenatorMsgSender
  case _ => pathsMsgSender
}

sender.send(path).map(_ => Nil)
```

pipeline/relation_embedder/path_concatenator/src/test/scala/weco/pipeline/path_concatenator/PathsServiceTest.scala
```diff
@@ -0,0 +1,193 @@
...
+                      childPath: String): Future[Option[String]] =
+    service.getParentPath(childPath)
+
+  def queryWorkWithPath(service: PathsService,
+                        path: String)(): Future[Work[Merged]] =
+    service.getWorkWithPath(path)
+
+  def queryChildWorks(service: PathsService,
+                      path: String): Future[Seq[Work[Merged]]] =
+    service.getChildWorks(path) // TODO: Add a test to show this works for more than 10 children
```
alexwlchan, line 191: Have you added this test, or is this a TODO for a separate PR?

pipeline/relation_embedder/path_concatenator/src/test/scala/weco/pipeline/path_concatenator/PathsModifierTest.scala
```diff
@@ -0,0 +1,179 @@
...
+    withLocalMergedWorksIndex { index =>
+      insertIntoElasticsearch(index, works: _*)
+      testFunction(modifier(index))
+    }
+
+  private def assertDoesNothing(worksInDB: List[Work[Merged]],
+                                path: String): Assertion =
+    withContext(worksInDB, { pathsModifier =>
+      whenReady(pathsModifier.modifyPaths(path)) { resultWorks =>
+        resultWorks shouldBe empty
```
alexwlchan, line 51: nit:

```suggestion
      whenReady(pathsModifier.modifyPaths(path)) {
        _ shouldBe empty
```

pipeline/relation_embedder/path_concatenator/src/test/scala/weco/pipeline/path_concatenator/PathConcatenatorWorkerServiceTest.scala
```diff
@@ -0,0 +1,160 @@
...
+      case (workId, work) =>
+        (workId, work.data.collectionPath.get.path)
+    } should contain theSameElementsAs pathMap
+
+  private def assertQueueContainsPaths(sender: MemoryMessageSender,
+                                       expectedPaths: List[String]) =
+    sender.messages.map(_.body) should contain theSameElementsAs expectedPaths
+
+  private def storeWorks(index: Index, works: List[Work[Merged]]): Assertion =
+    insertIntoElasticsearch(index, works: _*)
```
alexwlchan, line 126: I'm not a massive fan of this sort of very light helper, because it adds a layer of obfuscation – every test will have a slightly differently named method doing the same thing, and I find it harder to follow.

Not a must change, just a note.

alexwlchan (approved):
Some minor comments, but overall looking pretty good now.

I should have been more on top of this as you went along; breaking this into smaller PRs of individually mergeable units might have been easier to review (e.g. adding a skeleton app, then the models, then the service workers) – but I should have suggested that when this PR started to grow.

pipeline/relation_embedder/path_concatenator/src/test/scala/weco/pipeline/path_concatenator/PathsServiceTest.scala
paul-butcher: Sorry, yes, I had a couple of TODOs that I did, but forgot to actually remove the todo.

pipeline/relation_embedder/path_concatenator/src/test/scala/weco/pipeline/path_concatenator/PathsServiceTest.scala
```diff
@@ -0,0 +1,193 @@
...
+        work(path = "grandfather/parent"),
+        work(path = "parent/child")
+      )
+      withLocalMergedWorksIndex { index =>
+        insertIntoElasticsearch(index, works: _*)
+
+        queryParentPath(
+          service(index),
+          childPath = "parent/child"
+        ).failed.futureValue shouldBe a[RuntimeException]
```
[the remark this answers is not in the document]
paul-butcher: Thanks, I did look for an example of this to see what the convention was in the codebase, but somehow didn't find it.

pipeline/relation_embedder/path_concatenator/src/test/scala/weco/pipeline/path_concatenator/PathConcatenatorWorkerServiceTest.scala
paul-butcher: I agree.  This was actually copied from the Relation Embedder, in an attempt at fostering consistency.

PR #2092, opened 2022-05-16
Add aggregatable values to works, like images

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5530

This follows the pattern established for images. I haven't tested it with the API yet, but I assume this will work.

### DIFF
--- a/pipeline/ingestor/ingestor_common/src/main/scala/weco/pipeline/ingestor/common/models/AggregatableValues.scala
+++ b/pipeline/ingestor/ingestor_common/src/main/scala/weco/pipeline/ingestor/common/models/AggregatableValues.scala
@@ -66,7 +66,8 @@ trait AggregatableValues {
       workData.production
         .flatMap(_.dates)
         .flatMap(_.range)
-        .map(range => LocalDateTime.ofInstant(range.from, ZoneId.systemDefault()).getYear)
+        .map(range =>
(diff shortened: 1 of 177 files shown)


pipeline/ingestor/ingestor_common/src/main/scala/weco/pipeline/ingestor/common/models/AggregatableValues.scala
```diff
@@ -0,0 +1,114 @@
...
+  *
+  * See https://github.com/wellcomecollection/docs/tree/main/rfcs/049-catalogue-api-aggregations-modelling
+  *
+  */
+trait AggregatableValues {
+  implicit class WorkDataOps(workData: WorkData[DataState.Identified]) {
+    def genreAggregatableValues: List[String] =
+      workData.genres
+        .map(DisplayGenre(_, includesIdentifiers = false))
+        .asJson(_.update("concepts", Json.fromValues(List())))
```
alexwlchan, line 28: This was a bug in the first version of this code – it would preserve the `concepts` field, so we'd (in theory) be aggregating over genres plus concepts, which isn't what we want.

The existing behaviour is to have an empty list of concepts in the genre aggregation; arguably it's not useful and we could look at changing it, but I don't want to be changing the external behaviour at the same time as internal refactoring.

jamieparkinson (approved):

PR #2098, opened 2022-05-30
Add the index config to support querying by subject identifier

alexwlchan:
The new `WorkQueryableValues` is where we'll store values we want to query on the work. It does flatten it out, but this is very similar to what Elasticsearch is already doing internally – e.g. we lose the connection between a subject ID and its label, but we'll keep it in the `display` field, and we can't query across them right now.

I'm torn about how to name the fields. `subjects.label` seems pretty clear cut, but the ID and identifiers field less so.

1. `subjects.id` and `subjects.identifiers.value` is more explicit about where we get that value on the work
2. `subjects` and `subjects.identifiers` matches what we do in API filters

I'm drawn to (1) because we'll use this object for anything we want to query, including values that don't support filtering (e.g. `title`), so I don't want it to be driven by the needs of filtering – but maybe having them match means the API can have even less knowledge of filtering, and just prepend `query.*` to user queries?

Suggestions welcome.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
@@ -227,6 +227,14 @@ object WorksIndexConfig extends IndexConfigFields {
       // to index it -- it's just an arbitrary blob of JSON.
       val display = objectField("display").withEnabled(false)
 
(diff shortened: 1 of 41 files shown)


jamieparkinson (approved):

jamieparkinson:
I think (1) is good - I think it's fine and even good for the API to have knowledge of filtering; we are doing more than providing a proxy to elasticsearch

PR #2110, opened 2022-06-06
Extract the title from the $s subfield for 773 series links

paul-butcher:
https://github.com/wellcomecollection/platform/issues/5526

Because of the different meanings of $s between 773 and 830, this has forced titleFromVarField to become a bit more specific in what it does for different fields.


pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraLinksTest.scala
```diff
@@ -150,7 +202,7 @@ class SierraLinksTest
         ),
         VarField(
           marcTag = Some(marcTag),
-          subfields = List(Subfield(tag = "t", content = "A Host"))
+          subfields = List(Subfield(tag = "a", content = "A Host"))
```
paul-butcher, line 205: because all of the fields can have $a, but not all can have $t

alexwlchan (approved):
🎫

PR #2129, opened 2022-06-17
Remove the now-unused derivedData field from images

alexwlchan:
We only use derivedData for two things:

* Getting the thumbnail for `DisplayImage`. This made more sense when the DisplayImage model was in a separate repo, and we wanted to keep the API as lightweight as possible; now it's in the same repo we can inline that logic in DisplayImage.

   It gets called at the same time (when an image is indexed in the `images-indexed` index), so it's a very small change.

* The source contributor agents, which we were using as a hacky form of aggregations, e.g. storing `Person:Henry Wellcome` and having API logic to unpack that. Now we have the `aggregatableValues` object storing full JSON strings, and the API is reading that instead, we can ditch this particular field.


alexwlchan:
This is a consequence of https://github.com/wellcomecollection/platform/issues/5550

common/display_model/src/test/scala/weco/catalogue/display_model/image/DisplayImageTest.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.scalatest.matchers.should.Matchers
+import weco.catalogue.display_model.locations.DisplayDigitalLocation
+import weco.catalogue.internal_model.generators.ImageGenerators
+
+class DisplayImageTest extends AnyFunSpec with Matchers with ImageGenerators {
+  it(
+    "sets the thumbnail to the first iiif-image location it finds in locations") {
+    val imageLocation = createImageLocation
+    val image = createImageDataWith(
+      locations = List(createManifestLocation, imageLocation)).toIndexedImage
```
paul-butcher, line 13: Is it important that it's the first iiif-image location?

Should this location list have more than one location in it to ensure that it does find the first?

alexwlchan: Good suggestion, will change.

paul-butcher (approved):

PR #2131, opened 2022-06-18
Add the 2022-06-18 pipeline

alexwlchan:
On Friday I tried to reindex into a 2022-06-17 pipeline, but I made it live prematurely and the API became flaky. We rolled back to the previous, known good pipeline, and I kicked off a fresh reindex 2022-06-18 with some safety guards to prevent making the same mistakes as on 2022-06-17.

As of Saturday afternoon:

* prod = 2022-06-09 ✅ 
* stage = 2022-06-17 ❌

This should allow us to roll forward into a good index.


pipeline/terraform/stack/elastic_pipeline.tf
```diff
@@ -10,7 +10,7 @@ locals {
   #
   # Once we stop reindexing and make the pipeline live, we want it to be
   # highly available, because it's serving API traffic.
-  es_node_count = var.reindexing_state.scale_up_elastic_cluster ? 1 : 3
+  es_node_count = var.reindexing_state.scale_up_elastic_cluster ? 2 : 3
```
alexwlchan, line 13: This shouldn't cost us much extra, but it means this cluster will be continuously available during resizing.

This is useful if we ever need to change the size of a cluster after we start using it for live traffic; not doing this is what's scuppered the 2022-06-17 pipeline. I wanted to turn the cluster back up after it was live, but I couldn't do that without causing an outage.

pipeline/terraform/stack/variables.tf
```diff
@@ -9,7 +9,7 @@ variable "min_capacity" {
 
 variable "max_capacity" {
   type        = number
-  default     = 10
+  default     = 12
```
alexwlchan, line 12: Moar power.

This number is chosen somewhat conservatively to avoid overwhelming the ES/ID minter clusters; with the 2×58GB Elasticsearch instances this was fine for the 2022-06-18 index.

This index also put less load on the `images-indexed` index (see https://github.com/wellcomecollection/catalogue-pipeline/pull/2130), which might have been a contributing factor.

pipeline/transformer/transformer_tei/TeiLocalTransformerTest.scala.txt
```diff
@@ -2,24 +2,23 @@ package weco.pipeline.transformer.tei
 
 import org.scalatest.funspec.AnyFunSpec
 import org.scalatest.matchers.should.Matchers
-import weco.storage.fixtures.S3Fixtures
-import weco.storage.store.s3.S3TypedStore
```
alexwlchan, line 6: Flagging for @paul-butcher's benefit: this is a test you can run against a local checkout of the TEI repo to find transformer errors. It's only updated when somebody actually uses it rather than being tested continuously, but it's quite handy when there's stuff on the DLQ and you do want to use it.

reindexer/get_reindex_status.py
```diff
@@ -243,9 +249,32 @@ def main(reindex_date):
...
+    }
+    if non_empty_queues:
+        print("The following queues still have messages:")
+        rows = [
+            [q_url.split("/")[-1].replace(f"catalogue-{reindex_date}_", ""), size]
+            for q_url, size in sorted(non_empty_queues.items())
+        ]
+        print(tabulate.tabulate(rows, tablefmt="plain", colalign=("left", "right")))
+    else:
+        print("There are no messages on queues")
```
alexwlchan, line 268: Here's what the new output looks like:

Screenshot 2022-06-18 at 17 07 57

This is to avoid a mistake I made on 2022-06-17; I assumed "index counts are correct" = "nothing left to do", but in fact there was a bunch of stuff on the relation embedder queues that hadn't gone through yet. Turning the pipeline down prematurely swamped the cluster and caused it to fall over.

reindexer/get_reindex_status.py
```diff
@@ -153,27 +153,33 @@ def list_queue_urls_in_account(sess):
         yield from page["QueueUrls"]
 
 
-def get_dlq_stats(sess, *, reindex_date):
+def get_queue_stats(sess, *, reindex_date):
     """
-    Get the size of the DLQs associated with this pipeline.
+    Get the size of the queues associated with this pipeline.
     """
     dlq_urls = [
```
paul-butcher, line 160: Probably `queue_urls` now.

reindexer/get_reindex_status.py
paul-butcher: Excellent!  I started doing something like this once, then just started looking web interface instead.

paul-butcher (approved):

PR #2132, opened 2022-06-20
Remember to index the ID field on subjects

alexwlchan:


alexwlchan:
> IntelliJ reported that the normaliser was in expected but not actual, which confused me at first. It looks like the order of arguments to the assertion is the wrong way round.

Yeah, some of the JSON comparison logic is a bit ropey.

Thanks for helping me solve it!

paul-butcher (approved):

PR #2133, opened 2022-06-20
Remove model compatibility check on deployment

paul-butcher:
This is no longer required, due to restructuring of the model as used by the API.

See: https://github.com/wellcomecollection/catalogue-pipeline/issues/2118


paul-butcher:
> Does this mean we'll always deploy the latest version of the pipeline services into the pipeline?

Yes, or rather, the version that has been checked out by buildkite whenever this build pipeline runs.

Do you think there are some safety checks that should remain?

We could do a two-stage lock-gated deploy to staging then prod, but given that both APIs point to the same index more often than not, I don't think it adds anything.

alexwlchan:
> Yes, or rather, the version that has been checked out by buildkite whenever this build pipeline runs.

👍 I don't think it's a bad thing, I just wanted to check it was intentional.

> Do you think there are some safety checks that should remain?

I can imagine plenty of checks we might need, but I'd rather hold off until we have an incident that proves we need them (allowable as the pipeline is non-critical infra).

alexwlchan (approved):

PR #2135, opened 2022-06-22
restructure the works index mapping for better relations search

harrisonpim:


alexwlchan (approved):
I have to trust these are correct changes to be making, index-wise, but from a Scala POV looks good.

PR #2138, opened 2022-06-27
Add handnote test

paul-butcher:
https://github.com/wellcomecollection/platform/issues/5559


paul-butcher:
This test was Included as part of #2140

PR #2140, opened 2022-06-30
Provenance and Acquisition

paul-butcher:
https://github.com/wellcomecollection/platform/issues/5558


alexwlchan (approved):
Really nice stuff, especially the comments and references. ✨ 

The `Datable` trait is a really nice touch. It might be worth adding some traits for it specifically, because I think we might want to tweak it in future. Having a standalone test harness would make that easier.

e.g. turning dates like `2001-10-20` or `2003-04` into more human-readable strings like `20 October 2001` or `April 2003`.

(To be clear: I think the trait as-written is fine and doesn't need changing in this PR, but putting in tests now might be useful for future changes.)

jamieparkinson (approved):

PR #2149, opened 2022-07-15
Identifying the unidentified

paul-butcher:
Implements: https://github.com/wellcomecollection/concepts-pipeline/issues/3

I think this is good to go now.  I'm off for a few days, but I'd appreciate some eyes on it.

This now creates a "label derived" identifier when no identifier could be extracted from a Concept field.
It also now adds the appropriate identifier to concepts within the concepts list.

The majority of this change, by line count, is updating tests to either now expect an id where there was none, or to not bother checking whether there is an id or not, if it is not relevant to the test. I'm not entirely satisfied with this part of the test suite yet, but I think it is sufficient for this task, and there is a good direction of travel here.  I intend to move the Matchers into the test side for internal_model eventually, if they prove valuable for other parts of the pipeline.

I have also refactored the Agent and Concept transformers to use a common base trait.  In those traits, there was some very similar code, performing the same function, that had started to suffer drift.

I have added logging to the cases where Concepts remain unidentified.  These are both situations where an identifier may be present, but the application does not know what to do with it (unknown scheme, multiple ids). This will help us flush out any situations where this might be happening.  At the moment, they are just silently unidentified.


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraAbstractConcepts.scala
```diff
@@ -0,0 +1,62 @@
...
+      case Seq(subfieldContent) =>
+        maybeAddIdentifier(
+          ontologyType = ontologyType,
+          varField = varField,
+          identifierSubfieldContent = subfieldContent
+        )
+      case Nil =>
+        addIdentifierFromVarfieldText(ontologyType, varField)
+      case _ =>
+        warn(s"$varField has multiple identifier subfields, unable to identify")
```
alexwlchan, line 38: minor: I'd flip this log around so the `$varField` is at the end; it means the common bit is at the start of the message.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraAbstractConcepts.scala
```diff
@@ -0,0 +1,62 @@
...
+        addIdentifierFromText(
+          ontologyType = ontologyType,
+          label = label.trimTrailingPeriod)
+      case None => IdState.Unidentifiable
+    }
+
+  private def addIdentifierFromText(ontologyType: String,
+                                    label: String): IdState.Unminted =
+    IdState.Identifiable(
+      SierraConceptIdentifier.withNoIdentifier(
```
alexwlchan, line 57: nit: I'd be tempted to inline `withNoIdentifier()` here to reduce the levels of indirection, but I don't think it's vital.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraAgents.scala
```diff
@@ -59,43 +62,60 @@ trait SierraAgents extends SierraQueryOps with ConceptsTransformer {
...
-    //
-    // which all refer to the same identifier.
-    //
-    // For consistency, we remove all whitespace and some punctuation
-    // before continuing.
-    val codes = subfields.collect {
-      case Subfield("0", content) => content.replaceAll("[.,\\s]", "")
+  def identify(varfield: VarField, ontologyType: String): IdState.Unminted =
+    varfield.indicator2 match {
+      // 0 indicates LCNames id in use
```
alexwlchan, line 73: Is this part of the MARC spec, or a Wellcome-specific convention?

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraConcepts.scala
```diff
@@ -58,23 +75,13 @@ trait SierraConcepts extends SierraQueryOps with ConceptsTransformer {
...
-    getIdentifierSubfieldContents(varField) match {
-      case Seq(subfieldContent) =>
-        maybeAddIdentifier(
-          ontologyType = ontologyType,
-          varField = varField,
-          identifierSubfieldContent = subfieldContent
-        )
-      case _ => IdState.Unidentifiable
-    }
+    getIdState(ontologyType, varField)
```
alexwlchan, line 80: Could we replace calls to `identifyConcept()` with `getIdState()`?

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraGenres.scala
```diff
@@ -62,18 +60,36 @@ object SierraGenres
...
+      ontologyType = "Concept",
+      varField = varField
+    )
+    wholeFieldConceptId match {
+      case identifiable: IdState.Identifiable =>
+        if (identifiable.sourceIdentifier.identifierType == IdentifierType.LabelDerived)
+          IdState.Unidentifiable
+        else identifiable
+      case other => other
+    }
```
alexwlchan, line 81: Minor: you can put conditions on pattern matches, which is a bit more idiomatic:

```suggestion
    wholeFieldConceptId match {
      case identifiable: IdState.Identifiable if (identifiable.sourceIdentifier.identifierType == IdentifierType.LabelDerived) =>
        IdState.Unidentifiable

      case other => other
    }
```

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/subjects/SierraBrandNameSubjects.scala
```diff
@@ -1,7 +1,10 @@
 package weco.pipeline.transformer.sierra.transformers.subjects
 
+import weco.catalogue.internal_model.identifiers.IdState
 import weco.catalogue.internal_model.work.{Concept, Subject}
-import weco.pipeline.transformer.sierra.transformers.SierraAgents
+import weco.pipeline.transformer.sierra.transformers.SierraConcepts
+import weco.pipeline.transformer.transformers.ConceptsTransformer
+//import weco.pipeline.transformer.sierra.transformers.SierraAgents
```
alexwlchan, line 7: Delete this line.

alexwlchan (commented):
I'm about halfway through reading this; dropping comments I've got so far.

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraGenresTest.scala
```diff
@@ -247,23 +306,51 @@ class SierraGenresTest
         )
       )
     )
+    //TODO: This would be clearer without looping
```
jamieparkinson, line 309: Can you use https://www.scalatest.org/user_guide/table_driven_property_checks?

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/matchers/SubjectMatchers.scala
```diff
@@ -0,0 +1,42 @@
...
+        case _: MatchError =>
+          fail(
+            s"Subject expected to have exactly one concept, found: ${subject.concepts}")
+      }
+    }
+  }
+
+}
+
+object SubjectMatchers extends SubjectMatchers
```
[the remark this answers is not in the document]
paul-butcher: Not in current use, but this is the pattern advised in the Scalatest docs (for the purpose you mention in your question): https://www.scalatest.org/user_guide/using_matchers#usingCustomMatchers

One thing I was concerned about was finding out the best way to include the right test traits, in particular HasId - Should the test language mirror the inheritance structure of the model?  If so, we would get diamond inheritance if we use mixins (e.g. where we check a Subject and its Concepts).  In the end, I decided to be explicit in the test classes themselves, so these objects are not needed.

This comment has made me realise it's probably best to remove them as they are unused.  That would force homogeneity in tests later on.

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraGenresTest.scala
paul-butcher: In this case, I think that just unrolling it and doing the assertions twice is the better plan.

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraGenresTest.scala
```diff
@@ -122,44 +143,70 @@ class SierraGenresTest
...
+    val List(conceptA, conceptV) = genre.concepts
+    conceptA shouldBe a[Concept[_]]
+    conceptA should have(
+      'label ("A Content"),
+      labelDerivedConceptId("A Content")
+    )
+    conceptV shouldBe a[Period[_]]
+    conceptV should have(
+      'label ("MDCCLXXXVII. [1787]"),
+      labelDerivedPeriodId("MDCCLXXXVII. [1787]"),
```
paul-butcher, line 199: I'm not convinced that Periods should have ids, and even if they do, I don't think deriving them from the unmodified label is correct.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/subjects/SierraMeetingSubjects.scala
```diff
@@ -23,29 +24,27 @@ import weco.sierra.models.marc.VarField
...
   val subjectVarFields = List("611")
 
   val labelSubfields = List("a", "b", "c")
 
   def getSubjectsFromVarFields(bibId: SierraBibNumber,
-                               varFields: List[VarField]) =
+                               varFields: List[VarField]): Output =
     varFields.flatMap { varField =>
       createLabel(varField, subfieldTags = List("a", "c", "d")) match {
         case "" => None
```
paul-butcher, line 38: This is the only one of these SubjectTransformers that does this.
Others look like they would create a subject with an empty label.

I think there is scope for refactoring this for harmony.  Probably so that getSubjectsFromVarFields is in the base trait, and it is only the creation of an individual subject that changes by object.

alexwlchan (approved):
I'm going to approve this, but I don't feel I've been able to go through it properly – I can feel my eyes glazing over as I read through 2.5 kloc. I think I've read all the "interesting" bits, but they're hard to find among the test changes.

Next time, it would be better to break up something large like this into smaller pieces that build on each other.

e.g. there's a substantial amount of test churn to add the new HaveMatchers. That’s useful, but maybe it could have been set up in a separate PR which could be reviewed quickly (as it was just replicating existing behaviour) and then subsequent PRs could have been smaller and reviewed in more detail.

PR #2155, opened 2022-08-03
Non unique sourceidentifiers

paul-butcher:
A race condition occurs when documents containing the same SourceIdentifiers are processed simultaneously.

Until now, this race has been vanishingly unlikely, because
- new ids are not encountered very often
- ids have tended to refer to something quite specific, so less likely to clash in this way e.g. compound subjects like Dentistry+Orthodontics+History

However, this has become a problem with the introduction of individual, label-derived concept identifiers.  
- many new ids will be encountered as they are now being generated from labels in the absence of a proper id
- very broad concepts are being identified, e.g. "History", so it is more likely that multiple simultaneous documents will contain the same id.

This change fixes the problem by retrying if the document in question has been beaten in a race.  I have set it to retry only once, as I think that should be sufficient.

The situations in which this comes into play are: 
* https://github.com/wellcomecollection/platform/issues/5475#issuecomment-1203914337 the race is almost guaranteed to occur exactly once for a given document.  This change fixes that.
* During the first full reindex, it will probably occur.  This should catch and fix most of those instances, but it is possible that some documents might need to be redriven if the "terribly bad luck" situation occurs.

### DIFF
--- a/pipeline/id_minter/src/test/scala/weco/pipeline/id_minter/steps/IdentifierGeneratorTest.scala
+++ b/pipeline/id_minter/src/test/scala/weco/pipeline/id_minter/steps/IdentifierGeneratorTest.scala
@@ -90,6 +90,42 @@ class IdentifierGeneratorTest
     }
   }
 
+  it("reconciles new identical identifiers") {
+    //Given a batch with multiple identical source identifiers
+    //Only generate an identifier for each unique source id.
+    val sourceIdentifier1 = createSourceIdentifier
+    val sourceIdentifier2 = createSourceIdentifier
+
+    val sourceIdentifiers = List(sourceIdentifier1, sourceIdentifier2, sourceIdentifier1, sourceIdentifier2)
+
+    withIdentifierGenerator() {
+      case (identifierGenerator, identifiersTable) =>
+        implicit val session = NamedAutoSession('primary)
+
+        val triedIds = identifierGenerator.retrieveOrGenerateCanonicalIds(
+          sourceIdentifiers
+        )
+
+        triedIds shouldBe a[Success[_]]
+
+        val ids = triedIds.get
+        ids.size shouldBe 2
+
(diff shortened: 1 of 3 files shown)


pipeline/id_minter/src/test/scala/weco/pipeline/id_minter/steps/IdentifierGeneratorTest.scala
```diff
@@ -90,6 +96,114 @@ class IdentifierGeneratorTest
...
+        Failure(IdentifiersDao.InsertError(Nil, new Exception("no."), Nil))
+      )
+
+    withIdentifierGenerator(Some(daoStub)) {
+      case (identifierGenerator, _) =>
+        val triedIds = identifierGenerator.retrieveOrGenerateCanonicalIds(
+          sourceIdentifiers
+        )
+
+        triedIds shouldBe a[Success[_]]
```
alexwlchan, line 199: Is it worth asserting that we retrieve the correct identifiers?

alexwlchan (approved):
🍡 

I've toyed with creating a trait for `IdentifiersDao` for a while, which would let you do this all in memory – I'd been looking at what it would take to switch away from RDS and onto DynamoDB, for consistency with what else we do. But I didn't, so mocking seems very reasonable in this case.

One suggestion for extra rigour in the tests, otherwise LGTM.

pipeline/id_minter/src/test/scala/weco/pipeline/id_minter/steps/IdentifierGeneratorTest.scala
paul-butcher: Yes.  Originally. I thought that making that meaningful would overcomplicate the test,  but I realise that on the first attempt, it will generate new ids, so we need to ensure that they don't get returned.

PR #2160, opened 2022-08-05
Add the 2022-08-04 pipeline

alexwlchan:
This patch:

* Adds the Terraform config for the 2022-08-04 pipeline
* Adds a script for analysing transformer failures
* Continues consolidating the boilerplate to reduce the size of `main.tf` for pipelines – this is opening the door to adding the Elasticstack config, which will have to be a separate top-level block in `main.tf`.
* Consolidate the way transformers are defined, to reflect the fact that each transformer is very similar – much less repetition now.


paul-butcher (approved):

pipeline/transformer/get_dlq_failures.py
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env python3
+"""
+This is a crude script for analysing errors in the transformer DLQs.
+
+It:
+
+*   saves a list of all messages from the DLQ
+*   queries the logging cluster for errors with the same ID
```
paul-butcher, line 8: This looks like a really handy tool.

alexwlchan: I think it's potentially extensible to other pipeline apps also, if we log the IDs correctly.

PR #2161, opened 2022-08-05
Normalize non-ascii in label-derived ids

paul-butcher:


pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/transformers/ConceptsTransformer.scala
```diff
@@ -19,12 +21,18 @@ trait ConceptsTransformer {
     currentState match {
       case currentAsIdentifiable: IdState.Identifiable => currentAsIdentifiable
       case _ =>
+        val normalizedLabel = Normalizer
+          .normalize(
+            label.toLowerCase,
+            Normalizer.Form.NFKD
+          )
+          .replaceAll("[^\\p{ASCII}]", "")
```
alexwlchan, line 29: Perhaps worth a note of the exact records that caused us to add this? So we have a real example and don't lose your detective work from earlier.

alexwlchan (approved):

alexwlchan (approved):

PR #2170, opened 2022-08-18
[WIP] Refactor the SierraItemAccess tests to use the more verbose style

alexwlchan:
For https://github.com/wellcomecollection/catalogue-pipeline/issues/2145

This is an experiment into what these tests might look like if we took the more verbose GivenWhenThen style that Paul showed off in Tuesday's review. These tests are definitely a bit unwieldy, and I think maybe this is nicer?

Example output:

```
Given an item in the closed stores with no holds
  And LOCATION = 'scmac' / 'Closed stores Arch. & MSS''
  And STATUS = '-' / 'Available''
  And OPACMSG = 'f' / 'Online request''
  When we create an access condition
  Then the access method is 'online request'
  And the access status is 'open'
  And there's no free-text terms or note

Given an item in the closed stores with no holds
  And LOCATION = 'scmac' / 'Closed stores Arch. & MSS''
  And STATUS = '-' / 'Available''
  And OPACMSG = 'c' / 'Restricted''
  When we create an access condition
  Then the access method is 'online request'
  And the access status is 'restricted'
  And there's no free-text terms or note
```

I'm using `it` instead of `Scenario` for now, but when I swap out AnyFunSpec I should be able to change that.

### DIFF
--- a/common/source_model/src/test/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccessTest.scala
+++ b/common/source_model/src/test/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccessTest.scala
@@ -18,7 +18,8 @@ class SierraItemAccessTest
     with Matchers
     with SierraDataGenerators {
 
-  def createFixedFieldWith(label: String)(value: String, display: String): FixedField = {
+  def createFixedFieldWith(label: String)(value: String,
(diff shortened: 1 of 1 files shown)


alexwlchan:
Closing in favour of #2171 and #2172.

PR #2171, opened 2022-08-18
Add a helper for creating FixedField values

alexwlchan:
This by itself doesn't do much except reduce boilerplate, but it opens the door to refactoring these into GivenWhenThen-style tests, where we put an And() clause in the shared helper, e.g.

```scala
And(s"fixed field $label = '$value' / '$display')
```

which gets us verbose tests without lots of repetition.

We can discuss whether to do that in the patch that adds it -- even if we don't go down that route, I think this is a useful change.

For https://github.com/wellcomecollection/catalogue-pipeline/issues/2145

### DIFF
--- a/common/source_model/src/test/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccessTest.scala
+++ b/common/source_model/src/test/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccessTest.scala
@@ -16,16 +16,22 @@ class SierraItemAccessTest
     with Matchers
     with SierraDataGenerators {
 
-  def createFixedFieldWith(label: String)(value: String, display: String = ""): FixedField =
+  def createFixedFieldWith(label: String)(value: String,
+                                          display: String = ""): FixedField =
     FixedField(
       label = label,
-      value = value, if (display.isEmpty) { Some(display) } else { None }
+      value = value,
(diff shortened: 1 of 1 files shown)


common/source_model/src/test/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccessTest.scala
```diff
@@ -15,24 +15,33 @@ class SierraItemAccessTest
     extends AnyFunSpec
     with Matchers
     with SierraDataGenerators {
+
+  def createFixedFieldWith(label: String)(value: String,
+                                          display: String = ""): FixedField =
+    FixedField(
+      label = label,
+      value = value,
+      if (display.isEmpty) { Some(display) } else { None }
```
paul-butcher, line 24: Should this b e !display.isEmpty?  IIUIC, this results in either "" or None

alexwlchan: Yes, it should. Oops.

I'm tempted to omit the `display` field entirely, rather than fix this – it's useful for the readability of the tests and debugging, but the downstream code only uses the single-letter Sierra codes, never the `display` value.

jamieparkinson (approved):

PR #2172, opened 2022-08-18
Switch to using have()-style matchers in access condition tests

alexwlchan:
For https://github.com/wellcomecollection/catalogue-pipeline/issues/2145; follows #2171

This switches the test to having the nice `have`-style matchers that Paul introduced in some recent work. It's a no-op, but I think they read a little nicer and the error messages are an improvement. 👍


paul-butcher (approved):
Nice.  I particularly like the way `have noTerms()` and `have noNotes()` has turned out.

PR #2189, opened 2022-09-01
Script for reingesting

jamieparkinson:
Reingesting visible works/images based on display model changes is a common enough use case that it's useful to have a script do it for us - this PR adds such a script.

I've also added a `requirements.in`/`requirements.txt`, and moved everything to a `scripts` folder (the latter is really just so that IntelliJ can understand the Python properly).


alexwlchan (approved):
♻️

PR #2190, opened 2022-09-01
Remove records from a bib if they were modified at the same time as the existing record

alexwlchan:
When we add records to a SierraTransformable, we add anything that was modified *after or at the same time as* the existing record.

When we remove records to a SierraTransformable, we were removing anything that was modified *after* the existing record.  This meant we were losing some real updates where the `modifiedDate` of the deleted record was somehow identical to a prior version.

This makes the behaviour consistent and hopefully will avoid future bugs where records aren't removed properly.


jamieparkinson (approved):

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5577

PR #2191, opened 2022-09-01
Sort out the performance of the reingest script

jamieparkinson:
This now takes 4 minutes rather than 2 hours thanks to a mix of batching and concurrency - I suspect tweaking some numbers could improve it further, but it's good enough for now.

### DIFF
--- a/reindexer/scripts/concurrently.py
+++ b/reindexer/scripts/concurrently.py
@@ -1,0 +1,39 @@
+import concurrent.futures
+import itertools
+
+
+def concurrently(fn, inputs, *, max_concurrency=5):
+    """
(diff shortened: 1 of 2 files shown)


reindexer/scripts/concurrently.py
```diff
@@ -0,0 +1,39 @@
+import concurrent.futures
+import itertools
+
+
+def concurrently(fn, inputs, *, max_concurrency=5):
```
jamieparkinson, line 5: Might look familiar @alexwlchan?

alexwlchan (approved):
🐎 

copying code from strange blogs is no way to build a codebase

PR #2196, opened 2022-09-15
Include 520 $c in the description on Sierra records

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5601


paul-butcher (approved):

paul-butcher:
Looks like the failing test is somewhere else that isn't expecting the c subfield to be included.

alexwlchan:
> Looks like the failing test is somewhere else that isn't expecting the c subfield to be included.

Yup, on it. I’m going to change the input to that test to pass a different subfield; I’m guessing the intention is "don't include subfields you're not meant to".

PR #2200, opened 2022-09-20
Speed up the script for reindexing the visible Works

alexwlchan:
Previously we were batching in two places:

* We'd get a batch of 100 IDs from Elasticsearch
* We'd split those into 10 batches of 10 to send to SNS in PublishBatch

This created a sawtooth-like set of work: we'd have to publish 10 batches before we could start working on the next batch of 100.  This is the failure mode described here:
https://alexwlchan.net/2019/10/adventures-with-concurrent-futures/#running-a-large-fixed-number-of-tasks

Screenshot 2022-09-20 at 16 29 29

This changes the script to work on a continuous stream of works, which improves the concurrency and takes the script from ~25m to ~2m.

### DIFF
--- a/reindexer/scripts/reingest_for_display.py
+++ b/reindexer/scripts/reingest_for_display.py
@@ -18,10 +18,9 @@ def chunked_iterable(iterable, size):
         yield chunk
 
 
-def get_work_ids(es, api_index, query, reingest_docs_count):
-    for hit in tqdm(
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):
If this is now known to be working we should merge it as I totally forgot about it when I reingested images on Weds

alexwlchan:
> If this is now known to be working

I think so; the mistake was wrapping `tqdm` around the wrong iterator.

PR #2201, opened 2022-09-21
Fix identifiers on Sierra contributors

alexwlchan:
Previously we were using a single method to mint identifiers for both

* Sierra agents as subjects, and
* Sierra agents as contributors

but these are drawn from different MARC fields and behave in different ways; this PR restores the old approach to getting identifiers for contributors.  This should restore the missing identifiers in the works index.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraContributors.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraContributors.scala
@@ -1,6 +1,10 @@
 package weco.pipeline.transformer.sierra.transformers
 
-import weco.catalogue.internal_model.identifiers.{IdState, IdentifierType, SourceIdentifier}
+import weco.catalogue.internal_model.identifiers.{
+  IdState,
+  IdentifierType,
+  SourceIdentifier
+}
 import weco.catalogue.internal_model.work._
 import weco.sierra.models.SierraQueryOps
 import weco.sierra.models.data.SierraBibData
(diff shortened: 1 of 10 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraContributors.scala
```diff
@@ -162,4 +162,45 @@ object SierraContributors
...
+    codes.distinct match {
+      case Seq(code) =>
+        IdState.Identifiable(
+          SourceIdentifier(
+            identifierType = IdentifierType.LCNames,
+            value = code,
+            ontologyType = ontologyType
+          )
+        )
+      case _ => addIdentifierFromText(ontologyType, label)
```
alexwlchan, line 203: This is the one change from the previous method – we'll now add a label-derived identifier to contributors rather than leaving them unidentified.

pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/subjects/SierraOrganisationSubjectsTest.scala
```diff
@@ -139,7 +139,7 @@ class SierraOrganisationSubjectsTest
       subject.id shouldBe IdState.Identifiable(
         SourceIdentifier(
           identifierType = IdentifierType.LCNames,
-          ontologyType = "Subject",
+          ontologyType = "Organisation",
```
paul-butcher, line 142: This is a change to the original behaviour.  The change is "more correct", but I think we intend to get rid of this entirely, so if it just comes along for the ride, that's fine, but it will result in any organisation subject to be given a new canonical id.

https://wellcome.slack.com/archives/C02ANCYL90E/p1658140051620059?thread_ts=1658132806.695909&cid=C02ANCYL90E

https://wellcome.slack.com/archives/C02ANCYL90E/p1658148261259319?thread_ts=1658132806.695909&cid=C02ANCYL90E

alexwlchan: Ah, oops, then I’ll revert this change.

I think this will require a full reindex – it’ll also mint new source identifiers for every Sierra contributor.

paul-butcher (approved):

PR #2203, opened 2022-09-22
Add label-derived contributor IDs for Calm, Miro and TEI

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5603

This should add label-derived contributor IDs for all contributors in the works API.

* Sierra: they're already added in #2201 so I don't need to add them here
* METS: we don't get any contributors from METS files, so we don't need to do anything for that


pipeline/transformer/transformer_miro/src/test/scala/weco/pipeline/transformer/miro/transformers/MiroContributorsTest.scala
```diff
@@ -20,68 +21,61 @@ class MiroRecordTransformerContributorsTest
...
     transformRecordAndCheckContributors(
       miroRecord = createMiroRecordWith(
-        creator = Some(List(Some(creator1), None, Some(creator2)))
+        creator = Some(List(Some("Beekeeper Brian"), None, Some("Dog-owner Derek")))
       ),
-      expectedContributors = List(creator1, creator2)
-    )
-  }
-
-  it("passes through multiple values in the image_creator field") {
```
alexwlchan, line 43: I remove this test because it felt superfluous with the `ignores null values` test.

paul-butcher (approved):

PR #2211, opened 2022-09-26
Add a bunch more fields to the query object

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5550

I looked at a bunch of fields in https://github.com/wellcomecollection/catalogue-api/blob/c3f23d912bd9e1b8ac3ff28f2c036d921409c172/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala, and added a bunch of the ones we're missing. There's still a fair amount of stuff in the `data` field, in particular where we copy certain values into synthetic fields – but this takes us in the right direction.


paul-butcher (approved):
I spent a while wondering whether all of these fields need to exist, or whether they should be merged to simplify the index, definition, and queries.  

It looks like they are currently all distinct fields, either because of different analysis or boosting, or filtering, but there may be scope in the future to merge some of them (but that's outside the scope of this change)

PR #2214, opened 2022-09-28
Finish migrating all the `data` fields on a Work to the new index structure

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5550

After this I think we'll be able to delete the `data` and `state` fields from an IndexedWork, but that's a sufficiently disruptive change I think we should do that in a new pipeline.

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
@@ -76,6 +76,9 @@ object WorksIndexConfig extends IndexConfigFields {
       val identifiersPath = List("search.identifiers")
       val titlesAndContributorsPath = List("search.titlesAndContributors")
 
+      val newIdentifiersPath = List("query.allIentifiers")
+      val newTitlesAndContributorsPath = List("query.titlesAndContributors")
+
       // Indexing lots of individual fields on Elasticsearch can be very CPU
       // intensive, so here only include fields that are needed for querying in the
       // API.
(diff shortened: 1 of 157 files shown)


alexwlchan:
I will be very glad when I no longer have to write the mapping in two places. 😰

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
```diff
@@ -76,6 +76,9 @@ object WorksIndexConfig extends IndexConfigFields {
       val identifiersPath = List("search.identifiers")
       val titlesAndContributorsPath = List("search.titlesAndContributors")
 
+      val newIdentifiersPath = List("query.allIdentifiers")
```
alexwlchan, line 79: I've had to rename this because `query.identifiers` already exists – it's the list of all source identifiers on a work, to match the field on the display model.

(Previously we distinguished between source/other identifiers in the work model, and indexed both separately – this has simplified that.)

common/internal_model/src/test/scala/weco/catalogue/internal_model/index/SearchIndexConfigJsonTest.scala
```diff
@@ -68,6 +72,14 @@ class SearchIndexConfigJsonTest
         indexName = index.name
       )
 
+      implicit val lcs: Patience[Json] = new Patience[Json]
+
+      println("Mapping diff:")
+      println(diff(fileMapping, indexMapping))
+
+      println("Settings diff:")
+      println(diff(fileMapping, indexMapping))
```
alexwlchan, line 81: This is so the test doesn't tell us "(1500 lines of JSON) is different to (1500 lines of JSON)". It's not super helpful, but it at least gives a clue as to where to look.

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
```diff
@@ -76,6 +76,9 @@ object WorksIndexConfig extends IndexConfigFields {
...
+      val newTitlesAndContributorsPath = List("query.titlesAndContributors")
```
jamieparkinson, line 80: Assuming these distinctions are temporary?

jamieparkinson (approved):

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksIndexConfig.scala
alexwlchan: Yeah, I’ll drop the `new` suffix when I remove the old fields, but I didn't want to delete the old `search.*` fields quite yet.

PR #2225, opened 2022-10-10
Reduce the message retention period and retries for the ingestor input queue

paul-butcher:
To fix https://github.com/wellcomecollection/catalogue-pipeline/issues/2224

### DIFF
--- a/pipeline/terraform/modules/fargate_service/variables.tf
+++ b/pipeline/terraform/modules/fargate_service/variables.tf
@@ -24,6 +24,18 @@ variable "queue_visibility_timeout_seconds" {
   default = 30
 }
 
+variable "message_retention_seconds" {
+  type = number
+  # The actual default on SQS is four whole days.
+  #   default = 345600
+  # This is sufficient to cope with normal bank holiday weekends.
+
+  # However, if a message is sitting on any main queue for more than a day,
+  # then something has gone rather awry, and it should be moved to a DLQ,
+  # rather than cluttering a main queue.
+  default = 86400
+}
+
 variable "max_receive_count" {
   type    = number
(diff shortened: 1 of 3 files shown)


pipeline/terraform/modules/fargate_service/variables.tf
```diff
@@ -24,6 +24,18 @@ variable "queue_visibility_timeout_seconds" {
...
+variable "message_retention_seconds" {
+  type = number
+  # The actual default on SQS is four whole days.
+  #   default = 345600
+  # This is sufficient to cope with normal bank holiday weekends.
+
+  # However, if a message is sitting on any main queue for more than a day,
+  # then something has gone rather awry, and it should be moved to a DLQ,
+  # rather than cluttering a main queue.
+  default = 86400
```
paul-butcher, line 36: This will change the behaviour of any queues relying on that default.  However, I don't think 4 days is a reasonable default for main queue retention (even one day is pretty long, we don't have anything that works like that).

pipeline/terraform/modules/fargate_service/variables.tf
```diff
@@ -24,6 +24,18 @@ variable "queue_visibility_timeout_seconds" {
...
+}
+
 variable "max_receive_count" {
   type    = number
   default = 4
```
paul-butcher, line 41: I wonder if max_receive_count should also default to something lower.

I fancy that two attempts is sufficient to cope with odd intermittent glitches.  Anything more and we're either looking at something truly broken, or something overwhelmed, and in either case, sending the message again without thinking about it is probably not going to help.

jamieparkinson (approved):

PR #2241, opened 2022-11-07
Remove subject type

paul-butcher:
This removes the Subject ontologyType (https://github.com/wellcomecollection/catalogue-pipeline/issues/2235)

The fix in SierraConceptSubjects could be refectored a little, but I intend to pick that up as part of this issue: 
https://github.com/wellcomecollection/catalogue-pipeline/issues/2240

* Reindex after this is merged* It will cause most, if not all, Sierra records to change, so will suffer the harvest problem if it is just merged and left to run as normal.


alexwlchan:
Why has so much code been reformatted differently? D:

jamieparkinson:
@alexwlchan I've wondered this too (as it happens to all of us intermittently) and my best guess is that these files were last formatted with an older version of scalafmt, as we only scalafmt changed files

paul-butcher:
Yes, I think it's scalafmt version change. It started reformatting every file I opened, so I switched it off.  Evidently not soon enough.

jamieparkinson (approved):

PR #2245, opened 2022-11-09
Always run our Lambda tests, and make sure they work

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5626

We have a bunch of Python Lambdas in this repo, and they're pretty old. They use all the clever "should we build" this logic, and because we haven't changed them, we haven't been testing them in a very long time. It turns out if you try to test them now, in 2022, they're broken. Oops.

The problem isn't the Lambdas themselves; it's poorly specified dependencies, e.g. `pip install moto` vs `pip install moto=1.2.3`. Our tests use a [Docker image for tox](https://github.com/wellcomecollection/dockerfiles/blob/main/tox/Dockerfile) that's using Python 3.6, which is EOL. When tox tries to resolve the dependencies, it gets a bunch of deps that aren't Python 3.6 compatible, and the tests break.

The long-term fix would be to upgrade all these Lambdas to newer versions of Python (and possibly ditch tox, which we don't use anywhere else). Or even get rid of these Lambdas, not all of them are still necessary, e.g. https://github.com/wellcomecollection/catalogue-pipeline/issues/2244

The short-term fix (in this patch) is to properly specify their dependencies with 3.6 compatible versions, so we can run the tests consistently and have them pass.


paul-butcher (approved):

paul-butcher:
> and possibly ditch tox, which we don't use anywhere else

Yes.  No point in running tox if it's only expected to run in a specific known environment.

.buildkite/pipeline.yml
```diff
@@ -205,8 +205,23 @@ steps:
...
+  - label: "{{ matrix }} (Lambda)"
+    command: |
+      ./builds/run_python_tests.sh "{{ matrix }}"
+
+      if [[ "$BUILDKITE_BRANCH" == "main" ]]
+      then
+        ./builds/publish_lambda_zip.py "{{ matrix }}"
+      fi
+
+    matrix:
```
paul-butcher, line 217: build matrices are cool

alexwlchan: yeah, they're really nice

PR #2246, opened 2022-11-09
Run all the tests, always

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5626

This means we're now building and publishing all the Scala and Scala-adjacent apps (=image inferrers) on every build, rather than trying to guess which builds should be run (and getting it wrong). Most of this is copied directly from the catalogue-api repo, unmodified.

This doesn't cover deploying the images into ECS – I'll do that separately, because that's a bit trickier with the number of services, and auto-deploying into the catalogue pipeline.


.buildkite/scripts/run_autoformatting.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
alexwlchan, line 1: copied from catalogue-api, see https://github.com/wellcomecollection/catalogue-api/pull/571

builds/get_sbt_project_directory.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
alexwlchan, line 1: copied from catalogue-api

builds/run_sbt_task_in_docker.sh
```diff
@@ -1,13 +1,40 @@
 #!/usr/bin/env bash
+<<EOF
```
alexwlchan, line 2: copied from catalogue-api

builds/run_sbt_tests.sh
```diff
@@ -1,21 +1,40 @@
 #!/usr/bin/env bash
+<<EOF
```
alexwlchan, line 2: copied from catalogue-api

builds/get_sbt_project_directory.sh
```diff
@@ -0,0 +1,85 @@
...
+<<EOF
+Finds the path to an sbt project directory.
+
+This script is mirrored in the catalogue-pipeline and storage-service repos.
```
paul-butcher, line 5: Should tis be `This script is mirrored across our Scala projects.`, as it is in https://github.com/wellcomecollection/scala-libs/pull/193/files#diff-459895d3f6e6deea111c2bd33ad23854299cf6a8555794299b97222fc496dcecR5

alexwlchan: Good point, fixed everywhere – that's the old comment from the API repo, until I realised these scripts would end up in more places than that. 😅

jamieparkinson (approved):

PR #2248, opened 2022-11-09
Replace lshEncodedFeatures with reducedFeatures

harrisonpim:
Elasticsearch 8 allows us to run [approximate-nearest-neighbour queries on dense vectors](https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#approximate-knn) directly, with a [maximum vector dimensionality of 1024](https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html). Previously, we've used [locality sensitive hashing](https://en.m.wikipedia.org/wiki/Locality-sensitive_hashing) as a workaround to make our image feature vectors comparable as text. Knn queries mean that's no longer necessary. 

This PR removes the LSH encoding of feature vectors and directly returns a subset of vectors from the feature extractor.


alexwlchan:
Is this ready for somebody to review?

harrisonpim:
I think so! @paul-butcher what do you reckon?

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/ImagesIndexConfig.scala
```diff
@@ -19,126 +16,22 @@ object ImagesIndexConfig extends IndexConfigFields {
...
-      val fields = Seq(
-        dateField("modifiedTime"),
-        display,
-        query,
-        aggregatableValues
-      )
-
       // Here we set dynamic strict to be sure the object vaguely looks like an
       // image and contains the core fields, adding DynamicMapping.False in places
       // where we do not need to map every field and can save CPU.
```
alexwlchan, line 21: There's a comment here about setting "dynamic strict" that doesn't seem to apply any more, should this be moved to IndexMapping?

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/IndexMapping.scala
```diff
@@ -0,0 +1,30 @@
...
+    * This essentially replaces the practice of calling methods on an elastic4s MappingDefinition
+    * to update a mapping with values for dynamic and _meta etc.
+    * (which do not work with a definition initialised with rawSource),
+    * allowing us to configure the properties in a mapping with plain JSON.
+    *
+    */
+  def apply(propertiesJson: String, buildVersion: String): String = {
+    val version = buildVersion.split("\\.").toList
+    val meta = toJson(
+      Map(s"model.versions.${version.head}" -> version.tail.head)).get
```
alexwlchan, line 22: This should be removed; I've ditched writing the model version on main and it's no longer available.

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/ImagesIndexConfig.scala
```diff
@@ -19,126 +16,22 @@ object ImagesIndexConfig extends IndexConfigFields {
...
-          .dynamic(DynamicMapping.Strict)
-          .meta(Map(s"model.versions.${version.head}" -> version.tail.head))
+        MappingDefinition(
+          rawSource = Some(
+            IndexMapping(
+              propertiesJson = Source
+                .fromInputStream(
+                  getClass.getResourceAsStream("/imagesIndexProperties.json")
+                )
+                .mkString,
```
alexwlchan, line 30: Minor: in scala-libs we have a LocalResources trait which wraps this operation in a function. You can't access it here, but maybe it would be clearer to create a function for reading a resource as a string?

common/internal_model/src/test/scala/weco/catalogue/internal_model/generators/ImageGenerators.scala
```diff
@@ -206,23 +206,23 @@ trait ImageGenerators
     } else {
       (1 to n).map(_ => randomVector(4096, maxR = 10.0f))
     }
-    val lshFeatures = if (similarFeatures) {
-      similarHashes(32, n)
+    val reducedFeatures = if (similarFeatures) {
+      similarVectors(1024, n)
```
alexwlchan, line 210: nit: not one for this PR, but it's hard to know what 1024/n are doing in this context (even with keyword arguments)

common/internal_model/src/test/scala/weco/catalogue/internal_model/generators/VectorGenerators.scala
```diff
@@ -7,19 +7,6 @@ trait VectorGenerators extends RandomGenerators {
 
   private val defaultSimilarity = math.cos(Math.PI / 64).toFloat
 
-  def randomHash(d: Int): Seq[String] =
```
alexwlchan, line 10: As you say, deleted code is debugged code. 👍

pipeline/inferrer/feature_inferrer/app/main.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
```
alexwlchan, line 1: nit: this is a third-party import, should be between fastapi and weco_datascience

pipeline/inferrer/feature_inferrer/app/main.py
```diff
@@ -7,27 +8,40 @@
...
+    normalised_vectors = sliced / np.linalg.norm(sliced, axis=1, keepdims=True)
+    return normalised_vectors
+
+
 def batch_infer_features(images):
     vectors = extract_features(images)
-    lsh_encoded = lsh_encoder(vectors)
-    return [{"vector": v, "lsh": l} for v, l in zip(vectors, lsh_encoded)]
+    reduced = feature_reducer(vectors)
+    return [{"vector": v, "reduced_vector": l} for v, l in zip(vectors, reduced)]
```
alexwlchan, line 44: nit:

```suggestion
    return [{"vector": v, "reduced_vector": r} for v, r in zip(vectors, reduced)]
```

pipeline/inferrer/inference_manager/src/test/scala/weco/pipeline/inference_manager/adapters/FeatureVectorInferrerAdapterTest.scala
```diff
@@ -43,15 +43,16 @@ class FeatureVectorInferrerAdapterTest
     it("augments InferredData with the data from the inferrer response") {
       val features = (0 until 4096).map(_ / 4096f).toList
       val featuresB64 = Encoding.toLittleEndianBase64(features)
-      val lshEncodedFeatures = ('a' to 'z').map(_.toString * 3).toList
+      val reducedFeatures = (0 until 1024).map(_ / 1024f).toList
+      val reducedFeaturesB64 = Encoding.toLittleEndianBase64(reducedFeatures)
```
alexwlchan, line 47: This makes me a bit uneasy, because we're doing both rounds of the base64-ing in Scala. I don't actually know what a feature inferrer response looks like; if we could save a real response into a JSON file and load it here, I'd find it a bit more useful.

Not blocking, but a thought.

alexwlchan (approved):
A couple of small suggestions and tweaks, but the bulk of this looks good from a Scala POV; obviously I can't speak to the data science aspects.

Happy for you to make the changes then merge; I don't need to review this again.

pipeline/terraform/main.tf
```diff
@@ -52,3 +52,30 @@ module "catalogue_pipeline_2022-11-19" {
     aws.catalogue = aws.catalogue
   }
 }
+
+module "catalogue_pipeline_2022-11-28" {
```
paul-butcher, line 56: I have not created this pipeline or populated it or anything at all.

By including the extra pipeline in this file, it blocks deployment of the new code to the current production pipeline.

Depending on when I actually do it, I will add the correct date to both here and the API.

alexwlchan:
The one failing test was caused by a flaky error somewhere inside Docker Compose in the relation embedder tests; I've re-run the test.

PR #2252, opened 2022-11-16
drive pipeline deployment from latest pipeline in terraform

paul-butcher:
There is a little trap in the way we deploy, which this PR is intended to fix:

*  To deploy, it must be on main (to put all the correct latest images on ECR)
*  Once on main, it gets automatically deployed to production

This means that we can't easily create a new pipeline with new code, whilst leaving the existing production pipeline unmodified.  

Leaving the old code running on the existing production pipeline is desirable when there are significantly widespread changes, and we wish to run a reindex on a new pipeline in preparation for that new pipeline becoming the production pipeline. (e.g. for the change I recently made, to avoid an overload during the overnight harvest)

Previously, the choice of pipeline to deploy to was driven by asking the API which pipeline it points to.

With this change, the choice of pipeline is the most recently dated pipeline in main.tf.

It still identifies the production pipeline so that it can print a warning that production is not necessarily up to date, but its target will always be the latest pipeline in terraform.

Scenarios:

1. latest pipeline is production, no change to existing behaviour
2. latest pipeline is not production, deploys to latest, production starts to lag behind main
    * Normally, the latest pipeline is expected to replace production soon, so scenario 1 will resume
    * If a change needs to be pushed to production, without promoting the latest pipeline, we can remove the later pipeline from the tf file on main.

### DIFF
--- a/.buildkite/scripts/deploy_latest_pipeline.py
+++ b/.buildkite/scripts/deploy_latest_pipeline.py
@@ -16,27 +16,7 @@ def get_current_index_name():
     return resp.json()["worksIndex"]
 
 
-if __name__ == "__main__":
-    index_name = get_current_index_name()
-    print(f"The current index name is {index_name}")
-
-    print()
-
-    # The works index name is a string that looks something like
-    #
-    #     works-indexed-2021-08-19
-    #
-    index_regex = re.compile(r"^works-indexed-(?P<date>\d{4}-\d{2}-\d{2})[a-f]*$")
-    pipeline_date = index_regex.match(index_name).group("date")
-    print(f"The current prod pipeline is {pipeline_date}")
-
-    print()
-
-    root = (
-        subprocess.check_output(["git", "rev-parse", "--show-toplevel"])
-        .decode("utf8")
-        .strip()
-    )
+def deploy_to(root, pipeline_date):
 
     os.environ.update({"PIPELINE_DATE": pipeline_date})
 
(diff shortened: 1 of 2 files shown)


paul-butcher:
Another benefit of this approach is that it encodes the "must do a reindex with this merge" prose into an actual change in the code.

If a change is significant enough to require a reindex, it should include a new pipeline in main.tf, which will prevent those changes going on production until the new pipeline becomes production.

jamieparkinson (approved):
Makes sense to me

.buildkite/scripts/deploy_latest_pipeline.py
```diff
@@ -16,34 +16,74 @@ def get_current_index_name():
...
-    )
+    candidate_pipelines = get_pipeline_names_from_terraform_file(root)
+    print(f"possible existing pipeline are: {', '.join(candidate_pipelines)}")
+    latest_pipeline = sorted(candidate_pipelines, reverse=True)[0]
+    print(f"most recent pipeline is: {latest_pipeline}")
+    if latest_pipeline != prod_pipeline:
+        print(
+            "WARNING: The most up to date pipeline is not the current production pipeline "
+        )
+        print(f"production:{prod_pipeline} latest:{latest_pipeline}")
```
alexwlchan, line 88: you can make the dates line up:

```suggestion
        print(f"production:\t{prod_pipeline}\nlatest:\t\t{latest_pipeline}")
```

alexwlchan (approved):

PR #2271, opened 2022-12-05
Tweak the way we use due dates on Sierra items

alexwlchan:
1.  We don't display the due date on the public catalogue record, because often they'll just cause awkward questions for LE&E.  These dates don't seem to be accurate or enforced.

    e.g. "The catalogue says this was due for return 2019. Why can't I find it?"

    A generic message that takes somebody to the desk avoids extra hassle.

3.  If an item has a due date, assume it's checked out to somebody and mark it as unavailable in the catalogue.  I don't totally understand how to detect if items are checked out, but this seems like a good starting heuristic.

    As noted in the comment, if this is drastically wrong we'll hear about it from LE&E; sending people to the desk is the more conservative approach and should reduce confusion.

For https://github.com/wellcomecollection/platform/issues/5632


pollecuttn:
@alexwlchan Yes, the item's due date is the place to check. If there's a date there then the item has been checked out.

PR #2279, opened 2022-12-15
Remove the 2022-11-19 pipeline

paul-butcher:


alexwlchan (commented):
This diff looks very odd.

Also, we seem to have two pipelines labelled 2022-11-28 in main – is that going to be an issue? https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/terraform/main.tf#L4-L5

alexwlchan (approved):
🦒

PR #2295, opened 2023-01-12
Harmonise id ontology types

paul-butcher:


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/OntologyTypeOps.scala
```diff
@@ -0,0 +1,212 @@
...
+    /**
+      * Return a copy of this Contributor with the Agent having been assigned the correct type.
+      */
+    def copyWithNewType(newType: String): Contributor[IdState.Unminted] = {
+      // This should never be called with an unidentifiable agent, because the
+      // point of this is to fix references that have been identified, but have the wrong type,
+      // so let this error out if called on a subject with an inappropriate state
+      val oldAgent =
+        original.agent
+          .asInstanceOf[AbstractRootConcept[IdState.Identifiable]]
```
jamieparkinson, line 122: Could we do this with a pattern match and a "this should never happen" error message?

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/OntologyTypeOps.scala
```diff
@@ -0,0 +1,212 @@
...
+      val newSourceIdentifier =
+        mainSourceIdentifier.get.copy(ontologyType = newType)
```
jamieparkinson, line 124: How do we know this will always be a `Some`? Can we reflect that knowledge in the types?

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/OntologyTypeOps.scala
```diff
@@ -0,0 +1,212 @@
...
+    * There are two "base" or "vague" types, and all the others are equally as "specific".
+    * In real data, this function will not be called for many documents, and when it is, it
+    * will almost always be to resolve disharmony between one of the vague types and one more specific one.
+    */
+  private def mostSpecificType(types: List[String]): String = {
+    // Filter out the vague types.
+    val specificTypes = types.filterNot(Seq("Agent", "Concept").contains(_))
+    specificTypes match {
+      // if there are no more specific types, Agent is more specific than Concept.
+      case Nil => if (types.contains("Agent")) "Agent" else "Concept"
```
jamieparkinson, line 173:
```suggestion
      case Nil if types.contains("Agent") => "Agent"
      case Nil                            => "Concept"
```

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/AbstractRootConcept.scala
```diff
@@ -6,11 +6,36 @@ sealed trait AbstractRootConcept[+State] extends HasId[State] {
...
+            IdState.Identifiable
+          ]] = Map(
+    "Concept" -> (new Concept[IdState.Identifiable](_, _)),
+    "Agent" -> (new Agent[IdState.Identifiable](_, _)),
+    "Place" -> (new Place[IdState.Identifiable](_, _)),
+    "Period" -> (new Period[IdState.Identifiable](_, _)),
+    "Meeting" -> (new Meeting[IdState.Identifiable](_, _)),
+    "Organisation" -> (new Organisation[IdState.Identifiable](_, _)),
+    "Person" -> (new Person[IdState.Identifiable](_, _))
+  )
```
jamieparkinson, line 23: This is annoying isn't it - is there any way round it?

jamieparkinson (approved):
Overall nit: prefer to use `Seq` (which is the interface) over `List` (which is the linked-list implementation) unless there's a strong reason not to

common/internal_model/src/main/scala/weco/catalogue/internal_model/work/AbstractRootConcept.scala
paul-butcher: I wondered about an enum, but I think that's no less annoying than this.

Strictly, none of this is necessary.  At this point, I don't think there is a difference in the behaviour of any of these types (there is when constructing them from parsed MARC, but if we reach the point where we are changing one type for another, it will not have the extra properties associated with the new type).

However, downstream of here, (in the display format stuff), they derive the type value in the JSON from the Scala type.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/OntologyTypeOps.scala
paul-butcher: That was the path I originally tried to take, but got stuck due to type erasure.  I'm going to try again and see if I can get around that sensibly.

PR #2307, opened 2023-01-23
Be more permissive about MeSH ids marked as LoC

paul-butcher:


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraConceptIdentifier.scala
```diff
@@ -55,6 +56,11 @@ object SierraConceptIdentifier {
       // LCSH in the Works Catalogue, but these appear to be errors, so an exact match is
       // enforced here in order to guard against such typos re-emerging.
       case "sh" => IdentifierType.LCSubjects
+      // MESH is not a Library of Congress scheme, but a common error found in source documents
+      // is to mark a MeSH id as a LoC id. Fix it, but also log a warning
+      case "D" =>
+        warn(s"MeSH identifier found masquerading as LoC identifier: $idValue")
+        IdentifierType.MESH
```
alexwlchan, line 63: Do we know how many IDs this affects? I'd lean towards omitting the identifier rather than "fixing" it, and making it a Collections Information problem to get the data right.

paul-butcher: There are 475 ids across 232 documents.  I have submitted a list to CI for them to have a look at.

paul-butcher:
These have all been fixed at source now.  We don't have to worry about them.

PR #2310, opened 2023-01-27
Report sbt warnings

paul-butcher:
One problem I have noted on buildkite is that it (rightly) conceals all the output from the stages, and simply reports pass/fail. You can look for the detail if you wish, but it takes two clicks for each step, so if you don't have any specific concerns, you probably won't look.

This means that warnings (such as dependency evictions) can go unnoticed if they do not result in a command failure.

This PR annotates the build with warnings found during the scalafmt run.  As that is the stage that looks at all the files/projects, it is probably not necessary to extend this to extracting warnings from all stages, but it would be simple to do so if required.

e.g. https://buildkite.com/wellcomecollection/catalogue-pipeline/builds/7590


kenoir (approved):
👏

PR #2317, opened 2023-02-02
Add indexedTime to images, to match works

jamieparkinson:
This field is used by the snapshot reporter for works - I thought I might as well do this properly and add it to images as well. Part of https://github.com/wellcomecollection/platform/issues/5478


alexwlchan (approved):
Good idea!

PR #2319, opened 2023-02-03
Include language from $l subtag

paul-butcher:
This mitigates https://github.com/wellcomecollection/catalogue-pipeline/issues/2308 (though this issue is more widespread than just that document)

Some of those Iliads use $s to differentiate themselves, but I think this is the only place where that construction is used in that way,  Including $s would be wrong in other situations.

Also, these Iliads look wrong anyway, so we shouldn't try too hard to accommodate them.


alexwlchan (approved):
🏮

PR #2327, opened 2023-02-10
[WIP] Run tests on the new scala2 queue

alexwlchan:
This is a set of instances running the latest Buildkite version, so I can test how the batcher behaves without breaking other Scala tests.

For https://github.com/wellcomecollection/platform/issues/5656

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,7 +5,7 @@ steps:
       ./builds/run_linting.sh
     label: "autoformat"
     agents:
-      queue: "scala"
+      queue: "scala2"
 
   - label: "{{ matrix }} (library)"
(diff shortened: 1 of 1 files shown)


alexwlchan:
This is bizarre: it works if internal model itself is in the dependency tree for the batcher, but not any of the internal model dependencies?

alexwlchan:
I think maybe the changes here are significant? https://github.com/wellcomecollection/catalogue-pipeline/pull/2327/commits/c3f01de5151230d55b3cee852e950facc04b59a6

alexwlchan:
Okay, so here's an interesting thing: the same commit ran twice, did/didn't pass. What's up with that? 🤔  https://buildkite.com/wellcomecollection/catalogue-pipeline/builds/7723#01864a90-5b45-4154-9cb5-b761c2857afe

PR #2329, opened 2023-02-14
Mapping for increased precision

harrisonpim:
Adds new some new analysers and subfields to help improve the precise ordering of catalogue search results

linked to https://github.com/wellcomecollection/catalogue-api/pull/619

closes https://github.com/wellcomecollection/catalogue-api/issues/501 and closes https://github.com/wellcomecollection/catalogue-api/issues/607

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksAnalysis.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksAnalysis.scala
@@ -8,6 +8,11 @@ object WorksAnalysis {
   val slashesCharFilter =
     MappingCharFilter("slashes_char_filter", mappings = Map("/" -> " __"))
 
+  // This analyzer "keeps" the hyphen, by removing it and treating hyphenated
+  // tokens as a single token.
+  val hyphensCharFilter =
+    MappingCharFilter("hyphens_char_filter", mappings = Map("-" -> ""))
+
   val asciiFoldingTokenFilter = AsciiFoldingTokenFilter(
     "asciifolding_token_filter",
     preserveOriginal = Some(true)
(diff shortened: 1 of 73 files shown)


common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksAnalysis.scala
```diff
@@ -8,6 +8,15 @@ object WorksAnalysis {
   val slashesCharFilter =
     MappingCharFilter("slashes_char_filter", mappings = Map("/" -> " __"))
 
+  // This analyzer "keeps" the hyphen, by removing it and treating hyphenated
+  // tokens as a single token.
```
alexwlchan, line 12: Does this mean that, for example, "e-mail" would be tokenised as "email"?

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksAnalysis.scala
```diff
@@ -78,12 +87,31 @@ object WorksAnalysis {
...
-    charFilters = Nil
+    charFilters = List(hyphensCharFilter.name)
+  )
+
+  val shingleCasedAnalyzer = CustomAnalyzer(
+    "shingle_cased_analyzer",
+    tokenizer = "standard",
+    tokenFilters = List(shingleTokenFilter.name, asciiFoldingTokenFilter.name),
+    charFilters = List(hyphensCharFilter.name)
   )
```
alexwlchan, line 115: What does this analyser do?

alexwlchan (approved):
Looks fine. Makes me wonder how we preserve these useful comments if we ditch the Scala DSL. 🤔

harrisonpim:
> Makes me wonder how we preserve these useful comments if we ditch the Scala DSL. 🤔

I've been thinking the same thing. We have [some documentation of the query structure](https://github.com/wellcomecollection/catalogue-api/blob/main/docs/search/current_queries/works.md), but there's an equal amount of thought that has gone into the structure of the mapping which is only implicitly described there. Might follow up this pair of PRs with another one which explains those choices in prose.

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksAnalysis.scala
harrisonpim: exactly that! NB that only applies to (sub)fields which use the `hyphens_char_filter`, so we'll still be indexing a lot of "e"s and "mail"s, but we now have a better chance of retrieving docs which mention "email" over docs which mention "mail" when people search for "e-mail"

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksAnalysis.scala
```diff
@@ -78,12 +87,31 @@ object WorksAnalysis {
     charFilters = Nil
   )
 
+  // now also include a case sensitive version of the english analyzer
```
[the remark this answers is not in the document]
harrisonpim: right again!

common/internal_model/src/main/scala/weco/catalogue/internal_model/index/WorksAnalysis.scala
harrisonpim: In elastic land, shingles are spans of multiple words which are indexed as one token, while preserving their order. As a simple example, a 3-gram shingles analyser would index "the quick brown fox jumped over the lazy dog" as 
- "the quick brown"
- "quick brown fox"
- "brown fox jumped"
- "fox jumped over"
- "jumped over the"
- "over the lazy"
- "the lazy dog"

The existing `shingleAsciifoldingAnalyzer` lowercases and asciifolds all of those tokens (and the queries which it compares them to). This new `shingleCasedAnalyzer` does the same thing, but doesn't apply a lowercase token filter. That should make us even more likely to retrieve strings of text where both casing and position are important, eg when people search for "hearing aids", a work called "hearing the stories of AIDS survivors" should appear lower in the results than "a story about hearing aids".

PR #2338, opened 2023-03-09
Genres as concepts

paul-butcher:


common/display_model/src/main/scala/weco/catalogue/display_model/work/DisplayAbstractRootConcept.scala
```diff
@@ -38,7 +38,17 @@ case object DisplayAbstractConcept extends GetIdentifiers {
     abstractConcept match {
       case Concept(id, label) =>
         DisplayConcept(
-          id = id.maybeCanonicalId.map { _.underlying },
+          id = id.maybeCanonicalId.map {
+            _.underlying
+          },
```
alexwlchan, line 43: minor, not for this PR: I wonder if we need a case class `DisplayCanonicalId` to handle this for us.

alexwlchan (approved):

PR #2340, opened 2023-03-10
Use Concept ontologyType in Miro Subjects

paul-butcher:
This is the "Divination problem".  There are two canonicalIds being assigned to the Concept "Divination" (among others), due to there being some "Subject" Divinations and some "Concept" Divinations.  The Subject ones coming from Miro.

In addition, this makes Miro Subjects look the same as Sierra Subjects. The sole Concept in a Subject in Works from both systems now carry an identifier.


alexwlchan (approved):
🔮

PR #2348, opened 2023-04-14
Tweak the memory available for the 1x image inferrer

alexwlchan:
It looks like the amount of memory available has changed slightly at some point, and now image inferrer tasks on the 1x instances are failing to start because they don't have enough memory.  Correcting the amount of memory available should fix it.


jamieparkinson (approved):
I wonder if the default version of the ECS agent / something in an AMI / ??? changed so as to use more memory...

PR #2358, opened 2023-05-17
Add a new Sierra reader that picks up deleted records

alexwlchan:
For https://github.com/wellcomecollection/catalogue-pipeline/issues/2325

## The problem

Our queries for Sierra are only picking up modified records, not deleted records. I don't know if this was always broken or has broken since we originally wrote it; either way our existing queries aren't sufficient.

This is causing problems for Collections Information's workflows, and is obviously the wrong behaviour for the pipeline.

## Idea #1: modify the Scala apps

I had a brief look at modifying the existing Sierra reader code, but it's somewhat gnarly and I think it would need significant changes to support this new type of query.

In particular, it writes records in bulk an S3 bucket, which then get "demultiplexed" by a Lambda function that splits them into individual SNS messages. This was originally done so that we could resume a partially-completed window if we got interrupted, but I'm not sure this has ever happened in practice. Modifying this to support multiple streams of updates would be a fairly substantial change.

## Idea #2: stop using those Scala apps

The existing Sierra reader has felt somewhat large and overengineered for a while now, and I had a Python Sierra API client just lying around…

This PR creates a new Lambda function which replaces the Sierra reader (and the demultiplexer). It uses the new SNS PublishBatch API and sends individual records directly to SNS, bypassing the bulk records in S3. This is what the revised architecture looks like:

```mermaid
graph TD
   W[window generator] -. "❌❌❌" .-> R["Sierra reader (ECS, Scala)"]
   W --> N["Sierra reader (Lambda, Python, new)"]
   R -->|uploads bulk records to| S3[(S3 bucket)]
   S3 -->|S3 bucket notification| D["S3 demultiplexer (Lambda, Python)"]
   N -->|sends individual Sierra records| L["Sierra linker (ECS, Scala)"]
   D -. "❌❌❌" .-> L

    classDef externalNode fill:#e8e8e8,stroke:#8f8f8f
    class W,R,S3,D,L externalNode

    classDef newNode fill:#9bc0af,stroke:#4f7d68,stroke-width:2px
    class N newNode
```

This new reader is deployed and sending messages to the rest of the Sierra adapter, but I haven't removed the old Sierra reader yet – it's still in place ~and processing windows, those messages just aren't going anywhere~. We can roll back easily if necessary (check out main, apply Terraform).

I'd like to leave both readers in-place for a while to check the new one is working correctly, and then delete the old one if everything turns out to be okay.

## Benefits of this approach

For Collections:

* We actually get deleted records, which we didn't before.
* Updates will appear on the site much faster. We now fetch updates every minutes, whereas before we were waiting ~7m between bibs and ~15m between items/holdings/orders (plus the latency of the catalogue pipeline itself).

For us:

* Less code to maintain. This is roughly half the size of the old Sierra reader, and a lot of it is comments explaining the weird behaviours of Sierra.
* It's a step towards the Lambda-everywhere universe we're heading towards.
* We can run the new reader locally – I've added a bash script that lets you test the reader (see the README for an example).


sierra_adapter/sierra_reader_new/sierra_reader.py
```diff
@@ -0,0 +1,173 @@
...
+    what the Scala app expects, but for now this matches the messages sent
+    by the existing Sierra reader.
+    """
+    try:
+        return json.dumps(
+            {
+                "id": record["id"],
+                "data": json.dumps(record),
+                "bibIds": [],
+                "modifiedDate": record["deletedDate"] + "T23:59:59Z",
```
paul-butcher, line 80: What is the significance of marking deletions as happening at the end of the day?

alexwlchan: Arbitrary choice, but maybe feeds into the comment above.

sierra_adapter/sierra_reader_new/sierra_reader.py
```diff
@@ -0,0 +1,173 @@
...
+        params={
+            "fields": sierra_fields,
+            "updatedDate": f'[{window["start"]},{window["end"]}]',
+        },
+    ):
+        yield record
+
+    dates = {window["start"].split("T")[0], window["end"].split("T")[0]}
+
+    # Because filtering by updatedDate doesn't catch deleted records, we
```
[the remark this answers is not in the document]
alexwlchan: Yeah, this is a tricky one. I think we do want to pick up deletions on the same interval as regular windows – from a Collections Information perspective, they want deletions to propagate as quickly as updates – but this model means we'll repeatedly process a deleted record, the first time we see it and then every time after.

I'm thinking of maybe tracking deletions somewhere in the Lambda context, so the Lambda doesn't repeatedly re-send them. And then log the time when something was initially spotted as deleted, so we know when in the day it occurred – rather than the end of the day.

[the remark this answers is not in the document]
paul-butcher: I agree it's important that we process them promptly, and if it's the same interval as updates, then it might as well be in the same process.

I think re-processing deletions is fine, as a deletion will be idempotent (unless attempting to delete an absent record yields an exception) and NOOP deletions should be pretty speedy.

I'd also expect deletions to be low volume. If we end up deleting the same few records several times in a day, then it's not going to break the bank or fill up logs or anything.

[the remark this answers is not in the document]
alexwlchan: > I think re-processing deletions is fine, as a deletion will be idempotent (unless attempting to delete an absent record yields an exception) and NOOP deletions should be pretty speedy.

Good point, the whole idea of this rewrite was to reduce the premature optimisations in the Sierra reader. 😅 

I'll put some comments about this in the code.

sierra_adapter/sierra_reader_new/sierra_reader.py
jamieparkinson: Probably worth a comment in the code to make clear it's arbitrary

alexwlchan: This was a good question, because I'd already forgotten (but remember when I went to write the comment) that this isn't arbitrary after all!

alexwlchan:
Okay, I started having issues with the Sierra progress reporter, so I went and tidied that up too.

* It now gets the new "deploy from a shell script" and "run the Lambda locally" niceties.
* We were doing some very funky date serialisation when we cross the Scala/Python app boundary – this simplifies that by just stuffing an entire JSON object in the S3 key:

Screenshot 2023-05-19 at 14 46 21

I've used this to backfill deletions – I ran a local copy of the reader just fetching deletions, and had it write updates into a fresh prefix. This will purge any previously-unnoticed deletions.

.buildkite/pipeline.deploy-adapters.yml
```diff
@@ -59,6 +59,28 @@ steps:
...
+  - label: deploy Sierra progress reporter
+    command: |
+      ./sierra_adapter/sierra_progress_reporter/deploy.sh
+
+    plugins:
+      - wellcomecollection/aws-assume-role#v0.2.2:
+          role: "arn:aws:iam::760097843905:role/platform-ci"
+
+    agents:
+      queue: nano
```
alexwlchan, line 82: Ideally I'd do this with something like a matrix build (and maybe we will as more stuff goes to Lambda), but for now this is fine.

infrastructure/modules/window_generator/cloudwatch.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_cloudwatch_event_rule" "window_generator_rule" {
   name                = "${var.source_name}_window_generator_rule"
   description         = "Starts the ${var.source_name}_window_generator lambda"
-  schedule_expression = "rate(${var.trigger_interval_minutes} minutes)"
+  schedule_expression = var.trigger_interval_minutes == 1 ? "rate(1 minute)" : "rate(${var.trigger_interval_minutes} minutes)"
```
alexwlchan, line 4: This isn't just me, apparently CloudWatch is also picky about this!

alexwlchan:
This is ready for another review.

* I've fixed all the bugs, so it actually sends updates to the rest of the pipeline.
* I've backfilled all the deleted records since we last used the Sierra adapter, and re-fetched all the updates for the last week, so anything that's been dropped is now in our database. I've seen those updates get to the Sierra merger and beyond.
* I've changed the way the reader/progress reporter serialise window information, so it's easier to read and less weird – just Python's built-in `isoformat()` function.
* I've fixed the script for building missing windows, to make sure it works with the new progress reporter.

sierra_adapter/terraform/sierra_reader_new/untitled.txt
alexwlchan, line 1: delete this

paul-butcher (approved):

PR #2360, opened 2023-05-24
Precreate indices

paul-butcher:
This is the terraform part of reducing our dependence on E4S.

This creates new, configured indices using terraform, but does not change the practice of also creating/updating them at runtime.


pipeline/terraform/local/stack/locals.tf
```diff
@@ -0,0 +1,14 @@
...
+  es_works_merged_index       = "works-merged-${var.local_pipeline_name}"
+  es_works_identified_index   = "works-identified-${var.local_pipeline_name}"
+  es_works_denormalised_index = "works-denormalised-${var.local_pipeline_name}"
+  es_works_index              = "works-indexed-${var.local_pipeline_name}"
+
+  es_images_initial_index   = "images-initial-${var.local_pipeline_name}"
+  es_images_augmented_index = "images-augmented-${var.local_pipeline_name}"
+  es_images_index           = "images-indexed-${var.local_pipeline_name}"
+
+  es_config_path = "${path.root}/../../../index_config"
```
alexwlchan, line 12: what I’d give for a `git.root` Terraform provider

pipeline/terraform/modules/es_index/main.tf
```diff
@@ -0,0 +1,26 @@
+locals {
+  full_name = var.name
+
+  mappings_file = "${var.config_path}/mappings.${var.mappings_name}.json"
+  analysis_name = var.analysis_name != "" ? var.analysis_name : var.mappings_name
+
+  analysis_json = jsondecode(local.analysis_name != "empty" ? file("${var.config_path}/analysis.${local.analysis_name}.json") : "{}")
```
alexwlchan, line 7: This was a little confusing at first, because it wasn't clear to me where `analysis_name` was being set to `"empty"` – I can see it's coming via `mappings_name`, but I wonder if we could document this expectation in the variable descriptions? Just a thought.

pipeline/terraform/modules/es_index/main.tf
```diff
@@ -0,0 +1,26 @@
...
+
+  analysis = {
+    analyzer : try(jsonencode(local.analysis_json["analyzer"]), "{}")
```
alexwlchan, line 10: TIL: Terraform has a `try` function!

Do we need `try(…)` here? Is there any case when we wouldn't want this to fail if passed invalid JSON?

alexwlchan (approved):
Overall looks really nice. 👍 

Couple of minor suggestions inline; the other would be about the naming of the index config files – `v1`, `v2`, `v3` overlap with the names of the APIs, which feels like it could get confusing. Could we use dates instead, perhaps?

pipeline/terraform/modules/es_index/main.tf
paul-butcher: The problem I was overcoming here is that I don't want to have to specify all the analysis properties in the JSON even if we don't want them.

I think this is the clearest way to express "get the property or return an empty object".

The JSON is known to be well-formed, having been decoded into an object above, so this is just about whether the requested property exists at the top level.  It's possible that the property could be present but inappropriate (e.g. a number or a string), but that will be caught by `elasticstack_elasticsearch_index`, and wouldn't be caught here even if we didn't `try`.

TF is a bit annoying about optional or default properties and `try` expects both results to be the same type, meaning that `{}` is not the same as `{"x":{"y":"z"}}` because it doesn't have an `x` property, hence encoding it to a string.   (it may be possible to do something by defining a `map` of `maps` and telling it how to decode the JSON in the first place, instead of letting it do type magic on its own, but since we just want a string representation, that seemed like a bit of a faff)

pipeline/terraform/modules/es_index/main.tf
paul-butcher: Good point.  I wanted to avoid the situation where authors would have to type in two identical values, given that that would be the most common case, so set it to default to match mappings (then failed to describe that defaultness adequately).

PR #2374, opened 2023-06-02
Ignores the identifier in 650 ǂ0 if we can't deduce the identifier scheme

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5699

This is causing 14 of the 23 failures that are currently on the Sierra transformer DLQ.


jamieparkinson (approved):

alexwlchan:
Looking at what tests have failed elsewhere; I think this is actually cataloguing errors.

PR #2375, opened 2023-06-02
Remove runtime index creation

paul-butcher:
Indices are already created at deploy time.  With this change, they no longer get modified with create-or-update whenever a Worker starts.

As with other changes related to https://github.com/wellcomecollection/catalogue-pipeline/issues/2356, this leaves much of the existing Elastic4s stuff in the application, but makes it easier to remove. I've also been trying to gradually consolidate bits of E4S for this purpose.

The main thrust of this change is the removal of the index config parameter from ElasticIndexer, and instead of creating an index in `ElasticIndexer.init()`, it now raises an exception if the index does not exist.

Being a popular class, this has involved touching quite a few files.  Also, as this index creation behaviour was relied upon in some tests to create the temporary test index, it has caused some churn in tests.


common/pipeline_storage/src/test/scala/weco/pipeline_storage/elastic/ElasticIndexerTest.scala
```diff
@@ -26,49 +25,47 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
     with JsonAssertions
-    with IndexConfigFields
-    with SampleDocumentGenerators {
+    with SampleDocumentGenerators
+    with EventuallyInElasticsearch {
 
-  import weco.pipeline_storage.generators.SampleDocument._,
-  weco.pipeline_storage.generators.SampleDocument.{
-    canonicalId => sampleDocumentCanonicalId
-  }
```
paul-butcher, line 38: This was the cause of a bizarre error about "ambiguous implicit values"

common/pipeline_storage/src/test/scala/weco/pipeline_storage/EitherIndexerTest.scala
```diff
@@ -32,56 +34,77 @@ class EitherIndexerTest
...
 
-        whenReady(indexer(works.map(Left(_)))) { result =>
-          result shouldBe a[Right[_, _]]
-          assertElasticsearchEventuallyHas(workIndex, works: _*)
+  it("returns successfully when there are neither works nor images") {
+    withEitherIndexer[Work[Merged], Image[Initial]] {
+      case (indexer, _, _) =>
+        whenReady(indexer(Nil)) {
+          result =>
+            result shouldBe a[Right[_, _]]
```
paul-butcher, line 105: I spotted a couple of notable omissions from this suite - only images and nothing, so here they are.

There is also no test for failures, but that wouldn't follow the same pattern and would add even more change to an already rather large PR.

jamieparkinson (approved):
Lovely

Issue #2386, opened 2023-06-26
Don't transform genres from Sierra that don't have a label


alexwlchan:
If you look at [z4qvbzeh](https://api.wellcomecollection.org/catalogue/v2/works/z4qvbzeh?include=genres,identifiers), we have a genre that has no identifier and an empty label:

```
"genres": [
  {
    "label": "",
    "concepts": [],
    "type": "Genre"
  }
],
```

This is because there's an almost-empty 655 in the underlying MARC (b2317884):

```
{
  "fieldTag": "m",
  "marcTag": "655",
  "ind1": " ",
  "ind2": "7",
  "subfields": [
    {
      "tag": "2",
      "content": "rbgenr"
    }
  ]
},
```

Also uue6a6bt/b17652686.

We can ask Collections Information to delete the empty MARC fields, but we should consider discarding them in our transformer – this is the second/third time this issue has occurred.

PR #2400, opened 2023-07-19
filter out genres with empty labels

georgiaewhitney:
For #2386 

Genres with empty labels are filtered out

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraGenres.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraGenres.scala
@@ -48,18 +48,21 @@ object SierraGenres
   def apply(bibData: SierraBibData): List[Genre[IdState.Unminted]] =
     bibData
       .varfieldsWithTag("655")
-      .map {
+      .flatMap {
         varField =>
           val (primarySubfields, subdivisionSubfields) =
             getLabelSubfields(varField)
 
-          val label = getLabel(primarySubfields, subdivisionSubfields)
-          val concepts = getPrimaryConcept(
-            primarySubfields,
-            varField = varField
-          ) ++ getSubdivisions(subdivisionSubfields)
+          getLabel(primarySubfields, subdivisionSubfields) match {
+            case "" => None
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraGenres.scala
```diff
@@ -48,18 +48,21 @@ object SierraGenres
...
-            primarySubfields,
-            varField = varField
-          ) ++ getSubdivisions(subdivisionSubfields)
+          getLabel(primarySubfields, subdivisionSubfields) match {
+            case "" => None
+            case label =>
+              val concepts = getPrimaryConcept(
+                primarySubfields,
+                varField = varField
+              ) ++ getSubdivisions(subdivisionSubfields)
```
alexwlchan, line 62: nit: I would keep the two variable assignments from the original code, then add the `match` block as a new thing – there's a lot going on in a couple of lines that make it a bit tricky to follow.

alexwlchan (approved):
One readability suggestion, not blocking.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraGenres.scala
georgiaewhitney: Thanks, Alex! This is such a great learning point. 

Something like this?

```scala
val (primarySubfields, subdivisionSubfields) = getLabelSubfields(varField)
      val label = getLabel(primarySubfields, subdivisionSubfields)
      label match {
        case "" => None
        case _ =>
          val concepts = getPrimaryConcept(
            primarySubfields,
            varField = varField
          ) ++ getSubdivisions(subdivisionSubfields)
```

alexwlchan: I'd move the `concepts` variable out as well, so the `match` arms are both quite short. You'll only use the variable in one of the arms, but then you can glance over the `match` to see what each arm is doing.

PR #2403, opened 2023-08-02
Fix the lookup of bib IDs on order records

alexwlchan:
On order records, the linked bibs are supplied as a list of URLs to the resources in the Sierra API, and not a list of individual bib numbers. This patch updates the reader to extract the bib IDs from the URLs, which should fix the issue where order records aren't being linked correctly.

For https://github.com/wellcomecollection/platform/issues/5709

I've already deployed this change and I'm refetching all the order records right now.


paul-butcher (approved):
🔨

PR #2405, opened 2023-08-02
override Miro licences with METS

paul-butcher:
This is a solution to https://github.com/wellcomecollection/wellcomecollection.org/issues/10074 and  https://github.com/wellcomecollection/catalogue-pipeline/issues/2404  (will require a reindex)

This now works by modifying the image data, rather than just the query/aggregation stuff.


jamieparkinson:
I might just be completely braindead, but is there a reason this logic doesn't belong in the merger?

paul-butcher:
This resolves the superficial issue whereby the fields serving filters and aggregations are different.  That causes the result count after filtering to be different to the expected result count shown beside the filter option, and for the filter option to be duplicated with a different count when applied.

The main reason I put it here is because this is where the problem currently arises.  We copy the source licences into the image's source aggregation without considering that this does not match the image's queryable licence value.

This fixes that point - the queryable licence value and the aggregatable licence value now match.

Another reason I put it here is because there is something much deeper going on which I do not understand. There are a lot of mismatches between the licences, including some very incompatible ones (CC0 + In Copyright).  I'm more comfortable with searching and filtering returning some items whose licence does not match the filter, than I am with "correcting" the item and finding that it now has the wrong licence.

It is absolutely incorrect that we have records that are both CC0 and In Copyright, so something is definitely screwy upstream of the ingestor, and I suspect it's in the source data.  That can be investigated under https://github.com/wellcomecollection/catalogue-pipeline/issues/2404.

jamieparkinson (approved):
Very nice - might suggest adding a test case in `MergerIntegrationTest` for documentation purposes?

PR #2420, opened 2023-08-17
Prevent unexpected hierarchies due to circular paths

paul-butcher:
https://github.com/wellcomecollection/platform/issues/5714

There are two aspects to this change, in something of a belt-and-braces approach. 

1. The Sierra Transformer will not a path where the parent and child are identical.
2. The Path Concatenator will not create path where any section of the path is circular.

I believe that the problem, as reported, is caused by the former, but not also fixing the latter could cause strange editing order effects, where a node in a hierarchy could move around depending on which documents were edited first.


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraCollectionPath.scala
```diff
@@ -122,14 +133,18 @@ private object HostEntryFieldCollectionPath
...
+    if (hostEntryField.isDefined && bibId.nonEmpty) {
+      collectionPathString(hostEntryField.get, bibId.trim) match {
+        case Some(path) =>
+          Some(
+            CollectionPath(
+              path = path,
+              label = None
+            )
+          )
+        case None => None
```
jamieparkinson, line 145: Just my usual non-obligatory monadic nit-conjuring
```suggestion
      collectionPathString(hostEntryField.get, bibId.trim) map { path =>
        CollectionPath(
          path = path,
          label = None
        )
```

jamieparkinson (approved):
🎉

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraCollectionPath.scala
paul-butcher: Thanks, I always seem to miss this.  I start with a match because I want to do something different in the None case, then realise I don't need to, and never take that final step of changing it to map

PR #2447, opened 2023-09-22
Eager global ordinals

paul-butcher:
This sets eager_global_ordinals on the fields we use for aggregation.  In doing so, this will push the work of creating the global aggregation data structures to index time (strictly, refresh time), rather than query time.

I'll do a reindex on Monday.


paul-butcher:
This change now incorporates some fixes to get the API Key stuff working (following some unfortunate discoveries when running it in anger)

pipeline/terraform/2023-09-29/main.tf
```diff
@@ -0,0 +1,30 @@
...
+
+  index_config = {
+    works = {
+      identified = "works_identified.2023-05-26"
+      merged     = "works_merged.2023-05-26"
+      indexed    = "works_indexed.2023-09-12"
+    }
+    images = {
+      indexed        = "images_indexed.2023-09-12"
+      works_analysis = "works_indexed.2023-05-26"
```
agnesgaroux, line 20: I don't quite understand why it's not using the new version here, ie `works_indexed.2023-09-12`.

pipeline/terraform/modules/es_index/main.tf
```diff
@@ -23,6 +23,7 @@ resource "elasticstack_elasticsearch_index" "the_index" {
   analysis_filter      = local.analysis.filter
   analysis_char_filter = local.analysis.char_filter
   analysis_tokenizer   = local.analysis.tokenizer
+  deletion_protection  = false
```
agnesgaroux, line 26: Is it to do with the api key and secret? ie. deleting them when an old/unused pipeline is taken down?

pipeline/terraform/2023-09-29/main.tf
paul-butcher: That's a mistake, but it's OK, because they are actually identical.  I created the new one because it was simpler for the Works (which currently only specify one date)

pipeline/terraform/modules/es_index/main.tf
paul-butcher: This is also a mistake.  I added it because I needed to destroy a pipeline that had gone funny, then forgot to remove it.

I'll raise a ticket about this, as I think this should be something we pass in from the top level to be very explicit about what we are doing.

jamieparkinson (approved):

agnesgaroux (approved):

PR #2451, opened 2023-09-27
Use a common base for the inference manager builds

alexwlchan:
Following discussion at standup today, this patch switches all the inferrer apps to a Python 3.8 base. I considered upgrading to a newer Python, but that would mean upgrading our libraries/dependencies.

I additionally considered trying to get the Python dependencies onto common versions, but we have multiple versions of numpy/scipy which are fiddly to upgrade and probably the most expensive bit of the build.

For https://github.com/wellcomecollection/catalogue-pipeline/issues/2445


alexwlchan:
Okay, this doesn't work because you need to upgrade tox first. Argh.

PR #2462, opened 2023-10-13
correctly find digmiro

paul-butcher:
Fixes part of https://github.com/wellcomecollection/catalogue-pipeline/issues/2460

### DIFF
--- a/pipeline/merger/src/test/scala/weco/pipeline/merger/rules/ImageDataRuleTest.scala
+++ b/pipeline/merger/src/test/scala/weco/pipeline/merger/rules/ImageDataRuleTest.scala
@@ -180,6 +180,30 @@ class ImageDataRuleTest
     }
   }
 
+  it(
+    "correctly identifies a digmiro Sierra work regardless of other digcodes"
+  ) {
+    val metsWork = createInvisibleMetsIdentifiedWorkWith(numImages = 1)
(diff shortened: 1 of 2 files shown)


pipeline/merger/src/test/scala/weco/pipeline/merger/rules/ImageDataRuleTest.scala
```diff
@@ -180,6 +180,30 @@ class ImageDataRuleTest
     }
   }
 
+  it(
+    "correctly identifies a digmiro Sierra work regardless of other digcodes"
```
agnesgaroux, line 184: What does this mean? 😅

paul-butcher: A Sierra Work might be marked with digmiro to show that its Miro content has been superseded by a METS version.

The problem this fixes is that the application could only identify a digmiro work if the `digmiro` is the first `digcode` on that Work.

paul-butcher: This test is essentially a copy of the preceding one, but with some extra digcodes

agnesgaroux (approved):

PR #2466, opened 2023-10-19
Only merge Miro images into the imageData field

paul-butcher:
This is the second part of https://github.com/wellcomecollection/catalogue-pipeline/issues/2460

When I merge this and start running data through it, I will need to consult with Ashley about the results.

Although the change by line count is non-trivial, it should be very easily reversible, as the only file with lines that have truly "changed" is [pipeline/merger/src/main/scala/weco/pipeline/merger/services/Merger.scala](https://github.com/wellcomecollection/catalogue-pipeline/pull/2466#diff-c7be255dbff9e530f074c88db38e45cd5afc10046edc67314f49427dd4df71e2) whereas the rest is additive.


jamieparkinson (approved):
lovely tests!

PR #2468, opened 2023-10-20
updates to account for changes in auth and APIs

agnesgaroux:
A few changes came up when I was suppressing the Brought to Life images
- authenticating to ES with api key rather than username/password. The python `Elasticsearch` client want it decoded so it can encode it back again in the request... 
- the works index has changed and has no `allIdentifiers` in the `query`. Now found at `query.identifiers.value` 
- dlcs server only responds with 204 and no body in case of success


jamieparkinson (approved):
Lovely

PR #2470, opened 2023-10-30
Authenticate the TEI updater to GitHub

jamieparkinson:
Also fixes Lambda package uploads!


agnesgaroux (approved):

tei_adapter/tei_updater/src/tei_updater.py
```diff
@@ -93,10 +93,19 @@ def main(event, _ctxt=None, s3_client=None, sns_client=None, session=None):
     bucket_name = os.environ["BUCKET_NAME"]
     key = os.environ["TREE_FILE_KEY"]
     github_api_url = os.environ["GITHUB_API_URL"]
+    github_token_secret = os.environ.get("GITHUB_TOKEN_SECRET", None)
 
     s3_client = s3_client or boto3.client("s3")
     sns_client = sns_client or boto3.client("sns")
 
+    if github_token_secret and not session:
```
paul-butcher, line 101: Personally, I would have made this a separate function.  Partly for SRP, but also so it could look a bit like lines 98-99 above `session = session or authenticatedGithubSession(github_token_secret)` (but that's just my own preference)

paul-butcher (approved):

tei_adapter/tei_updater/src/tei_updater.py
jamieparkinson: Good suggestion

PR #2472, opened 2023-10-31
Merger refactor

paul-butcher:
* Replaces the Cats State stuff with simply returning a list of all the redirects (all sources other than the target were redirects, so the state business where it keeps the first retain/redirect value for each source was unnecessary)
* A little bit of renaming and shuffling around
* A shortcut for when there is only one source, which is therefore the target. (Saves the chore of reading through all of merge working out which bits are relevant)


jamieparkinson (approved):
Definitely improved - has this clarified any of your thoughts for a rewrite?

PR #2483, opened 2023-11-17
Pipeline with new mapping

jamieparkinson:


pipeline/terraform/modules/es_index/main.tf
```diff
@@ -16,6 +16,7 @@ locals {
 }
 
 resource "elasticstack_elasticsearch_index" "the_index" {
+  deletion_protection  = false
```
paul-butcher, line 19: This should probably be omitted - I keep committing this as well (https://github.com/wellcomecollection/catalogue-pipeline/issues/2463)

paul-butcher (approved):

pipeline/terraform/modules/es_index/main.tf
jamieparkinson: Agh I managed to avoid committing it right until my very last commit!

PR #2486, opened 2023-11-23
New image test documents

jamieparkinson:
Following on from https://github.com/wellcomecollection/catalogue-pipeline/pull/2483


pipeline/ingestor/update_image_documents.py
```diff
@@ -0,0 +1,100 @@
...
+    host = get_secret_string(session, secret_id=f"{secret_prefix}/public_host")
+    port = get_secret_string(session, secret_id=f"{secret_prefix}/port")
+    protocol = get_secret_string(session, secret_id=f"{secret_prefix}/protocol")
+    username = get_secret_string(
+        session, secret_id=f"{secret_prefix}/read_only/es_username"
+    )
+    password = get_secret_string(
+        session, secret_id=f"{secret_prefix}/read_only/es_password"
+    )
+    return Elasticsearch(f"{protocol}://{host}:{port}", basic_auth=(username, password))
```
agnesgaroux, line 60: What's the auth changed to use api key? or was it something else?

agnesgaroux (approved):

pipeline/ingestor/update_image_documents.py
jamieparkinson: Nah you're right but we still have the `read_only` user as a one-off for scripts: https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/terraform/modules/stack/elastic_pipeline.tf#L213-L232

PR #2494, opened 2023-12-06
METS: Stop premature normalisation

paul-butcher:
Ideally (If I understand the existing pattern properly), the MetsXML class pulls relevant data as-is out of the XML file, and then the transformer transforms it for use downstream.

This is a place where that wasn't the case.  The MetsXML class was transforming URLs from their source data form into one compatible with DLCS. 

This PR sorts that out and pushes that transformation a bit to the right.

### DIFF
--- a/pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformers/MetsImageDataTest.scala
+++ b/pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformers/MetsImageDataTest.scala
@@ -1,0 +1,95 @@
+package weco.pipeline.transformer.mets.transformers
+
+import org.scalatest.funspec.AnyFunSpec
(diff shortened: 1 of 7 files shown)


pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformers/MetsImageDataTest.scala
```diff
@@ -0,0 +1,95 @@
...
+          recordIdentifier = "b30246039",
+          fileReference = FileReference("a", "objects/b30246039_my_location")
+        ).locations.head.url should include(
+          "/image/b30246039_my_location/"
+        )
+      }
+    }
+  }
+
+  private val nowhere = DigitalLocation(url = "", locationType = OnlineResource)
```
jamieparkinson, line 80: poignant

jamieparkinson (approved):

PR #2510, opened 2024-01-04
New pipeline with more appropriate hardware

jamieparkinson:
We had been using the default "Storage Optimized" hardware profile for our pipeline instances. Elasticsearch [describe](https://www.elastic.co/guide/en/cloud/current/ec_selecting_the_right_configuration_for_you.html) the use case for this as 

> Good for most ingestion use cases with 7-10 days of data available for fast access. Also good for light search use cases without heavy indexing or CPU needs.

Which is very much _not_ our use case. We have also seen latency issues which are very clearly CPU-bound. This PR creates a pipeline which uses a smaller CPU-optimized instance:

> Suitable for ingestion use cases with 1-4 days of data available for fast access and for search use cases with indexing and querying workloads. Provides the most CPU resources per unit of RAM.

Because this is a smaller (and ARM-based) instance, it's actually about 17% cheaper than the current cluster 🤑


paul-butcher (approved):
LGTM!

PR #2513, opened 2024-01-09
improve precedence tests

paul-butcher:
## What does this change?

This makes the the TargetPrecedenceTests more precise.  

Previously, 
- the tests could be passed by an implementation of `getTarget` that simply returned `works.headOption`
- there was no test for when two records are tied for priority.

I have shuffled the order of the input Seqs so that the expected return value for each test is in an arbitrary position in the input, and added a tiebreaker test.

## How to test

Run the tests.  This is a change to the unit test suite and not the behaviour of the application.  You could try changing the behaviour as suggested above and see the tests fail.

The addition of the tiebreaker test could crystallise the existing semi-arbitrary behaviour. As a mitigation, I have begun a Slack discussion about it, and included commentary pointing out that this is a descriptive, rather than prescriptive test.


jamieparkinson (approved):
Useful, thank you

PR #2514, opened 2024-01-09
New pipeline 2024-01-09

jamieparkinson:
This pipeline does not make use of a shingle analyzer, which addresses latency issues https://github.com/wellcomecollection/catalogue-api/issues/734.

This PR also changes the default cluster to be a 4gb CPU-optimized one, and uses CPU-optimized hardware for both reindexes and normal operation. It also resolves https://github.com/wellcomecollection/catalogue-pipeline/issues/2463.


pipeline/terraform/2024-01-09/main.tf
```diff
@@ -2,22 +2,22 @@ module "pipeline" {
   source = "../modules/stack"
 
   reindexing_state = {
-    listen_to_reindexer      = false
-    scale_up_tasks           = false
-    scale_up_elastic_cluster = false
-    scale_up_id_minter_db    = false
-    scale_up_matcher_db      = false
+    listen_to_reindexer      = true
```
paul-butcher, line 5: Does all this need to be scaled back down now?

pipeline/terraform/modules/stack/variables.tf
```diff
@@ -74,3 +70,8 @@ variable "index_config" {
     })
   })
 }
+
+variable "allow_delete_indices" {
```
paul-butcher, line 74: 🎉

pipeline/terraform/2024-01-09/main.tf
jamieparkinson: Yes - will do that now

kenoir:
Not blocking for merge, but what metrics did you use to validate that the latency is caused by the shingling (it seems very likely from the context and linked conversations that it is, but is there a smoking gun somewhere that makes it very clear it's this specifically?).

paul-butcher (approved):

jamieparkinson:
Good question @kenoir - I used the [Kibana search profiler](https://www.elastic.co/guide/en/kibana/current/xpack-profiler.html) with a query that I knew was very slow. It showed (a) uniform high latency across all shingle-analyzed fields that was (b) distributed across a huge number of shingled terms, so I guessed that the shingling was the problem.

I then added [this](https://github.com/wellcomecollection/rank/pull/101/commits/b2bed81e5305a277c12ae94bd140cc10b1051288) to the relevance tests and confirmed that it failed with a timeout, removed the shingles from the analyzers and saw that it started passing.

PR #2515, opened 2024-01-10
Tweak getMatcherGraph script

kenoir:
## What does this change?

Some small changes to improve working with getMatcherGraph:

- Give a helpful suggestion if dot isn't installed
- Allow specifying a pipeline date
- Synchronously exec commands and fail if there's an error (so we don't show success, but really fail to generate a graph)
- Add some logging to say what we're querying and where the output is

## How to test

- [ ] Run this locally, does it behave as expected?

Those working with this script can more easily get started, and can specify pipeline dates not yet in use.

We're presuming users have OSX in our suggestion to use brew, but all our devs are at the moment.


kenoir:
Triggered by reviewing https://github.com/wellcomecollection/catalogue-pipeline/pull/2511

jamieparkinson (approved):

PR #2518, opened 2024-01-15
Calm archivematica links

paul-butcher:
## What does this change?

This allows CALM and Archivematica METS records to merge.

1. It sets the objectIdentifier (UUID) of the Archivematica record as its sourceIdentifier
2. It creates a mergeCandidate in a CALM Work using the SDB_URL, which contains the matching UUID
3. It makes UUID an acceptable format for METS identifiers.

## How to test

Attached is a CSV of a bunch of CALM records that should link to Archivematica records [Untitled%20discover%20search.csv](https://github.com/wellcomecollection/catalogue-pipeline/files/13949687/Untitled.20discover.20search.csv)

Once this is deployed to a new pipeline and reindexed, the CALM records should be in works-merged, the archivematica records should be merged into them so that the archivematica IIIF resources are now in the CALM work.

We start seeing born-digital records coming through. 

1. CALM works may stall in the merger.  This could happen if 
    * A UUID-only SDB_URL in a CALM record does not match a corresponding Archivematica record.
    * Corresponding Archivematica records do not successfully pass through the pipeline
2. Once this runs on all of the Archivematica records, it may expose some edge cases I hadn't considered
    * This will surface as failures in the METS transformer. We can fix any that arise.
4. Non-publishable Born-Digital material might get published
    * I don't believe that this can actually happen, and would most likely be the result of a fault in the source data if it does.

To mitigate these, I have added a new latest pipeline to deploy and reindex into.


pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
```diff
@@ -56,7 +56,8 @@ object SourceIdentifierValidation {
      * see: `https://search.wellcomelibrary.org/iii/encore/record/C__Rb1971204?marcData=Y`
      */
     private def tryParseUUID(str: String): Try[UUID] = Try(UUID.fromString(str))
-
+    private def isValidMetsId(str: String): Boolean =
+      sierraSystemNumber.toPredicate(str) || tryParseUUID(str).isSuccess
```
kenoir, line 60: Feels a little strange that we need to do this ... did we think about having a sierra identifier type and an archivematica identifier type? I'm guessing there is something awkward about the adapter data model that makes this difficult?

pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformer/GoobiMetsXmlTest.scala
```diff
@@ -25,6 +25,13 @@ class GoobiMetsXmlTest
     MetsXml(xml).value.recordIdentifier shouldBe Right("b30246039")
   }
 
+  it("returns the same value as the recordIdentifier fro the metsIdentifier") {
```
kenoir, line 28:
```suggestion
  it("returns the same value as the recordIdentifier for the metsIdentifier") {
```

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/identifiers/SourceIdentifierValidation.scala
paul-butcher: I'll put some commentary in about it.

That relates to [this ramble](https://wellcome.slack.com/archives/C02ANCYL90E/p1705060396703809?thread_ts=1705058777.597079&cid=C02ANCYL90E).

I think it's perfectly possible to have two distinct METS ID types here. The METS Transformer already knows whether it's a b number or a UUID because of where it comes from.  Similarly, the CALM transformer will already know which type it is when creating the mergeCandidate for the same reason. However, I don't think it's a valuable distinction.

It boils down to this - 
1. The existing behaviour works this way because:
    * Goobi METS files only have the Sierra B number as their identifier
    * That B Number is also the identifier for the Sierra record it corresponds to
    * The identifier for a document must be unique across the whole system
    * Therefore, we put these two identical identifiers into two distinct namespaces
2. The behaviour I have added simply puts the METS UUID into the METS database because:
    * It comes from METS, and we have an existing METS namespace
    * That it is from METS is the only practical/important distinction here. This is used in the mergeCandidate code
    * The namespace must reflect the source of the id, otherwise the b number problem could arise
    * I think the only thing that would be gained from having two distinct METS id types, would be to simplify this one line at the expense of (admittedly trivial) complexity elsewhere.
    * The existing METS type is already named with the unmarked form METS, and not SierraMets or METSBnumber, meaning that METSUUID would be an unfairly marked form.  The existing name cannot bet changed without significant messing about in the ID Minter.

kenoir (approved):

PR #2530, opened 2024-01-24
i-numbers can have a dotted suffix

paul-butcher:
## What does this change?
See https://github.com/wellcomecollection/wellcomecollection.org/issues/10565

There are some iconographic numbers (e.g. `3329263i.104`  on [vu3mwczx](https://wellcomecollection.org/works/vu3mwczx)) which have a numeric suffix after a dot.  These are currently being erroneously ignored by the validation that expects a number followed by `i`.

This fixes that bug.

## How to test

This is pretty well isolated, so the unit tests should cover it quite well.  When this has been deployed and the relevant records indexed, then records like vu3mwczx should show a `Reference` beneath the title (`Reference: 3329263i.104` in this case), in the same way as https://wellcomecollection.org/works/g8j7y2xh

It is syntactically possible that a Sierra record might have multiple valid `001` fields.  If a dotted one appears ahead of an undotted one, then the reference displayed for that record will change. 

1. I don't think that is likely
2. If it does happen, it will probably be more correct
3. I think this number is only for display, and no machines are relying on it to preferentially point to a dotless i-number

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraIconographicNumber.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraIconographicNumber.scala
@@ -9,7 +9,7 @@ object SierraIconographicNumber
     with SierraQueryOps {
   override type Output = Option[String]
 
-  private val IconographicNumberMatch = "^([0-9]+i)$".r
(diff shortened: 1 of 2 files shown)


rcantin-w:
I approve of the new regex, I've tested our known scenarios with a regex testing tool and it, but I can't approve the rest of the scala code 😬  so assigning others as reviewers.

kenoir (approved):

PR #2531, opened 2024-01-24
Fix the Sierra adapter deployment

kenoir:
## What does this change?

The sierra adapter deployment has been failing for a long time, the cause from some time in may seeming to be updating the sierra_reader & sierra_progress_reporter lambdas. Specifically they've been looking for the upload zip artefact in the wrong place. 

Broken since:
Screenshot 2024-01-24 at 16 24 15

Fixed briefly:
Screenshot 2024-01-24 at 16 44 51

This change fixes the build scripts for those lambdas to find the zip file where it's built and add some more output.

## How to test

These scripts can be run locally:

```sh
./sierra_adapter/sierra_progress_reporter/deploy.sh
./sierra_adapter/sierra_reader/deploy.sh
```

They run successfully when the correct AWS profile is assumed (when testing I commented out the lambda update step, so that will be exercised in CI when this is merged).

These lambdas are not being kept up to date, and incorrectly show the failure of the adapter build in buildkite. This change should fix those problems.

This will be the first time these lambdas are updated since May 2023, so if there have been code changes since then we won't have tried them out anywhere. That does unfortunately seem to be the case (https://github.com/wellcomecollection/catalogue-pipeline/pull/2358), and the change was significant although I suspect the deployment was done manually and the code is unchanged since this update.

We can confirm by comparing the currently in use lambdas to the deployment artefact we generate.

### DIFF
--- a/sierra_adapter/sierra_progress_reporter/deploy.sh
+++ b/sierra_adapter/sierra_progress_reporter/deploy.sh
@@ -6,22 +6,26 @@ set -o errexit
 set -o nounset
 
 ROOT=$(git rev-parse --show-toplevel)
+S3_BUCKET="wellcomecollection-platform-infra"
+S3_KEY="lambdas/sierra_adapter/sierra_progress_reporter.zip"
+FUNCTION_NAME="sierra-adapter-20200604-sierra_progress_reporter"
 
+echo "Building sierra_progress_reporter.zip"
 pushd "$ROOT/sierra_adapter/sierra_progress_reporter"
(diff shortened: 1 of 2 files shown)


agnesgaroux (approved):
So it was just the missing ../ 🙃

kenoir:
The zip files this builds are different (at least for the progress reporter):

Screenshot 2024-01-24 at 16 57 42

So worth some further investigation before trying to deploy this.

PR #2538, opened 2024-01-26
allow matcher graph to cope with suppressed images

paul-butcher:
## What does this change?

The current incarnation of the matcher graph script cannot cope with graphs containing suppressed images.  This fixes that problem.

The problem has two causes:

1. The ".jpg" file that gets downloaded in this situation is actually a JSON description of the error
    * This is fixed by copying in a placeholder "No Image" image if the response is unsuccessful
2. [html labels in .dot files](https://graphviz.org/doc/info/shapes.html#html) can be either a table or some text, not both.  When generating the label for a suppressed image, we were appending "(suppressed)" to a table containing the image.
      * I have fixed this by appending "(suppressed)" to the text part of the label, rather than the whole thing.

I noticed this whilst trying to diagnose https://github.com/wellcomecollection/catalogue-pipeline/issues/2536

## How to test
Run the matcher graph generator on a graph that contains a suppressed image, e.g. `yarn getMatcherGraph szjj6gqy`

Before this change, you would have seen an error complaining about a surprise `(`

After this change, you will see this:
[szjj6gqy.pdf](https://github.com/wellcomecollection/catalogue-pipeline/files/14065920/szjj6gqy.pdf)

We'll be able to more reliably generate matcher graphs to help diagnose potential merge problems.

This is an internal tool used by developers to help diagnose problems.  It does not interfere with the running of anything.  The only risk I can imagine is if we genuinely have an image in the collection that looks like the placeholder image, and someone gets confused.  I think that highly unlikely.


agnesgaroux:
The changes are only in /pipeline/matcher/scripts/graphAttributes.ts + the 404 jpg?

agnesgaroux:
I'm getting 
```
Querying catalogue-2024-01-09_works-graph for szjj6gqy
Warning: "_images/MiroImageNumber/L0005170.jpg" was not found as a file or as a shape library member
Error: No or improper image file="_images/MiroImageNumber/L0005170.jpg"
in label of node tsqvr466
```
and the graphs are generated properly, except I'm not seeing the 404.jpg 😕

paul-butcher:
Whoops - I included a change to scalafmt that made everything go awry.  I've recreated this PR properly here: https://github.com/wellcomecollection/catalogue-pipeline/pull/2539

PR #2544, opened 2024-02-05
Don't discard Sierra Items when merging similar Sierras

paul-butcher:
### Background
There was an assumption that whenever two Sierra records merge, it is due to a direct MARC 856 link between a physical and a digital Sierra item.

That assumption has turned out to be false.

The assumption was strong, and only the happy path (where this is true and error-free) was considered.  When it turned out to be false, the target's item list was being emptied.  Although there was some consideration given to the possibility that more than one candidate merge source existed, the repercussions of this were also not fully considered.  

As a result, whenever two Sierra records merge due to an indirect link (e.g. they both point to the same Miro record, and the `mergeSingleMiroIntoSingleOrZeroItemSierraTarget` rule had not been invoked, then the "source" record would become a redirect, and all the items on both records would be discarded.

This goes part way towards fixing https://github.com/wellcomecollection/catalogue-pipeline/issues/2536, however, because of https://github.com/wellcomecollection/catalogue-pipeline/issues/1561, it cannot guarantee that all of the expected items will be present on the resulting work.  It will ensure that _some_ of the items will be there, rather than none of them.  Which ones will depend on which record is treated as the target in the given merge.
 
### Changes

I have tightened up the condition so that this merge rule is only applied when it is true (I.e. it will only attempt a physical/digital merge on a target with a physical/digital merge candidate).  If this rule is not applied, then it falls through to the default NOOP rule of leaving the item list unchanged on the target.

If the corresponding source cannot be found (e.g. if the source record has failed transformation due to an error), then the target will still retain its own item list.

The list of merged Works (which is used to create the redirects) now only includes those Sierra source records that contributed to the merge, and not just all of them that _could_ have contributed to the merge.

Filled in the holes in the test suite surrounding this merge rule.

## How to test

Because of the problem of nondeterministic merges, I cannot give a definitive result to look for with production data.  The test for this change is to examine the test suite and see that it corresponds to your expectations.

Once merged and the troublesome records in https://github.com/wellcomecollection/catalogue-pipeline/issues/2536 have been reingested, then we will see that those records that had _no_ items will now have _some_ items.

This PR corresponds to a scenario that probably _should_ not occur. However, it does. This represents a robustness against unexpected data.

Perhaps Nilling the items list is desirable for some reason.  I don't think it is.  In any case, the new additions to the test suite offer an opportunity to record and discuss this behaviour in prose, whereas before, it was only in the application code.  I have included some log warnings for when the data is not exactly as I would expect.


paul-butcher:
This may be unnecessary, as this is a scenario that [should not happen](https://wellcome.slack.com/archives/C8X9YKM5X/p1707476527733699?thread_ts=1706107574.100929&cid=C8X9YKM5X).  However, we have seen some instances where this does happen and causes unexpected results.

This change adds some logging for when these unexpected situations occur, as well as making the result a bit more logical when it does, so it is still a good improvement.

jamieparkinson (approved):
Nice tests

PR #2553, opened 2024-02-09
Split miro from reindexer

paul-butcher:
## What does this change?

This adds a new option to start_reindex, which is to run everything apart from Miro (you are then expected to run Miro only after everything else has reached the merger)

The reason for this is to avoid the timing problem where Miro records hit the Merger before their associated Sierra records and incorrectly become standalone images.

It also adds a new pipeline that has been populated by that method.

See https://github.com/wellcomecollection/catalogue-pipeline/issues/2549

## How to test

Run start_reindex.py. It will warn you if you try to reindex "all"

The number of Images in images-indexed should be around 136,000.  

When deploying the _next_ pipeline to the API (i.e. whatever comes after 02-09), the row "/catalogue/v2/images" should get a tick.

This is a much more laborious process than the fire-and-forget reindex all that it replaces.  We should try to progress with a proper solution as soon as possible.


kenoir (approved):
👍 contingent on this having the expected results in a real reindex.

paul-butcher:
Some notes about the real reindex here: https://wellcome.slack.com/archives/C02ANCYL90E/p1707487982445049

I'm not entirely happy about the result.  It works, but it's a faff and it is slow.  However, it does fix the issue at hand and hopefully we'll be able to implement a better approach in the future.

This does not remove the ability to run an "all" reindex, it just warns against it

PR #2554, opened 2024-02-12
Allow the AMI for the ECS EC2 instance cluster host to be set

kenoir:
## What does this change?

This change allows setting of the AMI for EC2 ECS cluster host (used by the inferrer task) in order to fullfil: https://github.com/wellcomecollection/platform-infrastructure/issues/410

## How to test

Apply this change to a running pipeline, and ensure tasks are placed in the instance as required.

This instance can be monitored via the CrowdStrike & Qualys agents.

### DIFF
--- a/infrastructure/modules/scaling_service/variables.tf
+++ b/infrastructure/modules/scaling_service/variables.tf
@@ -114,7 +114,6 @@ variable "queue_config" {
     visibility_timeout_seconds = number
     message_retention_seconds  = number
     max_receive_count          = number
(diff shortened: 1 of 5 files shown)


jamieparkinson (approved):

kenoir:
The plan diff here is huge, i'm going to hold off until I get someone to help me understand why it's so big.

kenoir:
> The plan diff here is huge

Because I unpinned the terraform version, without this only the launch template is updated as expected.

PR #2555, opened 2024-02-16
Put mergeCandidates in the indexed debug information

jamieparkinson:
This is in service of building a matcher-graph-visualiser which doesn't need access to DynamoDB (which also leads me to wonder whether the next version of the matcher/merger even needs DynamoDB itself)


paul-butcher (approved):
I wonder if we might end up indexing redirectSources as well.

PR #2567, opened 2024-02-29
Common marc

paul-butcher:
IN PROGRESS.

## What does this change?

https://github.com/wellcomecollection/catalogue-pipeline/issues/2563

This should be strictly a refactoring of existing code.  The end result when processing real data will be unchanged.

Previously, all MARC content came from Sierra in JSON format.  In preparation for MARC from EBSCO in XML format, I am pulling common  field transformations out from the Sierra transformer into a set of common MARC transformers.

In doing so, I am defining a common interface for both JSON and XML formats of MARC.  A problem arising from the difference between the formats is that they both divide the fields into different groups, which are both meaningless to the pipeline.

JSON divides fields between fixed and variable fields, whereas the XML format presents control fields and data fields as the important distinction.

The transformers just want fields with a given tag, they don't care whether the field in question is fixed, data, variable or control.

Also related to this is the potential removal of code from scala-libs that is currently only being used in the Sierra transformer.

## How to test

There should, be no change in behaviour.

Accepting a new format of MARC data should be as simple as implementing the new trait to extract the fields. Accepting it from a new source will still require us to do all the work around that, including deciding exactly which fields we want to extract.

This is potentially quite a large refactor, and it may turn out that I make some mistake.  To mitigate the risks, I am leaving the Sierra field transformers in situ, but now they do nothing but call their common MARC counterpart.  I am also trying to distinguish between Sierra-specific tests (e.g. those derived from real data), which will be left in-situ, as they are a useful record and tool for describing the behaviour to librarians, and more internally functional or abstract tests, which are being moved and/or enhanced in the new suite.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/data/BibDataAsMarcRecord.scala
+++ b/pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/data/BibDataAsMarcRecord.scala
@@ -1,0 +1,48 @@
+package weco.pipeline.transformer.sierra.data
+
+import weco.pipeline.transformer.marc_common.models.{
+  MarcField,
+  MarcRecord,
+  MarcSubfield
+}
+import weco.sierra.models.SierraQueryOps
+import weco.sierra.models.data.SierraBibData
+import weco.sierra.models.marc.{Subfield, VarField}
+
+trait SierraMarcDataConversions {
+  import scala.language.implicitConversions
+
+  /* TODO:
+   * This implicit conversion stuff exists to temporarily implement SierraBibData
+   * as a subclass of MarcRecord until I move it out of the scala-libs project
+   * work out exactly what we _do_ need it to look like, and implement the
+   * interface directly on it.
+   * */
+  implicit def bibDataToMarcRecord(bibData: SierraBibData): MarcRecord =
+    new BibDataAsMarcRecord(bibData)
+
+  implicit def varFieldToMarcField(varField: VarField): MarcField =
+    MarcField(
+      marcTag = varField.marcTag.get,
+      subfields = varField.subfields.map(sierraSubfieldToMarcSubField),
+      content = None,
+      fieldTag = varField.fieldTag,
+      indicator1 = varField.indicator1.getOrElse(" "),
+      indicator2 = varField.indicator2.getOrElse(" ")
+    )
+
+  implicit def sierraSubfieldToMarcSubField(subfield: Subfield): MarcSubfield =
+    MarcSubfield(tag = subfield.tag, content = subfield.content)
+}
+object SierraMarcDataConversions extends SierraMarcDataConversions {}
+
+class BibDataAsMarcRecord(bibData: SierraBibData)
+    extends MarcRecord
+    with SierraQueryOps {
+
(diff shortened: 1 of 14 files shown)


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraTitle.scala
```diff
@@ -23,55 +24,6 @@ object SierraTitle extends SierraDataTransformer with SierraQueryOps {
   //
   // http://www.loc.gov/marc/bibliographic/bd245.html
   def apply(bibData: SierraBibData): Option[String] = {
-    val marc245Field = bibData
-      .nonrepeatableVarfieldWithTag("245")
```
paul-butcher, line 27: This is the only place where this function from SierraQueryOps in scala-libs is being used.  I have inlined it instead.

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/data/BibDataAsMarcRecord.scala
```diff
@@ -0,0 +1,48 @@
...
+}
+import weco.sierra.models.SierraQueryOps
+import weco.sierra.models.data.SierraBibData
+import weco.sierra.models.marc.{Subfield, VarField}
+
+trait SierraMarcDataConversions {
+  import scala.language.implicitConversions
+
+  /* TODO:
+   * This implicit conversion stuff exists to temporarily implement SierraBibData
```
kenoir, line 16: I cannot remember which other project made use of `SierraBibData` necessitating it being in `wellcomecollection/scala-libs` ... but wherever it was it's [not there anymore](https://github.com/search?q=org%3Awellcomecollection+SierraBibData&type=code).

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraTitle.scala
kenoir: Gotcha, shifting [this](https://github.com/wellcomecollection/scala-libs/blob/6070b4d97408044eac1445836ae2e03e77fd980d/sierra/src/main/scala/weco/sierra/models/SierraQueryOps.scala#L31), over [here](https://github.com/wellcomecollection/catalogue-pipeline/pull/2567/files#diff-896a4df57aa2b93311fde85fa143bed6509fb829f5ed8ca343e736a3981d9ec5R31).

pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/data/BibDataAsMarcRecord.scala
paul-butcher: Yes, There remains a small amount of code (to do with Items) in this part of scala-libs that is shared between Pipeline and API, but not enough to warrant the current shape of it all.

kenoir (approved):

kenoir:
Part of https://github.com/wellcomecollection/platform-infrastructure/issues/426

PR #2570, opened 2024-03-04
Add EBSCO merge candidate

kenoir:
## What does this change?

This change adds an `EbscoAltLookup` identifier type and extracts merge candidates in the sierra transformer from the MARC control number: https://www.loc.gov/marc/bibliographic/bd001.html, as a step towards receiving e-journal data directly from EBSCO rather than via Sierra.

There is an extra check on MARC 003 the "control number identifier" field that identifies the source system as "EBZ" where we want to merge to as an extra check, protect us against any future records having a matching identifier pattern.

Part of https://github.com/wellcomecollection/platform-infrastructure/issues/426

## How to test

- [x] Run the tests!

Works created from Sierra bib records imported as EBSCO e-resources have merge candidates including the EBSCO lookup number, enabling us to redirect from Sierra records to EBSCO records in the future.

We must pick a unique identifier reliably available on all records we wish to merge from both systems. Investigating the data available in Sierra and from EBSCO indicate the `EbscoAltLookup` field meets this criterion.

### DIFF
--- a/pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraMergeCandidatesTest.scala
+++ b/pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraMergeCandidatesTest.scala
@@ -311,6 +311,25 @@ class SierraMergeCandidatesTest
         )
       )
     }
+
+    it("only creates an EBSCO merge candidate if the control number matches the expected format") {
+      // These are all control numbers that we've seen in Sierra records
+      val bibDataA = bibDataWith001("23734725")
+      getMergeCandidates(bibDataA) should be(empty)
+
+      val bibDataB = bibDataWith001("EPH607A:215")
+      getMergeCandidates(bibDataB) should be(empty)
+
+      val bibDataC = bibDataWith001("SA/MWF/E")
+      getMergeCandidates(bibDataC) should be(empty)
+
+      val bibDataD = bibDataWith001("978-1-4939-0320-7")
+      getMergeCandidates(bibDataD) should be(empty)
+
+      // Hypothetical control numbers with the right prefix and suffix, but not the right format
+      val bibDataE = bibDataWith001("ebse")
+      getMergeCandidates(bibDataE) should be(empty)
+    }
   }
 
   describe("physical/digital and single-page merges across works") {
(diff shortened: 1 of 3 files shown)


pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/SierraMergeCandidatesTest.scala
```diff
@@ -284,9 +296,46 @@ class SierraMergeCandidatesTest
     }
   }
 
+  describe("EBSCO/Sierra e-resource") {
```
kenoir, line 299: This is the interesting bit, the rest is down to auto-formatting :(

jamieparkinson (approved):

paul-butcher:
In itself, LGTM, but I'm a little concerned about[this line](https://github.com/wellcomecollection/catalogue-pipeline/blob/97e4a43ad57b26d7e933c1f2708716a46eedd3ff/pipeline/merger/src/main/scala/weco/pipeline/merger/services/IdentifiedWorkLookup.scala#L41) in the merger.

I can't see this message in any logs (maybe expired, so it doesn't prove anything), but I fear that if a record has a mergeCandidate that does not yet exist, it could stall in the merger.

I think that's a normal and correct thing to do during a reindex, as it doesn't make sense for a set of records to progress down the pipeline until they are all there, as the work would have to be done again when the last record in the match appears.  Under normal circumstances, I wouldn't expect this to be a problem in normal running either, because the other end of the link is almost certain to be there.

That said, we have [a test](https://github.com/wellcomecollection/catalogue-pipeline/blob/58e30172de8bed1511af0de2e327cb8b002a6a24/pipeline/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala#L738) that implies that a missing MIRO record is not fatal to merging the corresponding CALM/Sierra pair, so perhaps it doesn't wait for them all to exist.

kenoir:
> In itself, LGTM, but I'm a little concerned about[this line](https://github.com/wellcomecollection/catalogue-pipeline/blob/97e4a43ad57b26d7e933c1f2708716a46eedd3ff/pipeline/merger/src/main/scala/weco/pipeline/merger/services/IdentifiedWorkLookup.scala#L41) in the merger.

@paul-butcher I think this is safe because in the case that the matcher has not yet seen an EBSCO sourced work, it will not emit a `MatcherResult` with a set of `MatchedIdentifiers` (as it can't know the Work id for a work that does not exist) although the matcher will start building a graph that includes it.

See: https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala#L141

PR #2573, opened 2024-03-06
Common marc notes

paul-butcher:
## What does this change?

This brings the implementation of Notes transformations out of Sierra and into Common Marc.

This one is notable, as there are some 

In terms of inputs and outputs, this should be a straight refactor of the existing SierraNotes.

## How to test

Run tests.

This demonstrates a model for dealing with a transformation target (Note) whose transformation is mostly covered by the Common transformer, but has some source-specific behaviours (In this case, 787, which is specifically about linking b-numbers; 59x fields, which are explicitly for "local notes"; and the 514 condition triggered on mattype=k, which allows me to delay thinking about [finding and harmonising materialTypes across systems](https://github.com/wellcomecollection/catalogue-pipeline/issues/2572))


pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcNotes.scala
```diff
@@ -0,0 +1,132 @@
...
+}
+import weco.pipeline.transformer.text.TextNormalisation._
+
+import java.net.URL
+import scala.util.Try
+trait MarcNotes extends Logging {
+
+  protected val globallySuppressedSubfields: Set[String] = Set("5")
+
+  val notesFields: Map[String, MarcField => Note] = Map(
```
paul-butcher, line 17: This list is yanked from SierraNotes.  The significant differences being that I've moved the privacy condition into createOwnershipNote, and removed (moved) the Sierra-specific entries.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcNotes.scala
```diff
@@ -0,0 +1,132 @@
...
+    "561" -> createOwnershipNote,
+    "562" -> createNoteFromContents(NoteType.GeneralNote),
+    "563" -> createNoteFromContents(NoteType.BindingInformation),
+    "580" -> createNoteFromContents(NoteType.GeneralNote),
+    "581" -> createNoteFromContents(NoteType.PublicationsNote),
+    "585" -> createNoteFromContents(NoteType.ExhibitionsNote),
+    "586" -> createNoteFromContents(NoteType.AwardsNote),
+    "588" -> createNoteFromContents(NoteType.GeneralNote)
+  )
+  def toNotes(
```
paul-butcher, line 53: I didn't like the way apply worked before, with the map creating field/transformer pairs due to the need to check a few conditions beforehand.

I think this is neater.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcTitle.scala
```diff
@@ -0,0 +1,84 @@
+package weco.pipeline.transformer.marc_common.transformers
+import grizzled.slf4j.Logging
+import weco.pipeline.transformer.exceptions.ShouldNotTransformException
+import weco.pipeline.transformer.marc_common.models.{MarcRecord, MarcSubfield}
+
+object MarcTitle extends MarcDataTransformer with Logging {
+
+  type Output = Option[String]
+
+  // Populate wwork:title.  The rules are as follows:
```
kenoir, line 10: Nit:
```suggestion
  // Populate work:title.  The rules are as follows:
```

kenoir (approved):
I can see folk contributing to this in the future perhaps having difficulty pinning down what is Sierra specific and what is not, but this is perhaps unavoidable and this PR does a good job of making that split.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcTitle.scala
paul-butcher: Apparently this means Wellcome Work, from the days of the original ontology development.  However, I can't find that written down anywhere, and I'm not sure that it's at all pertinent.  However, it is ubiquitous (https://github.com/search?q=org%3Awellcomecollection%20wwork&type=code)

PR #2574, opened 2024-03-07
Add safeguarded access status

jamieparkinson:
This is in service of https://github.com/wellcomecollection/catalogue-api/issues/753

As well as handling the Sierra access status, it adds a new one to our own model in order to represent the collections access policy. It shouldn't change any behaviour - next I will need to update the rules for requesting in the items API, then work out how/whether to reflect the new status on the website.

### DIFF
--- a/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
+++ b/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
@@ -276,6 +276,20 @@ object SierraItemAccess extends SierraQueryOps with Logging {
           note = Some(message)
         )
 
+        //
+      case (
+            _,
+            Some(Status.Safeguarded),
+            Some(OpacMsg.ByApproval),
+            NotRequestable.SafeguardedItem(message),
+            _
+          ) =>
+        AccessCondition(
+          method = AccessMethod.NotRequestable,
+          status = Some(AccessStatus.Safeguarded),
+          note = Some(message)
+        )
+
       // If an item is on hold for another reader, it can't be requested -- even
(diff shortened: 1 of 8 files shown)


common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
```diff
@@ -276,6 +276,20 @@ object SierraItemAccess extends SierraQueryOps with Logging {
           note = Some(message)
         )
 
+      //
+      case (
+            _,
+            Some(Status.Safeguarded),
+            Some(OpacMsg.ByApproval),
```
kenoir, line 283: Is the correct opac message a requirement to result in this access condition? Can we safely omit it and rely solely on the status? I'm guessing the answer is _yes_ and i've missed some context, so not blocking.

kenoir (approved):
One comment asking for a little bit more information, but this looks good. I'm guessing this blocks a re-index until we do the work in the catalogue items API?

common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
jamieparkinson: No this is a good question - I put the opac message in initially because we match on both for all the other access statuses. I thought about it a bit more and I reckon the reason we do that is so that we control very tightly on how we map access statuses; the opac message and the status should always come together and so if they don't I think we'd regard it as a data issue.

jamieparkinson:
This doesn't block a reindex fortunately because (a) I'm pretty sure it wouldn't break anything and (b) there is only one record with the status currently!

PR #2577, opened 2024-03-08
Remove tox to test lambdas, use docker-compose to orchestrate

kenoir:
## What does this change?

This change removes the need to use `tox` to test the EBSCO adapter lamdba, which was causing issues with pinned versions of dependencies not being available for certain versions of python. 

Removing the dependency on `tox` simplifies the build process and using docker & docker-compose improves this (and allows us to test with the version of python the lambda will be deployed at).

This change should be rolled out to the other python lambdas in this repository in a future PR, and expanded to incorporate the build & publish step. Using the `content-api` build as inspiration, we could move the `Dockerfiles` and compose further up and share them for all projects of a similar type.

## How to test

- [x] Does this build pass?

Less failures during the build process, and a better developer experience in general.


ebsco_adapter/ebsco_adapter/test.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.8
```
jamieparkinson, line 1: We are liable to get bitten by Docker Hub rate limits, elsewhere we use the ECR Public Gallery mirrors of docker hub official images
```suggestion
FROM public.ecr.aws/docker/library/python:3.8-slim
```

jamieparkinson: Also - can we use a more recent version (for the Lambda as well)?

jamieparkinson (approved):

PR #2578, opened 2024-03-08
Use the given access note for safeguarded items

jamieparkinson:
The first attempt at this used a hardcoded access note, but the items should have more specific notes on them (as with other statuses), so this PR changes the logic to use the given access note for Safeguarded items.

Explained by the text - these items will have specific instructions regarding access which will live in the item display note. We already filter out irrelevant display notes so this will work much like other manual request items.


kenoir (approved):
It might be useful to have a bit more context in the PR description, even if it's restating what would appear obvious.

PR #2579, opened 2024-03-11
Transformer marc xml

paul-butcher:
## What does this change?

This introduces a MarcXML transformer library that exercises the common MARC transformations.

The immediate intent is just to prove that the abstraction and interface defined in transformer.marc.common are appropriate to allow us to transform data from both formats.

There are a few lines that are EBSCO-specific in here, which relate to the creation of a document identifier.  These should eventually be moved to the proper EBSCO transformer when it is implemented.

## How to test

This work is still abstract, in the sense that there is no application wired up to exercise it with live data and a real pipeline.

As such, it is only testable through the test suite..

As more fields are moved from the Sierra transformer to the common MARC transformer, it should be simple to incorporate them into the MARC XML transformer by calling them from `MarcXMLRecordTransformer` and devising some new tests against XML data. 

This work is a mitigation of risk in itself.  There was a risk that refactoring Sierra transformer objects and moving them into `marc.common` could bring along some Sierra~ or JSON~ specific practices, thereby not making the EBSCO/XML transformer as simple to implement as hoped for.  Similarly, if implemented directly as part of the upcoming EBSCO transformer, we may have had to do some of this extraction work again in the future when we start ingesting other sources of MARC XML.

### DIFF
--- a/pipeline/transformer/transformer_marc_xml/src/main/scala/weco/pipeline/transformer/marc/xml/transformers/MarcXMLRecordTransformer.scala
+++ b/pipeline/transformer/transformer_marc_xml/src/main/scala/weco/pipeline/transformer/marc/xml/transformers/MarcXMLRecordTransformer.scala
@@ -1,0 +1,43 @@
+package weco.pipeline.transformer.marc.xml.transformers
+
+import weco.catalogue.internal_model.identifiers.{
+  DataState,
+  IdentifierType,
+  SourceIdentifier
+}
+import weco.catalogue.internal_model.work.WorkState.Source
+import weco.catalogue.internal_model.work.{Work, WorkData}
+import weco.pipeline.transformer.marc.xml.data.MarcXMLRecord
+import weco.pipeline.transformer.marc_common.transformers.MarcTitle
+
+import java.time.Instant
+
+object MarcXMLRecordTransformer {
+
+  def apply(record: MarcXMLRecord): Work.Visible[Source] = {
+    // TODO: The state stuff here is EBSCO-specific.  Move it when the EBSCO transformer is implemented
+    val state = Source(
+      sourceIdentifier = SourceIdentifier(
+        identifierType = IdentifierType.EbscoAltLookup,
+        ontologyType = "Work",
+        value = record.controlField("001").get
+      ),
+      // TODO: I don't think we get sourceModifiedTime in the XML records from EBSCO,
+      // but we might be able to work something out
+      sourceModifiedTime = Instant.now
+    )
+    Work.Visible[Source](
+      version = 0,
+      state = state,
+      data = workDataFromMarcRecord(record)
+    )
+  }
+
+  def workDataFromMarcRecord(
+    record: MarcXMLRecord
+  ): WorkData[DataState.Unidentified] = {
+    WorkData[DataState.Unidentified](
+      title = MarcTitle(record)
+    )
+  }
+}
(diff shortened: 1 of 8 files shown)


pipeline/transformer/transformer_marc_xml/src/main/scala/weco/pipeline/transformer/marc/xml/transformers/MarcXMLRecordTransformer.scala
```diff
@@ -0,0 +1,43 @@
...
+import weco.catalogue.internal_model.work.{Work, WorkData}
+import weco.pipeline.transformer.marc.xml.data.MarcXMLRecord
+import weco.pipeline.transformer.marc_common.transformers.MarcTitle
+
+import java.time.Instant
+
+object MarcXMLRecordTransformer {
+
+  def apply(record: MarcXMLRecord): Work.Visible[Source] = {
+    // TODO: The state stuff here is EBSCO-specific.  Move it when the EBSCO transformer is implemented
```
kenoir, line 18: Perhaps we should just deal with this as an "EBSCO Transformer" until we have cause to deal with another MARCXML source?

kenoir:
An aside: Do you think we should validate the MARCXML against [a schema](https://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd)? If so should this be at the adapter or transformer? My initial feeling is that we can be fairly permissive and define any restrictions in code, and that we may run into unforeseen issues that are hard to resolve (short of chucking a schema) using one if we encounter source data that doesn't validate.

kenoir (approved):

paul-butcher:
> An aside: Do you think we should validate the MARCXML against [a schema](https://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd)? If so should this be at the adapter or transformer? My initial feeling is that we can be fairly permissive and define any restrictions in code, and that we may run into unforeseen issues that are hard to resolve (short of chucking a schema) using one if we encounter source data that doesn't validate.

I'm in two minds about this. The format is so simple that a provider would have to be doing something rather odd to fail validation, so perhaps we should.  Then again, the simplicity of the format and the way we use it means that I don't think we'd get any significant benefit from validating against a schema.

Because we are looking in specific fields and subfields for certain data, we can happily accept and discard any old gibberish, as long as the stuff we want is in there.  It's possible that a schema would make it easier to spot homographs in tag names or  a malformed leader (if we start trying to process it), but I don't think we need it for the overall element structure.

pipeline/transformer/transformer_marc_xml/src/main/scala/weco/pipeline/transformer/marc/xml/transformers/MarcXMLRecordTransformer.scala
paul-butcher: I thought of this, and it might turn out that way.  I started writing this as EBSCO-specific, then made it purely source-independent thing, but then realised I needed an id with an IdentifierType, so this bit of EBSCO had to stay.

Ideally, as with the Sierra changes, I'd like us to understand the difference between the quirks of a source and the general behaviour of a format.

kenoir: > so this bit of EBSCO had to stay.

Interestingly I wondered whether pulling the source identifier from control field 001 is something that might be source independent if we had multiple sources. Given the field definition as:

> For interchange purposes, documentation of the structure of the control number and input conventions should be provided to exchange partners by the organization initiating the interchange.

We don't do this for Sierra but it looks like the bnumber is in varField 001 for those records created in Sierra.

> understand the difference between the quirks of a source and the general behaviour of a format

Would we treat MARCXML any differently from MARC21 in terms of general behaviour? My mental model is we are parsing MARC-ish data from multiple sources (Sierra/EBSCO) into our internal representation of MARC and we can pick which fields from a source need special treatment otherwise falling back to the "common" MARC transform for that field. 

Hope that makes sense.

paul-butcher: I think pulling stuff out of 001 should be source-independent, but calling it an EbscoAltLookup is (obvs) EBSCO-specific.   I also think that trying to find the last modified time somewhere and falling back to some default that isn't in the record itself (Now, or file creation time) should also be source-independent.

PR #2582, opened 2024-03-11
Update & test all python lambdas with docker-compose

kenoir:
## What does this change?

Follows https://github.com/wellcomecollection/catalogue-pipeline/pull/2577, uses the same pattern to test all python lambdas, and bumps the EBSCO adapter python version to 3.10.

We also split the test & build steps so we can containerise them seperately.

This change is intended to start aligning the test methods of our python lambdas and removing tox as preparation for further work on the EBSCO lambda. 

> [!NOTE]
> 🧠 malfunction withe the EBSCO adapter, 3.12 is the highest lambda runtime AWS supports, we should move to that for all lambdas in the near future.

## How to test

- [x] Do the tests pass?

We should test our python lambdas the same way everywhere with the simplest setup we can, making it easier to bump python version.

Some existing lambdas are not well tested, we could break things. To mitigate we should add at least skeleton tests where non exist.


calm_adapter/calm_window_generator/src/test_window_generator.py
```diff
@@ -0,0 +1,34 @@
...
+    publish_calls = None
+
+    def publish(self, TopicArn, Message):
+        # create or append to list of publish calls
+        if self.publish_calls is None:
+            self.publish_calls = []
+
+        self.publish_calls.append((TopicArn, Message))
+
+        pass
```
paul-butcher, line 16: This pass looks extraneous.

paul-butcher (commented):
just one nit, otherwise LGTM.

paul-butcher (approved):

PR #2587, opened 2024-03-13
Marc edition

paul-butcher:
## What does this change?

Transforms the Edition Statement from both MarcXML and Sierra with the same code.


jamieparkinson (approved):
Started writing a comment suggesting you use a typeclass and then got a grip on reality

paul-butcher:
> Started writing a comment suggesting you use a typeclass and then got a grip on reality

TBH it might end up going that way eventually.  Part of this work is to reduce the dependency on SierraQueryOps in Sierra-libs.  Where some functionality over there is only used in one place over here, I'm inlining it or otherwise putting it a bit closer to the action.  Other than that, I'm currently trying to minimise the changes to actual lines of code, otherwise this all gets a bit confusing.

PR #2589, opened 2024-03-15
Marc electronic resources

paul-butcher:
## What does this change?

This creates a common MARC transformer for ElectronicResources


pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcElectronicResources.scala
```diff
@@ -0,0 +1,143 @@
...
+      .fieldsWithTags("856")
+      .flatMap(field => toItem(field))
+
+  protected def getLabel(field: MarcField): String =
+    field.subfields.filter(_.tag == "y").map(_.content).mkString(" ")
+
+  // Return title as Left or linkText as Right
+  protected def getTitleOrLinkText(
+    field: MarcField
+  ): Either[String, String] = Left(getLabel(field))
```
kenoir, line 39: Does this need a path for `Right` too?

pipeline/transformer/transformer_marc_common/src/test/scala/weco/pipeline/transformer/marc_common/transformers/MarcElectronicResourcesTest.scala
```diff
@@ -0,0 +1,68 @@
...
+import weco.pipeline.transformer.marc_common.models.{MarcField, MarcSubfield}
+
+class MarcElectronicResourcesTest
+    extends AnyFunSpec
+    with Matchers
+    with LoneElement {
+
+  private implicit val ctx: LoggingContext = LoggingContext("")
+
+  describe("extracting Electronic resourced from MARC 856 fields") {
```
kenoir, line 20: Nit:
```suggestion
  describe("extracting Electronic resources from MARC 856 fields") {
```

pipeline/transformer/transformer_marc_xml/src/main/scala/weco/pipeline/transformer/marc/xml/transformers/MarcXMLRecordTransformer.scala
```diff
@@ -8,8 +8,15 @@ import weco.catalogue.internal_model.identifiers.{
 import weco.catalogue.internal_model.work.WorkState.Source
 import weco.catalogue.internal_model.work.{Work, WorkData}
 import weco.pipeline.transformer.marc.xml.data.MarcXMLRecord
+<<<<<<< HEAD
```
kenoir, line 11: Think these need cleaning up.

pipeline/transformer/transformer_marc_xml/src/test/scala/weco/pipeline/transformer/marc/xml/transformers/MarcXMLRecordTransformerTest.scala
```diff
@@ -57,5 +66,13 @@ class MarcXMLRecordTransformerTest extends AnyFunSpec with Matchers {
     it("extracts the edition statement") {
       work.data.edition.get shouldBe "Director's cut"
     }
+
+    it("extracts an electronic resource") {
+      val resource = work.data.items.loneElement
+      resource.title.get shouldBe "Hampster Dance"
+      resource.locations.loneElement
+        .asInstanceOf[DigitalLocation]
+        .url shouldBe "https://example.com/hampsterdance"
```
kenoir, line 75: Think this will fail as ham`p`ster.

kenoir (approved):
Looks good, couple of nits, I think i've reviewed while work is still in progress so the couple of notes here are probably moot already.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcElectronicResources.scala
paul-butcher: Not yet.

The left/right thing replaces a previous 2-tuple in the Sierra transformer that decided whether to put this label in the title or the linkText based on whether the label contained an imperative verb. All the labels in the sample data appear to be the name of the resource, so they are all "titles". (Actually, according to MARC, the label is the Link Text, but since we have imperative link text, I'm calling it a title)

This is based on a historic quirk of the existing catalogue, where the label has been used for multiple purposes. As EBSCO are selling this data, I expect it will be consistent.

When we start working with actual data, we may find that we need distinct Left and Right paths in the EBSCO

PR #2594, opened 2024-03-19
Marc designation

paul-butcher:
## What does this change?

Moves Sierra Designation to the common Marc Designation

This is not quite a straight refactor, it will now reject invalid 362 fields.  This should not affect any existing records, as they are all (according to a search in reporting) currently valid.


pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcDataTransformer.scala
```diff
@@ -1,9 +1,18 @@
...
+import weco.pipeline.transformer.marc_common.logging.LoggingContext
 import weco.pipeline.transformer.marc_common.models.MarcRecord
 
 trait MarcDataTransformer {
   type Output
 
   def apply(record: MarcRecord): Output
 }
+
+trait MarcDataTransformerWithLoggingContext {
```
paul-butcher, line 12: I think I will actually just add the implicit context argument onto MarcDataTransformer.apply, for consistency, but I wanted to keep the changes localised to the current task, so I've created a separate base trait for now.

kenoir (approved):

PR #2598, opened 2024-03-21
Marc alternative titles

paul-butcher:
## What does this change?

Moves Sierra Alternative Titles to the common Marc Alternative Titles

There's a small part of this that is kinda Sierra-specific, (the UkLW stuff), but it would only be problematic in a very unlikely scenario:
* we start ingesting MARC data from another publisher that uses the string "UkLW" in subfield 246$5 for a different meaning 
* and we want to include that value in alternative titles from them, but not from Sierra

Even in that scenario, it's not a significant problem, and we'd simply have to create a source-specific transformer for that field, which is what I'm not bothering to do now in mitigation.


kenoir (approved):
🎼 👌

PR #2600, opened 2024-03-22
Send the Scala dependency graph

kenoir:
## What does this change?

Follows https://github.com/wellcomecollection/catalogue-api/pull/766 this change pushes the scala dependency graph to the GitHub API to surface vulnerabilities.

We don't need access to S3 as with other similar pull requests as we're not pulling our own deps (we build them here).

Part of https://github.com/wellcomecollection/platform-infrastructure/issues/431

Depends on https://github.com/wellcomecollection/aws-account-infrastructure/pull/19

## How to test

- [ ] Does the build go green, can we see the dep graph after merging?

Visibility on dependencies and vulnerabilities. 

This change should help manage / reduce risk.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -348,7 +348,7 @@ lazy val tei_adapter = setupProject(
 s3CredentialsProvider := {
   _ =>
     val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
-      "arn:aws:iam::760097843905:role/platform-ci",
+      "arn:aws:iam::760097843905:role/terraform-20210811133135108800000001",
       UUID.randomUUID().toString
     )
     builder.build()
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -348,7 +348,7 @@ lazy val tei_adapter = setupProject(
 s3CredentialsProvider := {
   _ =>
     val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
-      "arn:aws:iam::760097843905:role/platform-ci",
+      "arn:aws:iam::760097843905:role/terraform-20210811133135108800000001",
```
jamieparkinson, line 351: obviously not for this PR but is there a way of giving this role a nicer name?

jamieparkinson (approved):

PR #2606, opened 2024-04-08
Read MARCXML from EBSCO FTP server

kenoir:
## What does this change?

This change implements part of the EBSCO MARCXML adapter, syncing files from their FTP server and breaking up work into MARCXML records identified by the value of control field 001.

Part of https://github.com/wellcomecollection/catalogue-pipeline/issues/2552

## How to test

- [x] Run locally with developer credentials and appropriate developer permissions
- [x] Run the tests

We are able to send notifications for new e-resource records provided by the EBSCO FTP feed.

### Remaining work

There are a number of improvement we could make in future PRs:

- Extract notification flags into their own module
- A number of modules here could be turned into classes
- Extract main execution outside lambda_handler and test end-to-end
- Generally improve test coverage


jamieparkinson (commented):
Interesting for me to review "serious" Python - not knowing the type signature of functions makes it a rather different experience

ebsco_adapter/ebsco_adapter/src/compare_uploads.py
```diff
@@ -0,0 +1,134 @@
...
+def find_updated_records(current_records, previous_records):
+    for id, record in current_records.items():
+        if id not in previous_records:
+            yield id, record
+        elif record["sha256"] != previous_records[id]["sha256"]:
+            yield id, record
+
+
+def find_deleted_records(current_records, previous_records):
+    for id, record in previous_records.items():
```
paul-butcher, line 60: This could be a generator expression, and we can use .keys() instead of .items, because we are ignoring the record at this point.

```suggestion
   return (id for id in previous.keys() if id not in current)
```

ebsco_adapter/ebsco_adapter/src/compare_uploads.py
```diff
@@ -0,0 +1,134 @@
...
+    available_files_dates = list(available_files.keys())
+    dates_list = sorted(available_files_dates, reverse=True)
+
+    print(f"Available uploads: {dates_list}")
+
+    # Check if we've sent a notification for the date
+    notified_completion_flag = "notified.flag"
+    date_list_with_notified_flag = []
+    for date in dates_list:
+        notified_completion_flag_path = os.path.join(
```
paul-butcher, line 14: I'd be tempted to put the body of this loop into a function so that date_list_with_notified_flag could be populated like this:
```
date_list_with_notified_flag = [{"date": date, "notified_completed": is_completed(date)} for date in dates_list]
```

paul-butcher (approved):
LGTM, just a couple of minor suggestions you can take or leave.

PR #2612, opened 2024-04-19
Add EBSCO transformer skeleton

kenoir:
## What does this change?

This change builds on https://github.com/wellcomecollection/catalogue-pipeline/pull/2606, to implement an EBSCO transformer that consumes messages from the adapter. Here we only add a skeleton for the application so it is deployable but does no work.

Part of https://github.com/wellcomecollection/catalogue-pipeline/issues/2552

## How to test

- [x] Do the tests pass, does it compile?
- [x] Does the terraform associated with this PR apply cleanly?
- [ ] Does this PR have everything needed to deploy the transformer application and to receive messages from the adapter?

A foundation to start processing messages from the EBSCO adapter and start processing them using the recent MARCXML work.

This change should not impact any other parts of a deployed pipeline. We should ensure it does not send messages on, or modify any data in an existing deployed pipeline.


paul-butcher (approved):

paul-butcher:
LGTM as a skeleton.  As it does nothing, it might be good to merge it and then all the work that actually does stuff isn't mixed in with the boilerplatey bits.

PR #2615, opened 2024-04-23
EBSCO adapter message sending fix

kenoir:
## What does this change?

This change adds a test for the `update_notifier` and fixes a bug where we were sending too many messages due to an indentation error around a loop.
 
## How to test

- [x] Run this in production, does it send the expected number of messages?

A functioning adapter sending the expected number of messages.

### DIFF
--- a/ebsco_adapter/ebsco_adapter/src/test_fixtures.py
+++ b/ebsco_adapter/ebsco_adapter/src/test_fixtures.py
@@ -1,4 +1,33 @@
+import json
 import os
+
+
+class FakeSnsClient:
+
+    def __init__(self):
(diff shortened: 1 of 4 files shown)


ebsco_adapter/ebsco_adapter/src/test_fixtures.py
```diff
@@ -1,6 +1,31 @@
+import json
 import os
 
 
+class FakeSnsClient:
```
jamieparkinson, line 5: Worth looking at `moto` if we start writing more of these I imagine

jamieparkinson (approved):

PR #2620, opened 2024-04-26
Marc concepts

paul-butcher:
## What does this change?

Part of (part of) https://github.com/wellcomecollection/catalogue-pipeline/issues/2563

This introduces the transformation of subjects to the common and XML MARC libraries.

This has been extracted from a larger piece of work that also implements the corresponding Sierra transformations by using the common transformer, but due to the scale of the inconsistencies there, I thought it would be better to split the PR into two main parts.  First, this one - implementing the common and xml versions, then the other to use the common transformer in the Sierra. 

There are still some inconsistencies present here. This is because I want to preserve the existing behaviour as much as possible in order to be able to use the existing test suite to have confidence in the refactoring.  There remain some questions to be answered as to how these inconsistencies are to be resolved.

## How to test

Run the test suite.  The suite for this part of the common marc library is not yet very comprehensive. I have mostly been working on it using the Sierra suite before isolating these changes.  This will be dealt with in an upcoming PR.

* We will have subjects in documents extracted by the XML transformer
* It will be relatively easy to understand the upcoming changes to the Sierra transformer
* It will be relatively simple to maintain a consistent set of behaviours WRT Concepts, Genres, and Agents.

Leaving the inconsistencies unresolved may be seen as simply kicking the can down the road.  However, I think I have isolated them in such a way that once we decide the correct single behaviour, they will be easy to resolve.

### DIFF
--- a/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
+++ b/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
@@ -1,0 +1,32 @@
+package weco.pipeline.transformer.marc_common.transformers
+import weco.catalogue.internal_model.identifiers.IdState
+import weco.catalogue.internal_model.work.Subject
+import weco.pipeline.transformer.marc_common.models.MarcRecord
+import weco.pipeline.transformer.marc_common.transformers.subjects.{
+  MarcConceptSubject,
+  MarcMeetingSubject,
+  MarcOrganisationSubject,
+  MarcPersonSubject,
+  MarcSubject
+}
+
+object MarcSubjects extends MarcDataTransformer {
+
+  override type Output = Seq[Subject[IdState.Unminted]]
+
+  private val transformers: Map[String, MarcSubject] = (
+    Seq("650", "648", "651").map(marcTag => marcTag -> MarcConceptSubject)
+      ++ Seq(
+        "600" -> MarcPersonSubject,
+        "610" -> MarcOrganisationSubject,
+        "611" -> MarcMeetingSubject
+      )
+  ).toMap
+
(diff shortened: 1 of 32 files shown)


pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
```diff
@@ -0,0 +1,32 @@
...
+import weco.pipeline.transformer.marc_common.models.MarcRecord
+import weco.pipeline.transformer.marc_common.transformers.subjects.{
+  MarcConceptSubject,
+  MarcMeetingSubject,
+  MarcOrganisationSubject,
+  MarcPersonSubject,
+  MarcSubject
+}
+
+object MarcSubjects extends MarcDataTransformer {
```
paul-butcher, line 13: This behaviour differs from Sierra, which does something weird WRT order of subjects - see [Slack](https://wellcome.slack.com/archives/C02ANCYL90E/p1713877204246529)

pipeline/transformer/transformer_marc_common/src/test/scala/weco/pipeline/transformer/marc_common/transformers/MarcOrganisationTest.scala
```diff
@@ -64,7 +64,7 @@ class MarcOrganisationTest
           )
       }
 
-      val nonLabelSubfields = Seq("n", "0", "u") map {
+      val nonLabelSubfields = Seq("n", "2", "u") map {
```
paul-butcher, line 67: Using 0 here caused its value to be interpreted as an identifier, clouding the actual point of this test.

kenoir (approved):
I won't pretend that this is a thorough review, but I can follow what is happening here and understand why it's being done, so gets a +1 from me.

PR #2622, opened 2024-04-26
6xx bottom up pt 2

paul-butcher:
## What does this change?

This moves the (I think) final common fields - Subjects and Genres into the common MARC library.

## How to test

It may be prudent to run this on a new pipeline and compare with the existing one.  This has been a bit of a large piece of work, and as I have worked through it, I am not 100% confident that the existing Sierra test suite is without holes.  That said, I do believe this has preserved all relevant behaviour (save for one, see comment inline, where the conflicting behaviour was too awkward to accommodate)

Genres and Subjects will now be extracted from MARC XML data.

Due to the intertwined nature of Subjects and Genres, along with the various inconsistencies in the way diverse Subjects were implemented, this is quite a large change.  I have already extracted and harmonised a lot of this content in previous PRs (https://github.com/wellcomecollection/catalogue-pipeline/pull/2620 and https://github.com/wellcomecollection/catalogue-pipeline/pull/2611), but this is where the "rubber hits the road", as it now implements the Sierra transformer in terms of the Marc transformer.

There is still potential to refactor this further, as well as taking decisions on conflicting behaviours.  I believe that I have isolated these sufficiently that once we choose one consistent way, we can make the changes easily


pipeline/transformer/transformer_sierra/src/test/scala/weco/pipeline/transformer/sierra/transformers/subjects/SierraOrganisationSubjectsTest.scala
```diff
@@ -228,20 +230,12 @@ class SierraOrganisationSubjectsTest
     // throw and have somebody inspect the record manually.  It's not at all
     // clear what we should do, as we don't have enough to populate the
     // Organisation label.
-    it("errors if there's nothing in subfield a or b") {
+    it("returns zero subjects if there's nothing in subfield a or b") {
```
paul-butcher, line 233: This is a substantive change to the behaviour.  I couldn't see a simple way to reconcile throwing here with returning Nil elsewhere, and I decided it wouldn't be wroth the effort.

Given that I've never seen this error, and our normal modus operandi is to warn-and-ignore, I thought I'd go with this approach.

kenoir (approved):

PR #2626, opened 2024-04-29
fix calling of reindex script

paul-butcher:
## What does this change?

There was a typo in the reindeer script (exit/exist) and the mode option did not prompt for input, so needed to be provided as a CLI argument with the initial call.

## How to test

Run the reindexer with no arguments, it should eventually prompt you to provide a mode (complete, partial, specific)


jamieparkinson (approved):
oops

PR #2627, opened 2024-04-29
enable adapter to find born-digital METS files

paul-butcher:
## What does this change?

The METS file in a born digital bag uses a different filename pattern to the digitised bags.  Born-digital bags were therefore being rejected because their METS file could not be found.  This fixes that. (https://github.com/wellcomecollection/catalogue-pipeline/issues/2623)

As a result of this fix, I expect born-digital records to start successfully passing through the adapter.  

Once deployed, we will need to repopulate those records that were erroneously rejected, e.g. 

`python populate_mets/populate_mets.py specific born-digital/PPMDM/A/5`

By doing this manually for a handful of records, we will be able to see whether this bug has been masking any other issues downstream.


kenoir (approved):

paul-butcher:
[Slack](https://wellcome.slack.com/archives/C02ANCYL90E/p1714476767428289)

PR #2630, opened 2024-04-30
Better handling of long queries

jamieparkinson:
The diff for this PR is confusing so I will outline the changes below. There were some defunct duplicate JSON files which I've also removed (hence the uneven diff).

- Changes the analysis for `images.identifiers.value` from text to lowercase-normalised keyword, hopefully addressing https://github.com/wellcomecollection/catalogue-api/issues/773
- Configures the following analyzers with a [token limit](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-limit-token-count-tokenfilter.html) of 75 tokens. This applies when both queries _and_ documents are analyzed (ie document fields are truncated after 75 tokens).
    - german
    - spanish
    - swappable characters _(this is used for things like vv => w, v => u in old manuscripts etc)_
    - italian
    - persian
    - cased
    - arabic
    - bengali
    - hindi
    - french
    - base
- Adds duplicates of the remaining analyzers to be used _at query time only_ with a token limit of 75 tokens:
  - english
  - lowercase

The effect of the latter 2 changes is that all fields are fully analyzed in `english` and `lowercase`, but for (a) queries and (b) all other analyzers, analysis is limited to 75 tokens. The intention of this is to achieve a compromise which accommodates:
- A need for a limit on analyzed query length to preserve search performance
- Manageable and minimally-duplicated index config (ie not having a token-limited duplicate of each analyzer for use as a `search_analyzer`)
- An understanding that there needs to be _some_ analysis of the entirety of every field, as for long fields (eg descriptions) there are likely to be keywords after the 75th token.
- An additional understanding that complete analysis (particularly with various language stemmers) of the entirety of every field is unnecessary
 
The 75-token limit was found empirically, decreasing it until I found that queries averaged < 10 seconds.

Relates to https://github.com/wellcomecollection/catalogue-api/issues/772 and https://github.com/wellcomecollection/catalogue-api/issues/773

Tested by https://github.com/wellcomecollection/rank/pull/102


common/internal_model/src/test/scala/weco/catalogue/internal_model/fixtures/index/WorksIndexFixtures.scala
```diff
@@ -31,8 +31,8 @@ trait WorksIndexFixtures extends IndexFixturesBase {
   def withLocalWorksIndex[R]: Fixture[Index, R] = {
     withLocalElasticSearchIndex[R](config =
       getConfig(
-        mappings = "mappings.works_indexed.2023-11-09.json",
-        analysis = "analysis.works_indexed.2023-11-09.json"
+        mappings = "mappings.works_indexed.2024-01-09.json",
```
kenoir, line 34: Do these need to be `2024-04-30`?

jamieparkinson: Oops, yes

paul-butcher (approved):
Looks good.  I like the fact that the whole query is still used.  This obviates the need to declare the truncation to the user.  This is nicely defined so that we can tweak it as needed.

PR #2631, opened 2024-05-02
Add reindex functionality to EBSCO adapter

kenoir:
## What does this change?

This changes add the ability to reindex items from the EBSCO adapter by invoking the lambda directly with a JSON payload.

It can be invoked locally using the CLI, directly or by invoking the lambda directly using a payload like:

# Partial reindex
```json
{
  "reindex_type": "partial",
  "reindex_ids": "ebz123e"
}
```

# Full reindex
```json
{
  "reindex_type": "full",
}
```

The intention is to modify the reindex script to invoke this lambda alongside the usual re-index service in a future PR.

## How to test

- [x] Run the tests
- [x] Test this locally against prod, does it behave as expected?

We can run a full re-index of the pipeline including EBSCO data.

This changes the previously untested `main.py`. We've added end to end tests here in an attempt to provide better end to end test coverage. We should also test in production with a manually deployed lambda to be sure this behaves as expected (as well as invoking the deployed lambda).

### DIFF
--- a/ebsco_adapter/ebsco_adapter/src/compare_uploads.py
+++ b/ebsco_adapter/ebsco_adapter/src/compare_uploads.py
@@ -1,13 +1,14 @@
 import os
 
 
-def find_uploads_to_compare(available_files, xml_s3_prefix, s3_store):
+def find_notified_and_completed_flag(available_files, xml_s3_prefix, s3_store):
     available_files_dates = list(available_files.keys())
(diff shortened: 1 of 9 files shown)


jamieparkinson (approved):

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/5738

PR #2633, opened 2024-05-03
Link by refno

paul-butcher:
## What does this change?

This links Archivematica METS files to CALM records via the CALM id.  It also changes the source of the identifier for Archivematica METS files to use the CALM Refno instead of the unstable UUID that gets regenerated by Archivematica on each ingest.

Should fix https://github.com/wellcomecollection/catalogue-pipeline/issues/2623 after a reindex
## How to test

In a new pipeline (this change requires a reindex), push through a CALM record and corresponding Archivematica METS record (e.g. ARTCOO/A/3/1).

The records should merge

We should get Born Digital content on the website (for those records with the appropriate access)

Because I am changing the identifier for these Archivematica-sourced records, their URLs will change.  However, they would not have been reachable anyway as METS files only result in a record at the end of the pipeline in consort with whatever they merge into, and never on their own.

The ID I have chosen is (I think) guaranteed to be unique in our context. I don't believe it is possible for two records to come through the pipeline with the same RefNo, because it corresponds to a position in a directory tree (I suppose it may be possible to put a record in the "wrong" position in the tree, but I think that would be a different problem). 

This ID is not guaranteed to be stable.  It is a value managed by humans which is meaningful - it governs the position of a record in a hierarchy - and as such _may_ change if someone decides it belongs in a different position.  However, it is considerably more stable than the UUID, and only ever changed deliberately, so if someone changes it on one side, it is not unreasonable to expect them to change it on the other.

### DIFF
--- a/pipeline/transformer/transformer_calm/src/main/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidates.scala
+++ b/pipeline/transformer/transformer_calm/src/main/scala/weco/pipeline/transformer/calm/transformers/CalmMergeCandidates.scala
@@ -41,7 +41,7 @@ object CalmMergeCandidates extends CalmRecordOps {
       .flatMap {
         id =>
           SourceIdentifier(
-            identifierType = IdentifierType.METS,
+            identifierType = IdentifierType.CalmRefNo,
             ontologyType = "Work",
             value = id
           ).validatedWithWarning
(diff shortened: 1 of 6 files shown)


kenoir:
I think this change means we no longer really use `metsIdentifier` for either `GoobiMetsXml` or `ArchivematicaMetsXML` so it may be worth just removing it entirely and refactoring so the `metsIdentifiers is removed from `MetsXml`?

kenoir (approved):

PR #2635, opened 2024-05-08
Create EBSCO holdings and perform merge

kenoir:
## What does this change?

This change creates holdings instead of items on an EBSCO sourced work and adds the EBSCO identified works to the target precedence in the merger so that matched Sierra and EBSCO works will get merged.

> [!NOTE]
> This change does not transform the `CollectionPath`, which is needed for Ebsco sourced records to have parity with Sierra source e-resource records. That should be the next change that's worked on before connecting the Ebsco transformer to the id minter in a prod pipeline.

Part of https://github.com/wellcomecollection/catalogue-pipeline/issues/2618

## How to test

- [x] Run the tests
- [ ] Can we test this in non-prod pipeline before deploying?

EBSCO sourced e-resource work pages look more similar to their current Sierra counterparts and where sierra and EBSCO works are matched, Sierra sourced works are redirected to EBSCO works.

If run in an pipeline that a production catalogue API is using this will change results for works in unexpected ways. We should test this before exposing it to public view.

### DIFF
--- a/pipeline/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
+++ b/pipeline/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
@@ -1061,6 +1061,26 @@ class PlatformMergerTest
     )
   }
 
+  describe("merging EBSCO & Sierra works") {
+    it("merges a Sierra digital work with an EBSCO work") {
+      val merger = PlatformMerger
+      val (sierraDigitalWork, ebscoWork) = sierraEbscoIdentifiedWorkPair()
+
+      val result = merger
+        .merge(works = Seq(sierraDigitalWork, ebscoWork))
+        .mergedWorksWithTime(now)
+      val redirectedWorks = result.collect {
+        case w: Work.Redirected[Merged] => w
+      }
+      val visibleWorks = result.collect { case w: Work.Visible[Merged] => w }
+
+      redirectedWorks should have size 1
+      visibleWorks should have size 1
+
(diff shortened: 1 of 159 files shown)


project/Dependencies.scala
```diff
@@ -265,8 +265,9 @@ object CatalogueDependencies {
     ExternalDependencies.jsoupDependencies ++
       ExternalDependencies.parseDependencies
 
-  val ebscoTransformerDependencies: Seq[ModuleID] =
+  val ebscoTransformerDependencies: Seq[ModuleID] = {
```
paul-butcher, line 268: All the other Dependencies vals here are unbraced.

pipeline/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
```diff
@@ -1061,6 +1061,26 @@ class PlatformMergerTest
...
+        .mergedWorksWithTime(now)
+      val redirectedWorks = result.collect {
+        case w: Work.Redirected[Merged] => w
+      }
+      val visibleWorks = result.collect { case w: Work.Visible[Merged] => w }
+
+      redirectedWorks should have size 1
+      visibleWorks should have size 1
+
+      visibleWorks.head.state.canonicalId shouldBe ebscoWork.state.canonicalId
```
paul-butcher, line 1080: Could collapse this to `visibleWorks.loneElement.state...`, but I know the `x should have size 1` ... `x.head` pattern is used elsewhere in this file, so it might be best to just put that in the TODO/Tidying bank for another time.

PR #2643, opened 2024-05-16
Add MarcParents transformer

kenoir:
## What does this change?

This change is needed to enable the Series relationship present in EBSCO sourced works. It adds a `MarcParents` transformer to replace the underlying transform in `SierraParents`. The Sierra tests for this transformer were previously called "SierraLinksTest", this PR renames for clarity.

Part of:

## How to test

- [ ] Run the tests

This change will impact existing records as it modifies the logic used by `SierraParents`. The change here removes the fallback to use the "content" of a `VarField` where the required subfields are missing, and modifies existing tests to respect that change.

However on investigation it appears that the content fallback logic is unlikely to ever be exercised. See: https://wellcome.slack.com/archives/CGXDT2GSH/p1715763567400779

We should also run this into a test pipeline and observe the difference to impacted records before going live.


paul-butcher (commented):
Should these fields be exercised in MarcXMLRecordTransformerTest.scala as well?

PR #2644, opened 2024-05-16
Add MarcProduction transformer

kenoir:
## What does this change?

This change adds a transformer for MarcProduction, and updates SierraProduction to use that. It also shifts some prrsing and exception logic that go with it.

Depends on:
- https://github.com/wellcomecollection/catalogue-pipeline/pull/2638

Part of:

## How to test

- [ ] Run the tests

TBD ... probably some!


pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcProduction.scala
```diff
@@ -0,0 +1,309 @@
...
+    marc264fields
+      .map { _.subfields.map(_.content).mkString("") }
+      .forall { _ matches "^[:,]*$" }
+
+  /** Populate the production data if both 260 and 264 are present.
+    *
+    * In general, this is a cataloguing error, but sometimes we can do something
+    * more sensible depending on if/how they're duplicated.
+    */
+  private def getProductionFromBothFields(record: MarcRecord) = {
```
paul-butcher, line 211: I know this is just a straight copy from its previous location, but having read it just now, I fancy it is ripe for a refactor at some point in the future.

1. all the (successful) branches return the same thing, (i.e. the 260) so it might be preferable to OR all these conditions, possibly in a separate "shouldDiscard264" function
2. Because they only pull 260, it's not really get from both,
3. this is only called in a place where we already have the output, this is really just choosing whether to keep the output or throw an exception, but it's regenerating the output

paul-butcher (approved):
🪢➡️🧵

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcProduction.scala
kenoir: I will have a go at this in a fresh PR.

PR #2649, opened 2024-05-22
move matcher and merger into a common parent folder

paul-butcher:
## What does this change?

This moves matcher and merger into a matcher_merger folder.  There should be no actual changes.  The resulting applications should be identical.

https://github.com/wellcomecollection/catalogue-pipeline/issues/2528

This makes it clearer that the Matcher and the Merger form one subsystem. 

This will make it simpler to move the path concatenator into that subsystem where it really belongs.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -113,14 +113,14 @@ lazy val ingestor_images = setupProject(
 
 lazy val matcher = setupProject(
   project,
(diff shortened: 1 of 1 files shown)


kenoir (approved):
👌 

Did you consider the folder name `matger`?

PR #2650, opened 2024-05-22
Add MarcFormat transformer

kenoir:
## What does this change?

This change adds a transformer for MarcFormat, only used by the EBSCO transformer as the Sierra transformer reads material type from the bibdata for this purpose. 

The material types used in format are specific to the Wellcome Collection and do not directly relate to the MARC data in our records and requires some assumptions about how to do that encoded here.

We currently only matches E-book & E-Journal formats as that's all that is required for the EBSCO source data, but the change here tries to make a start a transformer that could be used more widely to identify format from MARC in other sources.

Part of:

## How to test

- [x] Run the tests

This should only impact EBSCO sourced works.

### DIFF
--- a/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormat.scala
+++ b/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormat.scala
@@ -1,0 +1,173 @@
+package weco.pipeline.transformer.marc_common.transformers
+
+import weco.catalogue.internal_model.work.Format
+import weco.pipeline.transformer.marc_common.models.MarcRecord
+
+/** https://www.loc.gov/marc/bibliographic/bd008s.html "23 - Form of item
+  * (006/06)" The spec for field 008 contains a list of material specific
+  * details including the form of item, the code for which matches the value
+  * found in the 006 field at position 6.
+  */
+sealed trait FormOfItem
+object MarcO08Field {
+  object FormOfItem {
+    case object Online extends FormOfItem
+  }
+}
+
+// Used to transform data from a MARC field where the data is positional
+// e.g. the leader and some control fields
+sealed trait MarcPositionalElement extends MarcDataTransformer {
+  val position: Int
+  val length: Int = 1
+}
+
+/** https://www.loc.gov/marc/bibliographic/bd006.html "Additional Material
+  * Characteristics" Marc 006 field is a control field that contains additional
+  * material characteristics, based on the form of material.
+  *
+  * For example, for continuing resources, the 006 field contains the form of
+  * item at position 6, but for visual materials it's at position 12.
+  *
+  * We infer the form of material based on values in the leader field.
+  *
+  * At present we only care about the form of item for books and continuing
+  * resources, in order to indentify e-books and e-journals (o in position 6).
+  * This could be extended to other forms of material in the future.
+  */
+object MarcO06Field {
+  trait FormOfItemField extends MarcPositionalElement {
+    type Output = Option[FormOfItem]
+
+    val position = 6
+
+    def apply(record: MarcRecord): Option[FormOfItem] = {
+      record
+        .controlField("006")
+        .map(_.content)
+        .map(_.slice(position, position + length))
+        .flatMap {
+          case "o" => Some(MarcO08Field.FormOfItem.Online)
+          case _   => None
+        }
+    }
+  }
+
+  case object Books {
+    case object FormOfItemField extends FormOfItemField
+  }
+
+  case object ContinuingResource {
+    case object FormOfItemField extends FormOfItemField
+  }
+}
+
+/** https://www.loc.gov/marc/bibliographic/bdleader.html "Leader" The leader
+  * field contains information about the record as a whole. We use it to
+  * determine the type of record and the bibliographic level, which is used to
+  * determine the format of the material.
+  *
+  * For example the type of record for a continuing resource is 'a' (language
+  * material) in position 6 (type of record), and 's' (serial) in position 7
+  * (bibliographic level).
+  *
+  * At present we only care about the type of record and bibliographic level for
+  * books and continuing resources, in order to indentify e-books and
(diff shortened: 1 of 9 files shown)


pipeline/transformer/transformer_marc_common/src/test/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormatTest.scala
```diff
@@ -0,0 +1,56 @@
...
+        leader = "00000cam a22000003a 4500",
+        controlFields = Seq(
+          MarcControlField("006", "m\\\\\\\\\\o\\\\d\\\\||||||")
+        )
+      )
+
+      MarcFormat(record).value shouldBe EBooks
+    }
+
+    // We only care about the format ofe-journals and e-books at the moment
```
paul-butcher, line 40:
```suggestion
    // We only care about the format of e-journals and e-books at the moment
```

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormat.scala
```diff
@@ -0,0 +1,173 @@
...
+import weco.catalogue.internal_model.work.Format
+import weco.pipeline.transformer.marc_common.models.MarcRecord
+
+/** https://www.loc.gov/marc/bibliographic/bd008s.html "23 - Form of item
+  * (006/06)" The spec for field 008 contains a list of material specific
+  * details including the form of item, the code for which matches the value
+  * found in the 006 field at position 6.
+  */
+sealed trait FormOfItem
+object MarcO08Field {
```
paul-butcher, line 12: Same as the `O06` question - should the O be 0?

pipeline/transformer/transformer_marc_common/src/test/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormatTest.scala
```diff
@@ -0,0 +1,56 @@
...
+
+class MarcFormatTest extends AnyFunSpec with Matchers with OptionValues {
+  describe(
+    "Extracting format information from MARC Leader/07, 006/06 & 008 material specific details"
+  ) {
+    info("https://www.loc.gov/marc/bibliographic/bdleader.html")
+    info("https://www.loc.gov/marc/bibliographic/bd006.html")
+    info("https://www.loc.gov/marc/bibliographic/bd008s.html")
+
+    it("returns an e-journal format when metadata indicates it") {
```
paul-butcher, line 18: I'd like these test titles to be a bit more descriptive of the metadata that causes the result.  Particularly because the wire feed style of data makes it hard to immediately tell what is pertinent.  Perhaps something like this:

```
describe("when the seventh character of the leader is 's'")
it("returns an e-journal when the sixth character in field 006  'o'"){
}
it("returns nothing for any other value of 006[6]"){
}
}
```
etc.

It would also be possible to drive this from TableDrivenPropertyChecks, which might also help because something like `leader= s"00000ca{$bibLevel} a22000003a 4500"` would show which character(s) are relevant.  However, I'm not sure whether that would be better without actually writing it out and looking at it.

pipeline/transformer/transformer_marc_common/src/test/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormatTest.scala
```diff
@@ -0,0 +1,56 @@
...
+        )
+      )
+
+      MarcFormat(record).value shouldBe EBooks
+    }
+
+    // We only care about the format ofe-journals and e-books at the moment
+    // so we don't need to test for other formats. As this transformer
+    // is only used in the context of the EBSCO transformer.
+    it(
```
paul-butcher, line 43: IIUIC, This exercises the scenario where it is a book, but it is not online, so it's not an e-book.  We should also exercise the scenario where 006 indicates that it would be online (`006[6] == o`), were it a book, but it isn't so it's not.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormat.scala
```diff
@@ -0,0 +1,173 @@
...
+  * For example, for continuing resources, the 006 field contains the form of
+  * item at position 6, but for visual materials it's at position 12.
+  *
+  * We infer the form of material based on values in the leader field.
+  *
+  * At present we only care about the form of item for books and continuing
+  * resources, in order to indentify e-books and e-journals (o in position 6).
+  * This could be extended to other forms of material in the future.
+  */
+object MarcO06Field {
```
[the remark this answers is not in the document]
kenoir: eep.

pipeline/transformer/transformer_marc_common/src/test/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormatTest.scala
kenoir: I think the table tests are nicer, will do that.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcFormat.scala
```diff
@@ -0,0 +1,173 @@
...
+  * Characteristics" Marc 006 field is a control field that contains additional
+  * material characteristics, based on the form of material.
+  *
+  * For example, for continuing resources, the 006 field contains the form of
+  * item at position 6, but for visual materials it's at position 12.
+  *
+  * We infer the form of material based on values in the leader field.
+  *
+  * At present we only care about the form of item for books and continuing
+  * resources, in order to indentify e-books and e-journals (o in position 6).
```
paul-butcher, line 35:
```suggestion
  * resources, in order to identify e-books and e-journals (o in position 6).
```
Whoops. I missed this typo first time round

PR #2651, opened 2024-05-23
Refine subject transform based on feedback from collections information

kenoir:
## What does this change?

This change refines the rules for transforming subjects from MARC based on conversations with Collection Information.

> Our cataloguing practise encodes the following rules with respect to which headings we choose from the MARC record in 6xx fields, which we ignore and what concept type they map to.
> 
> See: https://www.loc.gov/marc/bibliographic/bd6xx.html - Subject Access
> Fields:
> - https://www.loc.gov/marc/bibliographic/bd650.html650 - Subject Added Entry - Topical Term (`MarcConceptSubject`)
> - https://www.loc.gov/marc/bibliographic/bd651.html651 - Subject Added Entry - Geographic Name (`MarcConceptSubject`)
> - https://www.loc.gov/marc/bibliographic/bd648.html648 - Subject Added Entry - Chronological Term (`MarcConceptSubject`)
> - https://www.loc.gov/marc/bibliographic/bd600.html600 - Subject Added Entry - Personal Name (`MarcPersonSubject`)
> - https://www.loc.gov/marc/bibliographic/bd610.html610 - Subject Added Entry - Corporate Name (`MarcOrganisationSubject`)
> - https://www.loc.gov/marc/bibliographic/bd611.html611 - Subject Added Entry - Meeting Name (`MarcMeetingSubject`)
> 
> We catalogue using LoC (2nd indicator 0) and MeSH (2nd indicator 2) and other (2nd indicator 7), so for the above Subject Added Entry's we do not take any headings with 2nd indicators 1, 3-6, there are particular rules on 2nd indicator 7 headings: 
> 
> We currently keep/use the following 650_7 ǂ2: local, homoit, indig, enslv

See: https://wellcome.slack.com/archives/C02ANCYL90E/p1716217517115789

## How to test

- [ ] Run the tests

This does change a transform used elsewhere, but the Sierra subject transform overrides `getSubjects`, so those records should not be impacted.

### DIFF
--- a/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
+++ b/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
@@ -54,9 +54,6 @@ object MarcSubjects extends MarcDataTransformer with MarcFieldOps {
     "651" -> MarcConceptSubject
   )
 
-  val validIndicator2Values: Seq[String] = Seq("0", "2")
-  val validSubjectMarcTags: Seq[String] = transformMap.keys.toSeq
-
   private lazy val marcTags = transformMap.keys.toSeq
   override def apply(record: MarcRecord): Output = {
     record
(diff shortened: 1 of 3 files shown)


pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
```diff
@@ -1,32 +1,87 @@
...
-      field => transformers(field.marcTag)(field)
-    }
+    record
+      .fieldsWithTags(marcTags: _*)
+      .filter {
+        field =>
+          {
+            (field.marcTag, field.indicator2) match {
+              case (tag, i2)
+                  if validSubjectMarcTags.contains(tag) && validIndicator2Values
```
paul-butcher, line 69: Do we need to check `validSubjectMarcTags.contains(tag)`, given that this is already a list only containing fields with those tags?

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
```diff
@@ -1,32 +1,87 @@
...
+                    .contains(i2) =>
+                true
+              case (tag, "7") if validSubjectMarcTags.contains(tag) =>
+                field
+                  .subfieldsWithTag("2")
+                  .exists(_.content match {
+                    case "local" | "homoit" | "indig" | "enslv" => true
```
paul-butcher, line 76: It might be nice to have some commentary about what these strings mean and why we care.

also, would something like `.exists(Seq("local", "homoit", "indig" , "enslv").contains(_.content))` be clearer?  It would allow us to name this group of strings in a way that might demonstrate its purpose.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
kenoir: Nope, we do not!

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcSubjects.scala
kenoir: Yes, and also because they're quite interesting.

paul-butcher (approved):

PR #2653, opened 2024-05-24
Always prefer MARC 260 if 260 & 264 are present transforming ProductionEvent

kenoir:
## What does this change?

This change removes the logic for throwing a `CataloguingException` in the case that a record contains valid MARC 260 & 264 fields when attempting to transform a `ProductionEvent`. Tracing back the [origin of this logic to 2018, when this code was in the platform monorepo](https://github.com/wellcomecollection/platform/pull/2205) it seems like this change was made as expected correct behaviour, but there's no link to a justification for it.

Removing this logic considerably simplifies the transformer, enables MARC records from EBSCO to be transformed without error.

> [!NOTE]
> We should verify that this behaviour is acceptable with the Collections Information team before merging. See https://wellcome.slack.com/archives/C02ANCYL90E/p1716555606347529

Following this we can [perform an EBSCO re-index](https://github.com/wellcomecollection/catalogue-pipeline/pull/2631) to bring the Sierra sourced records into the new model.

Part of:

Future work should address where we source production events from for Sierra records (see the conversation linked above).

## How to test

- [ ] Run the tests!

Simpler codebase & the EBSCO transformer can process all records without error.

There may be Sierra records that will now be accepted that were not before, however it is likely we would have noticed these during re-indexes and there is a good chance this does not impact any records.

### DIFF
--- a/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcProduction.scala
+++ b/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcProduction.scala
@@ -2,11 +2,24 @@ package weco.pipeline.transformer.marc_common.transformers
 
 import grizzled.slf4j.Logging
 import weco.catalogue.internal_model.identifiers.IdState
-import weco.catalogue.internal_model.work.{Agent, Concept, Period, Place, ProductionEvent}
+import weco.catalogue.internal_model.work.{
+  Agent,
+  Concept,
+  Period,
+  Place,
+  ProductionEvent
+}
 import weco.pipeline.transformer.marc_common.exceptions.CataloguingException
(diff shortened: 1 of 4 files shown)


paul-butcher (approved):

kenoir:
> This is definitely better if we intend to continue using 260, but from [this conversation](https://wellcome.slack.com/archives/C02ANCYL90E/p1716561577413829?thread_ts=1716555606.347529&cid=C02ANCYL90E), it looks like we should be moving to preferring (or even only using) `264`. If that is the case, I assume we would still need to filter out unwanted `264`s by the same criteria currently used in `shouldDiscard264`
> 
> I know that could potentially be a larger piece of work, with us needing to examine and analyse the real data to see if we would lose anything by discarding `260`.
> 
> I'm happy to approve this with that caveat.

👍 I'll see about getting work prioritised to look at this in more detail.

PR #2654, opened 2024-05-28
Do not show reserve notes

### LINKED ISSUE #2584: Don't show Reservation Notes
See this example, from https://wellcomecollection.org/works/uxrgqnek

In the Note within the "Where to find it" section, only the final line, which is the Shelfmark 999 should be shown.  The lines starting with "Exhibited" are all Reservation Notes, which contain internal information that is not expected to make sense outside of the organisation.

Screenshot 2024-03-12 at 14 41 56

The internal information is not confidential, so that does not put any extra urgency on this fix. It is just that the field there should only tell the user where it is, not where it has been.  That information is correctly conveyed by the Exhibition Note, which is intended for public consumption:

Screenshot 2024-03-12 at 14 48 24

StepanBrychta:
## What does this change?

Fixes #2584 

This modifies the content of the `note` field (stored under `accessConditions`) on Sierra items. Instead of using the reserve note field(s), it now uses the 999 MARC field.

I also removed the sanitisation code because the text it removes (e.g. `RESERVE FOR`) doesn't seem to be present in the 999 MARC field.

I based the code on the following Sierra item (only relevant fields included):

```json
{
  "id": "1607327",
  "varFields": [
    {
      "fieldTag": "c",
      "marcTag": "999",
      "ind1": " ",
      "ind2": " ",
      "subfields": [
        {
          "tag": "a",
          "content": "Pre-Raphaelites, Musei di San Domenico, Forli, Italy, 23 February - 30 June 2024"
        }
      ]
    }
  ]
}
```

Additionally, the implementation should be able to handle the following two cases (though I'm not sure there are any items where either of these two edge cases applies):
1) Multiple var fields with the `999` MARC tag
2) A var field with the `999` MARC tag with multiple subfields

In both of these cases the contents of individual var fields/sub fields are concatenated and separated with `<br />`.

## How to test

I modified the existing unit tests.

The `note` section under `Where to find it` should stop showing reservation notes and instead show the contents of the 999 MARC field  on Sierra items (e.g. https://wellcomecollection.org/works/uxrgqnek). 

The logic I implemented uses the contents of _all_ subfields on all MARC 999 fields but I wonder if we should only filter for subfields with a specific tag (e.g. `a`).

### DIFF
--- a/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
+++ b/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala
@@ -367,11 +367,11 @@ object SierraItemAccess extends SierraQueryOps with Logging {
       // the contents of this field under the "a" subfield of the "MARC 999" field.
       case (_, _, _, _, Some(LocationType.OnExhibition))
           if itemData.varFields.exists(_.marcTag.exists(_ == "999")) =>
-        // Get all subfields inside of all "MARC 999" fields
+        // Get all "a" subfields inside of all "MARC 999" fields
         val marcTag999SubfieldContent =
           itemData.varFields
             .filter(_.marcTag.exists(_ == "999"))
-            .flatMap(varField => varField.subfields.map(sub => sub.content))
+            .flatMap(varField => varField.subfields.filter(_.tag == "a").map(sub => sub.content))
 
         // Concatenate all of them together and separate them with "<br />"
         AccessCondition(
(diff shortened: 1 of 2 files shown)


kenoir:
@StepanBrychta You can check the contents of MARC field 999 in the [Sierra sourced data](https://reporting.wellcomecollection.org/s/sierra/app/r/s/7vJiv) here, "RESERVE FOR" doesn't seem to exist in any of these fields as you've noted.

> The logic I implemented uses the contents of all subfields on all MARC 999 fields but I wonder if we should only filter for subfields with a specific tag (e.g. a).

I'm hoping [#collections-info](https://wellcome.slack.com/archives/CGXDT2GSH) should be able to give us some documentation for use of these "locally defined" fields and we should ask but querying the data in the reporting cluster it always looks to be subfield `a`.

kenoir (approved):
This looks good to me, but worth following up with collections information to ensure we are following their rules / can link to supporting documentation on our cataloguing practise.

kenoir:
Probably worth synthesising an inline comment here from this Slack conversations describing [why this field exists and where it comes from in Sierra](https://wellcome.slack.com/archives/CGXDT2GSH/p1716973238250079).

PR #2674, opened 2024-06-18
fix problem with blank subjects

paul-butcher:
## What does this change?

This allows Sierra Works with blank subjects to pass through the pipeline.  The problematic subjects do not pass through.

## How to test

The corresponding test has been updated.  The previous version of `returns nothing if the only relevant fields are invalid`, which appeared to be checking for empty labels was passing because the field had an undesirable value of indicator2 (it was empty, rather than 0,2 or 7).

 
More Sierra records make it through the pipeline.

There are currently over 400 messages on the Sierra transformer DLQ.  Some (maybe all) of these are there because of this problem.  Once deployed, redrive the DLQ and we should see fewer (I hope none) on the DLQ afterwards.

This looks like a cataloguing error.  We could end up masking an issue in source data (e.g. from a supplier) that would be better fixed at source.  However, even with perfectly clean source data, we should program defensively here.


paul-butcher:
Deployed and redriven.  Nothing left on the DLQ.

PR #2675, opened 2024-06-24
Move EBSCO Adapter to ECS

kenoir:
## What does this change?

This change moves the EBSCO adapter from lambda to ECS in order that it can run for a longer period than 15 minutes.

It packages the code as a container image which is then run using the ECS RunTask API via an AWS EventBridge schedule. Logging now works via the standard fluentbit sidecar, but there is custom monitoring to alarm and send a message to Slack if the lambda does not send success metrics within a 24 hour period.

> [!NOTE]
> The scheduled event is currently disabled so we can manually check for some potential incoming large changes.

## How to test

- [x] Use the `run_local.sh` script to test the task running locally.
- [ ] Deploy this and observe to being executed every 24 hours.

The EBSCO adapter can run to completion at any reasonable scale.

As this change uses custom alerting, it could be confusing or alerts could be missed. This is mitigate by testing and making the alerting quite simple.


StepanBrychta (approved):
This looks good to me! Tested locally.

PR #2679, opened 2024-07-17
Migrate from Akka to Pekko

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5754

See https://github.com/wellcomecollection/scala-libs/pull/246 for more information.

## How to test

See https://github.com/wellcomecollection/scala-libs/pull/246.

Akka is fully replaced with Pekko, functionality stays exactly the same.


sierra_adapter/sierra_merger/src/test/scala/weco/pipeline/sierra_merger/fixtures/RecordMergerFixtures.scala
```diff
@@ -20,18 +20,18 @@ import weco.storage.streaming.Codec._
 
 import scala.concurrent.ExecutionContext.Implicits.global
 
-trait RecordMergerFixtures extends Akka with SQS with SourceVHSFixture {
+trait RecordMergerFixtures extends Pekko with SQS with SourceVHSFixture {
 
-  def withRunningWorker[Record <: AbstractSierraRecord[_], R](
```
StepanBrychta, line 25: I renamed the `Record` type parameter to `SierraRecord` here and in a few other places because the compiler was complaining:

> type parameter Record defined in method withRunningWorker shadows class Record defined in package lang. You may want to rename your type parameter, or possibly remove it.

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/RelationsService.scala
```diff
@@ -36,8 +36,16 @@ class PathQueryRelationsService(
...
+      search = request,
+      maxItems = Long.MaxValue,
+      fetchThreshold = affectedWorksScroll,
+      warm = true
+    )
     Source
-      .fromPublisher(elasticClient.publisher(request))
+      .fromGraph(
+        new ElasticSource(elasticClient, sourceSettings)(as.dispatcher)
+      )
```
StepanBrychta, line 48: This is part of the migration from `elastic4s-http-streams` to `elastic4s-streams-pekko`. The latter provides a slightly different interface (e.g. it exposes a Graph instead of a Publisher), so a few changes in the code were needed.

project/Dependencies.scala
```diff
@@ -2,7 +2,7 @@ import sbt._
 import scala.language.reflectiveCalls
 object WellcomeDependencies {
 
-  val defaultVersion = "32.41.1" // This is automatically bumped by the scala-libs release process, do not edit this line manually
+  val defaultVersion = "32.42.0" // This is automatically bumped by the scala-libs release process, do not edit this line manually
     "32.40.4" // This is automatically bumped by the scala-libs release process, do not edit this line manually
```
agnesgaroux, line 6: I don't think it's a problem because it's been this way for a few commits on main but it's worth removing this duplicate line

agnesgaroux (approved):

PR #2681, opened 2024-07-30
Prefer MARC tag 264 over 260 for Sierra

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/catalogue-pipeline/issues/2662

Change the logic for extracting production information for Sierra items. 

The `apply` method of the `MarcProduction` object now has a new `prefer264Field` boolean parameter. This parameter has a default value of `false`, preserving the current behaviour of preferring tag 260 over tag 264.

This parameter is hard-coded to `true` in the `SierraProduction` object,  preferring tag 264 over tag 260 for all Sierra items. 

See [here](https://wellcome.slack.com/archives/C02ANCYL90E/p1722250892840619) for more information.

## How to test

I added unit tests to cover the new logic.

If implemented correctly, this change should result in no currently visible changes in the frontend. 

For all future Sierra items which contain both 260 and 264 tags, the frontend should only display the contents of the 264 tag under the `Publication/Creation` section.

The largest risk involves losing potentially useful information by preferring 264 over 260. This would happen if a Sierra item included both tag types but the information stored under the 260 tag was more useful.

Luckily, no such Sierra items currently exist. There used to be 3 items where the 264 tag only stored punctuation, but these were fixed at source (see [here](https://wellcome.slack.com/archives/CGXDT2GSH/p1722332790619849)).

This was determined using a Python script which pulled all 264 and 260 Sierra varfields from the reporting cluster and compared their values for all Sierra items:

```py
from elasticsearch import Elasticsearch
from collections import defaultdict

ES_URL = "<ES_URL>"
ES_API_KEY = "<ES_API_KEY>"
ELASTIC_SEARCH = Elasticsearch(ES_URL, api_key=ES_API_KEY)


def _get_hits(marc_tag: str, search_after: list):
    response = ELASTIC_SEARCH.search(index="sierra_varfields", size=1000, query={"match": {"varField.marcTag": marc_tag}}, sort=[{"parent.id": "asc"}, {"position": "asc"}], search_after=search_after)
    return response['hits']['hits']


def get_marc_tag_hits(tag: str):
    """
    Gets all distinct varfields with the given `tag` number for all Sierra items.
    """
    all_hits = []
    search_after = None

    print(f"Getting all items with MARC tag {tag}.")

    while len(hits := _get_hits(tag, search_after)) > 0:
        all_hits += hits
        search_after = hits[-1]["sort"]

        if len(all_hits) % 20000 == 0:
            print(f"    Current count: {len(all_hits)} items.")
    
    print(f"Total count: {len(all_hits)}")

    return all_hits
   
# Get all varfields tagged with the 264 tag or the 260 tag
marc_tag_264_hits = get_marc_tag_hits("264")
marc_tag_260_hits = get_marc_tag_hits("260")


def extract_subfields(all_hits: list):
    """
    Returns a dictionary storing Sierra item IDs as keys. Each value in the dictionary is a list with values in this shape:
    {
        "ind2": <indicator 2 value>,
        "subfields": <A list of subfields>
    }
    """
    subfields = defaultdict(list)

    for hit in all_hits:
        parent_id = hit['_source']['parent']['id']
        ind2 = hit['_source']['varField']['ind2']
        curr_subfields = tuple(sorted([(i["tag"], i["content"]) for i in hit["_source"]["varField"]["subfields"]], key=lambda i: i[0]))

        subfields[parent_id].append({
            "ind2": ind2,
            "subfields": curr_subfields
        })
    
    return subfields


subfields_264 = extract_subfields(marc_tag_264_hits)
subfields_260 = extract_subfields(marc_tag_260_hits)


# Get a set of Sierra IDs which have both a 264 tag and a 260 tag
overlapping_items = set(subfields_264).intersection(set(subfields_260))

print(f"Number of items with both 260 and 264 MARC tags: {len(overlapping_items)}")

for key in overlapping_items:
    _260, _264 = subfields_260[key], subfields_264[key]

    # Remove all 264 tags which have an ind2 value of "4" or " " because these are filtered out in the catalogue pipeline logic.
    _264 = [item for item in _264 if item["ind2"] not in [" ", "4"]]

    # Ignore items which have the same value under both tags
    if _260 == _264:
        pass
    # Ignore items where the 264 tag has no valid values
    elif len(_264) == 0:
        pass
    # Print all problematic items
    else:    
        print(key, _260, _264)
```


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/transformers/SierraProduction.scala
```diff
@@ -17,5 +17,5 @@ object SierraProduction
     bibId: SierraBibNumber,
     bibData: SierraBibData
   ): List[ProductionEvent[IdState.Unminted]] =
-    MarcProduction(bibData)
+    MarcProduction(bibData, prefer264Field = true)
```
kenoir, line 20: I might have inverted this because EBSCO is probably the odd case, but not blocking for merge.

pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcProduction.scala
```diff
@@ -22,28 +22,26 @@ import weco.pipeline.transformer.transformers.{
...
       case (Nil, from264) => from264
-
-      // If both 260 and 264 are present we prefer the 260 fields
-      case (from260, _) => {
-        warn(
-          s"Record ${record.controlField("001")} has both 260 and 264 fields. Using 260 fields."
-        )
-        from260
-      }
+      case (from260, from264) =>
```
kenoir, line 42: 👌

kenoir (approved):

PR #2689, opened 2024-08-15
Loop suppress miro

paul-butcher:
## What does this change?

Instead of passing an id as an argument to suppress_miro, it now accepts a list on stdin.  The current behaviour can be mimicked by simply echoing a single id and piping it, thus:

`echo L999999 | python suppress_miro.py`

This allows us to suppress multiple images in one call to suppress_miro, which previously took a long time due to the suppression list update being unnecessarily inside the loop (the script only worked on one id at a time, end-to-end, so the previous solution was to use `xargs`)

## How to test

This script operates on live data, so running it For Real Life is not really an option when testing.
I have added a dry-run flag so you can run it without doing anything, thus:
`python suppress_miro.py --id_source=suppress_list --message=hello --dry-run`
(assuming you have a file, suppress_list, that contains one id per line, e.g. :
```
L0025992
L0026124
L0026100
```

Running a miro suppression for multiple ids will be much quicker (it used to take hours)

Ideally, we'd have dummy endpoints to call in a test situation, rather than bypassing all that with a print command.

### DIFF
--- a/scripts/suppress_miro.py
+++ b/scripts/suppress_miro.py
@@ -22,9 +22,7 @@ miro_id_regex = re.compile("^[A-Z][0-9]{7}[A-Z]{0,4}[0-9]{0,2}$")
     required=True,
 )
 @click.option(
-    "--dry-run",
-    help="Show what will happen, without actually doing it",
-    is_flag=True
+    "--dry-run", help="Show what will happen, without actually doing it", is_flag=True
 )
 def suppress_miro(id_source, message, dry_run):
     """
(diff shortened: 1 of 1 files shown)


scripts/suppress_miro.py
```diff
@@ -2,49 +2,71 @@
...
+    type=click.File("r"),
+    default=sys.stdin,
+)
 @click.option(
     "--message",
     help="Why the image was removed, a link to a Slack message, etc.",
     required=True,
 )
-def suppress_miro(id, message):
+@click.option(
```
kenoir, line 24: 👏

kenoir (approved):
Neat and tidy 👌

PR #2691, opened 2024-08-16
Update EBSCO reindex pattern

kenoir:
## What does this change?

When we moved to using ECS tasks from a Lambda implementation [we neglected to updated reindexer](https://github.com/wellcomecollection/catalogue-pipeline/pull/2675). This change updates the invocation method to use AWS EventBridge, and an update to the EBSCO adapter to send reindex events to the reindex topic created by the reindexer.

The sequence of events that occur when `start_reindex.py` is run is as follows:

- `start_reindex.py` checks it's run with the "ebsco" source specified and "complete" type
- An EventBridge event is sent to the default event bus with a pattern matching:

```json
{
  "detail": {
    "ReindexTargets": ["ebsco"]
  },
  "source": ["weco.pipeline.reindex"]
}
```
- The EventBridge rule "ebsco-adapter-reindex-rule" should be invoked targeting the EBSCO adapter ECS task with the following input specifying the start command:
```
  input = jsonencode({
    containerOverrides = [
      {
        name    = "ebsco-adapter-ftp"
        command = ["--reindex-type", "full"]
      }
    ]
  })
```
- The updated lambda which has `REINDEX_TOPIC_ARN` specified in an environment variable uses updated logic to direct reindex events to the specified topic.
- If the reindex topic has been connected to a pipeline as per the [pipeline configuration block](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/terraform/2024-08-15/main.tf#L4) that pipeline will perform a reindex of the EBSCO source data.

Discovered while working on: https://github.com/wellcomecollection/catalogue-pipeline/issues/2659

## How to test

- [ ] Run this script locally, is a reindex perfomed as expected?

Reindexes can proceed without special regard to the EBSCO source, future reindexers can be implemented with this pattern easily.

This change is hard to test as there are not good tests around the reindex scripts. Local modifications have been made to perform no-op reindexes exercising the necessary code paths to ensure it will behave as expected.


ebsco_adapter/ebsco_adapter/src/main.py
StepanBrychta, line 1: **This is a nitpick and does not block merging** but I find the arguments and the surrounding logic a bit confusing.

I think it's because the arguments are mutually exclusive:
* Passing some `--reindex-ids` makes all other arguments redundant
* Passing `--scheduled-invoke=true` also makes all other arguments redundant

This might lead to unexpected behaviour when mutually exclusive arguments are passed together. For example, looking at the logic, it looks like passing `--reindex-type=full` but also passing in some `--reindex-ids` results in a partial reindex.

I wonder if this could be improved by:
1. Raising an exception if the user passes mutually exclusive arguments
2. Potentially reducing the number of arguments to prevent the user from passing nonsensical combinations of arguments. For example, we could have 2 arguments like this:
* `--process-type` (required):  Two valid values — `scheduled` or `reindex`
* `--reindex-ids` (optional): If specified, only the passed IDs would be reindexed. Otherwise a full reindex would be done.

StepanBrychta (approved):
Looks good!

ebsco_adapter/ebsco_adapter/src/main.py
kenoir: Agreed, I'll address this in a future PR. 👍

PR #2692, opened 2024-08-20
Add script to revert Miro suppressions

paul-butcher:
## What does this change?

This allows us to unsuppresss previously suppressed images.

## How to test

1. Suppress an image using suppress_miro.py
2. verify that it can no longer be seen
3. unsuppress it using unsuppress_miro.py
4. verify that it can be seen.

This process has been carried out on [L0071319](https://wellcomecollection.org/works/ua87equq/images?id=ce8jsnzs) and [L0024908](https://github.com/wellcomecollection/catalogue-pipeline/issues/2144).  Images which were invisible and now can be seen.

It should now be (almost) as easy to reinstate an image as it is to suppress it.

Unsuppressions are, almost by necessity, to be performed one at a time, as a user still needs to work out exactly which image is to be reinstated, whereas suppressions can be performed with the identifier alone.

If, at some point, we get a CSV full of records to reinstate, then we might want to consider a similar loop to the suppression script


scripts/miro_updates.py
```diff
@@ -264,6 +263,27 @@ def suppress_image(*, miro_id, message: str):
...
+        )
+        exit(1)
+
+    # First, make the DDS record reflect that the image should be visible
+    _set_image_availability(miro_id=miro_id, message=message, is_available=True)
+
+    # Now the actual image must be registered on DLCS so that it can be seen
+    register_on_dlcs(origin_url=origin, miro_id=miro_id)
+    # Finally, reindex the Miro record. This makes the Image record available to the API.
+    _request_reindex_for(miro_id=miro_id)
```
kenoir, line 284: I think this is done inside `_set_image_availability` also, so perhaps redundant?

scripts/miro_updates.py
```diff
@@ -461,3 +481,28 @@ def update_miro_image_suppressions_doc():
...
+                {
+                    "space": "8",
+                    "origin": origin_url,
+                    "id": miro_id,
+                    "mediaType": "image/jpeg",
+                }
+            ],
+        },
+    )
+    print(dlcs_response.text)
```
kenoir, line 501: Is this text useful, or left here for debug purposes?

kenoir (approved):
Looks good.

scripts/miro_updates.py
paul-butcher: It's useful.  It includes the batch's URL in the DLCS API so you can look at it to know when it has completed, rather than rushing straight over to the website and getting upset that it isn't there.

I'll make that clearer.

PR #2693, opened 2024-08-20
Add shelfmark to index mapping

kenoir:
## What does this change?

This change adds an index mapping and works ingestor update to index the shelfmark on works in order to make it easier for library staff and others to query on shelfmark.

The updated mapping includes this for items on the `query` property:

```
        "items": {
          "properties": {
            "id": {
              "type": "keyword",
              "normalizer": "lowercase"
            },
            "identifiers": {
              "properties": {
                "value": {
                  "type": "keyword",
                  "normalizer": "lowercase"
                }
              }
            },
            "shelfmark": {
              "properties": {
                "value": {
                  "type": "keyword",
                  "normalizer": "lowercase",
                  "fields": {
                    "path": {
                      "type": "text",
                      "analyzer": "path_analyzer",
                      "search_analyzer": "lowercase_whitespace_tokens"
                    }
                  }
                }
              }
            }
          }
        },
```

We have used a `path_analyzer` analyzer, as we do for collection path as shelfmarks similarly can use slash separated characters allowing more granular searching where available.

Working towards resolving https://github.com/wellcomecollection/catalogue-pipeline/issues/2453

Part of https://github.com/wellcomecollection/platform/issues/5757

## How to test

- [ ] Run the test, do they pass?

Collection staff are more easily able to find 

We should test that things are indexed as expected and that the mapping can be queried on to mitigate any risk from this change before releasing.

### DIFF
--- a/index_config/mappings.works_indexed.2024-08-20.json
+++ b/index_config/mappings.works_indexed.2024-08-20.json
@@ -1,0 +1,804 @@
+{
+  "dynamic": "strict",
+  "properties": {
+    "aggregatableValues": {
+      "properties": {
+        "availabilities": {
+          "type": "keyword",
+          "eager_global_ordinals": true
+        },
+        "contributors": {
+          "properties": {
+            "agent": {
+              "properties": {
+                "label": {
+                  "type": "keyword",
+                  "eager_global_ordinals": true
+                }
+              }
+            }
+          }
+        },
+        "genres": {
+          "properties": {
+            "label": {
+              "type": "keyword",
+              "eager_global_ordinals": true
+            }
+          }
+        },
+        "items": {
+          "properties": {
+            "locations": {
+              "properties": {
+                "license": {
+                  "type": "keyword",
+                  "eager_global_ordinals": true
+                }
+              }
+            }
+          }
+        },
+        "languages": {
+          "type": "keyword",
+          "eager_global_ordinals": true
+        },
+        "production": {
+          "properties": {
+            "dates": {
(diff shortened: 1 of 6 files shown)


index_config/mappings.works_indexed.2024-08-20.json
StepanBrychta, line 1: Out of curiosity, what's the process for creating these index mapping files? My current understanding is that it involves identifying the mapping file with the latest date, manually duplicating it and renaming the duplicate to today's date, and then making the necessary changes. Is this correct?

pipeline/ingestor/ingestor_common/src/main/scala/weco/pipeline/ingestor/common/models/ValueTransforms.scala
```diff
@@ -33,4 +34,11 @@ object ValueTransforms {
       .flatten
       .map(_.value)
   }
+
+  // Shelfmarks are only available on physical locations, so
```
StepanBrychta, line 38: Is this comment finished?

StepanBrychta (approved):
Looks good! I ran the tests locally.

index_config/mappings.works_indexed.2024-08-20.json
kenoir: I duplicated the file and gave it a date manually and added the mapping changes before referencing the updated date in the pipeline terraform config. 

I'll add some documentation describing the process and files in the `index_config` folder.

PR #2694, opened 2024-08-27
Update EBSCO reindexer args for simplicity

kenoir:
## What does this change?

This change addresses review comments in https://github.com/wellcomecollection/catalogue-pipeline/pull/2691/files#r1723159361, in order to simplify the CLI for the EBSCO reindexer script.

This change also updates the terraform to allow the scheduled invocation mode to be started by sending an event bridge event as well as on a schedule.

> [!Note]
> This terraform change is applied.

## How to test

- [ ] Try and run the adapter locally, does it behave as documented?
- [x] Apply the terraform change and observe the apply is a no-op.

More consistent less surprising interface for developers.

This change requires co-ordinating a terraform change with a release to ECR, if this happens in the wrong order that should be tolerable, but worth checking the process proceeds in AWS as expected after deployment.


StepanBrychta (approved):
Looks good!

PR #2695, opened 2024-08-30
Read license information from licenseNote in premis access condition

kenoir:
## What does this change?

This PR updates how license information can be derived from METS where the PREMIS element is used to provide access conditions. 

Currently where the `rightsBasis` is "License" we ignore the `licenseNote` field when parsing METS that contains the PREMIS element (i.e. Works sourced from Archivematica AKA anything "Born Digital"). This change addresses that problem to ensure the correct license information is available on Works.

See this Slack thread: https://wellcome.slack.com/archives/C02ANCYL90E/p1724408305952489

Addresses https://github.com/wellcomecollection/wellcomecollection.org/issues/11016

> [!Note]
> This change is currently deployed manually to a `dev` tag [here](https://eu-west-1.console.aws.amazon.com/ecs/v2/clusters/catalogue-2024-08-15/services/transformer_mets/update?region=eu-west-1). Ensure it is returned to the correct tag after release.

## How to test

- [x] Deploy a dev METS transformer image into the pipeline and reindex some works with METS that match the license pattern this addresses ([qd84awx8](https://api.wellcomecollection.org/catalogue/v2/works/qd84awx8)). Do they get the expected license?

Here is an [example of a work with the CC-BY-NC license added](https://wellcomecollection.org/works/ecnf4mvj) by this process.

When the born digital toggle is turned on, this is what it looks like in the front-end:
Screenshot 2024-09-02 at 11 41 22

Born Digital items have the correct license information and can be made available to the public for viewing.

This change updates how we read existing licenses, referring to the `rightsBasis` for which field to read, this opens the possibility existing correctly licensed works will be changed. This risk is mitigated by falling back to existing behaviour and spot checking METS to see that the behaviour will remain as expected.

### DIFF
--- a/pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformers/PremisAccessConditionsTest.scala
+++ b/pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformers/PremisAccessConditionsTest.scala
@@ -19,12 +19,14 @@ class PremisAccessConditionsTest
           Some("Open")
         ).parse.right.get.accessStatus.get shouldBe AccessStatus.Open
       }
+
       it("translates the copyrightNote into a Licence") {
         PremisAccessConditions(
           Some("In Copyright"),
           None
         ).parse.right.get.licence.get shouldBe License.InCopyright
       }
+
       it("has no access conditions if none are given") {
         val conditions = PremisAccessConditions(
           None,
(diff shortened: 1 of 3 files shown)


pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformers/PremisAccessConditionsTest.scala
```diff
@@ -50,19 +52,40 @@ class PremisAccessConditionsTest
...
           openInCopyrightRightsMD
         ).useRightsGrantedNote shouldBe Some("Open")
       }
 
+      it("pulls out the copy for the access status") {
+        PremisAccessConditions(
+          openInCopyrightRightsMD
+        ).useRightsGrantedNote shouldBe Some("Open")
+      }
+
```
StepanBrychta, line 88: This is identical to the test above. I assume `openInCopyrightRightsMD` should be replaced with `openCCBYNCRightsMD`?

StepanBrychta (approved):
Looks good!

pipeline/transformer/transformer_mets/src/test/scala/weco/pipeline/transformer/mets/transformers/PremisAccessConditionsTest.scala
kenoir: I think this test might be left in accidentally after a copy / paste. I'll remove it.

PR #2698, opened 2024-09-03
Update rules for requesting (13/08/24)

kenoir:
## What does this change?

This change updates the [rules for requesting](https://github.com/wellcomecollection/docs/blob/main/rfcs/042-requesting-model/README.md?plain=1#L119) which need to match those in Sierra, following an update from LS in Collection Product Lines:

> Manual request: Sierra code = n
> By appointment Sierra code = a
> By approval Sierra code = p
> Unavailable Sierra code = u

See [similar changes](https://github.com/search?q=org%3Awellcomecollection+%22rules+for+requesting%22&type=pullrequests).

Resolves: https://github.com/wellcomecollection/catalogue-pipeline/issues/2697

## How to test

- [ ] Deploy this change, are the items specified un-requestable?

Folk can't request things matching this pattern

This change is fairly self contained, and does not need a [corresponding catalogue-api update](https://github.com/wellcomecollection/catalogue-api/blob/main/common/stacks/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraRulesForRequesting.scala) as these are permanently unrequestable.

### DIFF
--- a/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
+++ b/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/RulesForRequestingResult.scala
@@ -7,13 +7,18 @@ case object Requestable extends RulesForRequestingResult
 sealed trait NotRequestable extends RulesForRequestingResult
 
 object NotRequestable {
-  case class NeedsManualRequest(message: String) extends NotRequestable
+  val defaultManualRequestMessage =
+    "This item cannot be requested online. Please place a manual request."
+  val defaultItemUnavailableMessage =
+    "This item is unavailable."
+
+  case class NeedsManualRequest(message: String = defaultManualRequestMessage) extends NotRequestable
 
   case class ItemClosed(message: String) extends NotRequestable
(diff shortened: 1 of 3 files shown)


agnesgaroux:
`NotRequestable.NeedsManualRequest()` is used in a few cases and results in all cases, if I understand correctly, in the works page showing "Access: Manual request" for the item in the Where To Find It section
Is there any reason why the string can be either
- `This item cannot be requested online. Please place a manual request.` or
- `Please complete a manual request slip.  This item cannot be requested online.`
Could we possibly make it the same everywhere `NotRequestable.NeedsManualRequest()` is used for consistency?

kenoir:
> Could we possibly make it the same everywhere `NotRequestable.NeedsManualRequest()` is used for consistency?

I've dried up these messages a bit. As far as I can tell they are not used, and the not requestable message only appears in [very specific circumstances](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/common/source_model/src/main/scala/weco/catalogue/source_model/sierra/rules/SierraItemAccess.scala#L256), (look for instances where the message is taken from the NonRequestable object, I see only when missing, withdrawn or "top item").

PR #2704, opened 2024-09-09
update script requirements

paul-butcher:
## What does this change?
This updates httpx in order to avoid the [input validation vulnerability](https://github.com/wellcomecollection/catalogue-pipeline/security/dependabot/197)

It also adds requests, which is needed for [get_miro_links.py](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/scripts/get_miro_links.py#L17), but was not in the requirements file

## How to test

I have tried this out by running get_miro_links, download_images_on_work, and suppress_miro (with dry run). Unfortunately, most of the scripts in here actively modify live data so I have been unable to check whether everything actually still works as expected.

[Dependabot shuts up about this issue](https://github.com/wellcomecollection/catalogue-pipeline/security/dependabot/197)

As stated above, the live nature of the scripts makes it hard to be certain that everything works without there being an actual change we want to make.

Given that we do not extensively use lots of httpx features, sticking only to the main, presumably stable interface, I suspect there is little to go wrong.

The scripts that rely on this dependency are all run by us, so there isn't a service to fail if this goes wrong.  It may be the case that next time we come to suppress some miro images or add_created_date, then we will have to do some minor surgery to get it to work.  I think the likelihood of that is minimal, and the impact fairly low.  If it comes to that and we have any difficulty rolling forwards, it should be simple enough to simply use the older version of httpx (0.18.2)


paul-butcher:
When testing this, I discovered https://github.com/wellcomecollection/catalogue-pipeline/issues/2705

kenoir (approved):

PR #2708, opened 2024-09-13
Increment version number in CALM VHS on deletion

kenoir:
## What does this change?

This change updated the version number of a CALM record stored in the hybrid store when it is deleted in order to notify downstream systems there has been a change to the record properly. 

We've seen errors where deletion do not pass the matcher service [because of this check](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/workgraph/WorkGraphUpdater.scala#L61). We attempt to prevent this error occurring with this change.

See this slack conversation for context: https://wellcome.slack.com/archives/C02ANCYL90E/p1726233083142129

Follows: https://github.com/wellcomecollection/catalogue-pipeline/issues/2701

## How to test

Less confusing results for cataloguers and collection information people on deletion of records in CALM.

There is some risk in making this change has unknown impact elsewhere in the pipeline that is not easily testable. The "update version when deleted" behaviour is implemented in the sierra adapter so if it works there, it should here 🤞

### DIFF
--- a/calm_adapter/calm_deletion_checker/src/main/scala/weco/pipeline/calm_deletion_checker/DeletionMarker.scala
+++ b/calm_adapter/calm_deletion_checker/src/main/scala/weco/pipeline/calm_deletion_checker/DeletionMarker.scala
@@ -22,6 +22,9 @@ class DeletionMarker(sourceTable: String)(implicit client: DynamoDbClient)
             .when(attributeExists("id"))
             .update(
               "id" === record.id,
+              // Set isDeleted = true and increment the version
+              // We must increment the version to ensure downstream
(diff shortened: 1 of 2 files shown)


paul-butcher (commented):
This might benefit from some prose in the code or test descriptions about the version increment. Without the PR context, it looks like an irrelevant detail.  

Something along the lines of what you put in the PR comment about allowing it past the duplicate processing filter, or explaining that marking a record as deleted is in itself a change, so its version must be updated to reflect that...

paul-butcher (approved):

PR #2710, opened 2024-09-13
Matcher version bug

kenoir:
## What does this change?

This change fixes a bug with the matcher where it passes along incorrect version to the matcher. 

It is possible for the version of a work in the identified works index to get out of sync with that in the matcher graph store if updates are missed sent from the id_minter. In this case the out of sync work would always be excluded from graphs of which is a part when passed to the merger, resulting in bad merges.

We update this logis here, choosing between:

1. A complete failure where no works are merged
2. A partial failure where only the works with matching versions are merged
3. A partial success where all works are merged, accepting the risk of inconsistency

We choose (3) rather than (2) where works are excluded from a merge rather than included with the risk of inconsistency presuming this is less disruptive than a complete exclusion.

See this slack conversation for context: https://wellcome.slack.com/archives/C02ANCYL90E/p1725547846066509

Follows: https://github.com/wellcomecollection/catalogue-pipeline/issues/2701

## How to test

A reduced risk of confusing outputs from the merger in the event of a work version de-sync.

This will change the behaviour of the matcher / merger subsystem to merge works that otherwise would have been excluded. We propose that having seen a real issue in production caused by the exclusion, resolving that problem has better outcomes than protecting against the version inconsistency that we have not yet seen.

### DIFF
--- a/pipeline/matcher_merger/merger/src/main/scala/weco/pipeline/merger/services/IdentifiedWorkLookup.scala
+++ b/pipeline/matcher_merger/merger/src/main/scala/weco/pipeline/merger/services/IdentifiedWorkLookup.scala
@@ -1,5 +1,6 @@
 package weco.pipeline.merger.services
 
+import grizzled.slf4j.Logging
 import weco.catalogue.internal_model.work.WorkState.Identified
 import weco.catalogue.internal_model.work.Work
 import weco.pipeline.matcher.models.WorkIdentifier
@@ -9,7 +10,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class IdentifiedWorkLookup(retriever: Retriever[Work[Identified]])(
   implicit ec: ExecutionContext
-) {
+) extends Logging {
   def fetchAllWorks(
     workIdentifiers: Seq[WorkIdentifier]
   ): Future[Seq[Option[Work[Identified]]]] = {
@@ -28,14 +29,21 @@ class IdentifiedWorkLookup(retriever: Retriever[Work[Identified]])(
             .map {
               case WorkIdentifier(id, version) =>
                 val work = works(id.toString)
-                // We only want to get the exact versions of the works specified
-                // by the matcher.
-                //
-                // e.g. if the matcher said "combine Av1 and Bv2", and we look
-                // in the retriever and find {Av2, Bv3}, we shouldn't merge
-                // these -- we should wait for the matcher to confirm we should
-                // still be merging these two works.
-                if (work.version == version) Some(work) else None
+                // Being asked to merge non-matching versions is incorrect
+                // but it is possible for the version in the identified index
+                // to be higher than the version in the graph store.
+                // Choosing between:
+                // (a) a complete failure where no works are merged
+                // (b) a partial failure where only the works with matching versions are merged
+                // (c) a partial success where all works are merged, accepting the risk of inconsistency
+                // We choose (c) as the least disruptive option.
+                if (work.version != version) {
+                  warn(
+                    "Matching version inconsistent! " +
+                      s"Found work $work with version $version, expected ${work.version}"
+                  )
+                }
+                Some(work)
             }
         case RetrieverMultiResult(_, notFound) =>
           throw new RuntimeException(s"Works not found: $notFound")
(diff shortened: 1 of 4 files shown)


pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/services/IdentifiedWorkLookupTest.scala
```diff
@@ -41,7 +41,7 @@ class IdentifiedWorkLookupTest
     }
   }
 
-  it("returns None if the stored version has a higher version") {
+  it("returns Some even if the stored version has a higher version") {
```
paul-butcher, line 44:
```suggestion
  it("returns the stored version of a work if it has a higher version than the requested version") {
```

pipeline/matcher_merger/merger/src/main/scala/weco/pipeline/merger/services/IdentifiedWorkLookup.scala
```diff
@@ -28,14 +29,21 @@ class IdentifiedWorkLookup(retriever: Retriever[Work[Identified]])(
...
+                // Being asked to merge non-matching versions is incorrect
+                // but it is possible for the version in the identified index
+                // to be higher than the version in the graph store.
+                // Choosing between:
+                // (a) a complete failure where no works are merged
+                // (b) a partial failure where only the works with matching versions are merged
+                // (c) a partial success where all works are merged, accepting the risk of inconsistency
+                // We choose (c) as the least disruptive option.
+                if (work.version != version) {
+                  warn(
```
paul-butcher, line 41: 🪵👏

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/services/MergerWorkerServiceTest.scala
```diff
@@ -136,7 +137,7 @@ class MergerWorkerServiceTest
     }
   }
 
-  it("always sends the highest version of a Work") {
+  it("send versions of a Work even if they are not the latest") {
```
paul-butcher, line 140: I think this is a pre-existing issue with this test, but it took me a few reads-through to realise that although `olderWork` and `newerWork` are obviously versions of the same thing, `work` is something else entirely, and that was why the result now has two identifiers in it, when previously one of them was discarded.

Does this make sense?

```
it("sends a Work even if the matched version is not the latest") {
    withMergerWorkerServiceFixtures {
      case (retriever, QueuePair(queue, dlq), senders, _, index) =>
        val workA = identifiedWork()
        val workB = identifiedWork()
        val newerWorkB =
          identifiedWork(canonicalId = workB.state.canonicalId)
            .withVersion(workB.version + 1)
        val matcherResult = createMatcherResultWith(Set(Set(workA, workB)))

        retriever.index ++= Map(workA.id -> workA, workB.id -> newerWorkB)

        sendNotificationToSQS(
          queue = queue,
          message = matcherResult
        )

        eventually {
          assertQueueEmpty(queue)
          assertQueueEmpty(dlq)
          getWorksSent(senders) should contain allOf (workA.id, workB.id)
          index shouldBe Map(
            workA.id -> Left(
              workA.transition[Merged](matcherResult.createdTime)
            ),
            workB.id -> Left(
              newerWorkB.transition[Merged](matcherResult.createdTime)
            )
          )
        }
    }
  }
```

kenoir: Yes, this one also puzzled me for a bit so I think worth updating the test as described.

agnesgaroux: This is the one I found confusing! To check my understanding, what's being tested here is that if we have two works that should merged, we send them to the queue even if their versions are different. Previously workA would have been held up because it has an older version?

kenoir: > we send them to the queue even if their versions are different. Previously workA would have been held up because it has an older version?

Yes, I think so. It might be clearer to say "we send them to the queue even if the version are different when we compare the version of a work we get sent by the matcher and what we find in the identified index for that work".

paul-butcher (approved):

PR #2717, opened 2024-09-19
Use sonatype scala libs

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/scala-libs/pull/250

Update the sbt project to get all future `scala-libs` releases from Sonatype.

* All `scala-libs` packages are obtained from Sonatype instead of S3
* Scala Steward automatically bumps the version number whenever a new version of `scala-libs` is released


project/Common.scala
```diff
@@ -6,7 +6,7 @@ object Common {
     scalaVersion := "2.12.16",
     organization := "weco",
     resolvers ++= Seq(
-      "Wellcome releases" at "s3://releases.mvn-repo.wellcomecollection.org/"
+      "Wellcome releases" at "https://oss.sonatype.org/content/repositories/snapshots/"
```
kenoir, line 9: I don't think this line is necessary.

kenoir (approved):
I think we can also remove the `s3CredentialsProvider` as in https://github.com/wellcomecollection/storage-service/commit/937fa5d6a14bcce6de5b3003701685a954e3dc71#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R195

PR #2726, opened 2024-10-04
Migrate id_minter database to 8.0.mysql_aurora.3.07.1

kenoir:
## What does this change?

Addressing https://github.com/wellcomecollection/platform/issues/5785. 

This change moves us to Aurora v3 (MySQL 8 compatible) Serverless V2 in order to prevent needing to move to extended support but also to offer us:

- Improved autoscaling with [Aurora Serverless V2,](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-serverless-v2.html) so we can avoid manual scaling for reindexes
- Access to the [RDS Data API ](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/data-api.html)to enable easier querying of the id_minter database.

This change requires a managed migration with a temporary pause in catalogue pipeline updates.

### Migration process
You should follow these steps to ensure a successful migration:

- [x] Ensure we have sent comms to collection staff indication catalogue updates will be paused
- [x] Disconnect the id_minter from all pipelines by running the current pipeline terraform (scale to zero)
- [x] Snapshot the current id_minter database and record the snapshot ARN
- [x] Update the critical terraform to reference the new snapshot ARN
- [x] Plan / Apply the critical terraform, using the latest snapshot identifier
- [x] Follow instructions in the critical terraform to migrate to serverless (manual failover)
- [x] Check you can connect to both the old and new database locally, do they have the same number of records?
- [x] Run the pipeline stack terraform after uncommenting the correct secret block to use the new database
- [x] Manually trigger the id_minter to process a record
- [x] ~Run a small reindex (perhaps EBSCO) to review scaling process~ edit: skipping, we'll review this when update the scaling logic (it looks quick enough processing the backlog)
- [x] Send comms indicating the migration process has completed

### Clean-up process
After you are sure the migration has completed successfully? If so, follow these steps (another PR to follow):

- [ ] Delete the old database, ensuring you keep a final snapshot
- [ ] Clean up the scaling infra code for the id_minter

## How to test

- [x] Can you connect to the new cluster locally? Does it have the same number of records as the old cluster.
- [x] Can you run the id_minter locally against the new cluster, does it process messages successfully?
- [x] Is the deployed id_minter processing messages successfully?
- [x] ~Does the new database scale to meet demand during a reindex?~ Skipping as described above.

- Costs related to extended support are avoided
- Infrastructure & process is simplified by removing manual scaling during reindexes

- We have reserved cost (~$40 a month) for the previous instance type which we will not be able to take advantage of. This is acceptable to simplify process & avoid other costs.
- If there is some unknown issue we should keep a final snapshot of the existing database to allow us to rollback to existing infrastructure

### DIFF
--- a/infrastructure/critical/.terraform.lock.hcl
+++ b/infrastructure/critical/.terraform.lock.hcl
@@ -2,8 +2,23 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/hashicorp/aws" {
-  version = "4.61.0"
+  version = "5.69.0"
   hashes = [
-    "h1:GrboLwI2hok/i8xfq87uS9kUKPqAE8b/1iJTEupRthY=",
+    "h1:unGIj/eLOrl42LQm7u0fjtjQHp+FHKinpSxR1ZuWsfI=",
+    "zh:123af8815a80abfd62eab5f9fc3d9226735cfea3627e834a1b48321cd8d391a6",
+    "zh:1298f312e239768c1846541e89b4fbec7eb21769c4a488c87181909049219fbe",
+    "zh:4edc950b39f3653beb8cd3e0b86a7dc9b6a77e90e543ed7be72639107bbc48a9",
+    "zh:5f24c916d6d2ce51e18210628b3b1aca8b85b383982a920b2a6adc259bdbd4e9",
+    "zh:66f0b2f5869a4dfed7154444c272022c6d9350dc4dfa0fc6d87ccbfc983ec560",
+    "zh:67e3be60863cf1c51c5be866d8646d433cc31e07514b9121611f812e73f2400d",
+    "zh:884672345a1d0362644a4d1588085fd4c4f56d3ca61b10c0d25cd1940d828fec",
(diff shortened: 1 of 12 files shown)


paul-butcher:
Have you already run plan/apply with the critical terraform that should match the current configuration?

I would put that as a preliminary drift test step to make it easier to be sure that all the changes proposed by TF when we run the real apply are pertinent.

kenoir:
> Have you already run plan/apply with the critical terraform that should match the current configuration?

Yes, but I think having things set up so when the TF here is run for the first time it will add the new database pre-migration fiddling to switch it to serverless is a better starting point.

I've updated the PR so it's now in a "starting configuration" to re-add the database when run for the first time, requiring a valid snapshot name.

kenoir:
To follow this PR: 
- Terraform change to remove old database infra.
- Potential scaling modifications when we review performance at the next reindex.

kenoir:
> Could some comments be removed now that the migration is done?

Yes, I think there is at least one hanging about that doesn't make sense. Will do this in one of the downstream PRs to remove the old DB.

PR #2732, opened 2024-10-11
Make ingestor-images tests more reliable

paul-butcher:
## What does this change?

Fixes https://github.com/wellcomecollection/platform/issues/5793

The test for whether a failed ingest leaves the message on the queue is unreliable.  Sometimes the message is still in flight when the test declares it a failure.

The call to `eventually` was unexpectedly quitting after about five seconds, rather than the requested ten.  I did not get to the bottom of why this happened, but one part of the problem here was that the test was frantically examining the queue every 15 milliseconds even when we know that it won't pass until it has done about 66 of them (minus whatever for the minuscule gap between running the ingestor and entering the eventually loop.

I suspect that, in line with the eventually correct nature of real SQS API requests, localstack is not always telling the exact truth about in-flight messages.  It may be that, by continually asking it at a high frequency, we are preventing it from getting round to updating the queue states.

## How to test

This is tricky.  The problem is intermittent.  I think we just need to keep running and retrying the tests.  If we get a bunch of greens in a row, then I guess the problem is solved.

We should see fewer bogus ephemeral test failures on Buildkite.

When flaky tests get solved by fiddling with timeouts and intervals, then there is always a chance that this is just kicking the problem down the road.  Waiting longer can also make the test runs take longer, a perennial problem in UI test suites.

However, (IMO) It is certainly correct that we shouldn't start looking at the queue until the message is expected to timeout, so this is a valid solution for this specific scenario.

### DIFF
--- a/pipeline/ingestor/ingestor_images/src/test/scala/weco/pipeline/ingestor/images/ImagesIngestorFeatureTest.scala
+++ b/pipeline/ingestor/ingestor_images/src/test/scala/weco/pipeline/ingestor/images/ImagesIngestorFeatureTest.scala
@@ -88,7 +88,12 @@ class ImagesIngestorFeatureTest
         withImagesIngestor(queue, existingImages = Nil) {
           index =>
             assertElasticsearchEmpty(index)
-            eventually(timeout(Span(10, Seconds)), interval(Span(1, Seconds))) {
+            eventually(
+              timeout(Span(10, Seconds)),
+              // The message does not expire until a second has elapsed, so don't bother
+              // looking until at least then.
+              interval(Span(1, Seconds))
+            ) {
               assertQueueEmpty(queue)
               assertQueueHasSize(dlq, size = 1)
             }
--- a/pipeline/ingestor/ingestor_images/src/test/scala/weco/pipeline/ingestor/images/ImagesIngestorFeatureTest.scala
+++ b/pipeline/ingestor/ingestor_images/src/test/scala/weco/pipeline/ingestor/images/ImagesIngestorFeatureTest.scala
@@ -32,7 +32,8 @@ class ImagesIngestorFeatureTest
     extends AnyFunSpec
     with ImageGenerators
     with IndexFixtures
-    with IngestorFixtures {
+    with IngestorFixtures
+     {
 
   it("reads an image from the queue, ingests it and deletes the message") {
     val image = createImageData.toAugmentedImage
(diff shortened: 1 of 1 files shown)


pipeline/ingestor/ingestor_images/src/test/scala/weco/pipeline/ingestor/images/ImagesIngestorFeatureTest.scala
```diff
@@ -88,7 +88,12 @@ class ImagesIngestorFeatureTest
         withImagesIngestor(queue, existingImages = Nil) {
           index =>
             assertElasticsearchEmpty(index)
-            eventually(Timeout(Span(10, Seconds))) {
+            eventually(
```
agnesgaroux, line 91: Is the test description right? "does not delete a message from the queue if it fails processing it" then `assertQueueEmpty(queue)` is a bit confusing, unless I'm missing something. Would it be better as "move the message to the DLQ if it fails processing it"?

agnesgaroux: Also I don't quite understand what the interval does that the timeout wasn't already doing 😣

paul-butcher: The test description is right, but I agree it's misleading.  I started writing a comment to that end, but then I noticed that it was failing due to in-flight messages, so it became less relevant, also, I couldn't think of the right way to say it ("a message fails if it fails" is a bit redundant).

In SQS, a message gets marked invisible until either its visibility timeout expires (it has failed) or it is deleted by the thing that pulled it (it's been successfully processed).  That's what is being tested here - that it does not get deleted (marked successful).

paul-butcher: The timeout says "keep trying this until it either succeeds or the timeout is reached"
The interval says "wait this long before trying again"

I still don't know why it seemed to be timing out or failing after 5 seconds instead of 10, but it was inspecting the queue over 40 times in those 5 seconds.  It is already known that it will not succeed for at least one second, because that's the value of the visibility timeout on the queue.  That's why I set it to 1s.

paul-butcher: A reason for not naming this test as you suggest, is that the application does not move it to the DLQ, that's a side effect of it being left on the original queue after the processing attempt.

agnesgaroux (approved):

PR #2733, opened 2024-10-14
Remove old RDS database

kenoir:
## What does this change?

> [!Note]
> This has been applied on 2024-10-18.

Follows https://github.com/wellcomecollection/catalogue-pipeline/pull/2726, removes old infrastructure now unused.

We aim to deploy this change on **Wednesday the 16th of October** to allow some time to ensure the new database is functioning as expected.

Steps to deploy:
- [x] Ensure there are no issues reported related the use TEI & id_minter database in the current pipeline
- [x] Ensure the new databases have been auto-snapshotted successfully as part of regular daily snapshots
- [x] Manually delete the old [identifiers-delta-cluster](https://eu-west-1.console.aws.amazon.com/rds/home?region=eu-west-1#database:id=identifiers-delta-cluster;is-cluster=true), and [tei-adapter-cluster](https://eu-west-1.console.aws.amazon.com/rds/home?region=eu-west-1#database:id=tei-adapter-cluster;is-cluster=true) which should  both be temporarily stopped
  - [x] When prompted ensure that a final snapshot is taken for both databases
- [x] Run the terraform in this PR on the critical stack, then pipeline and tei to confirm

## How to test

- [x] Does the infrastructure disappear? Yes, good?

No old infra hanging about costing money confusing people.

The old database should be turned off and demonstrably not in use. As long as we have a snapshot of the old data in case of unseen corruption we can restore the old database with some loss of data and ID churn. Risk of that is mitigated by seeing the pipeline functioning as expected before deletion.


infrastructure/critical/rds_id_minter.tf
```diff
@@ -76,7 +57,6 @@ module "identifiers_serverless_rds_cluster" {
   db_security_group_id     = aws_security_group.database_sg.id
   aws_db_subnet_group_name = aws_db_subnet_group.default.name
 
-  # This is the snapshot from the last time we migrated the database.
   snapshot_identifier = "aurora-mysql-v3-pre-migration-24-10-08"
```
agnesgaroux, line 60: Will this need to be changed to the final snapshot id?

agnesgaroux: And add the tei one?

agnesgaroux (approved):

PR #2743, opened 2024-11-01
Born digital archives format

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/catalogue-pipeline/issues/2742
Born-digital archives currently have a format label of `Archives - Digital`
We want it to be `Born-digital archives` 
see https://wellcome.slack.com/archives/C3TQSF63C/p1730476730751429

## How to test

[This](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/transformer/transformer_calm/src/test/scala/weco/pipeline/transformer/calm/transformers/CalmFormatTest.scala) is probably test enough 

After the next reindex, all Born-digital archives (and the corresponding Format filter) should be labelled as such

None that I can think of, simple string replacement

image

### DIFF
--- a/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Format.scala
+++ b/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/Format.scala
@@ -106,7 +106,7 @@ object Format extends Enum[Format] {
     // for all CALM sourced works, hence the id here is prefixed
     // with a "h" to namespace it within "ArchivesAndManuscripts".
     override val id: String = "hdig"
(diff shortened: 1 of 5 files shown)


kenoir:
Looks good, but I think without a more recent pipeline this will be deployed to the current prod pipeline and cause both values to appear in the filters as some but not all born-digital things get updated. 

So I think this needs to wait for a new pipeline to be created to merge.

kenoir (approved):

kenoir:
Closes https://github.com/wellcomecollection/catalogue-pipeline/issues/2742

PR #2746, opened 2024-11-06
Index concept id for querying

StepanBrychta:


kenoir (approved):
Potentially we could also pull the external ontology id into an indexable field as with `sourceIdentifier` elsewhere, but we can do that in a future PR is we feel we need to.

PR #2748, opened 2024-11-06
Use the Concept ontology type in TEI

paul-butcher:
## What does this change?

This replaces the old Subject ontology type with Concept, to harmonise the data from TEI with the data from Sierra.

Resolves https://github.com/wellcomecollection/catalogue-pipeline/issues/2744

This fixes a problem where multiple canonical ids are minted for apparently identical concepts, where the subject of a Sierra record results in a Concept, but the subject of a TEI document results in a Subject, despite them both having the same source identifier and scheme (e.g. [LCSH:sh85008893](https://id.loc.gov/authorities/sh85008893)) 

The ID Minter distinguishes between identifiers based on three values - the value and scheme and also (inappropriately, IMO) the type.  

## How to test

This is a pretty well isolated and tested area, so the unit tests provide a good level of confidence.

The next step would be to run the TEI content through the transformer and see that no Subjects appear.

When we next run a full refresh of the Concept Store, the number of records will fall, and there will be fewer records with identifiers in their sameAs lists.

This will result in the `Subject` canonical ids disappearing from the ConceptStore, so the corresponding Concept pages will cease to exist.  This violates the [Cool URIs](https://www.w3.org/Provider/Style/URI) principle, but arguably, those URIs should not have existed in the first place, as IMO they were created in error.

However, someone may have bookmarked these URLs, and will now be displeased that the page has vanished.  I think this is a very unlikely scenario, given the research we have recently done on concept page usage.

### DIFF
--- a/pipeline/transformer/transformer_common/src/test/scala/weco/pipeline/transformer/generators/LabelDerivedIdentifiersGenerators.scala
+++ b/pipeline/transformer/transformer_common/src/test/scala/weco/pipeline/transformer/generators/LabelDerivedIdentifiersGenerators.scala
@@ -26,12 +26,12 @@ trait LabelDerivedIdentifiersGenerators {
       )
     )
 
-  def labelDerivedSubjectIdentifier(value: String): Identifiable =
+  def labelDerivedConceptIdentifier(value: String): Identifiable =
     IdState.Identifiable(
       sourceIdentifier = SourceIdentifier(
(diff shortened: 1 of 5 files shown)


kenoir:
> This violates the [Cool URIs](https://www.w3.org/Provider/Style/URI) principle, but arguably, those URIs should not have existed in the first place, as IMO they were created in error.

Agree, this is not something we should worry about in this case.

PR #2752, opened 2024-11-08
Add queue config for pipeline lambda

kenoir:
> [!Note]
> This terraform change is applied.

## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5794

Subscribes the batcher lambda queue to the correct upstream topics so that it can start processing work

## How to test

- [ ] Apply this terraform, can we see the queues subscribed to the correct topics?

- [ ] The batcher lambda can be seen to do work so we can test its performance.

The risks should be minimal as the batcher lambda is not influencing downstream services.

### DIFF
--- a/pipeline/terraform/2024-11-05/main.tf
+++ b/pipeline/terraform/2024-11-05/main.tf
@@ -2,7 +2,7 @@ module "pipeline" {
   source = "../modules/stack"
 
   reindexing_state = {
-    listen_to_reindexer      = false
+    listen_to_reindexer      = true
     scale_up_tasks           = false
     scale_up_elastic_cluster = false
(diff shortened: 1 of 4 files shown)


pipeline/terraform/2024-11-05/main.tf
```diff
@@ -2,7 +2,7 @@ module "pipeline" {
   source = "../modules/stack"
 
   reindexing_state = {
-    listen_to_reindexer      = false
+    listen_to_reindexer      = true
```
kenoir, line 5: This maintains the current state of reindexing in the november pipeline.

agnesgaroux (approved):

PR #2753, opened 2024-11-08
Remove redundant prefixes from MeSH IDs #5819

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5819

Implements a fix in the Sierra transformer to start removing redundant prefixes from MeSH IDs, including the `(DNLM)` prefix and the `https://id.nlm.nih.gov/mesh/` URL prefix.

This functionality was previously implemented in the `SierraHasRecordControlNumber` trait, but this trait was only utilised when transforming `genre` concepts, it was not used when transforming `subject` or `contributor` concepts, so the prefixes were not getting removed for these concept types.

This PR removes the `SierraHasRecordControlNumber` trait and instead moves the functionality to the parent `MarcHasRecordControlNumber` trait. This trait already implements similar functionality for removing LoC prefixes, so I think it makes sense from a maintainability perspective to modify this trait rather than maintain a separate one.

This PR also removes the unused `SierraAgents` trait. 

## How to test

Run the automated tests. 

I added new tests to reduce the probability of another regression.

All `(DNLM)` and URL prefixes removed from MeSH IDs in the production index.

There should be no risks associated with this change.


paul-butcher (approved):
🪩

PR #2754, opened 2024-11-08
make sure path is a string

agnesgaroux:
## What does this change?

`obj.get` returns an Option, which caused the collectionPath to be serialised as `Some("A/B")` 
Using flatmap to get the value of the Option 

## How to test

The `pathProcessor` logs the input paths and selectors. 
We can see with the change that they are values, not Some(value)
From
`Input paths (1-6): Some("A/B"), Some("A/B/C/D"), Some("E/F/G"), Some("F/G/H"), Some("Y/Z"), Some("Z")`
to
`Input paths (1-6): "A/B", "A/B/C/D", "E/F/G", "F/G/H", "Y/Z", "Z"`

The batcher service and lambda output the same values 

The batcher lambda output is not subscribed yet


paul-butcher (approved):
🫓🗺️

kenoir (approved):

agnesgaroux:
> 🫓🗺️

Flatbread world? 🤔

PR #2763, opened 2024-11-15
keep lambda warm to allow pekko to publish

agnesgaroux:
## What does this change?

Not waiting for the Future to complete caused failure to publish downstream notifications in some scenario 
https://wellcome.slack.com/archives/C02ANCYL90E/p1731662840990099
Also increased the `visibility_timeout_seconds` for the input queue. It was fine in reindexing mode (25 + 5) but scaled down, the `visibility_timeout_seconds` ended up being shorter (1 + 5) than the lambda timeout (10)

## How to test

The lambda batcher is still not wired to the relation embedder. We can deploy and test that it works in the live environment

The lambda batcher is able to Do The Thing even when it's dealing with low volume of input 

The lambda batcher is still not wired to the relation embedder

### DIFF
--- a/pipeline/terraform/modules/stack/service_work_batcher.tf
+++ b/pipeline/terraform/modules/stack/service_work_batcher.tf
@@ -59,7 +59,7 @@ module "batcher_lambda" {
       module.router_path_output_topic.arn,
       module.path_concatenator_output_topic.arn,
     ]
-    visibility_timeout_seconds = (local.wait_minutes + 5) * 60
+    visibility_timeout_seconds = (local.wait_minutes + 9) * 60
 
     maximum_concurrency     = 20
     batch_size              = 2500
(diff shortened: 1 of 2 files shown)


paul-butcher (approved):

pipeline/terraform/modules/stack/service_work_batcher.tf
```diff
@@ -59,7 +59,7 @@ module "batcher_lambda" {
       module.router_path_output_topic.arn,
       module.path_concatenator_output_topic.arn,
     ]
-    visibility_timeout_seconds = (local.wait_minutes + 5) * 60
+    visibility_timeout_seconds = (local.wait_minutes + 9) * 60
```
kenoir, line 62: Good catch. I think perhaps `wait_minutes` is actually not so useful in the context of the lambda, and this value should be the maximum runtime of the lambda? 

If we push the `lambda_timout` into locals, we can align both values, presuming the visibility timeout should always be at least as long the maximum execution time.

Also should we set `flush_interval_minutes` & `max_processed_paths` to zero for the lambda with the comment that they are not used in the lambda implementation?

PR #2776, opened 2024-11-29
Readonly api key

paul-butcher:
## What does this change?

This adds an api key with read only access to the Elasticsearch in the pipeline.

This is a prerequisite to allow us to run pipeline stages locally with access to their upstream live databases.

This is required because the existing read-only access method is via username and password, and all the pipeline applications have been switched to use an API key instead.

(This is yet to be applied on any pipeline)

## How to test

Run the terraform, this will create a new key and corresponding secret entry.
Use that API Key to read data from Elasticsearch

We can start running local pipeline stages that read the live data.  This could help us diagnose issues in the pipeline where logging has been insufficient for us to identify the problem.

This might also allow us to retire the the username/password access method.

This is an extra secret the need to store safely, so increases our attack surface.  In mitigation:

1. it is stored in the same manner as the existing username/password approach
2. It only has readonly access to data that is eventually to be exposed through the API anyway

### DIFF
--- a/pipeline/terraform/modules/stack/elastic_pipeline.tf
+++ b/pipeline/terraform/modules/stack/elastic_pipeline.tf
@@ -74,7 +74,7 @@ module "pipeline_storage_secrets" {
     # We could hard-code these values as "https" and "9243", but we can just as easily infer
     # them from the actual endpoint value.
     (local.pipeline_storage_protocol) = split(":", ec_deployment.pipeline.elasticsearch[0].https_endpoint)[0]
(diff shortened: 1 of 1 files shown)


kenoir (approved):
I think we _could_ continue to get away with letting local things read & write locally - that might also be necessary for some types of debugging. On balance it is safer to have local running things read-only be default to prevent worst case situations where we corrupt or delete data in a live pipeline, but outside of the id_minter where state is shared between pipelines (but not in ES) still recoverable.

PR #2777, opened 2024-12-02
Embedder reshuffle

paul-butcher:
## What does this change?

This moves the actual processing of a Batch out of the RelationEmbedderWorkerService, and into a new BatchProcessor.

Alongside that, it moves some of the tests that previously had to live in this integration-like layer, slowly chugging through the SQS Simulator, into a more unit-test like place.

## How to test

This is a straight refactor, it should continue to work in exactly the same way as before.  This area has good test coverage.

This will allow us to reuse the core RelationEmbedder code so we can use it in various calling scenarios - Lambda, CLI, SQS/ECS, whatever.

This should be a straight refactor, so it should not add any new risks.


pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/BulkWriter.scala
```diff
@@ -0,0 +1,87 @@
...
+        )
+      case Right(_) => Future.successful(works.toList)
+    }
+}
+
+/** Controls the indexing of Works by Elasticsearch Bulk dividing the incoming
+  * works into appropriate batches.
+  */
+
+class BulkSTDOutWriter(
```
kenoir, line 70: Is this for local testing while developing this branch or is the idea to have an option for writing downstream when running locally in the future that doesn't need to have ES to work?

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/BulkWriter.scala
```diff
@@ -0,0 +1,87 @@
...
+    extends BulkWriter {
+
+  // init checks whether we can connect to the index named in workIndexer
+  // There really is no point in doing anything else if it can't.
+  Await.result(workIndexer.init(), Duration.Inf)
+
+  protected def writeWorks(
+    works: Seq[Work[Denormalised]]
+  ): Future[Seq[Work[Denormalised]]] =
+    workIndexer(works).flatMap {
```
kenoir, line 57: I think you could avoid the await with this approach. I'm not sure which is cleaner (I think you factored this approach out from the previous implementation, so maybe this way is clearer?)

```suggestion
    workIndexer.init() flatMap {
      _ => workIndexer(works).flatMap {
```

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/BulkWriter.scala
paul-butcher: The latter.  As you mentioned elsewhere, we might want to be able to write to a downstream DB at some point in the future, but I think most use cases will be covered by "read from there and print it out" when running locally.

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/Downstream.scala
```diff
@@ -0,0 +1,34 @@
...
+    .buildSNSMessageSender(config, subject = "Sent from relation_embedder")
+
+  override def notify(workId: String): Try[Unit] = Try(msgSender.send(workId))
+}
+
+object STDIODownstream extends Downstream {
+  override def notify(workId: String): Try[Unit] = Try(println(workId))
+}
+
+object Downstream {
```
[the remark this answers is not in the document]
paul-butcher: Yes, my plan is to take the advanced DRY approach of Repeat Everything Once.

This version is better than the Batcher one, and would be pretty easy to change it to use generics.

I didn't want to include the creation of a common lib as part of this work

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/BulkWriter.scala
paul-butcher: Actually, I've decided that I think the Await is clearer.  It's a bit more RAII.

I don't like init methods, and I think this should actually be part of the construction of the Indexer, but I don't want to make changes outside of Relation Embedder as part of this work.

pipeline/relation_embedder/relation_embedder/src/test/scala/weco/pipeline/relation_embedder/BatchProcessorTest.scala
```diff
@@ -0,0 +1,217 @@
...
+                    )
+                }
+
+            }
+        }
+    }
+  }
+
+  describe("given a batch containing a list of selectors") {
+    withProcessedBatch(
```
kenoir, line 89: I appreciate how terse `withProcessedBatch` makes the actual tests 👍

pipeline/relation_embedder/relation_embedder/src/test/scala/weco/pipeline/relation_embedder/BatchProcessorTest.scala
```diff
@@ -0,0 +1,217 @@
...
+            workE.id -> Set.empty
+          )
+        }
+    }
+  }
+
+  describe("given a tree containing an invisible Work") {
+    val invisibleWork = work("a/2/invisible").invisible()
+
+    println(invisibleWork.id)
```
kenoir, line 175: nit: `println` ✂️

kenoir (approved):
👏 Great. You've also improved test coverage on the core business logic of the relation embedder 👍

PR #2780, opened 2024-12-04
Update new relation embedder submodule to apply cleanly

kenoir:
> [!Note]
> This change has been applied.

## What does this change?

Follows https://github.com/wellcomecollection/catalogue-pipeline/issues/2770

Some tweaks to allow the terraform to apply cleanly.

## How to test

- [ ] Run `terraform apply` does it apply without error?

New submodule with lambda batcher and embedder created!

This should only be applied to the `2024-11-18` pipeline which is not in production use!

### DIFF
--- a/pipeline/terraform/modules/stack/subsystem_relation_embedder.tf
+++ b/pipeline/terraform/modules/stack/subsystem_relation_embedder.tf
@@ -1,7 +1,7 @@
 module "relation_embedder_sub" {
   source = "./relation_embedder"
 
-  namespace        = "relation_embedder_sub"
+  namespace        = "r_embed"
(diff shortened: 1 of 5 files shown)


pipeline/terraform/modules/stack/subsystem_relation_embedder.tf
```diff
@@ -1,7 +1,7 @@
 module "relation_embedder_sub" {
   source = "./relation_embedder"
 
-  namespace        = "relation_embedder_sub"
+  namespace        = "r_embed"
```
kenoir, line 4: This namespace has been shortened to allow for maximum name lengths for some resources.

agnesgaroux (approved):

PR #2781, opened 2024-12-04
Relation embedder empty climain

paul-butcher:
## What does this change?

This adds a dummy CLIMain so that the build will create separate entrypoint scripts to match the entry point in task definition.

## How to test

* Run  `sbt "project batcher" ";stage"`, you should now see six entry point scripts (3 bash, 3 bat) in `pipeline/relation_embedder/relation_embedder/target/universal/scripts/bin` One should be called `main`

Revert the manually created task definition revisions, and once the result of this PR has been deployed to ECR, the relation embedder in the new pipeline should be able to start.

Relation embedder works.  

Making the change this way, instead of changing the `/opt/docker/bin/main` line allows us to keep moving forward, as there is currently work in progress to create a proper CLIMain.  

Changing the application code and/or docker images is better than changing the terraform code, as the latter requires manual action to apply, and can have wider reaching consequences if anything is awry.


paul-butcher:
No need for this, I'll carry on with the real thing.

PR #2783, opened 2024-12-10
Local relation embedder

paul-butcher:
## What does this change?

This adds a new Main object - CLIMain, which can be invoked on the command line to run the relation embedder
resolves  https://github.com/wellcomecollection/catalogue-pipeline/issues/2782
## How to test

With an appropriate AWS_PROFILE set, 

```
% sbt "project relation_embedder" ";stage"
% cat pipeline/relation_embedder/relation_embedder/scripts/batches.txt | python pipeline/relation_embedder/relation_embedder/scripts/cli.py 2024-11-18      
```

It should now print out a load of Json, some log entries, and all the work ids that the real relation embedder would be passing downstream.

This makes it easier to be confident of any changes we make to the embedder by allowing us to run local code against live upstream data. 

This is particularly important for the relation embedder over the other pipeline stages, as the behaviour of this stage is most dependent on the data in the upstream database, as it performs a search to find the relations.

Access is granted on a readonly basis.

We recently had to change the existing ECS task definitions manually to account for the lack of CLIMain.  We will have to revert those changes back to the properly deployed version.


pipeline/relation_embedder/relation_embedder/scripts/cli.py
```diff
@@ -0,0 +1,29 @@
...
+    Look up the value of a SecretString in Secrets Manager.
+    """
+    secrets = session.client("secretsmanager")
+    return secrets.get_secret_value(SecretId=secret_id)["SecretString"]
+
+
+def set_environment(pipeline_date):
+    session = boto3.Session()
+    prefix = f"elasticsearch/pipeline_storage_{pipeline_date}/"
+    os.environ["es_host"] = get_secret_string(session, f"{prefix}public_host")
```
kenoir, line 21: I think there may be room to have a `.env` file, or some means to generate one from which to load environment variables.  Now we may have a way to avoid referencing secrets directly in environment variables passed to an app i'm more comfortable about using that mechanism.

Not blocking, but related to https://github.com/wellcomecollection/catalogue-pipeline/pull/2785

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/lib/StdInStrings.scala
```diff
@@ -0,0 +1,34 @@
...
+  Framing,
+  Source,
+  StreamConverters
+}
+import org.apache.pekko.util.ByteString
+
+import scala.concurrent.Future
+
+trait StdInStrings {
+  private val stdinSource: Source[ByteString, Future[IOResult]] =
```
kenoir, line 16: I understand this is to accommodate the interfaces we've woven into the app elsewhere, but it does neatly highlight how what could be "take a string, parse it, hand that to the app" is somewhat obfuscated by using Akka Flow.

kenoir (approved):

pipeline/relation_embedder/relation_embedder/scripts/cli.py
paul-butcher: I think that once #2785 is done, there's a good opportunity to harmonise the local and lambda approaches.

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/lib/StdInStrings.scala
paul-butcher: Yes.  Part of the reason I pulled it out like this is that I think it makes it easier to think about those kinds of changes.

However, now that I have, and now you mention it, I think it would be pretty simple for me to replace this with a io.Source and only turn it into Akka Source just before it enters the bit that's already using Flow.  I'm going to give that a go right now.

PR #2784, opened 2024-12-11
Remove groupedWeightedWithin

paul-butcher:
## What does this change?
Resolves https://github.com/wellcomecollection/catalogue-pipeline/issues/2779, see description there for rationale.

## How to test

Unit tests pass in reasonable time, as does running a command line invocation `cat pipeline/relation_embedder/relation_embedder/scripts/batches.txt | python pipeline/relation_embedder/relation_embedder/scripts/cli.py 2024-11-18 `

This removes an unnecessary configuration parameter, marginally simplifying the code in the application, its tests, and the corresponding Terraform.

groupedWeightedWithin is provided by Pekko, and although groupedWeighted is also a Pekko function, it is one that would be simpler to replace with plain Scala, if we find that the rest of this application can be implemented without Pekko.  This may result in a lighter application with a faster startup.

I don't believe this change carries any risk. The timeout was redundant, as far as I could tell.

### DIFF
--- a/pipeline/relation_embedder/relation_embedder/src/main/resources/application.conf
+++ b/pipeline/relation_embedder/relation_embedder/src/main/resources/application.conf
@@ -6,7 +6,6 @@ es.merged-works.index=${?es_merged_index}
 es.works.scroll.complete_tree=${?complete_tree_scroll_size}
 es.works.scroll.affected_works=${?affected_works_scroll_size}
 es.works.batch_size=${?index_batch_size}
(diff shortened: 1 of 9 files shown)


kenoir:
```
Essentially, the ...within version is for sources where we cannot know whether more elements will appear when upstream is empty (e.g. consuming from a queue that is populated by external events). In this instance, as soon as upstream is empty, it will complete.
```

This is useful context. I was going to ask how we can be sure it will emit if we don't complete a group, but this makes sense.

kenoir (approved):

PR #2785, opened 2024-12-11
Allow reading of secrets to environment variable for containerised lambdas

kenoir:
## What does this change?

Adds a lambda extension to read variables from secrets manager, this is required as there is not a drop in replacement for the mechanism used in ECS where AWS will convert secrets manager references to secrets for us.

This approach uses the [Lambda Extension API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html), which is [available to containerised Lambdas](https://aws.amazon.com/blogs/compute/working-with-lambda-layers-and-extensions-in-container-images/) by adding files to `/opt/extensions/`.

For simplicity (by comparison to including some other language runtime to run extension code and/or providing a binary to package), we use a bash script extension, based on [the example here.](https://github.com/aws-samples/aws-lambda-extensions/tree/main/bash-example-wrapper).

## How it works:

- In terraform the `pipeline_lambda` module uses the existing map of environment variable names to secret names to generate a map of environment variable names to secret names prefixed `secret:`, in addition we also provision permissions for the lambda to access those secrets.
- `bash_secrets_extension.sh` looks for environment variables values that have been passed to the lambda that have been prefixed with `secret:` and retrieves them from AWS Secrets Manager
- The extension creates a file in the containerised environment called `/tmp/config` that will persist between invocations, the file uses the [Typesafe Config](https://lightbend.github.io/config/) format (HOCON).
- Inside the Scala application we layer global, application and the config loaded from `/tmp/config` to resolve a final set of configuration in [`LambdaConfiguration`](https://github.com/wellcomecollection/catalogue-pipeline/pull/2785/files#diff-c098649e55031c4b1982f9f61679545f609bd5514be0e9c3136265aed86c3e98R10), allowing the values from `/tmp/config`  to substitute for environment variables where they don't already exist.

## How to test

- [x] Run the relation_embedder lambda using RIE and docker-compose with a valid `.env` file.
- [ ] Deploy this change, and terraform update to a non-production pipeline, observe the lambda reading the secrets.

https://github.com/user-attachments/assets/8ba0de10-e2b0-426c-90a0-48a981b3b27b

This lambda, and all other ones that require secrets have a mechanism to do so that require no application or significant terraform changes.

Yes, this change involves handling and providing secrets to our services. We must take care not to log or reveal these in code or application logs. 

This change intends to keep secrets encrypted at rest within secrets manager until they are required by a lambda invocation. This follows an [AWS recommended pattern](https://aws.amazon.com/blogs/compute/using-the-aws-parameter-and-secrets-lambda-extension-to-cache-parameters-and-secrets/), that unfortunately we can't take direct advantage of as we are using containerised lambdas, but follows the same mechanism.

### DIFF
--- a/pipeline/relation_embedder/relation_embedder/bash_secrets_extension.sh
+++ b/pipeline/relation_embedder/relation_embedder/bash_secrets_extension.sh
@@ -35,6 +35,10 @@ get_secret() {
   echo "[${LAMBDA_EXTENSION_NAME}] Getting secret: $secret_name" > /dev/tty
 
   secret_value=$(aws secretsmanager get-secret-value --secret-id "$secret_name" --query SecretString --output text)
+  if [[ -z "$secret_value" ]]; then
+    echo "[${LAMBDA_EXTENSION_NAME}] Secret not found: $secret_name" > /dev/tty
+    exit 1
+  fi
   echo "$secret_value"
 }
 
--- a/pipeline/relation_embedder/relation_embedder/bash_secrets_extension.sh
+++ b/pipeline/relation_embedder/relation_embedder/bash_secrets_extension.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+set -euo pipefail
+
+OWN_FILENAME="$(basename $0)"
+LAMBDA_EXTENSION_NAME="$OWN_FILENAME" # (external) extension name has to match the filename
+TMP_FILE=/tmp/$OWN_FILENAME
+
+# Graceful Shutdown
+_term() {
+  echo "[${LAMBDA_EXTENSION_NAME}] Received SIGTERM"
+  # forward SIGTERM to child procs and exit
+  kill -TERM "$PID" 2>/dev/null
+  echo "[${LAMBDA_EXTENSION_NAME}] Exiting"
+  exit 0
+}
+
+forward_sigterm_and_wait() {
+  trap _term SIGTERM
+  wait "$PID"
+  trap - SIGTERM
+}
+
+# Initialization
+# To run any extension processes that need to start before the runtime initializes, run them before the /register
+echo "[${LAMBDA_EXTENSION_NAME}] Initialization"
+
+# Read secret from AWS Secrets Manager
+get_secret() {
+  local secret_name=$1
+  local secret_value
+  echo "[${LAMBDA_EXTENSION_NAME}] Getting secret: $secret_name" > /dev/tty
+
+  secret_value=$(aws secretsmanager get-secret-value --secret-id "$secret_name" --query SecretString --output text)
+  echo "$secret_value"
+}
+
+# Extract secret value from environment variable and write to config file
+extract_secret_value_to_config() {
+  local env_var=$1
+  local env_var_value
+  local env_var_key
+  local secret_key
+  local secret_value
+
+  env_var_value=$(echo "$env_var" | cut -d= -f2)
+  env_var_key=$(echo "$env_var" | cut -d= -f1)
+
+  if [[ $env_var_value == "secret:"* ]]; then
+    secret_key=$(echo "$env_var_value" | cut -d: -f2)
+    secret_value=$(get_secret "$secret_key")
+    echo "$env_var_key=\"$secret_value\"" >> /tmp/config
+  fi
+}
+
+# Create a configuration file with secrets
+create_config_file() {
+  local env_vars
+  local env_var
+
+  # Ensure the file exists, and is empty
+  touch /tmp/config
+  echo -n > /tmp/config
+
+  env_vars=$(printenv)
+  for env_var in $env_vars; do
+    extract_secret_value_to_config "$env_var"
+  done
+}
+
+create_config_file
+
+# Registration
+# The extension registration also signals to Lambda to start initializing the runtime.
+HEADERS="$(mktemp)"
+echo "[${LAMBDA_EXTENSION_NAME}] Registering at http://${AWS_LAMBDA_RUNTIME_API}/2020-01-01/extension/register"
+curl -sS -LD "$HEADERS" \
+  -XPOST "http://${AWS_LAMBDA_RUNTIME_API}/2020-01-01/extension/register" \
+  --header "Lambda-Extension-Name: ${LAMBDA_EXTENSION_NAME}" \
+  -d "{ \"events\": [\"SHUTDOWN\"]}" > "$TMP_FILE"
+
+RESPONSE=$(<"$TMP_FILE")
+# Extract Extension ID from response headers
+EXTENSION_ID=$(grep -Fi Lambda-Extension-Identifier "$HEADERS" | tr -d '[:space:]' | cut -d: -f2)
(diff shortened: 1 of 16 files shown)


pipeline/relation_embedder/relation_embedder/docker-compose.yml
```diff
@@ -1,5 +1,24 @@
...
+      - "9000:8080"
+    environment:
+      - AWS_PROFILE=platform-developer
+      - metrics_namespace=catalogue-dev_relation_embedder
+      - affected_works_scroll_size=50
+      - complete_tree_scroll_size=800
+      - index_batch_size=100
+      - index_flush_interval_seconds=60
+    env_file:
+      - .env
```
kenoir, line 20: The `.env` file contains the remaining pipeline specific variables, we could have a script to generate a file like this:

```
es_apikey=secret:elasticsearch/pipeline_storage_2024-11-18/relation_embedder/api_key
es_host=secret:elasticsearch/pipeline_storage_2024-11-18/public_host
es_denormalised_index=works-denormalised-2024-11-18
es_merged_index=works-merged-2024-11-18
es_protocol=https
es_port=443
```

pipeline/relation_embedder/relation_embedder/bash_secrets_extension.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+set -euo pipefail
+
+OWN_FILENAME="$(basename $0)"
```
kenoir, line 8: For use in other container images, this should be in a shared location eventually.

pipeline/relation_embedder/relation_embedder/bash_secrets_extension.sh
```diff
@@ -0,0 +1,106 @@
...
+  -XPOST "http://${AWS_LAMBDA_RUNTIME_API}/2020-01-01/extension/register" \
+  --header "Lambda-Extension-Name: ${LAMBDA_EXTENSION_NAME}" \
+  -d "{ \"events\": [\"SHUTDOWN\"]}" > "$TMP_FILE"
+
+RESPONSE=$(<"$TMP_FILE")
+# Extract Extension ID from response headers
+EXTENSION_ID=$(grep -Fi Lambda-Extension-Identifier "$HEADERS" | tr -d '[:space:]' | cut -d: -f2)
+echo "[${LAMBDA_EXTENSION_NAME}] Registration response: ${RESPONSE} with EXTENSION_ID $(grep -Fi Lambda-Extension-Identifier "$HEADERS" | tr -d '[:space:]' | cut -d: -f2)"
+
+# Event processing
```
kenoir, line 90: This event loop seems to be necessary to keep the Lambda running happily. The extension is invoked as a separate process from the main lambda, but the extension API will fail the start-up if we exit before the lifetime of the lambda container itself.

There may be scope to simplify this!

paul-butcher: Would sleeping for 15 minutes work here, or does the extension have to explicitly listen for the shutdown event to avoid keeping the Lambda awake?

pipeline/terraform/modules/stack/service_work_relation_embedder.tf
```diff
@@ -14,7 +14,7 @@ module "relation_embedder" {
   // Override entrypoint & command to dual use lambda container image
   // This should be removed once we have a dedicated batcher_lambda image
   entrypoint = [
-    "/opt/docker/bin/main"
+    "/opt/docker/bin/relation_embedder"
```
paul-butcher, line 17: Now that I've merged the CLI version, this should be main again

pipeline/relation_embedder/relation_embedder/bash_secrets_extension.sh
kenoir: After some experiments, removing the `while` loop here results in the following error at lambda initialization: 
Screenshot 2024-12-12 at 15 23 22

It looks like exiting early results in stopping the runtime interface emulator (RIE) from invoking the main lambda.

> Would sleeping for 15 minutes work here

Putting a `sleep 900` instead of the loop, does allow the lambda to be invoked and terminated (though the logs do mutter about a forced stop). However a 2nd invocation then fails with the following error:

Screenshot 2024-12-12 at 15 29 14

So I think that messes with some logic around invocation in the RIE in a way that may be reproduced in an in-situ invocation.

paul-butcher: That makes sense. I expect it starts when a lambda is first switched on and keeps going until there's nothing left to do, rather than running for each invocation.  This allows warm starts to benefit the most from that warmth.

pipeline/relation_embedder/relation_embedder/docker-compose.yml
kenoir: Script added, as part of `run_local.sh`

pipeline/relation_embedder/relation_embedder/scripts/run_local.sh
```diff
@@ -0,0 +1,30 @@
...
+set -euo pipefail
+
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 <PIPELINE_DATE>"
+  exit 1
+fi
+
+export PIPELINE_DATE=$1
+
+PROJECT_NAME="relation_embedder"
```
kenoir, line 12: This pattern is intended for re-use, in. combination with use of `template.env` and `local.docker-compose.yml` files. The intention is to work towards running any of the lambda services using the same script / pattern from the root with one script and a docker-compose.yml for all the services.

pipeline/relation_embedder/relation_embedder/scripts/run_local.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Build the project
+pushd ../../..
+sbt "project $PROJECT_NAME" ";stage"
+popd
+
+# Build the docker image
+docker compose -f local.docker-compose.yml \
+  build lambda
+
```
paul-butcher, line 27: I'd be tempted to call this a build script and end it here (or have a build and a run script).

Although recompiling with SBT is reasonably efficient, it still takes about 15 seconds to reach this point when there's nothing to do.

(not a blocker)

kenoir: I've added a `--skip-build` flag to do this optionally.

pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/lib/LambdaConfiguration.scala
```diff
@@ -0,0 +1,33 @@
+package weco.pipeline.relation_embedder.lib
+
+import java.io.File
+import com.typesafe.config.{Config, ConfigFactory}
+
+trait Configuration {
+  val config: Config
+}
+
+trait LambdaConfiguration extends Configuration {
```
paul-butcher, line 10: This kind of thing makes me happy.

Could we even go one step further and hide `val config` from the Lambda, instead translating the relevant config to a case class or trait so that all the string-based references to properties are in one place?

I suspect it has too many tendrils to do cleanly as part of this change, but might be a worthy ticket to raise for a future improvement

paul-butcher (approved):

PR #2792, opened 2024-12-18
Extract processEvent from TransformerMain

kenoir:
## What does this change?

This change is a small refactor to expose an interface that allows consumers of `TransformerEventProcessor[Payload, SourceData]` to call:

```scala
processEvent(
    payload: Payload
  ): Future[Result[Option[(Work[Source], StoreKey)]]]
```

Rather than having this tied into an Akka Flow that requires a `NotificationMessage` at the top level. This enables us to implement lambda-like functionality [as we have started to do elsewhere](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/LambdaMain.scala).

## How to test

- [x] Tests pass!

Easier implementation of transformer-lambdas, simpler pipeline.

This should be a no-op refactor.

### DIFF
--- a/pipeline/transformer/transformer_common/src/test/scala/weco/pipeline/transformer/TransformerWorkerTest.scala
+++ b/pipeline/transformer/transformer_common/src/test/scala/weco/pipeline/transformer/TransformerWorkerTest.scala
@@ -451,7 +451,7 @@ class TransformerWorkerTest
             val worker = new TransformerWorker(
               transformer = transformer,
               pipelineStream = pipelineStream,
-              retriever = retriever,
+              transformedWorkRetriever = retriever,
               sourceDataRetriever =
                 new ExampleSourcePayloadLookup(sourceStore = store)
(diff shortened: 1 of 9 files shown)


pipeline/transformer/transformer_common/src/test/scala/weco/pipeline/transformer/TransformerWorkerTest.scala
```diff
@@ -451,7 +451,7 @@ class TransformerWorkerTest
             val worker = new TransformerWorker(
               transformer = transformer,
               pipelineStream = pipelineStream,
-              retriever = retriever,
+              transformedWorkRetriever = retriever,
```
paul-butcher, line 454: Nice and explicit

paul-butcher (approved):

PR #2793, opened 2024-12-19
Decouple relation_embedder config from typesafe config

kenoir:
## What does this change?

This change implements [a suggestion in a previous PR to have one place where we reference string based paths for typesafe](https://github.com/wellcomecollection/catalogue-pipeline/pull/2785#discussion_r1884009296) config.

The suggestion is that services have classes that extend `LambdaConfigurable` and implement a `build` function that accepts typesafe config and provides a case class that extends `ApplicationConfig`.

## How to test

Simpler and clearer implementation of new lambda services that consume typesafe config.

This change should functionally be a no-op.


pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/BatchProcessor.scala
```diff
@@ -103,39 +102,38 @@ class BatchProcessor(
...
     ec: ExecutionContext,
     materializer: Materializer
   ): BatchProcessor = {
 
     val identifiedIndex =
-      Index(config.requireString("es.merged-works.index"))
+      Index(config.mergedWorkIndex)
 
-    val esClient = ElasticBuilder.buildElasticClient(config)
+    val esClient = ElasticBuilder.buildElasticClient(config.elasticConfig)
```
paul-butcher, line 115: Nice!  I think this is the line that really proves the value of the approach.

Previously, we were just handing all the config over to buildElasticClient, and letting it pick what it wants.  Now we can be much more targeted about it.

paul-butcher (approved):
Neat!

PR #2796, opened 2025-01-02
Make error handling easier to reason about

paul-butcher:
## What does this change?

This simplifies error handling in the Batcher by hiding the SQS Message inside a generic that gets passed around internally, rather than all that list index business.

## How to test

Externally, this is a straight refactor, so nothing should change.

I have tried to ensure that the pertinent tests look the same as before, to make it easier to verify that I have not changed the behaviour.

It should be easier to implement error handling in the lambda and CLI versions of this application.

As this should not result in a change in behaviour, I don't think there are any notable risks associated with this change

### DIFF
--- a/pipeline/relation_embedder/batcher/src/main/scala/weco/pipeline/batcher/BatcherWorkerService.scala
+++ b/pipeline/relation_embedder/batcher/src/main/scala/weco/pipeline/batcher/BatcherWorkerService.scala
@@ -33,14 +33,13 @@ class BatcherWorkerService[MsgDestination](
         source
           .map {
             case (msg: SQSMessage, notificationMessage: NotificationMessage) =>
-              (msg, notificationMessage.body)
+              PathFromSQS(notificationMessage.body, msg)
           }
           .groupedWithin(maxProcessedPaths, flushInterval)
-          .map(_.toList.unzip)
           .mapAsync(1) {
-            case (msgs, paths) =>
+            paths =>
               info(s"Processing ${paths.size} input paths")
-              processPaths(msgs, paths)
+              processPaths(paths)
           }
           .flatMapConcat(identity)
       }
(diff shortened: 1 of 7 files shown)


paul-butcher:
It still iterates over the input list filtering out those messages that did not fail, which is a little confusing.

I have a cool solution to that involving [partitionMap](https://www.scala-lang.org/api/2.13.x/scala/collection/immutable/List.html#partitionMap[A1,A2](f:A=%3EEither[A1,A2]):(CC[A1],CC[A2])), as soon as we upgrade to Scala 2.13

agnesgaroux:
> It still iterates over the input list filtering out those messages that did not fail, which is a little confusing.
> 
> I have a cool solution to that involving [partitionMap](https://www.scala-lang.org/api/2.13.x/scala/collection/immutable/List.html#partitionMap%5BA1,A2%5D(f:A=%3EEither%5BA1,A2%5D):(CC%5BA1%5D,CC%5BA2%5D)), as soon as we upgrade to Scala 2.13

The `BatcherWorkerService` will go away soonish anyway no?

agnesgaroux (approved):

paul-butcher:
> The `BatcherWorkerService` will go away soonish anyway no?

Yes, that's partly why I'm not in a hurry to change it, as I suspect we might not bother with partial failures when this is Lambda-only, and there may be a different way to deal with things even if we do.

Currently, the incongruity is that it returns the failed paths, but the Worker Service is interested in the succeeded ones (in order to emit them to get deleted from the input queue). In the Lambda context, I think the reverse is true.  All messages in the batch get deleted if the lambda is successful, but you can report that some of them failed.

So the way it is currently implemented is probably the right way to allow us to implement partial failures in the Lambda, but it might be even more correct for PathsProcessor to return something like a list of Eithers to indicate the success or failure of each path, rather than just returning the failed ones.

PR #2798, opened 2025-01-03
Pb/update/non aws

paul-butcher:
## What does this change?

This updates the dependencies found in this Scala Steward PR: https://github.com/wellcomecollection/catalogue-pipeline/pull/2727

Builds on #2727 fail because of Scala 2.13, the addSbtPlugin calls in [plugins.sbt](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/project/plugins.sbt) all fail because the relevant 2.13 versions of the plugins cannot be found.

We should still endeavour to upgrade to 2.13, but this PR leaves it at 2.12 to isolate that upgrade from all the libraries. 

## How to test

Does it build? Does it run? Do the test pass?

I haven't looked at what, if any, benefits there are in any of these updates. My goal here is just to isolate the automated library updates from the scala one.

Success, therefore, is that the next time Scala Steward creates an update PR, it won't be as big as this one.

Library updates can break things in. mysterious ways.  We'll have to cross that. bridge when we come to it.


paul-butcher:
I think this is a bust.  We'll need to update things in Scala-libs first.  At that point, the original PR this comes from might just work.

PR #2799, opened 2025-01-06
Tweaks for image surpression script

kenoir:
## What does this change?

This change allows failure in certain cases:
- The image is already missing from DLCS, in cases where some of a batch have already been deleted
- The CloudFront invalidation fails, this can happen when processing a large number of images and is arguably a non-critical error when invalidating a large batch. An error will appear in the output logs.

In addition it will quickly fail if the user does not have the GitHub CLI installed which is a pre-requisite for updating the surpression doc in the private repo.

## How to test

- [ ] Run locally, test using dry-run.

Easier image supressions.

Removing the fail on CloudFront invalidation is potentially a risk in some cases depending on the urgency of take downs. The output logs should be reviewed as part of running the script in cases where immediate take downs are required.

### DIFF
--- a/scripts/miro_updates.py
+++ b/scripts/miro_updates.py
@@ -13,6 +13,8 @@ import os
 import boto3
 import httpx
 
+from botocore.exceptions import ClientError
+
 from _common import (
     git,
     get_api_es_client,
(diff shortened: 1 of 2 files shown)


kenoir:
Part of https://github.com/wellcomecollection/platform/issues/5860

PR #2801, opened 2025-01-08
New scala libs

paul-butcher:
## What does this change?

- Adds changes from https://github.com/wellcomecollection/scala-libs/pull/264
- Bumps to https://github.com/wellcomecollection/scala-libs/releases/tag/v32.43.5
- Harmonises logback and bloop versions to match those used in scala-libs

## How to test

There should be no material change. This simply brings in updated version of dependencies

Fewer outdated dependencies get reported by dependabot/scalasteward et al.

Maybe something is incompatible, but forwards is the best direction.


kenoir:
Adds changes from https://github.com/wellcomecollection/scala-libs/pull/264
Bumps to https://github.com/wellcomecollection/scala-libs/releases/tag/v32.43.5

kenoir (approved):

PR #2802, opened 2025-01-08
Reinstate 2024-11-18 pipeline and fix deployment

kenoir:
## What does this change?

This change resolves:
- https://github.com/wellcomecollection/platform/issues/5869
- https://github.com/wellcomecollection/platform/issues/5870

## How to test

- [x] Apply the terraform, is the new pipeline created?
- [ ] Watch Buildkite deployment, [is this problem resolved](https://buildkite.com/wellcomecollection/catalogue-pipeline-deploy-pipeline/builds/682#0194467f-aacb-459c-bd8b-bf7ad0910f82)?

2024-11-18 pipeline is reinstated and ready to be reindexed into, enabling continued work in the catalogue-pipeline.

The production catalogue-api is not pointed at this index so should not be impacted.


builds/deploy_lambda_services.sh
```diff
@@ -7,10 +7,13 @@ set -o pipefail
 PIPELINE_NAMESPACE="catalogue-$PIPELINE_DATE"
 REPOSITORY_URI="760097843905.dkr.ecr.eu-west-1.amazonaws.com"
 
-for SERVICE_NAME in "$@"
+for FUNCTION_SUFFIX_IMAGE_NAME_PAIR in "$@"
 do
-  IMAGE_URI="${REPOSITORY_URI}"/uk.ac.wellcome/"${SERVICE_NAME}":"env.${PIPELINE_DATE}"
-  FUNCTION_NAME="${PIPELINE_NAMESPACE}"-"${SERVICE_NAME}"
+  FUNCTION_SUFFIX="${FUNCTION_SUFFIX_IMAGE_NAME_PAIR%:*}"
```
paul-butcher, line 12: I find this a tiny bit confusing because the function suffix comes before the image name in the input value, so this line splits the suffix off from the beginning of the input value (where it is a prefix) in order to add it on to the end of the function name (where it becomes a suffix).

Can we swap the order, so that the input values look like `batcher:r_embed_batcher` and `relation_embedder:r_embed_embedder`?

kenoir: Have done this and added some comments about the parameter expansion which is fairly new to me.

paul-butcher (approved):

PR #2803, opened 2025-01-09
Remove unused js

paul-butcher:
## What does this change?

This removes the Javascript and CSS linting stuff from the root of the project.

There is only one small part of this project where Javascript is used, and that is rarely touched and likely to be removed as its use has been superseded by the [datalink explorer](https://github.com/wellcomecollection/datalink-explorer)

The now-deleted package.json was a source of some critical vulnerability warnings (https://github.com/wellcomecollection/catalogue-pipeline/security/dependabot/220 and https://github.com/wellcomecollection/catalogue-pipeline/security/dependabot/211).  Although these are not real vulnerabilities because this package was unused, and certainly not on a live system, they are a distraction when we investigate actual vulnerabilities.

## How to test

Deleted code is debugged code!

Fewer critical vulnerabilities, fewer dependencies to keep up to date.

Maybe someone is using the eslint pre commit hook, and I've misjudged its usefulness in this repository.  If so, we can close this PR unmarked and I'll work on sorting out those transitive dependencies.


kenoir (approved):
👏

PR #2814, opened 2025-01-20
Remove pipeline 2024-11-05

kenoir:
> [!Note]
> This terraform change has been applied.

## What does this change?

This change removes an old pipeline, not in use in order to reduce costs and keep things tidy.

## How to test

- [ ] Run `./run_terraform.sh destroy` in the 2024-11-05 folder, does it go delete the pipeline?

Reduced costs!

Double checking that this pipeline is not referenced anywhere and that the new pipeline 2024-11-18 is in production use minimises risk.


paul-butcher (approved):
API says 2024-11-18, can't see any 11-05 queues.

PR #2818, opened 2025-02-11
Rds id generator

paul-butcher:
## What does this change?

A refactor that bundles up all the DB Initialisation stuff into one object, driven by config.

## How to test

This should (kinda) be a straight refactor.  The notable change is that buildDB happens inside workerService.run, rather than Main.

This is part of a refactor to facilitate [Creating the Lambda version](https://github.com/wellcomecollection/platform/issues/5888)

There may have been some reason why buildDB and provisionTable were kept separate, one in Main, the other in run, but I doubt it.


pipeline/id_minter/src/test/scala/weco/pipeline/id_minter/fixtures/WorkerServiceFixture.scala
```diff
@@ -54,40 +46,33 @@ trait WorkerServiceFixture
...
-      identifiersDao,
-      identifiersTableConfig,
-      mergedIndex,
-      identifiedIndex
-    ) {
-      service =>
-        testWith(service)
-    }
-  }
+//
```
kenoir, line 49: Is this left in intentionally?

kenoir (approved):

PR #2819, opened 2025-02-11
increase the search scroll keepAlive to 5 minutes

agnesgaroux:
## What does this change?

The relation_embedder is throwing a lot of messages on the DLQ
[This](https://github.com/wellcomecollection/catalogue-pipeline/pull/2816) fixed the timeouts 
Here we're increasing the length of time we can scroll, hoping to fix the `search_context_missing_exception` 

## How to test

Redrive the DLQ message to the source queue, do we still see this error?

The relation_embedder can scroll through all the ES search results to build large relation trees

None that I can think of


kenoir (approved):
Ahhh, good idea.

PR #2820, opened 2025-02-12
Add catalogue graph

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5903

Add catalogue graph code from the experimental [catalogue-graph](https://github.com/wellcomecollection/catalogue-graph) repository. 

For now, GitHub workflows are located in the nested `catalogue_graph/.github` directory and are not active as a result. They will be moved into the root `.github` directory (and slightly adjusted to reflect the fact that the code is now in a subfolder) in a separate PR.


builds/run_linting.sh
```diff
@@ -12,5 +12,5 @@ docker run --tty --rm \
 	--volume "$ROOT:/data" \
 	--workdir /data \
 	"$ECR_REGISTRY/wellcome/flake8:latest" \
-    --exclude .git,__pycache__,target,.terraform \
+    --exclude .git,__pycache__,target,.terraform,catalogue_graph \
```
StepanBrychta, line 15: We have a separate linting setup for catalogue graph code, so this removes the code from global linting to prevent conflicts.

kenoir (approved):

PR #2821, opened 2025-02-12
Update catalogue graph readme

StepanBrychta:
## What does this change?

Remove outdated content from the catalogue graph README.

### DIFF
--- a/catalogue_graph/README.md
+++ b/catalogue_graph/README.md
@@ -1,15 +1,11 @@
 # Catalogue graph pipeline
 
-Experimental repository for building a knowledge graph from catalogue concepts and works. The main goals are to:
-* Experiment with AWS Neptune and explore its features and potential limitations
-* Provision infrastructure for a knowledge graph pipeline
-* Write an initial pipeline in Python for populating the knowledge graph with concepts (LoC, MeSH, Wikidata)
-
-Eventually the contents of this repository might be merged into another repository, such as catalogue-pipeline or concepts-pipeline.
+Code and infrastructure for building the catalogue graph.
 
 See the following RFCs for more context:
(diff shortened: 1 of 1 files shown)


agnesgaroux:
WRT to Robert's comment
The repository variables need to be changed 👇
https://github.com/wellcomecollection/catalogue-pipeline/blob/2cb7c92f96799ea39ca7e572d072e60c23638a32/catalogue_graph/terraform/github/gha_role.tf#L5

https://github.com/wellcomecollection/catalogue-pipeline/blob/2cb7c92f96799ea39ca7e572d072e60c23638a32/catalogue_graph/terraform/github/gha_secret.tf#L11

I'm second guessing having named the tfstate "ci" because it's only going to be for the catalogue_grah and we'll have a different one for the pipeline

builds/run_formatting.sh
```diff
@@ -20,4 +20,4 @@ docker run --tty --rm \
 	--volume "$ROOT:/repo" \
   --workdir /repo \
 	"$ECR_REGISTRY/pyfound/black" \
-  black --exclude ".lambda_zips/|.terraform/|target/" .
+  black --exclude ".lambda_zips/|.terraform/|target/|catalogue_graph/" .
```
StepanBrychta, line 23: Exclude catalogue_graph from global black formatting to stop conflicts with ruff. In the future, we should consider switching to ruff globally for consistency.

catalogue_graph/.gitignore
StepanBrychta, line 1: Most of the contents of this .gitignore can be removed because they are nearly identical to the contents of the root .gitignore.

agnesgaroux (approved):

PR #2822, opened 2025-02-13
search scroll keepAlive 15 minutes + lambda memory 4096

agnesgaroux:
## What does this change?

As part of [this](https://github.com/wellcomecollection/platform/issues/5914) 
The relation_embedder fails a lot of messages, supposedly because the batches are not processed fast enough to fit into the ES keepAlive window
Increasing the lambda memory and the keepAlive 
The terraform has been applied 
 
## How to test

Re drive the messages to the source queue. Do they get processed successfully? 

No messages on the relation_embedder DLQ

Added cost will be minimal


kenoir (approved):
It's tempting to say we should parameterise the `scrollKeepAlive`, but not blocking for this.

PR #2823, opened 2025-02-14
ID Minter Lambda

paul-butcher:
## What does this change?

This creates a Lambda version of the ID Minter.

It also introduces a new base Lambda App that can handle partial failures.

## How to test

The local RIE lambda can be started with `run_local.sh` and invoked thus:
`cat source_ids.txt | python post_to_rie.py`

Currently, the only way to see the response is to modify the code to print it out.

One more lambda in the pipeline.  Less trouble 

This is a sensitive part of the pipeline, deployment needs to be handled carefully.

## Required in future PRs

* A bit of tidying up
   * Some of the invocation scripts were  copied and slightly modified from their relation_embedder counterparts.  It would be sensible to refactor the commonality between them and put them in a shared place.
* Write the Terraform to
   * deploy the lambda
   * set up the SQS trigger
   * remember to set ReportBatchItemFailures as described here https://docs.aws.amazon.com/lambda/latest/dg/services-sqs-errorhandling.html#services-sqs-batchfailurereporting


kenoir:
> Devise an end to end test that shows the Lambda pulling messages from the queue and leaving the failed ones on there.
> Ideally, for both the ID Minter Lambda, and SQSBatchResponseLambdaApp

My opinion on this is that dealing with failure when given a batch response is the responsibility of the SQS service, and while it would be pleasing to have a test for this, we'd be relying on localstack to faithfully implement SQS behaviour which is at a remove from testing the actual behaviour.

I'd be comfortable with testing the `SQSBatchResponseLambdaApp` functionality returns the expected response using the `SQSBatchResponse` from the AWS SDK.

In addition, for later consideration `SQSBatchResponseLambdaApp` could replace `SQSLambdaApp` as the default for simplicity, the behaviour of SQS depending on how we configure the SQS trigger (i.e. is "Report batch item failures" enabled), and whether we actually report individual failures in our services.

agnesgaroux (approved):

PR #2824, opened 2025-02-17
Fix RELATED_TO edges between MeSH concepts

Amvla:
## What does this change?

Fixed two bugs in the MeSH transformer:
* Updated xml tag to correctly identify IDs for `RELATED_TO` edges https://github.com/wellcomecollection/platform/issues/5920
* Updated xml tag to correctly identify alternative labels

## How to test

Added tests for MeSH concepts to check all attributes and edge types, and added tests for locations.

`RELATED_TO` edges between MeSH concepts in the graph.

n/a


catalogue_graph/src/transformers/mesh/concepts_transformer.py
```diff
@@ -42,6 +42,3 @@ def extract_edges(
             yield SourceConceptRelatedTo(
                 from_id=raw_concept.source_id, to_id=related_id
             )
-            yield SourceConceptRelatedTo(
-                from_id=related_id, to_id=raw_concept.source_id
-            )
```
StepanBrychta, line 47: Is this removal intentional? The `SourceConceptRelatedTo` edge model is annotated with `directed=false`, so wouldn't we want to create the edge in both directions?

StepanBrychta (approved):
Looks good!

catalogue_graph/src/transformers/mesh/concepts_transformer.py
Amvla: Well spotted! Yes this is intentional - I realised that the source data always contains these relations both ways, and since we're loading the full MeSH data both edges will be created anyway. We could keep this line in, but we'd end up with duplicates.

StepanBrychta: Good to know! It might be helpful to add a comment about this. Someone else might wonder why LoC `SourceConceptRelatedTo` edges are always created in pairs, and MeSH `SourceConceptRelatedTo` edges are not.

PR #2826, opened 2025-02-19
Ignore concept type in SourceIdentifier class #5910

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5911

Implement a change in the id minter to convert all concept types into a general type `Concept` when looking up (and creating) ids in the id minter. This will cause all concepts with the same source id and source system (e.g. LoC, MeSH) to have the same canonical id no matter their type, removing duplicates.

To prevent most existing concepts having their ids re-minted, this change will be accompanied with an equivalent change in the id minter database. This will involve running two SQL commands against the database.

The first command will delete all duplicate ids from the database. For example, if a given Library of Congress name is associated with two canonical ids (one categorised as an `Agent` and one categorised as a `Person`), one of them will be discarded:

```sql
DELETE FROM identifiers
WHERE CanonicalId IN (SELECT CanonicalId
       FROM (SELECT CanonicalId,
                    ROW_NUMBER() OVER (PARTITION BY SourceId, SourceSystem) AS duplicateRowNumber
             FROM identifiers
             WHERE OntologyType in
                   ('Person', 'Organisation', 'Place', 'Agent', 'Meeting', 'Genre', 'Period', 'Concept')) AS temp
       WHERE temp.duplicateRowNumber > 1)
```

The inner nested query selects all canonical identifiers generated for all concepts, partitioning the results based on a combination of source identifier and source system (`OntologyType` is omitted from the partitioning so that matching source identifiers are returned in the same partition regardless of their type). It then assigns a row number value to each element in each partition. For example, if a given LoC name has two canonical ids associated with it, they will both belong to the same partition — one of them will have row number 1, and the other one will have row number 2.

The outer nested query then filters for canonical ids which have row number above 1, only selecting duplicate identifiers. In the LoC example above, only the canonical id with row number 2 will be selected.

Finally, the outer DELETE statement removes all rows whose canonical ids match the selected ids.

The second command will replace all concept types with the general type `Concept`:

```sql
UPDATE identifiers
SET OntologyType = 'Concept' WHERE OntologyType IN ('Person', 'Organisation', 'Place', 'Agent', 'Meeting', 'Genre', 'Period', 'Concept')
```

(The first command needs to be executed first so that the second command does not create duplicate rows, breaking uniqueness constraints.)

## How to test

Id minter unit tests were expanded to better cover concepts.

The database change was tested on a database clone created in the platform account (`identifiers-testing`).

Concepts with the same combination of source (e.g. LoC, MeSH, label-derived) and source identifier should always have the same canonical id.

To mitigate risks  this change will be released via the following process:
1. The catalogue pipeline will be paused. Pending items will start queuing up in SQS.
2. This PR will be merged and deployed.
3. An RDS blue/green deployment will be initiated, creating a 'staging' environment for the id minter database.
4. The SQL commands will be executed against the staging environment and quick checks will be performed to verify that the commands executed successfully.
5. If step 4 is successful, the staging environment will be promoted to production.
6. The catalogue pipeline will be resumed.
7. At a later point, a full reindex will be performed to remove duplicate concept ids.


pipeline/id_minter/src/main/scala/weco/pipeline/id_minter/steps/IdentifierGenerator.scala
```diff
@@ -11,6 +11,17 @@ import scala.util.{Failure, Success, Try}
 class IdentifierGenerator(identifiersDao: IdentifiersDao) extends Logging {
   import IdentifiersDao._
 
+  private val conceptSubTypes = List(
```
kenoir, line 14: Should this align with: https://github.com/wellcomecollection/catalogue-pipeline/blob/main/common/internal_model/src/main/scala/weco/catalogue/internal_model/work/AbstractRootConcept.scala#L15?

pipeline/id_minter/src/test/scala/weco/pipeline/id_minter/steps/IdentifierGeneratorTest.scala
```diff
@@ -290,4 +290,80 @@ class IdentifierGeneratorTest
         )
     }
   }
+
+  it("converts concept ontologyType values to Concept") {
+    withIdentifierGenerator() {
+      case (identifierGenerator, identifiersTable) =>
+        implicit val session = NamedAutoSession('primary)
+
+        val sourceIdentifier = createSourceIdentifierWith(
```
kenoir, line 299: Is it useful for this to test to go over all the expected types? Even reproducing them here without reference to the source forcing a test change if things change?

kenoir:
A breakdown of the delete update queries would be useful in the nested SQL query (I had to research a bit to understand what's going in with `ROW_NUMBER`).

pipeline/id_minter/src/test/scala/weco/pipeline/id_minter/steps/IdentifierGeneratorTest.scala
StepanBrychta: Good idea, done.

StepanBrychta:
@kenoir I updated the description to include a breakdown of the first SQL statement.

pipeline/id_minter/src/main/scala/weco/pipeline/id_minter/steps/IdentifierGenerator.scala
kenoir: Thanks for the explanation, one non-blocking comment that it might be a good idea to add a comment to where the source is copied from (and vice-versa) if they need to be kept in sync.

kenoir (approved):
One non-blocking comment on referencing dependent data.

PR #2830, opened 2025-02-21
Consolidate lambda run_local.sh script

kenoir:
## What does this change?

Part of: https://github.com/wellcomecollection/platform/issues/5949

Depends on: https://github.com/wellcomecollection/catalogue-pipeline/pull/2823

This change consolidates the 3 `run_local.sh` scripts for the id_minter, relation_embedder & batcher into a single script at the root of the repository for simplicity.

We also update related documentation, and add a default for pipeline date by going to the catalogue API `_elasticConfig` endpoint.

## How to test

- [ ] Check out this change, does it work as expected?

Lack of proliferation of copy pasted scripts that are slightly different.

This should only impact the development workflow, documentation has been updated to mitigate.


agnesgaroux:
We like fewer scripts!

PR #2832, opened 2025-02-21
Terraform lambda id_minter and allow deployment

kenoir:
## What does this change?

Part of: https://github.com/wellcomecollection/platform/issues/5888

Depends on: https://github.com/wellcomecollection/catalogue-pipeline/pull/2831

This change adds the terraform required to provision the id_minter lambda, and deploy it as part of usual CI.

> [!Note]
> This change will incorporate testing and replacement of the id_minter, so will remain in draft until this change can be tested as part of the production pipeline and we've confirmed it behaves as expected.
>
> **Update:** We can keep a scaled down id_minter and queue around for a few days, and remove it in another PR. This PR will replace the id_minter functionality but keep the infra and queue as a buffer in case we need to replay it.

## How to test

- [x] Terraform this change, and observe that the id_minter lambda behaves as expected.

Replacement of the id_minter and reduction in the number of ECS services, simplifying the catalogue pipeline and improving developer experience.

If this service does not behave as expected it may cause the production pipeline to temporarily stop processing requests. It contains the same code for creating identifiers, so we should be safe in presuming it will not cause corruption of the id_minter database. 

A recent snapshot of the id_minter database is a good mitigation of any risk associated with using this new implementation of the service, so it can be restored in the unlikely event there is an issue.


pipeline/terraform/modules/stack/service_id_minter.tf
```diff
@@ -45,8 +52,46 @@ module "id_minter" {
...
+  max_capacity = 0 //local.max_capacity
 
   fargate_service_boilerplate = local.fargate_service_boilerplate
 }
+
+module "id_minter_lambda" {
+  source = "../pipeline_lambda"
+
+  pipeline_date = var.pipeline_date
+  service_name  = "${local.namespace}_batcher"
```
paul-butcher, line 67: Should this be id_minter?
```suggestion
  service_name  = "${local.namespace}_id_minter"
```

pipeline/terraform/modules/stack/service_id_minter.tf
```diff
@@ -11,6 +11,13 @@ module "id_minter" {
   name            = "id_minter"
   container_image = local.id_minter_image
 
+  // Override entrypoint & command to dual use lambda container image
+  // This should be removed once we have a dedicated batcher_lambda image
```
agnesgaroux, line 15: "id_minter" image?

agnesgaroux (approved):
Non blocking comment on a comment 🙃 jst for clarity

PR #2838, opened 2025-02-27
Fix issue with local script adding newlines to ids sent to id minter

StepanBrychta:


kenoir:
Replaced by https://github.com/wellcomecollection/catalogue-pipeline/pull/2848

PR #2839, opened 2025-02-28
Add a Concepts Ingestor service

kenoir:
## What does this change?

This change adds a concepts-ingestor service providing feature parity with the existing concepts-pipeline.

We've got the following lambdas:

- **Concept Ingestor Trigger:** Asks neptune how many records there are, generates jobs to process the data and put it in S3 as parquet files.
- **Concept Ingestor Loader (Map State):** Work to query Neptune, generate parquet files and write to S3 is parallelised.
- **Concept Ingestor Indexer (Map State):** Takes the output of the previous map state step and parallelises work to index the records in S3.

Here's what the state machine looks like:

Screenshot 2025-02-28 at 16 19 59

The state machine uses a [state map step](https://docs.aws.amazon.com/step-functions/latest/dg/state-map.html) for the loading and indexing step in order to parallelise the operation and allow different concurrency settings for writing to S3 and writing to Elasticsearch (we are writing to the production cluster so we want to be careful).

Running locally:

`ingestor_trigger.py`
```
AWS_PROFILE=platform-developer \
INGESTOR_SHARD_SIZE=20000 \
./src/ingestor_trigger.py --job-id=dev
```

`ingestor_loader.py`
```
AWS_PROFILE=platform-developer \
INGESTOR_S3_BUCKET=wellcomecollection-catalogue-graph \
INGESTOR_S3_PREFIX=ingestor \
./src/ingestor_loader.py --start-offset=0 --end-index=50
```

`ingestor_indexer.py`
```
INGESTOR_PIPELINE_DATE=2024-11-18 \
AWS_PROFILE=platform-developer \
./src/ingestor_indexer.py \
--s3-url s3://wellcomecollection-catalogue-graph/ingestor/dev/00000000-00000050.parquet
```

`ingestor_local.py` script to run the whole thing serially locally, if required:
```
AWS_PROFILE=platform-developer \
INGESTOR_SHARD_SIZE=20000 \
./src/ingestor_local.py --job-id=dev --pipeline-date 2024-11-18
```

## How to test

- [x] Do the tests pass?
- [x] Run the catalogue-ingestor state machine, does it succeed?
- [x] Test with local Elasticsearch, does it index things as expected?

We're able to replace the existing concepts-pipeline.

The new index must be as stable, and provide the same data as the previous one.


catalogue_graph/src/ingestor_loader.py
```diff
@@ -0,0 +1,133 @@
...
+    s3_object_key = f"{event.job_id}/{filename}"
+    s3_url = f"s3://{config.loader_s3_bucket}/{config.loader_s3_prefix}/{s3_object_key}"
+
+    load_data(
+        s3_url,
+        transform_data(
+            extract_data(event.start_offset, event.end_index, config.is_local)
+        ),
+    )
+    print("Data loaded successfully.")
```
StepanBrychta, line 88: I think it would be better to separate these nested function calls (in this Lambda and the other Lambdas). Something like this (possibly with more descriptive variable names):

```
extracted = extract_data(event.start_offset, event.end_index, config.is_local)
transformed = transform_data(extracted)
load_data(s3_url, transformed)
```

It's more compact and easier to debug (since it allows us to insert print statements between individual steps). I also think it's easier to read.

catalogue_graph/terraform/state_machine_ingestor.tf
```diff
@@ -0,0 +1,103 @@
...
+                    "Lambda.AWSLambdaException",
+                    "Lambda.SdkClientException",
+                    "Lambda.TooManyRequestsException"
+                  ],
+                  "IntervalSeconds" : 1,
+                  "MaxAttempts" : 3,
+                  "BackoffRate" : 2,
+                  "JitterStrategy" : "FULL"
+                }
+              ],
```
StepanBrychta, line 51: I think this is great. I think we should do something similar for the extractor and bulk loader state machines.

kenoir: These are actually the defaults that get generated when you use the visual designer in the console. They seemed to make sense so I kept them.

catalogue_graph/elasticsearch.docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      - discovery.type=single-node
+      - ES_JAVA_OPTS=-Xms512m -Xmx512m
+    ports:
+      - "9200:9200"
+    volumes:
+      - esdata:/usr/share/elasticsearch/data
+    networks:
+      - elastic
+
+  kibana:
```
StepanBrychta, line 17: Being able to see the indexed documents in Kibana locally is *very* nice.

catalogue_graph/src/ingestor_indexer.py
```diff
@@ -0,0 +1,110 @@
...
+def lambda_handler(event: IngestorIndexerLambdaEvent, context: typing.Any) -> int:
+    return handler(
+        IngestorIndexerLambdaEvent.model_validate(event), IngestorIndexerConfig()
+    )
+
+
+def local_handler() -> None:
+    parser = argparse.ArgumentParser(description="")
+    parser.add_argument(
+        "--s3-url",
```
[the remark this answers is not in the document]
kenoir: I think strictly the `https://` form and `s3://` form, are both URLs. It would definitely be clearer if the description had an example in it to prevent confusion though.

PR #2845, opened 2025-03-06
Test ingestor indexer

agnesgaroux:
## What does this change?

Adds test for ingestor_indexer, see https://github.com/wellcomecollection/catalogue-pipeline/pull/2839

## How to test

run tests, do they pass? 

Test pass

N/A


catalogue_graph/src/ingestor_indexer.py
```diff
@@ -52,17 +53,17 @@ def load_data(
...
-    es = elasticsearch.get_client(pipeline_date, is_local)
+    es = utils.elasticsearch.get_client(pipeline_date, is_local)
 
     def generate_data() -> Generator[dict]:
         for concept in concepts:
             yield {
                 "_index": index_name,
                 "_id": concept.query.id,
-                "_source": concept.model_dump_json(),
+                "_source": json.loads(concept.model_dump_json()),
```
agnesgaroux, line 63: Not 100% sure about this. Without it the `_source` is wrapped in an extra set of "" which doesn't match the expectation

paul-butcher: `model_dump_json` produces a json string, rather than an object. it is essentially turning the underlying data into a dict, and then converting it to a JSON string, which this line then converts back into a dictionary..

try this, which should give the same result
```suggestion
                "_source": concept.model_dump(),
```

agnesgaroux: Oh nice, much elegant, such class

PR #2846, opened 2025-03-06
Add new pipeline 2025-03-6

StepanBrychta:
## What does this change?

[ #5912](https://github.com/wellcomecollection/platform/issues/5912)

Following on https://github.com/wellcomecollection/platform/issues/5911.

After reindexing with the new pipeline, all concepts which reference the same source concept or have the same labels (for label-derived concepts) should have the same canonical id.

Risks should be minimal.


agnesgaroux:
Approved for reindex 👍
I assume it'll be made ready for review once it's all said and done and the reindexing_state variables are set to false?

pipeline/terraform/modules/stack/service_id_minter.tf
```diff
@@ -52,7 +52,7 @@ module "id_minter_lambda" {
   queue_config = {
     topic_arns          = local.transformer_output_topic_arns
     max_receive_count   = 3
-    maximum_concurrency = 10
+    maximum_concurrency = 30
```
StepanBrychta, line 55: When the ID minter concurrency is set to 10, it becomes a bottleneck during a reindex. Increasing the concurrency to 30 helps, without overwhelming the database.

kenoir (approved):

PR #2847, opened 2025-03-07
Close the scroll context in the id_minter

kenoir:
## What does this change?

This change copies and modifies the search iterator class from Elastic4S in order to properly close the [scroll context](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html). Following https://github.com/wellcomecollection/catalogue-pipeline/pull/2842 we we're seeing an error pattern that looks like this: 

image

@StepanBrychta & @paul-butcher suggest this is happening:

```
1. In a short period of time, we open lots of scroll contexts and successfully process some messages.
2. The scroll contexts are not automatically cleared, so we quickly reach the maximum number of 500 open scroll contexts. 
3. Any further Lambda executions fail (explaining the 0% success rate between the spikes).
4. After 5 minutes, the scroll contexts opened in step 1. are automatically cleared. Back to step 1.
```

See: https://wellcome.slack.com/archives/C02ANCYL90E/p1741274178988389?thread_ts=1741260150.022259&cid=C02ANCYL90E

Here we modify the `hasNext` function to notice when we have reached the end of the iterator, and then close the associated scroll context.

This change is difficult to write tests for but has been tested under load in production and seems to behave without errors.

## How to test

- [ ] Build this locally, push it to a dev tag, update the deployed lambda to use it. Does it error in the same pattern when run?

Reindexes that run in a timely manner without error.

This change couples us heavily to the current implementation of the elastic4s lib as it duplicates an internal implementation of `SearchIterator`. We should modify our usage to avoid use of the scroll API as recommended in the documentation.

### DIFF
--- a/pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/RelationsService.scala
+++ b/pipeline/relation_embedder/relation_embedder/src/main/scala/weco/pipeline/relation_embedder/RelationsService.scala
@@ -4,17 +4,15 @@ import org.apache.pekko.NotUsed
 import org.apache.pekko.stream.scaladsl.Source
 import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.circe._
-import com.sksamuel.elastic4s.{ElasticClient, HitReader, Index, RequestFailure, RequestSuccess}
+import com.sksamuel.elastic4s.{ElasticClient, Index}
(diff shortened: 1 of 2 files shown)


paul-butcher (approved):

paul-butcher:
I actually think this decouples us a bit from E4S.  We now have an object that is within our control that we can modify accordingly.

PR #2850, opened 2025-03-10
Concept-related changes

StepanBrychta:
## What does this change?

Related to https://github.com/wellcomecollection/catalogue-api/pull/847 and https://github.com/wellcomecollection/wellcomecollection.org/pull/11697.

* Adds utility functions to remove nodes from the catalogue graph. These are needed whenever we do a catalogue graph reindex which involves removing concepts/works. See https://github.com/wellcomecollection/platform/issues/5912.
* Adds the option to pass `CATALOGUE_SNAPSHOT_URL` as an environment variable when extracting catalogue concepts/works. This is needed so that we can extract works/concepts from a dev snapshot instead of a prod snapshot. See [this PR](https://github.com/wellcomecollection/catalogue-api/pull/847).
* Extracts source concept 'description' from the catalogue graph and indexes it into the concepts index. See https://github.com/wellcomecollection/platform/issues/5971.
* Fixes a bug in the catalogue concepts extractor which caused Subjects not to be extracted. 

## How to test

* Run the ingestor. Do the outputted parquet files contain the 'description' field with correct data?
* Run the catalogue concepts transformer locally. Does it output concepts of type 'Subject'?
* Run the catalogue concepts/works transformer with a different snapshot URL (there is a dev snapshot at https://data.wellcomecollection.org/catalogue_dev/v2/works.json.gz). Does it work?
* Check the ES concepts index (2025-03-06). Does it store descriptions?

* Concept 'descriptions' indexed in ES.
* Concepts of type 'Subject' correctly extracted. 

There should be no major risks associated with this change.


catalogue_graph/tests/transformers/test_catalogue_works_transformer.py
```diff
@@ -38,47 +48,71 @@ def test_catalogue_concepts_transformer_nodes() -> None:
...
+        "s6s24vd7",
+        WorkHasConcept(
+            from_type="Work",
+            to_type="Concept",
+            from_id="m4u8drnu",
+            to_id="s6s24vd7",
+            relationship="HAS_CONCEPT",
+            directed=True,
+            attributes=WorkHasConceptAttributes(referenced_in="subjects"),
+        ),
```
StepanBrychta, line 72: I changed the test to not care about the order in which edges are returned — this is an implementation detail which can change.

agnesgaroux (approved):

PR #2852, opened 2025-03-14
Parameterise the catalogue-graph ingestor trigger with pipeline date

kenoir:
## What does this change?

This change is in order to allow ingestion runs to be namespaced and monitored on a per pipeline basis, required for monitoring. We add some stats on file size and record count to the loader that will be used in reporting and monitoring in a later PR.

## How to test

Allow us to run multiple pipeline and monitor them seperately.

The catalogue graph is not quite in production use, so we should be able to catch errors before they are in public view, mitigating risk.

### DIFF
--- a/catalogue_graph/src/ingestor_local.py
+++ b/catalogue_graph/src/ingestor_local.py
@@ -4,14 +4,12 @@ import argparse
 
 from ingestor_indexer import (
     IngestorIndexerConfig,
-    IngestorIndexerLambdaEvent,
 )
 from ingestor_indexer import (
     handler as indexer_handler,
 )
 from ingestor_loader import (
     IngestorLoaderConfig,
-    IngestorLoaderLambdaEvent,
 )
 from ingestor_loader import (
     handler as loader_handler,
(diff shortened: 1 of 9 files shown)


catalogue_graph/src/ingestor_local.py
```diff
@@ -38,7 +36,13 @@ def main() -> None:
         "--pipeline-date",
         type=str,
         help="The date to use for the pipeline, required.",
-        required=True,
+        required=False,
+    )
+    parser.add_argument(
+        "--limit",
+        type=int,
+        help="The number of shards to process, will process all if not specified.",
```
paul-butcher, line 44: Is it clear enough what shard means here, given the context? (parquet shard vs. DB shard) I think it is.

paul-butcher (approved):

catalogue_graph/src/ingestor_local.py
kenoir: I'm not sure, this was a bit of an afterthought to allow this to be tested. I don't expect this will be used outside of testing but agree it's not super clear.

PR #2853, opened 2025-03-17
Monitor Catalogue Graph trigger output

kenoir:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5967

Adds a step after the ingestor_trigger that:
- Accepts the mapped output and sums the record_count
- Retrieves the last trigger status (record count) from the namespace in S3, and compares it to the current count
- Continues if no previous count, or count is within X%, fails otherwise
- Reports status as a CloudWatch metric
- Overwrites the current record count to the namespace in S3
- Forward the input (list of events) to the loader map state step

This change is intended to ensure prevent large changes from propagating through the pipeline unchecked.

## How to test

- [ ] Try and use the run local scripts, does it behave as expected?
- [ ] Run this deployed, does it behave as expected?

A reduction in risk of pushing unexpected changes into public view.

The catalogue graph ingestion pipeline is not in public view, so this should be relatively safe.

### DIFF
--- a/catalogue_graph/src/ingestor_trigger_monitor.py
+++ b/catalogue_graph/src/ingestor_trigger_monitor.py
@@ -1,0 +1,152 @@
+import typing
+
+import boto3
+import smart_open
+from pydantic import BaseModel
+
+from clients.metric_reporter import MetricReporter
+from config import INGESTOR_S3_BUCKET, INGESTOR_S3_PREFIX
+from ingestor_loader import IngestorLoaderLambdaEvent
+
+
+class MonitorStepEvent(BaseModel):
+    pipeline_date: str | None = None
+    force_pass: bool = False
+    report_results: bool = True
+
+
+class IngestorTriggerMonitorLambdaEvent(MonitorStepEvent):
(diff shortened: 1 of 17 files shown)


catalogue_graph/src/ingestor_trigger_monitor.py
```diff
@@ -0,0 +1,152 @@
...
+        "pipeline_date": event.pipeline_date or "dev",
+        "step": "ingestor_trigger_monitor",
+        "job_id": report.job_id,
+    }
+
+    print(f"Reporting results {report}, {dimensions} ...")
+    if config.report_results:
+        reporter = MetricReporter("catalogue_graph_ingestor")
+        reporter.put_metric_data(
+            metric_name="end_index", value=report.end_index, dimensions=dimensions
```
agnesgaroux, line 117: Could the metric be called something more descriptive? `graph_concepts_sum`, `total_ingestable_concepts` ?

agnesgaroux: `graph_records_count`

agnesgaroux (approved):

PR #2854, opened 2025-03-17
Monitor Catalogue Graph loader output

kenoir:
> [!Note]
> This PR should allow the [daily pipeline run to succeed](https://760097843905-2nxogsey.eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/v2/executions/details/arn:aws:states:eu-west-1:760097843905:execution:concepts-pipeline_daily:7d963918-a1b0-4e18-9724-994ba1a3d148).

## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5967

Adds a step after the ingestor_loader that:
- Accepts the mapped output of the loader and sums the record count / file sizes
- Retrieves the last load status from the namespace in S3, and compares it to the current summed count and sizes
- Continues if no previous count, or count is within X%, fails otherwise
- Reports status as a CloudWatch metric
- Overwrites the current load status to the namespace in S3
- Forward the input (list of events) to the indexer map state step

This change is intended to ensure prevent large changes from propagating through the pipeline unchecked.

## How to test

- [ ] Try and use the run local scripts, does it behave as expected?
- [ ] Run this deployed, does it behave as expected?

A reduction in risk of pushing unexpected changes into public view.

The catalogue graph ingestion pipeline is not in public view, so this should be relatively safe.


catalogue_graph/scripts/deploy_ingestor_lambdas.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
agnesgaroux, line 1: Why do the ingestor lambdas need their own separate deploy script?

catalogue_graph/src/ingestor_loader_monitor.py
```diff
@@ -0,0 +1,170 @@
...
+
+    report = None
+    try:
+        report = run_check(event, config)
+    except ValueError as e:
+        print(f"Check failed: {e}")
+        raise e
+
+    if report is not None and event.report_results:
+        report_results(report, config.report_results)
```
agnesgaroux, line 149: Can `report_results` be set on the event or the config? Can we do something like 
```
report_results = event.report_results or config.report_results
if report is not None and report_results:
        report_results(report)
```
then remove the `if report_results:` check on l.123?

catalogue_graph/terraform/lambda_catalogue_ingestor.tf
```diff
@@ -83,6 +128,11 @@ resource "aws_iam_role_policy" "ingestor_loader_lambda_read_secrets_policy" {
   policy = data.aws_iam_policy_document.allow_secret_read.json
 }
 
+resource "aws_iam_role_policy" "ingestor_loader_lambda_s3_read_policy" {
```
agnesgaroux, line 131: Does the ingestor_loader lambda need S3 read?

agnesgaroux (approved):

catalogue_graph/src/ingestor_loader_monitor.py
agnesgaroux: (same in the ingestor_trigger_monitor)

kenoir: Yes, can simplify - we still want to alwayca call report results as it will log the report even if we don't send the results to cloudwatch, but we can move the check _inside_ that function

catalogue_graph/terraform/state_machine_extractors.tf
```diff
@@ -46,9 +46,9 @@ resource "aws_sfn_state_machine" "catalogue_graph_extractors_monthly" {
           StateMachineArn = aws_sfn_state_machine.catalogue_graph_extractor.arn
           Input = {
             "stream_destination" : "s3",
-            "transformer_type.$" : "$$.Execution.Input.transformer_type",
-            "entity_type.$" : "$$.Execution.Input.entity_type",
-            "sample_size.$" : "$$.Execution.Input.sample_size"
+            "transformer_type" : "${task_input.transformer_type}",
+            "entity_type" : "${task_input.entity_type}",
+            # "sample_size.$" : "$$.Execution.Input.sample_size"
```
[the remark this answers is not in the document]
agnesgaroux: 👀 
```
"sample_size" : contains(keys(task_input), "sample_size") ? "${task_input.sample_size}" : null
```

catalogue_graph/terraform/lambda_catalogue_ingestor.tf
kenoir: Yes, reading `content_length` after it's been written by `smart_open` requires read.

PR #2855, opened 2025-03-19
More careful label matching

StepanBrychta:
## What does this change?

Match labels of label-derived concepts to source concepts more carefully to mitigate false positives. This includes:
* Prioritising matches on main labels: If a match is found on main source label, do not create additional matches on alternative labels
* Source prioritisation: Match on sources in the following order: MeSH, LoC SH, LoC Names. If a match is found on a higher-priority source, do not try matching on lower-priority sources.
* Not matching ambiguous concepts: If a label matches multiple alternative source labels, do not create any matches.
* Do not match things to agents and vice versa: If a concept is categorised as an Agent, Person, or Organisation, only match it to lc-names. 
* Blacklisting labels which previously matched to an incorrect source concept

Here are lists of matches that we are losing due to the some of the new rules:
[ambiguous_labels.txt](https://github.com/user-attachments/files/19346247/ambiguous_labels.txt)
[things_to_people.txt](https://github.com/user-attachments/files/19346256/things_to_people.txt)
[people_to_things.txt](https://github.com/user-attachments/files/19346255/people_to_things.txt)

Also addresses https://github.com/wellcomecollection/platform/issues/5973 

## How to test

Unit tests were added to cover the new logic.

* Lower likelihood of incorrect matches of label-derived concepts to source concept.
* A simple mechanism for manually adding concept labels into a deny list which should never be matched.

This work does not introduce any new risk — it adds stricter label-matching criteria to reduce existing risks. However, the risk is that the new label-matching criteria are still not strict enough.


catalogue_graph/src/transformers/catalogue/data/blacklisted_concept_labels.txt
```diff
@@ -0,0 +1,2 @@
+siamese
```
kenoir, line 1: I would use the terminology "allowlist" and "denylist" as it's clearer what the purpose is.

catalogue_graph/tests/transformers/test_id_label_checker.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Match on alternative label
+    assert id_label_checker.get_id("etching_s", "Concept") == "sh85045046"
+    assert id_label_checker.get_id("Some example concept", "Concept") == "sh85123237"
+
+
+def test_id_label_checker_denylist() -> None:
+    id_label_checker = _setup_id_label_checker()
+
+    # Do not match blacklisted concept labels
```
kenoir, line 32:
```suggestion
    # Do not match denylisted concept labels
```

kenoir (approved):

PR #2864, opened 2025-03-28
Increase memory for router

kenoir:
## What does this change?

Following some recent repeated failures of the router it looks like it might need more memory.

Screenshot 2025-03-28 at 08 45 15

This change increases the CPU and memory to 1vCPU & 8GB to ensure it has enough resources. This service is likely to be [subsumed into the merger soon](https://github.com/wellcomecollection/platform/issues/5985) so i'm not too worried about rightsizing, and want to guarantee the resource increase will be imapctful.

## How to test

- [ ] Apply the change, redrive the messages, does that solve the problem?

Screenshot 2025-03-28 at 08 47 54

The router does not fail processing messages.

Risks should be minimal as this is not a change that will impact output.

### DIFF
--- a/pipeline/terraform/modules/stack/relation_embedder/router.tf
+++ b/pipeline/terraform/modules/stack/relation_embedder/router.tf
@@ -49,8 +49,8 @@ module "router" {
 
   secret_env_vars = var.pipeline_storage_es_service_secrets["router"]
 
-  cpu    = 1024
-  memory = 2048
+  cpu    = 2048
+  memory = 8192
 
   min_capacity = var.min_capacity
   max_capacity = var.max_capacity


agnesgaroux (approved):
Out of curiosity, why do you memory x4 when the CPU is only x2, and was actually where the issue was (memory was well within tolerable levels)?

kenoir:
> Out of curiosity, why do you memory x4 when the CPU is only x2, and was actually where the issue was (memory was well within tolerable levels)?

There are only [certain valid memory/cpu combinations](https://gist.github.com/matthewjberger/6edfff5649d297bcc454a295c552dc5a).

PR #2865, opened 2025-03-28
Concepts pipeline report

agnesgaroux:
## What does this change?

Following the [alerting](https://github.com/wellcomecollection/platform/issues/5966), here's some reporting

Adds a lambda as a last step to the ingestor state machine. It
- picks up a few reports from s3://wellcomecollection-catalogue-graph/ingestor/<pipeline_date>
- formats the data much like the snapshot reporter
- posts to #wc-search-alerts Slack channel

## How to test
```
AWS_PROFILE=platform-developer python3.13 src/reporter.py --pipeline-date=2025-03-06 --job-id=20250327T1618 --success-count=570065
```

Reports are posted to #wc-search-alerts Slack channel every time the pipeline successfully runs

Lambda permissions are read only


agnesgaroux:
Screenshot 2025-03-28 at 11 41 32

catalogue_graph/src/reporter.py
```diff
@@ -0,0 +1,192 @@
...
+def publish_report(report: list[typing.Any], config: ReporterConfig) -> None:
+    secretsmanager = boto3.Session().client("secretsmanager")
+    slack_endpoint = secretsmanager.get_secret_value(SecretId=config.slack_secret)[
+        "SecretString"
+    ]
+
+    httpx.post(slack_endpoint, json={"blocks": report})
+
+
+def handler(event: ReporterEvent, context: typing.Any) -> None:
```
kenoir, line 144: I'd repeat pattern in the other ingestor lambdas where there is a `lambda_handler` function, that performs `model_validate` on the incoming event (which will also fill in the default values if required), and allows you to control the config handed to the handler function in testing.

catalogue_graph/src/reporter.py
```diff
@@ -0,0 +1,192 @@
...
+                "text": {
+                    "type": "mrkdwn",
+                    "text": "*Could not produce report*: job_id not supplied.",
+                },
+            },
+        ]
+
+    trigger_report_url = f"s3://{config.s3_bucket}/{config.ingestor_s3_prefix}/{pipeline_date}/{job_id}/report.trigger.json"
+    loader_report_url = f"s3://{config.s3_bucket}/{config.ingestor_s3_prefix}/{pipeline_date}/{job_id}/report.loader.json"
+    previous_loader_report_url = f"s3://{config.s3_bucket}/{config.ingestor_s3_prefix}/{pipeline_date}/report.loader.json"
```
kenoir, line 68: I think at this point the loader will have written both `/{pipeline_date}/report.loader.json"` and `/{pipeline_date}/{job_id}/report.loader.json"` so they won't differ ever.

agnesgaroux: Of course! 🙃 Originally I wanted the monitor lambdas to write a separate report and use that at the end. As things stands it's a bit tricky to find out about the previous pipeline run 🤔

catalogue_graph/src/reporter.py
```diff
@@ -0,0 +1,230 @@
...
+    ]
+
+
+def publish_report(report: list[typing.Any], config: ReporterConfig) -> None:
+    secretsmanager = boto3.Session().client("secretsmanager")
+    slack_endpoint = secretsmanager.get_secret_value(SecretId=config.slack_secret)[
+        "SecretString"
+    ]
+
+    httpx.post(slack_endpoint, json={"blocks": report})
```
StepanBrychta, line 173: I think we could use the `requests` package here instead of installing a new package.

catalogue_graph/src/reporter.py
```diff
@@ -0,0 +1,230 @@
...
+            es_record_count=current_report.record_count,
+            previous_es_record_count=latest_report.record_count,
+        )
+        with smart_open.open(
+            s3_url_final_report, "w", transport_params=transport_params
+        ) as f:
+            f.write(updated_final_report.model_dump_json())
+
+
+def dateTimeFromJobId(job_id: str) -> str:
```
StepanBrychta, line 81: snake_case is preferable

catalogue_graph/src/reporter.py
```diff
@@ -0,0 +1,230 @@
...
+        print(f"Report not found: {e}")
+
+    start_datetime = dateTimeFromJobId(job_id)
+    previous_start_datetime = dateTimeFromJobId(final_report.previous_job_id)
+    current_run_duration = round(
+        divmod(
+            (datetime.now() - datetime.strptime(job_id, "%Y%m%dT%H%M")).total_seconds(),
+            60,
+        )[0]
+    )
```
StepanBrychta, line 131: Since we're only extracting the quotient, I don't think we need to use `divmod`.

I think this simpler option would produce the same result:

```python
round((datetime.now() - datetime.strptime(job_id, "%Y%m%dT%H%M")).total_seconds() / 60)
```

Or, to always round down:

```python
int((datetime.now() - datetime.strptime(job_id, "%Y%m%dT%H%M")).total_seconds() / 60)
```

catalogue_graph/src/reporter.py
```diff
@@ -0,0 +1,230 @@
...
+    s3_url_final_report = f"s3://{config.ingestor_s3_bucket}/{config.ingestor_s3_prefix}/{current_report.pipeline_date}/{current_report.job_id}/{report_name}"
+    transport_params = {"client": boto3.client("s3")}
+
+    final_report = None
+    try:
+        with smart_open.open(s3_url_final_report, "r") as f:
+            final_report = FinalReport.model_validate_json(f.read())
+    # if file does not exist, ignore and create one
+    except (OSError, KeyError) as e:
+        print(f"No final report found: {e}. Creating report.")
```
StepanBrychta, line 48: This logic is now implemented in four separate places, so it might make sense to make a utility function for it.

catalogue_graph/src/reporter.py
agnesgaroux: Javascript habits die hard

PR #2866, opened 2025-04-02
Add graph remover lambda #5979

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5979

Adds a `graph_remover` Lambda function responsible for removing nodes/edges from the graph if they are no longer present in the corresponding bulk loader files (which means they were removed from their respective sources).

The removal works as follows:
1. On every bulk loader state machine run, the `graph_remover` Lambda is invoked as the last step.
2. The Lambda extracts IDs from the specified bulk loader file and compares them to a snapshot of IDs which was created during the previous `graph_remover` run (based on data from the previous bulk loader file).
3. The Lambda then makes a Cypher query to remove all nodes which exist in the snapshot but not in the latest bulk loader file. 
4. Finally, the Lambda overwrites the IDs snapshot with the latest IDs to be used in the next run. The Lambda also keeps an append-only log of added/removed IDs in S3 for monitoring and debugging purposes.

## How to test

The Lambda can be tested locally. Additionally, unit tests were introduced to cover the new logic.

Outdated/deactivated nodes and edges should stop accumulating in the catalogue graph.

The Lambda function uses the latest bulk loader file to determine which items to delete from the graph. this implies that if the bulk loader file contained incorrect/incomplete data, incorrect nodes/edges would be deleted. 

To mitigate this, a safety mechanism was added which activates if the difference between the two sets of IDs is over 5%.


catalogue_graph/src/clients/base_neptune_client.py
```diff
@@ -14,6 +14,8 @@
 
 DELETE_BATCH_SIZE = 10000
 
+ALLOW_DATABASE_RESET = False
```
agnesgaroux, line 17: Is this simply adding some friction around resetting the DB?

StepanBrychta: Yes, exactly. To prevent someone from calling the method accidentally.

catalogue_graph/src/graph_remover.py
```diff
@@ -0,0 +1,190 @@
...
+    try:
+        # Retrieve a list of all ids which were loaded into the graph as part of the previous run
+        previous_ids = get_previous_ids(transformer_type, entity_type)
+        is_first_run = False
+    except (OSError, KeyError):
+        print(
+            "File storing archived ids not found. This should only happen on the first run."
+        )
+        previous_ids = set()
+        is_first_run = True
```
agnesgaroux, line 112: Is this variable really useful? I think it might be confusing in the future because there'll only ever be one 1st run

agnesgaroux (commented):
I wouldn't mind a quick chat about the changes to the state machine, I find the input/output/query language combination a bit confusding

catalogue_graph/src/graph_remover.py
```diff
@@ -0,0 +1,190 @@
...
+# This is part of a safety mechanism. If two sets of IDs differ by more than 5%, an exception will be raised.
+ACCEPTABLE_DIFF_THRESHOLD = 0.05
+
+
+def get_previous_ids(
+    transformer_type: TransformerType, entity_type: EntityType
+) -> set[str]:
+    """Return all IDs from the latest snapshot for the specified transformer and entity type."""
+    s3_file_uri = f"s3://{config.INGESTOR_S3_BUCKET}/{IDS_SNAPSHOT_FOLDER}/{transformer_type}__{entity_type}.parquet"
+    df = df_from_s3_parquet(s3_file_uri)
```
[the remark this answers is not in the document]
StepanBrychta: I'm trying to make use of the `wellcomecollection-catalogue-graph` bucket, which is more general than the bulk loader bucket. (I think eventually it might be good to get rid of the `wellcomecollection-neptune-graph-loader` bucket completely — the bulk load files could be stored in a subdirectory in the `wellcomecollection-catalogue-graph` bucket).

But it would probably make sense to rename the `INGESTOR_S3_BUCKET` variable to something more general, like `CATALOGUE_GRAPH_S3_BUCKET`.

catalogue_graph/src/graph_remover.py
StepanBrychta: The main reason for this variable is to prevent all IDs from being added to the `added_ids` set on the first run.

On the first run, the snapshot of IDs won't be available and so the `previous_ids` set will be empty. As a result, computing the difference between `current_ids` and `previous_ids` would return all IDs in `current_ids`, and then all of them would be saved to the `added_ids` log file. This would be misleading because most of those IDs weren't added as part of that run — they were added before we started to keep track, and so we cannot know when we added them.

agnesgaroux (approved):

PR #2870, opened 2025-04-08
Increase bulk loader resilience

StepanBrychta:
## What does this change?

The latest monthly graph pipeline run failed several times during the bulk load stage. Reasons for these failures include:
* A single missing LoC node when inserting edges between two LoC names: This error happened due to LoC introducing a new URL prefix on a single ID.
* A single missing Wikidata node when inserting edges between Wikidata nodes: This error likely happened due to one Wikidata item changing during the pipeline run, which resulted in the item not being included in the nodes bulk load file (even though it was included in the edges bulk load file).
* Neptune running out of memory: This was a transient error which resolved itself when the bulk load was resumed

All of these caused the whole pipeline run to fail and had to be fixed before the pipeline could be resumed. This PR makes some changes so that when similar minor errors happen in the future, they do not stop pipeline execution.

To address the out of memory error (and other transient errors), a retry mechanism was added to the bulk loader state machine.

To address the other errors, a tolerance threshold was introduced into the bulk load poller Lambda function. If the number of bulk load insert errors is above 0 but below the threshold (currently set to 0.01%, or 1 in 10,000 records), the bulk load is still considered successful. 

In addition, the bulk load poller Lambda logs the result of the bulk load into a JSON file in S3, allowing us to include information about insert errors in the [concepts pipeline report](https://github.com/wellcomecollection/catalogue-pipeline/pull/2865) and investigate if needed.

## How to test

The bulk load poller change can be tested locally using an existing Neptune bulk load ID.

If a minor or a transient error occurs during a bulk load, it should not prevent the catalogue graph pipeline from completing its run. 

Even though the insert errors are minor and only affect a tiny percentage of nodes/edges, we do not want to completely silence them. Therefore, this change should be combined a a corresponding change in the concepts pipeline reporter to ensure we are aware of these errors.


catalogue_graph/src/bulk_load_poller.py
```diff
@@ -1,20 +1,102 @@
...
         }
 
-    if status == "LOAD_COMPLETED":
+    insert_error_count = overall_status["insertErrors"]
+    parsing_error_count = overall_status["parsingErrors"]
+    data_type_error_count = overall_status["datatypeMismatchErrors"]
+    formatted_time = datetime.timedelta(seconds=overall_status["totalTimeSpent"])
+
+    print(f"    Insert errors: {insert_error_count:,}")
+    print(f"    Parsing errors: {parsing_error_count:,}")
```
agnesgaroux, line 78: Why the colon after the variable?

catalogue_graph/src/bulk_load_poller.py
```diff
@@ -1,20 +1,102 @@
...
 
+INSERT_ERROR_THRESHOLD = 1 / 10000
+
+
+def log_payload(payload: dict) -> None:
+    """Log the bulk load result into a JSON file which stores all results from the latest pipeline run"""
+    # Extract the name of the bulk load file to use as a key in the JSON log.
+    bulk_load_file_uri = payload["overallStatus"]["fullUri"]
+    bulk_load_file_name = bulk_load_file_uri.split("/")[-1].split(".")[0]
+    log_file_uri = f"s3://{config.S3_BULK_LOAD_BUCKET_NAME}/bulk_load_log.json"
```
agnesgaroux, line 19: For consistency, can we use the same format for the file name as the ingestor reports, eg. report.neptune_loader.json?

catalogue_graph/src/bulk_load_poller.py
```diff
@@ -1,20 +1,102 @@
...
+    try:
+        with smart_open.open(log_file_uri, "r") as f:
+            bulk_load_log = json.loads(f.read())
+    except (OSError, KeyError):
+        # On the first run, the log file might not exist
+        bulk_load_log = {}
+
+    # Overwrite the existing result (from the last bulk load) with the current one
+    bulk_load_log[bulk_load_file_name] = payload
+
```
agnesgaroux, line 30: Do we end up with 1 file in the end when all have run with:
```
{
    "catalogue_concepts__edges": {
        // overallStatus object from load status response
    },
    // one such object for each source/entity type combination
}
```

StepanBrychta: Yes

catalogue_graph/src/bulk_load_poller.py
StepanBrychta: Sure, done!

agnesgaroux (approved):

PR #2871, opened 2025-04-09
Add concepts index remover

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5980

Follow-up on https://github.com/wellcomecollection/catalogue-pipeline/pull/2866

This introduces a new `index_remover` Lambda function, which removes unused documents from the concepts index. This is achieved by extracting a list of deleted concept IDs from the log file introduced as part of the [graph remover PR](https://github.com/wellcomecollection/catalogue-pipeline/pull/2866) and making a `DELETE` API call against the ES cluster.

This PR also modifies the `catalogue-graph-ingestor` state machine to call the new Lambda function as its last step.

Note: The Terraform changes in this PR have not been applied yet.

## How to test

The Lambda function code can be tested locally (if a pipeline_date argument is not specified, it will run against a local index).

Concepts are removed from the concepts index if they no longer exist in the catalogue graph.

This PR introduces similar risks to [this PR](https://github.com/wellcomecollection/catalogue-pipeline/pull/2866) and implements the same mitigation mechanism.


catalogue_graph/src/index_remover.py
```diff
@@ -0,0 +1,141 @@
...
+    deleted_count = response["deleted"]
+    print(f"Deleted {deleted_count} documents from the {index_name} index.")
+
+
+def get_ids_to_delete(pipeline_date: str | None) -> set[str]:
+    """Return a list of concept IDs marked for deletion from the ES index."""
+    s3_file_uri = f"s3://{config.INGESTOR_S3_BUCKET}/{DELETED_IDS_FOLDER}/catalogue_concepts__nodes.parquet"
+
+    # Retrieve a log of concept IDs which were deleted from the graph (see `graph_remover.py`).
+    df = df_from_s3_parquet(s3_file_uri)
```
kenoir, line 58: I think this will fail if the file does not exist, is that the behaviour we want here or would it be preferable to succeed but do no work? 

It would be unexpected for this to run without the graph remover having been run first so a failure is probably correct, but in that case perhaps we should have a test for it?

catalogue_graph/tests/test_index_remover.py
```diff
@@ -0,0 +1,92 @@
...
+
+from index_remover import lambda_handler
+
+
+def index_concepts(ids: list[str], index_name: str = "concepts-indexed") -> None:
+    for _id in ids:
+        MockElasticsearchClient.index(index_name, _id, {})
+
+
+def mock_deleted_ids_log_file() -> None:
```
kenoir, line 14: Because this file is so tiny and simple in contents, and the tests rely on what it contains perhaps generating it on the fly would be preferable so the contents are clearer? 

I don't think this is essential as the contents are fairly easily inspected but I think it would be a little clearer what the tests were doing.

kenoir (approved):

catalogue_graph/tests/test_index_remover.py
StepanBrychta: I think that's a good idea, and I also think we should consider doing this for all mock parquet files in the future so that updating fixtures to reflect changes in the code is easier. (Or maybe we could have a set of utility functions which generate the parquet fixtures based on some JSON/CSV input data so that we don't have to create them manually.)

catalogue_graph/src/index_remover.py
```diff
@@ -0,0 +1,141 @@
...
+    is_local: bool = False,
+) -> None:
+    ids_to_delete = get_ids_to_delete(pipeline_date)
+    current_id_count = get_current_id_count(pipeline_date, is_local)
+
+    if (
+        current_id_count > 0
+        and len(ids_to_delete) / current_id_count > ACCEPTABLE_DIFF_THRESHOLD
+        and not disable_safety_check
+    ):
```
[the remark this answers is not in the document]
StepanBrychta: Done

catalogue_graph/src/index_remover.py
StepanBrychta: Yes, I think we should let the code fail if the file doesn't exist. If that happens, something went wrong.

I added a new test to cover this.

kenoir (approved):

PR #2875, opened 2025-04-16
Merge merger router

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5987
The router currently takes the output of the merger, retrieves the works from the merged index and 
- if they don't have a `CollectionPath`, transition them to Denormalised and loads them into the denormalised index
- if they have a `CollectionPath` and a Sierra identifier, send them to the path concatenator service
- if they have a `CollectionPath` and an identifier other than Sierra, send them to the relation_embedder via the batcher

This is moving the routing upstream in the merger service:
Once the works have been merged and loaded into the merged index, if they don't have a `CollectionPath`, transition them to Denormalised, loads them into the denormalised index 

The workRouter then 
- sends Denormalised works to the ingestor
- sends Merged works with a Sierra identifier to the path concatenator service
- sends Merged works with an identifier other than Sierra to the relation_embedder via the batcher

EDIT:
Upon reflexion, the merged index is not strictly needed. Here's what the flow looks like with the latest commit:

Screenshot 2025-04-29 at 14 04 30

## How to test

Next step: 
- create a new pipeline and run a full reindex
- compare contents of the current/production denormalised index with that of the new denormalised index. They should be identical.

The Denormalised index contains all works

Silent failure 😣 where services don't error but the behaviour and results are changed in some hard-to-spot way


pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala
```diff
@@ -448,27 +449,27 @@ class MergerIntegrationTest
         processWorks(digitalSierra, physicalSierra, teiWork)
 
         Then("the Sierra works are redirected to the tei")
-        context.getMerged(digitalSierra) should beRedirectedTo(teiWork)
-        context.getMerged(physicalSierra) should beRedirectedTo(teiWork)
+        Right(context.getDenormalised(digitalSierra)) should beRedirectedTo(teiWork)
+        Right(context.getDenormalised(physicalSierra)) should beRedirectedTo(teiWork)
```
agnesgaroux, line 453: I'm double checking the logic here, I think IRL the "main" (not internal) TEI work would have a CollectionPath, which change things a bit: the TEI work would come out of the merger as a Work[Merged]

agnesgaroux: See [here](https://github.com/wellcomecollection/catalogue-pipeline/blob/26f16ef09f53d0942bb9bf8e9d8bc086f13299c5/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/TeiData.scala#L127)
I believe the test omitted adding the collectionPath to the top level TEI work because it was irrelevant to the case it was testing for.
In reality, the parts of a TEI manuscript are nested inside the top level one (see [here](https://github.com/wellcomecollection/wellcome-collection-tei/blob/main/Arabic/MS_Arabic_7.xml) and [here](https://github.com/wellcomecollection/wellcome-collection-tei/blob/main/Greek/MS_354.xml) for instance)
When it's transformed into a work, the TEI id is used as the root of the tree for the top level work

kenoir:
Closes https://github.com/wellcomecollection/platform/issues/5987

pipeline/matcher_merger/merger/src/main/scala/weco/pipeline/merger/services/MergerWorkerService.scala
```diff
@@ -25,8 +29,10 @@ class MergerWorkerService[WorkDestination, ImageDestination](
   msgStream: SQSStream[NotificationMessage],
   sourceWorkLookup: IdentifiedWorkLookup,
   mergerManager: MergerManager,
-  workOrImageIndexer: Indexer[Either[Work[Merged], Image[Initial]]],
-  workMsgSender: MessageSender[WorkDestination],
+  workOrImageIndexer: Indexer[
+    Either[Either[Work[Merged], Work[Denormalised]], Image[Initial]]
```
kenoir, line 33: A mythical EitherEither, rarely spotted in the wild 👹 .

Can we express this without the nested Either?

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
```diff
@@ -5,7 +5,8 @@ import org.scalatest.EitherValues
 import org.scalatest.matchers.{MatchResult, Matcher}
 import weco.catalogue.internal_model.identifiers.IdState
 import weco.catalogue.internal_model.image.ImageData
-import weco.catalogue.internal_model.work.WorkState.{Identified, Merged}
+import weco.catalogue.internal_model.work.WorkState.{Denormalised, Identified, Merged}
+//import weco.catalogue.internal_model.work.WorkState.{Identified, Merged}
```
kenoir, line 9: Can this be removed?

kenoir:
It would be good to have the PR description filled out a bit more - although it is linked to explanatory tickets, a bit more detail or even duplication is fine.

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala
kenoir: > I'm double checking the logic here, I think IRL the "main" (not internal) TEI work would have a CollectionPath, which changes things a bit: the TEI work would come out of the merger as a Work[Merged]

Is this a change to fix a separate issue?

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
agnesgaroux: Yes of course

pipeline/matcher_merger/merger/src/main/scala/weco/pipeline/merger/services/MergerWorkerService.scala
kenoir: I've had a go at refactoring the `EitherIndexer` which underlies this issue and I can't find a way that is straightforward, without making a bigger change, so I think it makes sense to leave this for a refactor following the release of the merged merger/router 

I do think it could do with a refactor as the logic is quite convoluted, but lets not put it in the way of releasing this change.

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
```diff
@@ -234,15 +269,34 @@ trait IntegrationTestHelpers
...
+          w.isInstanceOf[Work.Visible[Denormalised]],
+          s"${w.id} is not visible",
+          s"${w.id} is visible"
+        )
+      }
+
+    }
   }
 
+//
```
kenoir, line 290: I think this can be removed.

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
```diff
@@ -234,15 +269,34 @@ trait IntegrationTestHelpers
...
+//  def beVisible = new Matcher[Work[Merged]] {
+//    override def apply(left: Work[Merged]): MatchResult =
+//      MatchResult(
+//        left.isInstanceOf[Work.Visible[Merged]],
+//        s"${left.id} is not visible",
+//        s"${left.id} is visible"
+//      )
+//  }
```
kenoir, line 298: I think this can be removed.
```suggestion
```

pipeline/terraform/modules/stack/release_uris.tf
```diff
@@ -7,17 +7,13 @@ locals {
   repo_urls = [for repo_url in data.aws_ecr_repository.service.*.repository_url : "${repo_url}:env.${var.release_label}"]
   image_ids = zipmap(local.services, local.repo_urls)
```
kenoir, line 9: Why are the non-router images removed here? I think they are still needed, the lambdas use the container images they contain.

kenoir (commented):
~I think if we are removing the router image and service, we will also need to update the build scripts~, this will also mean that when we merge the existing router in the prod pipeline won't get updated if we make changes (this is probably acceptable for a short period before we switch to the new pipeline).

edit: I can see the router is removed in pipeline.yml 👍

kenoir (approved):

PR #2877, opened 2025-04-24
Multiple concept indexes within a stack

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5995

Allow a single pipeline stack to contain multiple concepts indexes, each with its own mappings/analysis.

## How to test

Run `terraform plan` and examine the output. Does it look reasonable?

After the terraform is applied, the existing concepts index should stay intact a second index should be created with the same mappings.

There are two risks associated with this change:
* The production concepts index being removed/modified — the Terraform `moved` block prevents this from happening. With this in place, `terraform plan` output shows that no changes will be made to the production index.
* The production concepts API losing access to the ES index due to its API key being destroyed — to prevent this from happening, this PR does not make any changes to the API key used by the concepts API. Instead, it introduces a new API key with the updated permissions as part of a multistage process to prevent downtime.


pipeline/terraform/modules/stack/elastic_pipeline.tf
```diff
@@ -132,66 +132,65 @@ module "pipeline_indices" {
 }
 
 locals {
-  indices = module.pipeline_indices.index_names
```
StepanBrychta, line 135: We don't need to retrieve the index names from the `pipeline_indices` module. They are defined locally and then inputted as variables into the `pipeline_indices` module, so we can just use the local names, which is more straightforward.

pipeline/terraform/modules/stack/elastic_pipeline.tf
```diff
@@ -233,7 +232,7 @@ resource "elasticstack_elasticsearch_security_role" "read_only" {
   name = "read_only"
 
   indices {
-    names      = [for idx in module.pipeline_indices.indices : idx.name]
+    names      = ["*"]
```
StepanBrychta, line 235: Instead of looping through all index names in the cluster (which gets more difficult with multiple concepts indexes), we can give the role read-only access to everything.

pipeline/terraform/2025-03-06/main.tf
```diff
@@ -20,7 +20,11 @@ module "pipeline" {
       works_analysis = "works_indexed.2024-11-06"
     }
     concepts = {
-      indexed = "concepts_indexed.2025-03-10"
+      # Define a set of concept indexes, each with its own config definition
+      indexed = {
+        "2025-03-06" = "concepts_indexed.2025-03-10"
+        "2025-04-24" = "concepts_indexed.2025-03-10"
```
StepanBrychta, line 26: This preserves the existing concepts index (2025-03-06) and also creates a new one (with the same mappings).

pipeline/terraform/modules/stack/elastic_pipeline.tf
```diff
@@ -132,66 +132,69 @@ module "pipeline_indices" {
...
     concepts_api = {
-      read  = [local.indices.concepts_indexed]
+      read  = ["${local.es_concepts_index_prefix}-2025-03-06"]
+      write = []
+    }
+    concepts_api_new = {
+      read  = ["${local.es_concepts_index_prefix}*"]
       write = []
     }
   }
```
StepanBrychta, line 200: To perform this update without any concepts API downtime, I don't think we can make changes to the existing API key. (Elasticsearch API keys are not mutable.)

However, we can use a slightly convoluted approach to change permissions without any downtime:
1. Creating a new API key with the updated permissions (`concepts_api_new`)
2. Switching the concepts API to the new key
3. Updating the permissions on the old key
4. Switching the concepts API back to the old key
5. Removing the new key

pipeline/terraform/2025-03-06/main.tf
```diff
@@ -33,3 +37,12 @@ module "pipeline" {
     aws.catalogue = aws.catalogue
   }
 }
+
+# To prevent the 2025-03-06 concepts index from being destroyed and recreated, we need to explicitly tell Terraform
+# it has moved to a different address.
+# This bit won't be necessary in any newly created pipeline stacks.
+moved {
```
kenoir, line 44: 🥇

pipeline/terraform/modules/stack/elastic_pipeline.tf
kenoir: So is the idea to perform this change in parts with multiple `apply`s?

pipeline/terraform/modules/stack/elastic_pipeline.tf
kenoir: Can this be `images*`, `concepts*`, `works*` so it's not a blanket read all? I think it's probably safe either way, but probably better to err on the side of limiting permissions if we can do it easily.

pipeline/terraform/modules/stack/elastic_pipeline.tf
StepanBrychta: Yes, there would be three `apply`'s in total with two concepts API deploys in between.

The steps don't need to be done in quick succession. For the purposes of switching to the new concepts index, only steps 1 and 2 are necessary. The remaining steps only involve 'cleaning up' and can be done later.

PR #2878, opened 2025-04-24
Choose concepts index from ingestor #5995

StepanBrychta:
## What does this change?

Follow-up on https://github.com/wellcomecollection/catalogue-pipeline/pull/2877, which added support for multiple concepts indexes within a cluster. This PR allows us to choose which concepts index to use when running the concepts ingestor.

To make sure there is separation between the different indexes, the folder structure storing ingestor files has changed from `ingestor/<PIPELINE_DATE>/JOB_ID` to `ingestor/<PIPELINE_DATE>/<INDEX_DATE>/JOB_ID`.

This PR also includes some unit test refactoring to make future changes easier.

## How to test

This change can be tested by running the ingestor locally against the new concepts index (2025-04-24).

> [!CAUTION]
> When testing, make sure not to write to the production index!

The concepts ingestor successfully writes to the chosen index.

There should be minimal risk associated with this change.


catalogue_graph/terraform/state_machine_ingestor.tf
```diff
@@ -31,6 +31,7 @@ resource "aws_sfn_state_machine" "catalogue_graph_ingestor" {
           FunctionName = module.ingestor_trigger_lambda.lambda.arn,
           Payload = {
             pipeline_date = local.pipeline_date
+            index_date    = local.concepts_index_date
```
StepanBrychta, line 34: I'm hardcoding the `index_date` as payload to the state machine to follow what we do with the `pipeline_date`. In the future, I think that it would be good to pass these as arguments from the daily/monthly run schedule instead for better configurability.

catalogue_graph/src/utils/safety.py
```diff
@@ -0,0 +1,24 @@
+def validate_fractional_change(
+        modified_size: int, total_size: int, fractional_threshold: float, force_pass: bool = False
+) -> None:
```
StepanBrychta, line 3: We now have 4 different Lambdas which implement a similar diff-based safety check, so I decided to make it into a general utility function.

catalogue_graph/src/graph_remover.py
```diff
@@ -132,13 +133,12 @@ def handler(
...
         )
 
-    if (
-        len(previous_ids) > 0
-        and len(deleted_ids) / len(previous_ids) > ACCEPTABLE_DIFF_THRESHOLD
-        and not disable_safety_check
-    ):
-        raise ValueError(
-            f"Attempted to remove {len(deleted_ids)} items (out of a total of {len(previous_ids)}), which is above the safety threshold."
+    if len(previous_ids) > 0:
```
kenoir, line 136: Can we push this check inside the `validate_fractional_change` function?

catalogue_graph/terraform/state_machine_ingestor.tf
kenoir: So is the outcome of this at present that we can only update one index in one pipeline at a time? Also that future refactoring in the scheduler is required to accommodate "simultaneous (ish)" updates to multiple indexes.

Presuming the intention is to deploy this and switch the catalogue-api to the new index close together so that the production index doesn't get out of date? 

Should we also create issues for (a) deleting the old index after release and (b) updating the scheduler as described above?

kenoir (approved):
This looks like it'll work :) Worth noting that the pipeline date & index date here increases the potential for confusion, but I don't think there is too much we can do about that while we are taking differing approaches for concepts & works/images, perhaps being clear in documentation is the best we can do for now?

A couple of question / potential improvements, not blocking for merge.

catalogue_graph/terraform/state_machine_ingestor.tf
StepanBrychta: > So is the outcome of this at present that we can only update one index in one pipeline at a time? 

Yes, we can only update one index at a time from the production state machine. (However, we now have the option to choose other indexes when running locally.) 

> Also that future refactoring in the scheduler is required to accommodate "simultaneous (ish)" updates to multiple indexes.

Yes. If we want to update multiple indexes simultaneously, we could also look into introducing a versioning mechanism so that we can run multiple versions of the pipeline in production (e.g. an older version updating a production index and a newer version updating a new dev index). 

> Presuming the intention is to deploy this and switch the catalogue-api to the new index close together so that the production index doesn't get out of date?

We could manually trigger a daily run right after merging this and the other PRs. If everything goes well, we could switch to the new index after the run completes. We might also have to do a monthly run to apply this PR: https://github.com/wellcomecollection/catalogue-pipeline/pull/2868. 

> Should we also create issues for (a) deleting the old index after release and (b) updating the scheduler as described above?

I'm happy to create those tickets.

PR #2881, opened 2025-05-01
Turn off listen to reindexer in 2025-03-06

kenoir:
> [!Note]
> The terraform in this PR has been applied.

## What does this change?

Before we create the new pipeline in https://github.com/wellcomecollection/catalogue-pipeline/pull/2875, we should turn off listening to the reindexer in the prod pipeline 

## How to test

- [ ] Apply the terraform, does it make the necessary updates?

We are able to reindex in the new pipeline without it impacting the old one.

The risk should be minimal (this is turning _off_ the reindexer).


kenoir:
Closing as change has gone in with merge to another branch.

PR #2892, opened 2025-05-06
Capture all concept types in graph #5946

StepanBrychta:
## What does this change?

* Implements https://github.com/wellcomecollection/platform/issues/5946 but takes a different approach than the one described on the ticket. Instead of storing concept types as labels (which would involve relying on an implementation detail of the Neptune bulk loader and making some queries more complex), this PR modifies the catalogue works extractor to store concept types as properties of `HAS_CONCEPT` edges. This ensures that if different works categorise a single concept under different types (e.g. `Genre` and `Concept`), all types are captured in the graph.
* Modifies the `ingestor_loader` Lambda to extract all concept types from `HAS_CONCEPT` edges and choose the most specific one to index to Elasticsearch (e.g. `Person` over `Agent`, `Genre` over `Concept` etc.).
* Adds a new data validation check which logs all concepts with inconsistent types (e.g. 'Person' and 'Place') into a CSV file in S3. Rather than running this check as part of the pipeline, it exists in its own `data_quality_checks.py` file to be run locally on an ad-hoc basis. We can add more checks to this file in the future.

## How to test

* Changes can be tested by running the modified extractor against an experimental Neptune cluster created for this purpose (`catalogue-graph-clone`).
* Unit tests were modified to reflect these changes.

With this change, theme pages should always display the most specific concept type (if more types are available for a given concept). This should fix an issue where the `Using this Type/Technique` tab is hidden on some theme pages. See https://wellcome.slack.com/archives/C8X9YKM5X/p1744889391879429.

Not storing concept types directly on `Concept` nodes in the graph makes determining all types of a concept more computationally expensive, since we now have to retrieve all `HAS_CONCEPT` edges leading to each `Concept` node. However, this is not a problem in practice — queries are not significantly slower than before.


catalogue_graph/src/data_validation/concept_types.py
```diff
@@ -0,0 +1,59 @@
...
+    count: int = response[0]["count"]
+    return count
+
+
+def get_concepts_with_inconsistent_types() -> Generator[dict]:
+    """Return all concepts whose combination of types is not consistent."""
+    client = get_neptune_client(True)
+
+    start_offset = 0
+    while start_offset < _get_concepts_count():
```
paul-butcher, line 47: Should the result of `_get_concepts_count` be stored as a variable outside the loop, or is it expected to keep updating with each iteration?  If the latter, it could do with a comment, because this is the kind of thing someone might come along and "fix"

catalogue_graph/src/utils/aws.py
```diff
@@ -85,6 +85,19 @@ def get_csv_from_s3(s3_uri: str) -> Generator[Any]:
         yield from csv_reader
 
 
+def write_csv_to_s3(s3_uri: str, items: list[dict]) -> None:
+    if len(items) == 0:
```
paul-butcher, line 89:
```suggestion
    if not items:
```
is the more Pythonic idiom.

StepanBrychta:
@paul-butcher I added some tests covering concept type precedence. I only added tests for the combinations where I think precedence is relevant. For example, it doesn't matter whether `Meeting` has higher precedence than `Period`, so even though the code arbitrarily chooses `Period` as having higher precedence, I did not add tests for this combination.

paul-butcher (approved):

catalogue_graph/src/models/catalogue_concept.py
```diff
@@ -101,6 +103,35 @@ def transform_related_concepts(
...
+    """If a concept is classified under more than one type, pick the most specific one and return it."""
+
+    # Prioritise concepts, with more specific ones (e.g. 'Person') above less specific ones (e.g. 'Agent').
+    # Sometimes a concept is classified under types which are mutually exclusive. For example, there are
+    # several hundred concepts categorised as both a 'Person' and an 'Organisation'. These inconsistencies
+    # arise upstream, and we cannot easily resolve them here. To mitigate this issue, the priority list below
+    # is ordered to maximise the probability of choosing the right type based on an analysis of current inconsistencies.
+    # (For example, when a concept is categorised as both an 'Organisation' and a 'Place', the 'Place' type is almost
+    # always the correct one, which is why 'Place' is higher in the priority list than 'Organisation').
+    concept_types_by_priority: list[ConceptType] = [
```
kenoir, line 116: This is a nice pragmatic way to solve this problem.

PR #2897, opened 2025-05-07
unpack nested either

agnesgaroux:
## What does this change?

This [change](https://github.com/wellcomecollection/catalogue-pipeline/pull/2875) introduced a nested Either flow which on ~reflexion~ reflection was unnecessary: the solution is to only have one index and not worry about having 2 work states getting in their respective, corresponding indices. (Both states shares the same properties)  

This also removes analysis and mappings for the merged index in index_config, as well as updating the index tests to reflect the fact that the merged index is no more

## How to test

Deploy to the new (non-live) pipeline before we switch the catalogue API to use it 

Things should work exactly as before

The denormalised indices on both pipelines will be compared before the API is switched to 2025-05-01


pipeline/matcher_merger/merger/src/main/scala/weco/pipeline/merger/Main.scala
```diff
@@ -87,18 +83,10 @@ object Main extends WellcomeTypesafeApp {
...
-          leftIndexer = new EitherIndexer[Work[Merged], Work[Denormalised]](
-            leftIndexer = new ElasticIndexer[Work[Merged]](
-              client = esClient,
-              index = Index(config.requireString("es.denormalised-works.index"))
-            ),
-            rightIndexer = new ElasticIndexer[Work[Denormalised]](
-              client = esClient,
-              index = Index(config.requireString("es.denormalised-works.index"))
-            )
+        new EitherIndexer[Work[Merged], Image[Initial]](
```
kenoir, line 86: 👏

kenoir (approved):

PR #2900, opened 2025-05-13
Document and refactor cypher queries #6001

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6001

* Add more comments to cypher queries. 
* Make improvements to the 'referenced together' query.

## How to test

This can be tested locally against a local ES index.

Queries should be easier to read and modify.

Risks are minimal. The 'referenced together' query tweaks will result in changes in the ES index, but those will only be visible behind a toggle.

### DIFF
--- a/catalogue_graph/src/ingestor_loader.py
+++ b/catalogue_graph/src/ingestor_loader.py
@@ -7,8 +7,6 @@ import typing
 import boto3
 import polars as pl
 import smart_open
-from pydantic import BaseModel
-
 from config import INGESTOR_S3_BUCKET, INGESTOR_S3_PREFIX
 from ingestor_indexer import IngestorIndexerLambdaEvent, IngestorIndexerObject
 from models.catalogue_concept import (
@@ -17,6 +15,7 @@ from models.catalogue_concept import (
     ConceptsQuerySingleResult,
 )
 from models.graph_node import ConceptType
+from pydantic import BaseModel
 from utils.aws import get_neptune_client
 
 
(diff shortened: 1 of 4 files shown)


catalogue_graph/src/ingestor_loader.py
```diff
@@ -36,6 +36,9 @@ class IngestorLoaderConfig(BaseModel):
 # Maximum number of related nodes to return for each relationship type
 RELATED_TO_LIMIT = 10
 
+# Minimum number of works in which two concepts must co-occur to be considered 'frequently referenced together'
+NUMBER_OF_SHARED_WORKS_THRESHOLD = 2
```
StepanBrychta, line 40: In the original `REFERENCED_TOGETHER_QUERY` query, this threshold was hardcoded to 10, so this might seem like a significant reduction. However, note that the original query did not properly deduplicate before counting works, which caused each work to be counted multiple times.

catalogue_graph/src/ingestor_loader.py
```diff
@@ -114,28 +135,69 @@ def get_related_query(
 
 # For every Wellcome concept, this query returns a list of concepts most frequently co‑occurring with it in Works.
 REFERENCED_TOGETHER_QUERY = """
```
StepanBrychta, line 137: There were some issues with the original query:
1. It wasn't correctly deduplicating works when counting them.
2. It wasn't considering 'same as' concepts when determining matches, causing [this page](https://wellcomecollection.org/concepts/txuwkbec) and [this page](https://wellcomecollection.org/concepts/eq7aud2q) to list different 'frequently referenced together' concepts, even though they represent the same concept.

This refactored query addresses both issues.

catalogue_graph/src/ingestor_loader.py
```diff
@@ -66,38 +61,148 @@ def get_related_query(
...
+             }})[0..$related_to_limit] AS related
+             
+        /* Return the ID of each concept and a corresponding list of related concepts. */
         RETURN 
             concept.id AS id,
             related
     """
 
 
+def get_referenced_together_query(
```
StepanBrychta, line 123: There were some issues with the original query:
1. It wasn't correctly deduplicating works when counting them.
2. It wasn't considering 'same as' concepts when determining matches, causing [this page](https://wellcomecollection.org/concepts/txuwkbec) and [this page](https://wellcomecollection.org/concepts/eq7aud2q) to list different 'frequently referenced together' concepts, even though they represent the same concept.

This refactored query addresses both issues (and also adds an optional concept type filter in preparation for https://github.com/wellcomecollection/platform/issues/6001).

catalogue_graph/src/ingestor_loader.py
```diff
@@ -36,6 +36,9 @@ class IngestorLoaderConfig(BaseModel):
 # Maximum number of related nodes to return for each relationship type
 RELATED_TO_LIMIT = 10
 
+# Minimum number of works in which two concepts must co-occur to be considered 'frequently referenced together'
+NUMBER_OF_SHARED_WORKS_THRESHOLD = 3
```
StepanBrychta, line 40: In the original `REFERENCED_TOGETHER_QUERY` query, this threshold was hardcoded to 10, so this might seem like a significant reduction. However, note that the original query did not properly deduplicate before counting works, which caused each work to be counted multiple times. 

In practice, the new threshold combined with proper deduplication somewhat *reduces* the number of returned concepts (so it's effectively a threshold increase).

kenoir (approved):
The documented queries are a great improvement.

PR #2902, opened 2025-05-14
force inferrer images onto amd64

paul-butcher:


kenoir:
Closing as GitHub runs are already (still?) x86_64.

PR #2903, opened 2025-05-14
Remove ECR login where broken and not needed

kenoir:
## What does this change?

The reference to logging in to ECR here is failing consistently in CI, but images are published anyway as buildkite runners are already logged into ECR, see [this hook](https://github.com/wellcomecollection/buildkite-infrastructure/blob/054c53fa48fc1fb26115849f80c081695432eaa3/buildkite_agent_hook.sh).

This removes the unnecessary login.

## How to test

- [ ] Deploy this change, ensure images are still published but no login error is visible.

Less confusing errors in build logs.

The ECR login command is currently failing, so should not impact deployment.

### DIFF
--- a/builds/publish_inferrer_image_to_ecr.sh
+++ b/builds/publish_inferrer_image_to_ecr.sh
@@ -29,8 +29,7 @@ else
 fi
 
 echo "*** Publishing Docker image to ECR"
-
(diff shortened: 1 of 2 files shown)


paul-butcher (approved):
Deleted code is debugged code!

PR #2905, opened 2025-05-15
Update reindex scripts

kenoir:
## What does this change?

This change removes the merged index which no longer exists, and updates the `./pipeline_inject_messages.py` script to work much faster.

## How to test

Verify the `./get_reindex_status.py` and `./pipeline_inject_messages.py` scripts behave as expected.

Only reference indices that exist and faster fixing of borked pipeline.

Risks should be minimal, this is used during reindexes only and changes are minor.

### DIFF
--- a/reindexer/scripts/pipeline_inject_messages.py
+++ b/reindexer/scripts/pipeline_inject_messages.py
@@ -7,16 +7,46 @@ See the section of the README "fixing leaks" for more info.
 """
 import click
 from tqdm import tqdm
+from uuid import uuid4
 from get_reindex_status import get_session_with_role
+from concurrent.futures import ThreadPoolExecutor
 
+
+def build_sns_batch(messages):
+    for m in messages:
+        yield {
+            "Id": str(uuid4()),
+            "Message": m,
+            "MessageStructure": "string"
+        }
+
+def publish_batch_to_sns(sns, topic_arn, messages):
+    batch_messages = build_sns_batch(messages)
+    response = sns.publish_batch(
+        TopicArn=topic_arn,
+        PublishBatchRequestEntries=list(batch_messages)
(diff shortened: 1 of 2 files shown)


reindexer/scripts/pipeline_inject_messages.py
```diff
@@ -7,16 +7,52 @@
...
-    for id in tqdm(ids):
-        sns.publish(
-            TopicArn=f"arn:aws:sns:eu-west-1:760097843905:catalogue-{reindex_date}_{destination_name}",
-            Message=id,
-        )
+
+    topic_arn = f"arn:aws:sns:eu-west-1:760097843905:catalogue-{reindex_date}_{destination_name}"
+
+    chunk_size = 10000
+    chunks = [ids[i : i + chunk_size] for i in range(0, len(ids), chunk_size)]
```
StepanBrychta, line 49: The [itertools.batched](https://docs.python.org/3.12/library/itertools.html#itertools.batched) utility would work here I think (Python 3.12+ only).

reindexer/scripts/pipeline_inject_messages.py
```diff
@@ -7,16 +7,52 @@
...
+    )
+    return response
+
+
+def publish_to_sns_threaded(sns, topic_arn, doc_ids, sns_batch_size=10):
+    with ThreadPoolExecutor(max_workers=10) as executor:
+        batched_messages = [
+            doc_ids[i : i + sns_batch_size]
+            for i in range(0, len(doc_ids), sns_batch_size)
+        ]
```
StepanBrychta, line 33: Same here

StepanBrychta (approved):
Looks good! Type annotations would be nice.

PR #2909, opened 2025-05-19
Increase concepts extractor concurrency

StepanBrychta:
## What does this change?

Increase the concurrency of the concepts extractor S3 load step to speed up the pipeline. In practice, this results in a 2.5x speed-up.


kenoir (approved):
👌

PR #2913, opened 2025-05-22
Lambda matcher router

paul-butcher:
## What does this change?

This introduces the Lambda version of the Matcher. (https://github.com/wellcomecollection/platform/issues/5988)

(terraform to follow)

## How to test

Run the lambda locally from the root of the repository:

1. sign in to AWS SSO, this will be using your account to connect to the matcher database.

2. Run it: `./scripts/run_local.sh matcher_merger/matcher`

3. Send an id to it: `echo cbdpxcxj | python scripts/post_to_rie.py`

Look at the logging from the lambda, it will have emitted ids for each record involved in the merge.  You can also send multiple ids to it (newline separated), and see each of them work as expected.

Using lambdas should make this pipeline step a bit simpler to manage.

When we deploy this, it will be to a new pipeline, so any snags that may occur can be resolved there before going live.


kenoir (approved):
This looks good, and works for me. 

I think the original ticket was to lambdify the new merger/router, and this moves the matcher, but it's matters not very much. 

It'd be good to do the merger/router (let's just call it the merger?) next though.

 I think this can't be merged yet without corresponding infra changes to ensure the SQS version continues to function.

PR #2914, opened 2025-05-23
Matcher lambda deployment

paul-butcher:
## What does this change?

This replaces the ECS version of the Matcher with the Lambda version.
(https://github.com/wellcomecollection/platform/issues/5988)

## How to test

Deploy this to a new pipeline, see if it works.

deploying other a new pipeline minimises the risk.


kenoir (approved):
Looks good, but see the note on https://github.com/wellcomecollection/catalogue-pipeline/pull/2913#pullrequestreview-2864342376 about merge order. 

I think we might be able to deploy to the current pipeline without worrying about dual running or testing in a new pipeline, but instead use the prod pipeline as a test - if it fails we can always roll things back.

PR #2918, opened 2025-05-28
fix downstream message type

agnesgaroux:
## What does this change?

See https://wellcome.slack.com/archives/C02ANCYL90E/p1748357694285859

## How to test

The code change has been deployed manually to unblock the pipeline. This updates the tests to match the type published downstream

Tests pass!

N/A


pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/MatcherSQSLambda.scala
```diff
@@ -40,14 +41,26 @@ trait MatcherSQSLambda[Config <: ApplicationConfig]
...
-        identifiers.foreach(downstream.notify)
-        findMissingMessages(messagesMap, identifiers)
+        // flatten sets of ids in MatcherResult into a single set of Strings.
+        val initialIdentifiers =
+          results.flatMap(_.allUnderlyingIdentifiers).toSet
+
+        results.foldLeft(initialIdentifiers) {
+          (identifiers, result) =>
+            downstream.notify(result)(MatcherResult.encoder) match {
+              case Success(_) => identifiers
```
kenoir, line 55: We should have a test for this behaviour.

pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/Main.scala
kenoir, line 1: It might be better to put removing the main and tests in a different PR as it's a separate change.

pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/MatcherSQSLambda.scala
```diff
@@ -40,14 +41,26 @@ trait MatcherSQSLambda[Config <: ApplicationConfig]
...
+    // 1. notify downstream with all the MatcherResults.
     // 2. filter out any successful ids from the messagesMap and return the bad messageIds.
     resultsFuture.map {
       results: Iterable[MatcherResult] =>
-        // as a prerequisite for both, flatten sets of ids in MatcherResult into a single set of Strings.
-        val identifiers = results.flatMap(_.allUnderlyingIdentifiers).toSet
-        identifiers.foreach(downstream.notify)
-        findMissingMessages(messagesMap, identifiers)
+        // flatten sets of ids in MatcherResult into a single set of Strings.
+        val initialIdentifiers =
```
kenoir, line 49: Not applying to this change set directly, but the extraction of message id and re-mapping of succeeded and failed messages by their contents is adding complexity here that we may be able to abstract.

pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/MatcherSQSLambda.scala
```diff
@@ -40,14 +41,26 @@ trait MatcherSQSLambda[Config <: ApplicationConfig]
...
+          results.flatMap(_.allUnderlyingIdentifiers).toSet
+
+        results.foldLeft(initialIdentifiers) {
+          (identifiers, result) =>
+            downstream.notify(result)(MatcherResult.encoder) match {
+              case Success(_) => identifiers
+              // remove from initialIdentifiers list any identifier that was not successfully sent downstream
+              case Failure(_) =>
```
kenoir, line 57: I don't think this behaves as expected, as the result of this operation is never returned.

PR #2919, opened 2025-05-28
Add neptune scaler

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6032

* Add a new `catalogue-graph-scaler` state machine, which sets the capacity of the Neptune cluster and polls the state of the cluster until the new capacity is applied.
* Modify the `catalogue-graph-ingestor` state machine to use `catalogue-graph-scaler`, scaling up the cluster before extracting data, and scaling it back down once done.

## How to test

The Terraform changes have been applied and the corresponding Lambda functions have ben deployed, so the new `catalogue-graph-scaler` state machine can be tested via the AWS console.

The main goal of this change is to decrease the runtime of the `catalogue-graph-ingestor` state machine without making it more error-prone.

We were able to decrease runtime in the past by increasing the concurrency of the 'Map load to S3' step, but this made the state machine fail more frequently due to Neptune not being able to scale up quickly enough. To work around this, we increase the minimum NCUs of the cluster before running the 'Map load to S3' step, ensuring the cluster is scaled up to handle multiple parallel requests.

Keeping the cluster scaled up would be expensive, so this also adds some error handling to make sure it always gets scaled down at the end of each run, even if the state machine fails.


catalogue_graph/terraform/state_machine_ingestor.tf
```diff
@@ -136,6 +185,27 @@ resource "aws_sfn_state_machine" "catalogue_graph_ingestor" {
       },
       Success = {
         Type = "Succeed"
+      },
+      # Any steps between 'Scale up cluster' and 'Scale down cluster' can fail. If this happens, we need
```
kenoir, line 189: 👌

kenoir (approved):

PR #2921, opened 2025-06-03
Concepts pipeline reports

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6036
- the graph loader has been set to [ignore missing edges if they're under a certain threshold
](https://github.com/wellcomecollection/catalogue-pipeline/commit/6e23b1810710019a521a9da6ef3ed86508748961) instead of failing the step. This posts a message to Slack (triggered from the individual bulk-loader lambdas) with the number of ignored nodes/edges if that number is below the failure threshold
- A reporter lambda is added to the end of the pipeline: it will post a report to Slack [#wc-search-alerts], see below
- the reporter lambda also posts a report about the deetions: both from the graph and the index
- the changes to the other lambdas revolve around compiling data into reports saved to S3, so that the reporter can pick it up at the end
Screenshot 2025-06-09 at 10 25 50

Screenshot 2025-06-09 at 10 23 35
Screenshot 2025-06-09 at 10 23 53

## How to test

Use the local handlers while logged in to AWS 
EDIT: can't run it locally yet: the pipeline needs to run to generate build the reports 

We get weekly and daily reports on concepts indexing and graph insert errors

Nothing we can't live with :)


kenoir (approved):
This looks good, but I think you could optionally combine the removal into the final report to simplify the summary?

PR #2923, opened 2025-06-10
make lambda name consistent with the others

agnesgaroux:
## What does this change?

Following https://github.com/wellcomecollection/catalogue-pipeline/pull/2921 the deploy uncovered an inconsistency in lambda naming
This fixes it
Applied

## How to test

N/A

All landas are deployed successfully and named consistently

none


kenoir (approved):
_-_-_

PR #2929, opened 2025-06-11
Lifespan events

paul-butcher:
## What does this change?

Changes the event-based startup/shutdown behaviour in the inferrers to use a lifespan context manager instead.

resolves https://github.com/wellcomecollection/catalogue-pipeline/issues/2925

## How to test

Run the inferrers, nothing has changed

We will be able to upgrade FastAPI whenever we want to.

It should just work. If it does not, we can roll it back.


kenoir:
With reference to https://fastapi.tiangolo.com/advanced/events/#lifespan-events, this looks good to me.

PR #2931, opened 2025-06-12
write indexer report in pipeline/index/job prefix

agnesgaroux:
## What does this change?

- write all reports in the job_id prefix as well as the root of the ingestor
- make sure reports are consistent, with pipeline_date, index_date and job_id
- report on graph deletions since last index_remover run, not just for today
- in order to be able to write the index_remover report in the job_id prefix, passes the ReportEvent to the "Remove documents" step of the state machine

Reports get published, easier to debug

### DIFF
--- a/catalogue_graph/src/index_remover.py
+++ b/catalogue_graph/src/index_remover.py
@@ -148,13 +148,12 @@ def handler(
 def lambda_handler(events: list[ReporterEvent], context: typing.Any) -> None:
     # we only want properties pertaining to the overall pipeline run, any event will do
     validated_event = [ReporterEvent.model_validate(event) for event in events][0]
-    override_safety_check = validated_event.force_pass if validated_event.force_pass is not None else False
     
     handler(
         validated_event.pipeline_date, 
         validated_event.index_date, 
         validated_event.job_id,
-        override_safety_check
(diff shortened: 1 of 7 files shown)


catalogue_graph/src/index_remover.py
```diff
@@ -134,15 +133,38 @@ def handler(
...
 
-def lambda_handler(event: dict, context: typing.Any) -> None:
-    pipeline_date = event["pipeline_date"]
-    index_date = event["index_date"]
-    override_safety_check = event.get("override_safety_check", False)
+def lambda_handler(events: list[ReporterEvent], context: typing.Any) -> None:
+    # we only want properties pertaining to the overall pipeline run, any event will do
+    validated_event = [ReporterEvent.model_validate(event) for event in events][0]
+    override_safety_check = (
+        validated_event.force_pass if validated_event.force_pass is not None else False
```
paul-butcher, line 159: I don't think this is necessary.  If you've defined force_pass, then it will be what has been defined.  If not, it will be None, which is false.

As handler is expecting a bool, you could do `bool(validated_event.force_pass)` to cast it.

agnesgaroux: Makes sense 👍

PR #2934, opened 2025-06-19
Initial reporting refactor, simplify and DRY

kenoir:
## What does this change?

Fix the issues with reporting failing when it can't find a previous run, various other refactors to simplify preceding larger change to add indexer monitor.

## How to test

Working reporter, more easy to parse reporting flow.

This should be thoroughly tested to ensure it behaves as expected.


catalogue_graph/src/utils/reporting.py
```diff
@@ -0,0 +1,129 @@
...
+
+class PipelineReport(BaseModel):
+    _report_type: ReportType
+    pipeline_date: str
+    index_date: str
+    job_id: str | None
+
+    @classmethod
+    def read(
+        cls,
```
paul-butcher, line 49: You can make this look a bit more like Scala Generics using TypeVar.  This will tell mypy that the return type matches `cls`
```
from typing import Type, TypeVar

T = TypeVar('T', bound='PipelineReport')


class PipelineReport(BaseModel):
    _report_type: ReportType
    pipeline_date: str
    index_date: str
    job_id: str | None

    @classmethod
    def read(
            cls: Type[T],
```

(obviously, putting the import with all the others)

paul-butcher (approved):

PR #2936, opened 2025-06-24
Downstream test fixture

agnesgaroux:
## What does this change?

Firebreak
Improve tests around pipeline lambdas.
This generalises the use of MemorySNSDownstream in id_minter, matcher, batcher and relation_embedder tests

## How to test

Run the test! :) 

Less duplication of test fixtures, less test code to write 

This doesn't affect the application logic

### DIFF
--- a/common/lambda/src/main/scala/weco/lambda/Downstream.scala
+++ b/common/lambda/src/main/scala/weco/lambda/Downstream.scala
@@ -16,11 +16,23 @@ trait Downstream {
   def notify[T](batch: T)(implicit encoder: Encoder[T]): Try[Unit]
 }
 
+object Downstream {
+  def apply(downstreamTarget: DownstreamTarget): Downstream = {
+    downstreamTarget match {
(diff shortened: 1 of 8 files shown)


common/lambda/src/main/scala/weco/lambda/Downstream.scala
```diff
@@ -39,24 +51,17 @@ sealed trait DownstreamTarget
...
 
-  def buildDownstreamTarget(config: Config): DownstreamTarget = {
+  def buildDownstreamTarget(
+    config: Config,
+    namespace: String = ""
+  ): DownstreamTarget = {
     config.getStringOption("downstream.target") match {
       case Some("sns") =>
-        val snsConfig = buildSNSConfig(config)
+        val snsConfig = buildSNSConfig(config, namespace)
```
agnesgaroux, line 64: Added this to account for the merger using 4 different Downstreams

kenoir (approved):
Neat 👍

PR #2937, opened 2025-06-24
Test common lambda behaviour in a common trait 

paul-butcher:
## What does this change?

Firebreak activity.

Moves the meat of the testing of lambda partial failure behaviour from individual applications into the base weco.lambda library.

This is not a direct test of SQSLambdaApp.  Each application will still need to inherit and implement its own tests based on this shared behaviour, as each one will have different conditions that lead to failure.

The point of this is to provide a common way to write tests that check that if a full/partial failure/success occurs within the application, then the external manifestation - in the form of messages marked as failed, and successful downstream messages posted - is correct.

Matcher and Id minter both now use the new Behaviour functions

## How to test

This is all about test refactoring.  As long as the same tests keep working, this is good.  The real test is whether it is now more maintainable and readable.

This should make it easier to move ECS Apps to Lambda, as there will now be a bunch of tests that are mostly already implemented for you.

Another aspect of this is that I want to make the tests tell a bit more of an easily readable story.  Currently (in general across the whole suite), there's a bit too much (admittedly necessary) guff in the tests - setup, dealing with Promises etc, making it a bit harder to see what the point is.  It's also all quite Scala-specific.  If I can start to change tests so that the body of the tests themselves are very clear about their purpose, and the framework becomes the thing that does all the work, then I hope it would be easier to port to a different language if we eventually do so.

This is only a change to test code, so it will not have a direct impact on applications in production.  If this format turns out to be too opinionated for some of the applications, then we do not have to use it, but it can still be used as inspiration to write more narrative tests.

### DIFF
--- a/common/lambda/src/test/scala/weco/lambda/behaviours/LambdaBehaviours.scala
+++ b/common/lambda/src/test/scala/weco/lambda/behaviours/LambdaBehaviours.scala
@@ -9,6 +9,11 @@ import weco.lambda._
 import weco.lambda.helpers.MemoryDownstream
 import weco.lambda.matchers.LambdaResultMatchers
 
+/** Implements tests for the three common scenarios in a Lambda invocation
+  *   1. Everything works 2. Nothing works 3. Some of it works, some of it
+  *      doesn't.
+  */
+
 trait LambdaBehaviours[
   InputMessageType,
   Config <: ApplicationConfig,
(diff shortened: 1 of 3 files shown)


kenoir (approved):
I think this is a good approach, it takes a little getting used to I think, but happy to merge this.

PR #2939, opened 2025-06-25
refactor Batcher lambda

agnesgaroux:
## What does this change?

Firebreak housekeeping
Refactor the Batcher to implement  `SQSBatchResponseLambdaApp`
Move notify downstream outside of the PathProcessor. 
PathProcessor returns a `Seq[String]` (collectionPath) in case of failures or a `Seq[Batch]` in case of success, to be sent downstream
Remove BatcherWorkerService

TO DO
Refactor the BatcherWorkerServiceTest to
- check that the PathProcessor generates the expected Batch, or sends the root downstream if too many selectors were found
- handles BatcherResponse failures 

## How to test

Tests refactor coming up!

More consistency, less dead code 

The logic of the pipeline step is unaffected. It might fail spectacularly but it will work as previously if it doesn't


pipeline/relation_embedder/batcher/src/test/scala/weco/pipeline/batcher/BatcherWorkerServiceTest.scala
```diff
@@ -1,188 +1,112 @@
...
+//    with Matchers
+//    with SQS
+//    with Pekko
+//    with Eventually
+//    with IntegrationPatience
+//    with MemoryDownstream {
+//
+//  import Selector._
+//
+//  it("doesn't delete paths where selectors failed sending ") {
```
agnesgaroux, line 33: This one to use Paul's LambdaBehaviours `aTotalSuccess` and `aFailingInvocation`

PR #2953, opened 2025-07-16
Concepts ingestor reporter refactor

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6055
It might be easiest to look at individual commits 
- 1a83af6369f1c30df7dc8e035206eb2c1d81ea06 : tf changes. Adds the indexer_monitor to the state machine. The rest is consolidating all the lambdas in the aforementioned state machine into lambda_catalogue_ingestor.tf. And arranging them in the order they get invoked, because it sparks joy
- b4049a0c26083c71ebc39f28a4dd1cd3a3138de7 : existing monitors don't need to mess with the indexer report, it has its own monitor now
- 08b564d15260fe407a823a3e007c3f9f07f69207: indexer returns a `IngestorIndexerMonitorLambdaEvent`, ingestor_indexer_monitor.py writes the indexer.report.json (with previous job id so we can make comparisons in the reporter), ingestor_report.py reads all the reports and posts to Slack
- 34698c0514dbdd80e0baf6633fe7169f99419512: updates tests and add fixtures
- f1aa5565ee143b8dc11f92fcb4950d1fc3016417: miscellaneous
- 7a7191814636dde7a0db4961a3f67609d39b7076: rename index_remover to ingestor_deletions

## How to test

The ingestor_reporter can be run locally, just make sure that you're invoking it with arguments that match what's in the ingestor bucket

Report gets posted to Slack daily M-T
Better separation of concern in the ingestor pipeline 

This doesn't touch the ingestor logic. Worse case scenario we don't get a report or the state machine fails a non-crucial step


catalogue_graph/src/ingestor_deletions.py
```diff
@@ -82,23 +82,29 @@ def get_ids_to_delete(pipeline_date: str | None, index_date: str | None) -> set[
...
-    # throw a RuntimeError here, to make it clear that the state of the index remover is unexpected.
-    if index_remover_report is None:
-        raise RuntimeError(
-            f"Unexpected error: IndexRemoverReport for pipeline_date={pipeline_date} and index_date={index_date} not found."
-        )
+    # We shouldn't get here as DeletionReport.read should throw its own error,
+    # throw a RuntimeError here, to make it clear that the state of the ingestor deletions is unexpected.
+    if ingestor_deletion_report is None:
+        # TO DO: remove this when we have a report.deletions.json report
+        s3_url = f"s3://{config.INGESTOR_S3_BUCKET}/{config.INGESTOR_S3_PREFIX}/{pipeline_date}/{index_date}/report.index_remover.json"
```
agnesgaroux, line 99: On the 1st run after this is merged, the existing report will be "report.index_remover.json", then on l.143 it will be written as "report.deletions.json". At this point we'll be able to remove this

catalogue_graph/src/ingestor_indexer_monitor.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def handler(events: list[IngestorIndexerMonitorLambdaEvent]) -> None:
+    print("Preparing concepts pipeline reports ...")
+
+    try:
+        build_indexer_report(events)
+    except ValueError as e:
+        print(f"Report failed: {e}")
+        raise e
```
StepanBrychta, line 44: I know this code is not new and instead reuses an existing pattern we use in other places. However, I think that the try/except block can be removed and replaced with just:

```python
build_indexer_report(events)
```

I think that intercepting the exception only to log it and raise it again makes the code verbose without adding much value — the exception, including the full traceback, will be automatically logged anyway.

agnesgaroux: Do you think it's superfluous in the trigger_monitor and the loader_?monitor as well?

catalogue_graph/src/ingestor_reporter.py
```diff
@@ -116,27 +134,21 @@ def get_indexer_report(
...
-    validated_events = [ReporterEvent.model_validate(event) for event in events]
+def lambda_handler(event: IngestorStepEvent, context: Any) -> None:
+    validated_event = IngestorStepEvent.model_validate(event)
     config = ReporterConfig()
 
-    handler(validated_events, config)
+    handler(validated_event, config)
 
 
 def local_handler() -> None:
```
StepanBrychta, line 154: I think that the `--success-count` argument in the local handler is no longer needed.

PR #2954, opened 2025-07-16
Extract catalogue graph data from the denormalised index

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6086

Change catalogue work/concept extractors to extract data from the denormalised index instead of the works snapshot. The resulting bulk load files should be exactly the same as before. 

> [!WARNING]  
> Terraform changes must be applied before merging.

## How to test

Run the catalogue_works and catalogue_concepts extractors locally and compare bulk load outputs with production bulk load outputs.

The catalogue graph stores exactly the same works and concepts as before but sourced from the denormalised index.

The main risk is a mismatch between current bulk load files (populated from the works snapshot) and new bulk load files (populated from the denormalised index). 

To mitigate this, I populated a set of bulk load files both from the denormalised index and from the works snapshot and wrote the following Python script to compare them:

```python
import csv

def _get_row_map(file_name: str):
    row_map = {}
    with open(file_name, newline='') as f:
        reader = csv.DictReader(f)

        for row in reader:
            row_map[row[':ID']] = row

    return row_map


def _check_ids(denormalised_map: dict, snapshot_map: dict):
    denormalised_ids = set(denormalised_map.keys())
    snapshot_ids = set(snapshot_map.keys())

    print(f"Missing IDs in snapshot: {len(denormalised_ids.difference(snapshot_ids))}")
    print(f"Missing IDs in denormalised: {len(snapshot_ids.difference(denormalised_ids))}")


def compare_denormalised_and_snapshot(transformer_type: str, entity_type: str):
    print(f"Checking {transformer_type}, {entity_type}.")
    snapshot_map = _get_row_map(f"{BASE_BULK_LOAD_FILES_PATH}/{transformer_type}__{entity_type}_snapshot.csv")
    denormalised_map = _get_row_map(f"{BASE_BULK_LOAD_FILES_PATH}/{transformer_type}__{entity_type}_denormalised.csv")
    _check_ids(denormalised_map, snapshot_map)

    mismatched_rows = 0
    for id in denormalised_map:
        if id not in snapshot_map:
            continue
        if denormalised_map[id] != snapshot_map[id]:
            mismatched_rows += 1

    
    print(f"Mismatched rows: {mismatched_rows}\n")
        
compare_denormalised_and_snapshot("catalogue_works", "nodes")
compare_denormalised_and_snapshot("catalogue_concepts", "nodes")
compare_denormalised_and_snapshot("catalogue_works", "edges")
compare_denormalised_and_snapshot("catalogue_concepts", "edges")
```

The script produced the following output:
```
Checking catalogue_works, nodes.
Missing IDs in snapshot: 49
Missing IDs in denormalised: 7
Mismatched rows: 1

Checking catalogue_concepts, nodes.
Missing IDs in snapshot: 16
Missing IDs in denormalised: 0
Mismatched rows: 13867

Checking catalogue_works, edges.
Missing IDs in snapshot: 77
Missing IDs in denormalised: 4
Mismatched rows: 0

Checking catalogue_concepts, edges.
Missing IDs in snapshot: 153
Missing IDs in denormalised: 163
Mismatched rows: 0
```

There are some small discrepancies between node/edge IDs (some IDs which exist in the current bulk load files are missing from the new ones and vice versa). This is expected since the denormalised index and the works snapshot are not perfectly in sync (the works snapshot can be up to 24 hours behind). In all 4 bulk load files, these discrepancies affect <0.01% of nodes/edges.

A relatively large number of concept node rows (2.5%) are mismatched (i.e. corresponding bulk load file rows have the same IDs but different content). This is also expected, since concepts with the same IDs can sometimes have different labels on different works. For example, a given concept can have the label `MEDICINE` on work A, and `Medicine` on work B, and the extractor always uses the first label it encounters. Therefore, streaming works in a different order will sometimes yield different concept labels.


catalogue_graph/src/extractor.py
StepanBrychta, line 1: Since data is now extracted from the denormalised index, we need to introduce the `pipeline_date` parameter, determining which denormalised index to use.

catalogue_graph/src/ingestor_deletions.py
```diff
@@ -23,7 +23,7 @@ def get_current_id_count(
     pipeline_date: str | None, index_date: str | None, is_local: bool
 ) -> int:
     """Return the number of documents currently stored in ES in the concepts index."""
-    es = utils.elasticsearch.get_client(pipeline_date, is_local)
+    es = utils.elasticsearch.get_client("concept_ingestor", pipeline_date, is_local)
```
StepanBrychta, line 26: Since we now connect to Elasticsearch from different services (extractor and ingestor), we need to pass a new `api_key_name` argument so that we can scope permissions for each service separately.

catalogue_graph/src/transformers/catalogue/concepts_transformer.py
```diff
@@ -38,6 +45,11 @@ def transform_node(self, raw_data: tuple[dict, WorkConceptKey]) -> Concept | Non
     def extract_edges(
         self, raw_data: tuple[dict, WorkConceptKey]
     ) -> Generator[ConceptHasSourceConcept]:
+        if self.id_label_checker is None:
```
kenoir, line 48: Why is `IdLabelChecker` only required in `extract_edges` now?

StepanBrychta: It was always only required when extracting edges, but we were unnecessarily creating it even when extracting nodes. I moved it for optimisation purposes.

catalogue_graph/src/transformers/catalogue/raw_work.py
```diff
@@ -16,35 +16,35 @@ class WorkConcept(TypedDict):
...
-        assert isinstance(label, str)
+        label: str = self.work_data.get("title", "")
         return label
 
     @property
     def type(self) -> WorkType:
-        concept_type: WorkType = self.raw_work["type"]
-        return concept_type
+        raw_work_type = self.work_data["workType"]
+        work_type: WorkType = "Work" if raw_work_type == "Standard" else raw_work_type
```
kenoir, line 35: Perhaps a comment about what this is? I think this denotes if it's a `Series` or a `Standard` work,  but the collisions of the many kinds of "type" possible here makes it a bit confusing.

catalogue_graph/src/sources/catalogue/concepts_source.py
```diff
@@ -14,27 +14,36 @@ def extract_concepts_from_work(
     # extracting these component concepts, since the frontend does not make use of them and the resulting
     # theme pages would be empty.
     for subject in raw_work.get("subjects", []):
+        subject["type"] = "Subject"
```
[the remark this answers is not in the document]
StepanBrychta: I agree, it's not a good solution. I'll come up with a better one, possibly as part of the `RawCatalogueConcept` class.

catalogue_graph/src/transformers/catalogue/concepts_transformer.py
```diff
@@ -1,21 +1,28 @@
...
+from .works_transformer import ES_FIELDS, ES_QUERY
 
 
 class CatalogueConceptsTransformer(BaseTransformer):
-    def __init__(self, url: str):
-        self.source = CatalogueConceptsSource(url)
-        self.id_label_checker = IdLabelChecker(
-            node_types=["concepts", "locations", "names"], sources=["loc", "mesh"]
+    def __init__(self, pipeline_date: str | None, is_local: bool):
+        index_name = get_standard_index_name(
```
kenoir, line 18: I wonder if we could move `get_standard_index_name` inside `ElasticsearchSource` so that `pipeline_date` doesn't need to be referenced here and also when handed to descendants of ``ElasticsearchSource`?

catalogue_graph/terraform/locals.tf
```diff
@@ -19,7 +19,7 @@ locals {
 
   # This is a hint that the ingestors might need to be in the pipeline stack!
```
kenoir, line 20: Not for addressing in this PR here, but I wonder if the above comment is becoming more true with the changes were making here.

PR #2955, opened 2025-07-17
Common python build

paul-butcher:
## What does this change?

This removes the common GitHub actions that were initially written for the UV-based inferrers, as they are now implemented in the .github repository.

## How to test

The inferrer workflows should work as expected.

The actions were written to be reusable across our Python projects, using the inferrers as a model . Success will be evident when we use them in other projects.

The intent of this work is to harmonise between various Python projects.  If it turns out that there is too much necessary difference, then we can cross that bridge when we come to it.


.github/workflows/inferrer-python-ci.yml
```diff
@@ -29,7 +34,12 @@ jobs:
         project: [ aspect_ratio_inferrer, feature_inferrer, palette_inferrer ]
     steps:
       - uses: actions/checkout@v3
-      - uses: ./.github/actions/docker-build-and-push
+      - uses: actions/checkout@v3
```
agnesgaroux, line 37: Should it be something like 
`uses: wellcomecollection/.github/.github/actions/docker-build-and-push/action.yml@main`
(@main once the PR in .github is merged)

.github/workflows/inferrer-python-ci.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         project: [ aspect_ratio_inferrer, feature_inferrer, palette_inferrer ]
     steps:
       - uses: actions/checkout@v3
-      - uses: ./.github/actions/docker-build-and-push
+      - uses: wellcomecollection/.github/actions/docker-build-and-push@main
```
agnesgaroux, line 32: Double `.github`! One for the repo, one for the folder

agnesgaroux: Oh wait you tried that

.github/workflows/inferrer-python-ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         project: [ aspect_ratio_inferrer, feature_inferrer, palette_inferrer ]
     steps:
       - uses: actions/checkout@v3
-      - uses: ./.github/actions/python_check
+      - uses: wellcomecollection/.github/actions/python-check@main
```
agnesgaroux, line 17: Ha! it's python_check
`wellcomecollection/.github/.github/actions/python_check@main`

agnesgaroux: Sorry I'm weirdly invested in this

PR #2965, opened 2025-07-23
Extractor bug fixes

StepanBrychta:

### DIFF
--- a/catalogue_graph/src/extractor.py
+++ b/catalogue_graph/src/extractor.py
@@ -104,9 +104,15 @@ def local_handler() -> None:
         type=int,
         help="How many entities to stream. If not specified, streaming will continue until the source is exhausted.",
     )
+    parser.add_argument(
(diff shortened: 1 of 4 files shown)


catalogue_graph/src/extractor.py
```diff
@@ -104,9 +104,10 @@ def local_handler() -> None:
         type=int,
         help="How many entities to stream. If not specified, streaming will continue until the source is exhausted.",
     )
+    parser.add_argument("--is-local", type=bool, required=False, default=False)
     args = parser.parse_args()
 
-    handler(**args.__dict__, is_local=True)
+    handler(**args.__dict__)
```
StepanBrychta, line 110: We were unintentionally running the extractor with the `is_local` flag set to `True`. This did not matter before but it does now since we want to connect to the private Elasticsearch endpoint.

PR #2966, opened 2025-07-23
Refactor concepts ingestor

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6091

Refactor the concepts ingestor service:
* Move logic for transforming raw Neptune data into Elasticsearch documents to the `elasticsearch_transformers` folder. 
* Remove the intermediate `CatalogueConcept` pydantic model. Transform raw Neptune data directly into the `IndexableConcept` model.

Implement a few other minor tweaks:
* Remove the `referencedTogether` field from the concepts index. This field has been superseded by other fields (`relatedTopics` and `frequentCollaborators`) and is no longer used in the frontend.
* Tweak the `frequentCollaborators` query to also include `Organisations` (so that we can show a list of frequent collaborators on organisation theme pages).

## How to test

Run the concepts ingestor locally. The resulting Elasticsearch index should be the same as the production index, with two differences:
* The `referencedTogether` field should no longer be stored in the index
* Some organisation concepts should now store a non-empty `frequentCollaborators` field

We should be able to reuse the patterns introduced in this PR when creating the new works ingestor.

Risks are minimal.

### DIFF
--- a/catalogue_graph/src/models/ingestor/concept.py
+++ b/catalogue_graph/src/models/ingestor/concept.py
@@ -1,4 +1,3 @@
-
 from models.graph_node import ConceptType
 from models.ingestor.display.identifier import get_display_identifier
 from models.ingestor.indexable import DisplayIdentifier


catalogue_graph/src/elasticsearch_transformers/concepts_transformer.py
```diff
@@ -0,0 +1,48 @@
...
+
+from .raw_neptune_concept import RawNeptuneConcept
+from .raw_neptune_related_concepts import RawNeptuneRelatedConcepts
+
+
+class ElasticsearchConceptsTransformer:
+    def transform_document(
+        self, concept: dict, related_concepts: dict
+    ) -> IndexableConcept:
+        neptune_concept = RawNeptuneConcept(concept, related_concepts)
```
kenoir, line 16: I think it would be better to push the responsibility for ensuring `neptune_concept` and `neptune_related` are of the correct type to the calling function.

PR #2975, opened 2025-07-29
Update ECR lifecycle policy to protect deployment tags

kenoir:
## What does this change?

This updates the ECR lifecycle policy for the catalogue_graph_extractor repository to ensure that images tagged with `dev`, `latest` or `prod`  are never automatically expired, while still maintaining cleanup of other images.

This change works with https://github.com/wellcomecollection/catalogue-pipeline/pull/2968, which increases the number of images that will be pushed by allowing PRs to push build artefacts.

### Changes made:
- **Priority 1**: Protect  tagged images (highest priority, only expire if >1 image)
- **Priority 2**: Protect  tagged images (only expire if >1 image) 
- **Priority 3**: Protect  tagged images (only expire if >1 image)
- **Priority 4**: Expire other tagged images (keep last 50)
- **Priority 5**: Expire untagged images (keep last 5, lowest priority)

Since each tag can only exist on one image at a time, setting  to 1 effectively prevents these deployment images from ever being expired.

## How to test

Deploy this change and verify that:
- Images tagged with `dev`, `latest` or `prod`  are never expired
- Other tagged images are cleaned up after 50 images
- Untagged images are cleaned up after 5 images

- ECR repository maintains deployment images indefinitely
- Repository size is still managed through cleanup of non-deployment images
- No unexpected deletion of critical deployment artifacts

- **Risk**: Protected tags could accumulate if somehow multiple images get the same tag
- **Mitigation**: The AWS ECR tagging system prevents this - only one image can have a specific tag at a time
- **Risk**: Repository could grow large with many non-deployment tagged images
- **Mitigation**: Other tagged images are limited to 50, and untagged images to 5, providing adequate cleanup

### DIFF
--- a/catalogue_graph/terraform/ecr.tf
+++ b/catalogue_graph/terraform/ecr.tf
@@ -52,9 +52,10 @@ resource "aws_ecr_lifecycle_policy" "expire_old_images" {
         rulePriority = 4
         description  = "Expire other tagged images, keep only the last 50"
         selection = {
-          tagStatus   = "tagged"
-          countType   = "imageCountMoreThan"
-          countNumber = 50
+          tagStatus      = "tagged"
+          tagPatternList = ["*"]
+          countType      = "imageCountMoreThan"
+          countNumber    = 50
         }
         action = {
           type = "expire"


kenoir:
Merging as this is now required after https://github.com/wellcomecollection/catalogue-pipeline/pull/2968 is merged.

PR #2979, opened 2025-08-05
Add works ingestor #6092

StepanBrychta:
## What does this change?

Depends on https://github.com/wellcomecollection/catalogue-pipeline/pull/2984

Implements https://github.com/wellcomecollection/platform/issues/6092

Adds a new works ingestor loader service, which creates final work documents, using the catalogue graph to replace the functionality of the relation embedder. 

At the moment, only visible works are processed. Other work types (e.g. deleted or redirected) will be handled in a separate PR.

## How to test

Run the works ingestor locally via:

```
AWS_PROFILE=platform-developer uv run src/ingestor/run_local.py --ingestor-type=works
```

Alternatively, run the loader service directly via
```
AWS_PROFILE=platform-developer python3.13 -m ingestor.steps.ingestor_loader --ingestor-type=works --limit=1
```

The final work documents should be the same as those produced by the current Scala-based ingestor, with a few minor differences (missing `succeededBy` and `precededBy` fields).

Risks are minimal, none of these changes affect production services.

### DIFF
--- a/catalogue_graph/src/ingestor/models/indexable_concept.py
+++ b/catalogue_graph/src/ingestor/models/indexable_concept.py
@@ -6,8 +6,8 @@ from utils.types import ConceptType
 
 class ConceptDescription(BaseModel):
     text: str
-    sourceLabel: str
-    sourceUrl: str
+    sourceLabel: str | None
+    sourceUrl: str | None
 
 
 class ConceptRelatedTo(BaseModel):
--- a/catalogue_graph/src/ingestor/transformers/concept_override.py
+++ b/catalogue_graph/src/ingestor/transformers/concept_override.py
@@ -1,0 +1,51 @@
+import csv
+import os
+from typing import TextIO
+
+from ingestor.models.indexable_concept import ConceptDescription
+
+from .raw_concept import RawNeptuneConcept
+from .raw_related_concepts import (
+    RawNeptuneRelatedConcept,
+)
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+
+
+class ConceptTextOverrideProvider:
+    overrides: dict[str, dict[str, str]] = {}
+
+    def __init__(self, overrides_csv: TextIO | None):
+        if overrides_csv:
+            self._load_overrides(overrides_csv)
+        else:
+            with open(
+                os.path.join(HERE, "label_description_overrides.csv")
+            ) as csv_file:
+                self._load_overrides(csv_file)
+
+    def _load_overrides(self, overrides: TextIO) -> None:
+        csv_reader = csv.DictReader(overrides)
+        self.overrides = {row["id"].strip(): row for row in csv_reader}
+
+    def display_label_of(
+        self, raw_concept: RawNeptuneConcept | RawNeptuneRelatedConcept
+    ) -> str:
+        override = self.overrides.get(raw_concept.wellcome_id)
+        if override and (override_label := override["label"].strip()):
+            return override_label
+        return raw_concept.display_label
+
+    def description_of(
+        self, raw_concept: RawNeptuneConcept
+    ) -> ConceptDescription | None:
+        override = self.overrides.get(raw_concept.wellcome_id)
+        if override:
+            override_description = override["description"].strip()
+            if override_description.lower() == "empty":
+                return None
+            if override_description:
+                return ConceptDescription(
+                    text=override_description, sourceUrl=None, sourceLabel=None
+                )
+        return raw_concept.description
--- a/catalogue_graph/src/ingestor/transformers/concepts_transformer.py
+++ b/catalogue_graph/src/ingestor/transformers/concepts_transformer.py
@@ -1,20 +1,58 @@
+from typing import TextIO
+
 from ingestor.extractors.concepts_extractor import GraphConceptsExtractor
 from ingestor.models.indexable_concept import (
     ConceptDisplay,
     ConceptQuery,
+    ConceptRelatedTo,
     IndexableConcept,
     RelatedConcepts,
 )
+from ingestor.transformers.concept_override import ConceptTextOverrideProvider
 from ingestor.transformers.raw_concept import RawNeptuneConcept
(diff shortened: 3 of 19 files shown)


catalogue_graph/src/ingestor/extractors/works_extractor.py
```diff
@@ -0,0 +1,97 @@
...
+
+        start_time = time.time()
+        result = es_client.mget(index=index_name, body={"ids": work_ids})
+        duration = round(time.time() - start_time)
+
+        work_mapping = {}
+        for work in result["docs"]:
+            work_id = work["_id"]
+            if not work["found"]:
+                print(f"Work {work_id} does not exist in the denormalised index.")
```
paul-butcher, line 41: Not necessarily for now, but I think it would be good to use a  logger, so we can use log.warning here and log.info for the print on 46. (and the handler "received event" comments)

catalogue_graph/src/ingestor/extractors/works_extractor.py
```diff
@@ -0,0 +1,97 @@
...
+    work: DenormalisedWork
+    hierarchy: WorkHierarchy
+    concepts: list[WorkConcept]
+
+
+class GraphWorksExtractor(GraphBaseExtractor):
+    def get_es_works(self, work_ids: list[str]) -> dict:
+        es_client = get_client("graph_extractor", "2025-05-01", True)
+        index_name = get_standard_index_name(
+            config.ES_DENORMALISED_INDEX_NAME, "2025-05-01"
```
paul-butcher, line 30: Can 2025-05-01 be derived from the environment? Or at least, as a member of the class, rather than hard-coded twice in this function

catalogue_graph/src/ingestor/extractors/works_extractor.py
```diff
@@ -0,0 +1,97 @@
...
+        duration = round(time.time() - start_time)
+
+        work_mapping = {}
+        for work in result["docs"]:
+            work_id = work["_id"]
+            if not work["found"]:
+                print(f"Work {work_id} does not exist in the denormalised index.")
+                continue
+
+            work_mapping[work_id] = work["_source"]
```
paul-butcher, line 44: I think this will fail if found is false.  A missing document has no _source.

catalogue_graph/src/ingestor/extractors/works_extractor.py
```diff
@@ -0,0 +1,97 @@
...
+        self,
+    ) -> Generator[ExtractedWork]:
+        work_ids = list(self._get_work_ids())
+        all_ancestors = self._get_work_ancestors()
+        all_children = self._get_work_children()
+        all_concepts = self._get_work_concepts()
+        all_es_works = self.get_es_works(work_ids)
+
+        for work_id in work_ids:
+            es_work = all_es_works[work_id]
```
paul-butcher, line 81: Currently, if not all works are present in ES, this cannot be reached (line 44).  However, if that is fixed, it may just push the problem further down to here.

catalogue_graph/src/ingestor/transformers/raw_concept.py
```diff
@@ -170,9 +172,15 @@ def identifiers(self) -> list[ConceptIdentifier]:
...
+                value=identifier.value,
+                identifier_type=Id(id=identifier.identifierType),
+                ontology_type="Concept",
+            )
             display_ids.append(
-                get_display_identifier(identifier.value, identifier.identifierType)
+                DisplayIdentifier.from_source_identifier(source_identifier)
             )
 
         return display_ids
```
paul-butcher, line 186: I think this could be a comprehension:
```suggestion
return [
        DisplayIdentifier.from_source_identifier(
            source_identifier = SourceIdentifier(
                value=identifier.value,
                identifier_type=Id(id=identifier.identifierType),
                ontology_type="Concept",
            )
        ) for identifier in self.identifiers
    ]
```

catalogue_graph/src/ingestor/extractors/works_extractor.py
StepanBrychta: There is a `continue` keyword on line 42, and so works for which `found=False` should be skipped.

catalogue_graph/src/ingestor/extractors/works_extractor.py
StepanBrychta: Well spotted. I changed the code to skip works which don't exist in the denormalised index and also added a comment explaining why we cover this scenario.

catalogue_graph/src/ingestor/extractors/works_extractor.py
StepanBrychta: I agree. At the moment, we use `print` statements across the codebase, and so I think that switching to a logger should probably be done as part of a separate ticket covering the whole codebase.

catalogue_graph/src/ingestor/extractors/works_extractor.py
```diff
@@ -0,0 +1,110 @@
...
+        self.pipeline_date = pipeline_date
+
+    def get_es_works(self, work_ids: list[str]) -> dict:
+        es_client = get_client("graph_extractor", self.pipeline_date, True)
+        index_name = get_standard_index_name(
+            config.ES_DENORMALISED_INDEX_NAME, self.pipeline_date
+        )
+
+        start_time = time.time()
+        result = es_client.mget(index=index_name, body={"ids": work_ids})
```
kenoir, line 40: Have you seen how long this takes for a chunk of 10k work ids?

catalogue_graph/src/ingestor/models/denormalised/work.py
```diff
@@ -0,0 +1,76 @@
...
+from ingestor.models.shared.serialisable import ElasticsearchModel
+from utils.types import WorkType
+
+
+class CollectionPath(BaseModel):
+    path: str
+    label: str | None = None
+
+
+class DenormalisedWorkData(ElasticsearchModel):
```
kenoir, line 26: Noting that there will eventually be some overlap between the work model being created here, and the necessary post-transform work model from the EBSCO adapter (and future adapter) steps.

PR #2980, opened 2025-08-07
Ensure new EBSCO adapter loader runs in lambda

kenoir:
## What does this change?

Closes https://github.com/wellcomecollection/platform/issues/6107

**TODO:** 

- Currently the operation fails in AWS due to a permissions issue, this needs resolving.

  ```
  "AccessDeniedException: Insufficient Lake Formation permission(s): Required Create Database on Catalog"
  ```

  The lambda can be run locally with: 

  ```
  AWS_PROFILE=platform-developer uv run src/steps/loader.py --use-glue-table s3://wellcomecollection-platform-ebsco-adapter/prod/ftp/ebz-s7451719-20250729-1.xml
  ```

- The loader also needs to place a flag for a given date in S3 on completion, and check when picking up a file that it hasn't already been processed.

## How to test

- [ ] Run the state machine in AWS, does it succeed?

The loader step can now run properly in AWS without issue, enabling the rest of the EBSCO adapter work to continue.

This is not yet in production use, risks should be minimal.

### DIFF
--- a/ebsco_adapter/terraform/iceberg_adapter/iam_policies.tf
+++ b/ebsco_adapter/terraform/iceberg_adapter/iam_policies.tf
@@ -24,7 +24,7 @@ data "aws_iam_policy_document" "iceberg_write" {
     ]
   }
 
-    statement {
+  statement {
     actions = [
       "glue:CreateTable",
       "glue:GetTable",


agnesgaroux:
Outstanding comments will be addressed in https://github.com/wellcomecollection/platform/issues/6119

PR #2982, opened 2025-08-11
ebsco adapter trigger

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6050

Ebsco adapter trigger lambda, as the 1st step in the ebsco adapter state machine:
- triggered by an eventbridge schedule, the `time` of the event being used to generate the job_id for the state machine execution (this will be uncommented when we move to the loader)
- we list files on the ftp server that match the expected pattern
- we sort them by most recent date in the file name 
- we check whether we already have this file in S3
-- if we don't, we download it from the ebsco server and upload it to S3, then return the S3 uri for the most recent file
-- if we do, we return the S3 uri for the most recent file
- added lifecycle policy to the bucket, so that files are deleted after 6 months

## How to test

Go to `s3://wellcomecollection-platform-ebsco-adapter/dev/ftp_v2/` and ensure there are no file there for August 2025 (202508xx in the filename). Delete if any

While logged in with aws, run
```
AWS_PROFILE=platform-developer uv run src/steps/trigger.py --local
```
You should see the downloaded xml file in the `dev/ftp_v2` prefix
The location sent downstream will be printed to the console: "Sending S3 location downstream: <s3_uri>" 

If you run the lambda again, you should see "File xxx already exists in S3. No need to download." with the same s3_uri printed before the process exits

Ebsco xml files get uploaded to the correct location in s3
The existing adapter continues to function (using `prod/ftp` prefix`)

There is no schedule in place for this lambda yet


ebsco_adapter/ebsco_adapter_iceberg/src/steps/trigger.py
```diff
@@ -1,60 +1,167 @@
...
     is_local: bool = False
 
 
-class EbscoAdapterTriggerEvent(BaseModel):
-    job_id: str | None = None
+class EventBridgeTriggerEvent(BaseModel):
+    time: str
+
+
+def get_most_recent_valid_file(filenames: list[str]) -> str | None:
```
kenoir, line 29: Suggest this is a bit simpler: https://github.com/wellcomecollection/catalogue-pipeline/commit/147ca4681a2eee6b13abb2b15932441098784677

That change will fail if it encounters a badly formed date, but arguably that is a stopping error (rather than a silent ignore) as it's an unexpected thing to have happen.

kenoir (approved):
A couple of nitpicks, looks good.

PR #2984, opened 2025-08-12
Refactor concepts ingestor

StepanBrychta:
## What does this change?

Final ingestor refactoring in preparation for https://github.com/wellcomecollection/platform/issues/6092:
* Move logic responsible for extracting catalogue graph data into a separate GraphConceptsExtractor class. This class yields concepts one by one to be processed by the transformer. (Works will be extracted using a separate class, with both classes subclassing from `GraphBaseExtractor`.)
* Add a new `ingestor_type` argument to all ingestor steps. (Once we start extracting works, the value of this argument will determine whether we ingest works or concepts.)
* Make testing the loader service easier, adding support for outputting documents into local files in Parquet or JSONL formats.
* Consolidate all step event models into one file to avoid repetition.
* Make all fields (pipeline_date, index_date, job_id) in the `IngestorStepEvent` model (and all of its subclasses) required. This results in simpler logic which doesn't require repeatedly checking if the fields are defined. When running locally, the values of these fields will always default to `dev` (currently they default to a combination of `None` and `dev` depending on the step service).

## How to test

Run the ingestor locally via 
```
AWS_PROFILE=platform-developer uv run src/ingestor/run_local.py --ingestor-type=concepts
```

* Adding support for loading to local JSONL files should make testing the loader service easier.
* Creating dedicated extractor classes should make it easier to add support for extracting works and promote code reuse.

While this PR should not result in any changes to indexed concepts, a new concepts index will be created (2025-08-13) to remove the risk of making unintended changes.

### DIFF
--- a/catalogue_graph/tests/test_ingestor_trigger_monitor.py
+++ b/catalogue_graph/tests/test_ingestor_trigger_monitor.py
@@ -51,6 +51,7 @@ def verify_s3_reports(record_count: int) -> None:
 
 
 def test_ingestor_trigger_monitor_success_no_previous() -> None:
+    loader_event = get_mock_ingestor_loader_event("123", 0, 1)
     event = IngestorTriggerMonitorLambdaEvent(
         ingestor_type="concepts",
         pipeline_date="2025-01-01",
@@ -58,13 +59,7 @@ def test_ingestor_trigger_monitor_success_no_previous() -> None:
         job_id="123",
         force_pass=False,
         report_results=True,
-        events=[
-            get_mock_ingestor_loader_event(
-                "123",
-                0,
-                1,
-            )
-        ],
+        events=[loader_event],
     )
 
     config = IngestorTriggerMonitorConfig(percentage_threshold=0.1)


catalogue_graph/src/ingestor/models/display/id_label.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import BaseModel
```
agnesgaroux, line 1: Not specifically about this file, but would it make sense to have the concepts models in `concepts/display/` or `display/concepts/` alongside `display/works`?

catalogue_graph/terraform/locals.tf
```diff
@@ -19,7 +19,7 @@ locals {
 
   # This is a hint that the ingestors might need to be in the pipeline stack!
   pipeline_date       = "2025-05-01"
-  concepts_index_date = "2025-07-21"
+  concepts_index_date = "2025-08-13"
```
agnesgaroux, line 22: That's going to cause a new concept index to be created correct? While the API still uses "2025-07-21"?
On a side note what are we doing with older concepts indices?

kenoir (approved):
To summarise what's going on:

Ingestors have `Loaders` and `Indexers`
- An ingestor is the top level abstraction, taking data from a source, transforming it into another data model and loading it to S3

`Loaders` have `Transformers` and load configuration
- A loader is responsible for extracting from a data source, transforming it to a target data model and loading it to a location for later indexing according to its configuration

`Transformers` have `Extractors` and transformations
- An extractor takes data from a source and hands it off for transformation
- A transformation converts from a “raw” data model to a target “display” data model

This looks like a good step towards abstracting things so we can do this also for works 👍

catalogue_graph/src/ingestor/transformers/base_transformer.py
```diff
@@ -0,0 +1,84 @@
...
+            "Each Elasticsearch transformer must implement a `transform_document` method."
+        )
+
+    def _load_to_file(self, file: IO, load_format: IngestorLoadFormat) -> int:
+        es_documents = list(self.stream_es_documents())
+        if len(es_documents) == 0:
+            raise ValueError("No documents to write.")
+
+        if load_format == "parquet":
+            df = pl.DataFrame(es_documents, infer_schema_length=None)
```
kenoir, line 36: I did wonder about whether we should hand a schema in here again, but we are essentially enforcing a schema with pydantic types ... I wonder if there is a way to make that _more_ enforced in the conversion and reading from parquet, [something like this.](https://github.com/simw/pydantic-to-pyarrow) 

I don't think this is necessary now, but it would be nice to have some guarantee beyond the model validation after loading that we are accessing the correct types.

PR #2987, opened 2025-08-14
Add new denormalised index mapping

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6111

Index the `mergedTime` field in the denormalised index as part of implementing [graph pipeline incremental mode](https://github.com/wellcomecollection/platform/issues/6105). Indexing the field will allow us to run filter queries to retrieve works which were updated within a given time window. 

> [!NOTE]  
> This change requires a reindex.

## How to test

This change can be tested via the following script (assuming Elasticsearch is running locally on port 9200):

```python
from elasticsearch import Elasticsearch

es = Elasticsearch(hosts=["http://localhost:9200"])

# Create a new index with the new denormalised mapping
index_name = "works-denormalised"
index_body = {
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 1,
    "analysis": {
      "analyzer": {
        "exact_path_analyzer": {
          "type": "custom",
          "tokenizer": "path_hierarchy"
        }
      },
      "normalizer": {
        "lowercase_normalizer": {
          "filter": [
            "lowercase"
          ],
          "type": "custom"
        }
      }
    }
  },
  "mappings": <insert new mapping>
}

es.indices.create(index=index_name, body=index_body)

# Index an example denormalised work
doc = <insert some denormalised work JSON>

response = es.index(index=index_name, document=doc)


# Use a range query to make sure the work is returned as part of the correct time window
query_body = {
    "size": 10,
    "query": {
        "range": {
            "state.mergedTime": {
                "gte": "2025-05-14T16:30:00Z",
                "lte": "2025-05-14T16:35:00Z"
            }
        }
    }
}

response = es.search(index=index_name, body=query_body)
print(response)
```

When running queries against the denormalised index, we ran filter for works updated within a given time window.

This change involves a full reindex, so risks are minimal.


index_config/mappings.works_denormalised.2025-08-14.json
```diff
@@ -0,0 +1,40 @@
...
+        }
+      }
+    },
+    "state": {
+      "properties": {
+        "mergedTime": {
+          "type": "date"
+        }
+      }
+    },
```
StepanBrychta, line 35: This is the only new field, other fields are copied from the previous mapping.

PR #2988, opened 2025-08-18
Add incremental mode to graph pipeline

StepanBrychta:
## What does this change?

* Add incremental mode support to graph pipeline services — extractor, bulk loader, and bulk load poller.
* Retire the `wellcomecollection-neptune-graph-loader` bucket and store bulk load outputs in the `wellcomecollection-catalogue-graph` bucket (along with ingestor and graph remover data).
* Add `pipeline_date` parameter to all graph pipeline services to separate bulk load files belonging to different pipelines.
* Remove unused step functions — `catalogue_graph_extractors` and `catalogue_graph_bulk_loaders` (we only use the daily/monthly variants).

## How to test

Run the following command to extract all concepts from works which were modified on 20/08/2025 between 15:30 and 15:45:

```
AWS_PROFILE=platform-developer python3.13 extractor.py --transformer-type=catalogue_concepts --entity-type=nodes --pipeline-date=2025-08-14 --stream-destination=local --window-start=2025-08-20T15:30 --window-end=2025-08-20T15:45 --is-local
```

Running the command above should result in a bulk load file being created in S3 under `graph_bulk_loader/2025-08-14/windows`.

* Support for incremental work/concept bulk loads into the catalogue graph
* Consolidation of graph-related data into a single bucket
* Isolation of bulk load files between different pipeline stacks

Risks are minimal — new logic for incremental operation should not affect the existing production pipeline.


catalogue_graph/src/transformers/catalogue/concepts_transformer.py
```diff
@@ -41,8 +49,9 @@ def extract_edges(
         self, raw_data: tuple[dict, WorkConceptKey]
     ) -> Generator[ConceptHasSourceConcept]:
         if self.id_label_checker is None:
+            transformers = get_args(MeshTransformerType) + get_args(LocTransformerType)
```
kenoir, line 52: 👌 neat.

catalogue_graph/src/utils/ontology.py
```diff
@@ -0,0 +1,56 @@
...
+    elif ontology == "wikidata_linked_loc":
+        transformers = tuple(t for t in get_args(WikidataTransformerType) if "loc" in t)
+    elif ontology == "wikidata_linked_mesh":
+        transformers = tuple(
+            t for t in get_args(WikidataTransformerType) if "mesh" in t
+        )
+    else:
+        raise ValueError(f"Unknown ontology {ontology}.")
+
+    return list(transformers)
```
kenoir, line 35:
```suggestion
# Build the mapping once at import time; reuse the wikidata tuple for derived subsets.
_wikidata = get_args(WikidataTransformerType)
TRANSFORMERS_BY_ONTOLOGY: dict[OntologyType, tuple[TransformerType, ...]] = {
    "wikidata": _wikidata,
    "loc": get_args(LocTransformerType),
    "mesh": get_args(MeshTransformerType),
    "catalogue": get_args(CatalogueTransformerType),
    "wikidata_linked_loc": tuple(t for t in _wikidata if "loc" in t),
    "wikidata_linked_mesh": tuple(t for t in _wikidata if "mesh" in t),
}

def get_transformers_from_ontology(ontology: OntologyType) -> list[TransformerType]:
    try:
        transformers = TRANSFORMERS_BY_ONTOLOGY[ontology]
    except KeyError:  # pragma: no cover - defensive; ontology is a Literal type
        raise ValueError(f"Unknown ontology {ontology}.")
    return list(transformers)
```

kenoir: I feel like you might be able to make the lookup simpler also (i.e. remove the tuple building) if you did something like this in `type.py`:

```python
# Wikidata-linked transformers split by the external ontology they link to.
WikidataLinkedLocTransformerType = Literal[
    "wikidata_linked_loc_concepts",
    "wikidata_linked_loc_locations",
    "wikidata_linked_loc_names",
]
WikidataLinkedMeshTransformerType = Literal[
    "wikidata_linked_mesh_concepts",
    "wikidata_linked_mesh_locations",
]
# Union of both; keep the original public name for backwards compatibility.
WikidataTransformerType = (
    WikidataLinkedLocTransformerType | WikidataLinkedMeshTransformerType
)
```

catalogue_graph/src/utils/bulk_load.py
```diff
@@ -0,0 +1,43 @@
...
+    """Return the file path of a bulk load file."""
+    file_name = f"{transformer_type}__{entity_type}.csv"
+
+    window_prefix = ""
+    if window is not None:
+        start = window.start_time.strftime("%Y%m%dT%H%M")
+        end = window.end_time.strftime("%Y%m%dT%H%M")
+        window_prefix = f"windows/{start}-{end}/"
+
+    return f"{pipeline_date}/{window_prefix}{file_name}"
```
kenoir, line 32:
```suggestion
from pathlib import PurePosixPath

def get_file_path(
    transformer_type: TransformerType,
    entity_type: EntityType,
    pipeline_date: str,
    window: IncrementalWindow | None = None,
) -> str:
    """Return the file path (object key) of a bulk load file."""
    file_name = f"{transformer_type}__{entity_type}.csv"

    parts: list[str] = [pipeline_date]
    if window is not None:
        start = window.start_time.strftime("%Y%m%dT%H%M")
        end = window.end_time.strftime("%Y%m%dT%H%M")
        parts += ["windows", f"{start}-{end}"]

    file_path = PurePosixPath(*parts) / file_name
    return str(file_path)
```

Maybe it would be preferable to use `pathlib` to avoid the string substitution construction of paths?

catalogue_graph/src/extractor.py
```diff
@@ -4,76 +4,69 @@
...
-    is_local: bool = False,
-) -> None:
+def handler(event: ExtractorEvent, is_local: bool = False) -> None:
     print(
-        f"Transforming {sample_size or 'all'} {entity_type} using the {transformer_type} "
-        f"transformer and streaming them into {stream_destination}."
+        f"Transforming {event.sample_size or 'all'} {event.entity_type} using the {event.transformer_type} "
+        f"transformer and streaming them into {event.stream_destination}."
     )
+    if event.pipeline_date == "dev":
```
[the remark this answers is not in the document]
kenoir: See https://github.com/wellcomecollection/platform/issues/6134

PR #2992, opened 2025-08-20
EBSCO Iceberg adapter, transformer reads from iceberg table

kenoir:
## What does this change?

Part of wellcomecollection/platform#6079.

This PR introduces a table-scoped client `utils.iceberg.IcebergTableClient` to encapsulate Iceberg operations and updates the loader/transformer to use it. It also removes legacy function shims and moves helper logic into the class.

Key changes:
- Add `IcebergTableClient` with:
  - `update(new_data, record_namespace?) -> str | None`
  - `get_records_by_changeset(changeset_id) -> pa.Table`
- Move helper functions into private methods on the class (static where appropriate): `_upsert_with_markers`, `_create_match_filter`, `_append_change_columns`, `_find_updates`, `_find_inserts`, `_get_deletes`.
- Remove legacy shims `update_table` and `get_records_by_changeset`.
- Migrate usages:
  - `src/steps/loader.py` constructs `IcebergTableClient` (with default namespace) and calls `update`.
  - `src/steps/transformer.py` reads records for a `changeset_id` via the client and reports `records_transformed` as the count.
- Tests updated to use the client directly; behavior unchanged.
- Add `ruff` to the dev dependency group; apply formatting/lint fixes and minor mypy cleanup.

## How to test

Run checks in the adapter:

```bash
uv run pytest
```

Manual flow:
1. Run the loader locally against a sample XML so it writes to the Iceberg table and returns a `changeset_id`.
2. Run the transformer locally with `--changeset-id=<id>` and verify `records_transformed` matches the number of records for that changeset.
3. If using Glue, ensure AWS credentials and `config.*` values are set; otherwise run with the local table option.

- Transformer can read the Iceberg table by `changeset_id` and reports a non-zero `records_transformed` when loader has written changes.
- Unit tests remain green (34 tests passing at time of writing).
- Lint/type checks pass cleanly (mypy, ruff).

- Removing shims may break external callers importing `utils.iceberg.update_table` or `get_records_by_changeset`. Mitigation: all in-repo usages have been migrated to the class; external consumers should update to the class API.
- Transformer’s Glue path requires appropriate AWS permissions/configuration; default config uses Glue unless `use_glue_table` is false.
- Minor dev tooling change (`ruff` added) updates `uv.lock`.

## Notes / Follow-ups

- Consider reconciling/removing the duplicated module `src/iceberg_updates.py` to avoid confusion.
- The client pattern enables future table-scoped operations (e.g. projections/pagination) and easier DI for time/UUID in tests.


ebsco_adapter/ebsco_adapter_iceberg/src/steps/loader.py
```diff
@@ -42,35 +41,29 @@ def load_xml(xmlfile: IO[bytes]) -> etree._Element:
...
-def extract_id(node: etree._Element) -> str:
-    id_field = node.find("./{http://www.loc.gov/MARC21/slim}controlfield[@tag='001']")
-    if id_field is None:
-        raise ValueError(f"id controlfield could not be found in {node}")
-    return id_field.text or ""
+    controlfield = cast(
+        list[str], node.xpath("./*[local-name()='controlfield' and @tag='001']/text()")
+    )
+    if not controlfield:
+        raise Exception("Could not find controlfield with tag 001")
```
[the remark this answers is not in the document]
kenoir: Maybe, but that should really be addressed in another PR as this one is not so connected to the loader.

ebsco_adapter/ebsco_adapter_iceberg/src/steps/loader.py
```diff
@@ -104,9 +97,13 @@ def handler(
 
 
 def lambda_handler(event: EbscoAdapterLoaderEvent, context: Any) -> dict[str, Any]:
-    return handler(
-        EbscoAdapterLoaderEvent.model_validate(event), EbscoAdapterLoaderConfig()
-    ).model_dump()
+    try:
+        config_obj = EbscoAdapterLoaderConfig()
+        transformer_event = handler(event, config_obj)
+        return {"changeset_id": transformer_event.changeset_id}
```
agnesgaroux, line 103: Isn't the EbscoAdapterTransformerEvent of the same type?

ebsco_adapter/ebsco_adapter_iceberg/tests/test_update.py
```diff
@@ -8,8 +8,8 @@
 from pyiceberg.expressions import EqualTo, In, IsNull, Not
 from pyiceberg.table import Table as IcebergTable
 
-from iceberg_updates import update_table
 from schemata import ARROW_SCHEMA
+from utils.iceberg import IcebergTableClient
```
agnesgaroux, line 12: Rename file `test_icebergTableClient.py`?

agnesgaroux: `test_utils_iceberg.py`?

ebsco_adapter/ebsco_adapter_iceberg/src/steps/loader.py
kenoir: Yes, it should do the same thing and `model_dump`, good catch.

PR #2999, opened 2025-08-26
use more specific concept Type is id same as subject

agnesgaroux:
## What does this change?

[#6094](https://github.com/wellcomecollection/platform/issues/6094)
See also https://wellcome.slack.com/archives/C02ANCYL90E/p1755769402477639

## How to test

Will add tests in upcoming PR
Still looking into local testing, which is made difficult by the fact that we can't transform specifc concepts by id

Concepts 
eg. `bjzcsb9x` work's subject "19th century" will be of type "period" as opposed to "subject" currently

Given the strict conditions under which the logic is applied, there may be cases where we still keep the more generic "subject" type

### DIFF
--- a/catalogue_graph/src/sources/catalogue/concepts_source.py
+++ b/catalogue_graph/src/sources/catalogue/concepts_source.py
@@ -14,7 +14,19 @@ def extract_concepts_from_work(
     # extracting these component concepts, since the frontend does not make use of them and the resulting
     # theme pages would be empty.
     for subject in raw_work.get("subjects", []):
-        yield subject, "subjects"
+        # caveat concerning the above comment:
+        # for non-composite concepts, identified as having the same id as the top-level subject,
+        # the nested "Type" is more specific than the top-level "Type"
+        # in that case, we copy the nested concept's "Type" onto the subject
+        # Check if the first concept in the concepts list has the same id as the subject
+        concepts = subject.get("concepts", [])
+        if concepts and concepts[0].get("id") == subject.get("id"):
+            # Replace the subject's type with the concept's type
+            subject_copy = subject.copy()
+            subject_copy["type"] = concepts[0].get("type", subject.get("type"))
+            yield subject_copy, "subjects"
+        else:
+            yield subject, "subjects"
 
     # Return all contributors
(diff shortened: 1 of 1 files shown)


catalogue_graph/src/transformers/catalogue/raw_concept.py
```diff
@@ -39,16 +39,12 @@ def label(self) -> str:
...
-        # not have a 'type' field in the denormalised index, so we need to add it here.
-        if self.referenced_in == "subjects":
-            return "Subject"
         # All concepts extracted from the 'genres' section are always of type 'Genre' (but the denormalised index uses
         # the term 'GenreConcept').
         if self.referenced_in == "genres":
             return "Genre"
 
-        concept_type: ConceptType = self.raw_concept["type"]
+        concept_type: ConceptType = self.raw_concept.get("type", "Subject")
```
kenoir, line 47: I think either the default value is not needed, or it should be `Concept` as the default root for anything that hasn't got a type.

agnesgaroux: https://github.com/wellcomecollection/catalogue-pipeline/blob/6ba727f6c3c62094d55e2d1af816ffc65c63b3d8/catalogue_graph/src/ingestor/transformers/raw_concept.py#L82 👍

catalogue_graph/src/sources/catalogue/concepts_source.py
```diff
@@ -13,16 +13,35 @@ def extract_concepts_from_work(
     # of concepts 'Milk' and 'Quality' (each with its own Wellcome ID). For now, we are not interested in
     # extracting these component concepts, since the frontend does not make use of them and the resulting
     # theme pages would be empty.
+    # CAVEAT concerning the above:
```
kenoir, line 16: You could replace this two part comment with:

```
# Generally, we ignore nested concepts within a subject because the frontend doesn't use them.
# However, an exception exists for simple, non-composite subjects where the nested concept 
# is the subject itself (identified by matching IDs). In this specific case, the nested 
# concept's "Type" is more specific, so we promote it to the top-level subject.
```

kenoir (approved):

PR #3005, opened 2025-09-04
Fix polars typing

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6139

* Add a utility function to convert Pydantic model definitions into PyArrow schemas.
* Use the function to create PyArrow schemas for ES documents as part of converting them into Parquet.
* Change the structure of the `partOf` field in the `display` part of work documents, moving from a recursive nested structure to a flat list. This makes it easier to convert work documents into Parquet (but will require a small frontend change before we can switch to the new works index).

## How to test

Run the works and concepts ingestors locally.

This removes our reliance on Polars type inference. As a result, it should now be possible to save all indexable works into Parquet files.

Risks are minimal.

### DIFF
--- a/catalogue_graph/tests/utils/test_parquet_roundtrip.py
+++ b/catalogue_graph/tests/utils/test_parquet_roundtrip.py
@@ -1,0 +1,116 @@
+import tempfile
+from pathlib import Path
+
+import polars as pl
+import pyarrow as pa
+import pyarrow.parquet as pq
+from pyarrow import types as pa_types
+from pydantic import BaseModel
+
+from utils.arrow import pydantic_to_pyarrow_schema
+
+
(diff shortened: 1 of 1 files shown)


catalogue_graph/pyproject.toml
```diff
@@ -17,7 +17,9 @@ dependencies = [
     "boto3-stubs",
     "types-requests",
     "python-dateutil",
-    "types-python-dateutil"
+    "types-python-dateutil",
+    "pyarrow~=20.0",
```
StepanBrychta, line 21: We need to pin PyArrow version to 20 because the latest version (21) does not include the required wheels for the Linux version we are using in the GitHub runner.

PR #3007, opened 2025-09-08
Graph add organisation connections

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6137
- adds HAS_INDUSTRY edges to wikidata-linked-loc concepts and names 
- adds HAS_FOUNDER edges to wikidata-linked-loc  concepts 

## How to test

Run
```
AWS_PROFILE=platform-developer uv run extractor.py --transformer-type wikidata_linked_loc_concepts --entity-type edges --stream-destination local --pipeline-date 2025-08-14 --sample-size 90000 --is-local

AWS_PROFILE=platform-developer uv run extractor.py --transformer-type wikidata_linked_loc_names --entity-type edges --stream-destination local --pipeline-date 2025-08-14 --sample-size 90000 --is-local
```
Expected
```
HAS_INDUSTRY:Q1770148-->Q11644505,Q1770148,Q11644505,HAS_INDUSTRY,,,,,,
HAS_FOUNDER:Q719620-->Q353723,Q719620,Q353723,HAS_FOUNDER,,,,,, // only for wikidata_linked_loc_concepts
```

Graph includes new edges, as described
Concepts ingestor can extract new edges and transform as
```
"relatedConcepts": {
   "relatedTo": [],
   "fieldsOfWork": [],
   "narrowerThan": [],
   "broaderThan": [],
   "people": [],
   "frequentCollaborators": [],
   "relatedTopics": [],
   "industry": [],   <- new property
   "foundedBy": []   <- new property
```

Graph part of the pipeline can be run before the ingestor changes go in: we can check the graph without affecting the index

### DIFF
--- a/catalogue_graph/src/sources/wikidata/linked_ontology_source.py
+++ b/catalogue_graph/src/sources/wikidata/linked_ontology_source.py
@@ -148,6 +148,12 @@ class WikidataLinkedOntologySource(BaseSource):
         yield from self._stream_all_edges_by_type("instance_of")
         yield from self._stream_all_edges_by_type("subclass_of")
 
+    def _stream_all_has_founder_edges(self) -> Generator[dict]:
+        yield from self._stream_all_edges_by_type("has_founder")
+
+    def _stream_all_has_industry_edges(self) -> Generator[dict]:
+        yield from self._stream_all_edges_by_type("has_industry")
+
     def _stream_filtered_wikidata_ids(self) -> Generator[str]:
         """Streams all wikidata ids to be processed as nodes given the selected `node_type`."""
         seen = set()
@@ -202,6 +208,20 @@ class WikidataLinkedOntologySource(BaseSource):
             if edge["from_id"] in streamed_wikidata_ids:
                 streamed_wikidata_ids.add(edge["to_id"])
                 yield {**edge, "type": "HAS_PARENT"}
+        
+        print("Streaming HAS_FOUNDER edges...")
+        for edge in self._stream_all_has_founder_edges():
+            if edge["from_id"] in streamed_wikidata_ids and is_id_in_ontology(
+                    edge["to_id"], "wikidata"
+                ):
+                yield {**edge, "type": "HAS_FOUNDER"}
+
+        print("Streaming HAS_INDUSTRY edges...")
+        for edge in self._stream_all_has_industry_edges():
+            if edge["from_id"] in streamed_wikidata_ids and is_id_in_ontology(
+                    edge["to_id"], "wikidata"
+                ):
+                yield {**edge, "type": "HAS_INDUSTRY"}
 
         # The following edges only apply to people (SourceName) nodes
         if self.node_type == "names":
--- a/catalogue_graph/src/sources/wikidata/linked_ontology_source.py
+++ b/catalogue_graph/src/sources/wikidata/linked_ontology_source.py
@@ -208,19 +208,12 @@ class WikidataLinkedOntologySource(BaseSource):
             if edge["from_id"] in streamed_wikidata_ids:
                 streamed_wikidata_ids.add(edge["to_id"])
                 yield {**edge, "type": "HAS_PARENT"}
-        
-        print("Streaming HAS_FOUNDER edges...")
-        for edge in self._stream_all_has_founder_edges():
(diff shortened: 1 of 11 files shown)


catalogue_graph/src/transformers/wikidata/names_transformer.py
```diff
@@ -40,20 +44,42 @@ def transform_node(self, raw_node: dict) -> SourceName | None:
...
             yield SourceNameRelatedTo(
-                from_id=raw_edge["from_id"],
-                to_id=raw_edge["to_id"],
+                from_id=raw_node["from_id"],
+                to_id=raw_node["to_id"],
                 attributes=attributes,
             )
         else:
-            yield from super().extract_edges(raw_edge)
+            raise ValueError(f"Unknown edge type {raw_node['type']}")
```
agnesgaroux, line 85: Because concepts now have an edge type (HAS_FOUNDER) that doesn't apply to names, they each have their own set of edges to extract

catalogue_graph/src/sources/wikidata/linked_ontology_source.py
```diff
@@ -234,6 +247,15 @@ def _stream_raw_edges(self) -> Generator[dict]:
                             "subtype": relationship_type,
                         }
 
+        # The following edges only apply to concepts (SourceConcepts) nodes
+        if self.node_type == "concepts":
```
StepanBrychta, line 251: I think we can simplify this by using a 'duck typing'-like approach — i.e. remove the `"self.node_type == "concepts"` check and stream the edge regardless of node type. (This will also allow us to simplify the Wikidata transformers by reducing duplicate code in `extract_edges` methods.)

In downstream queries, we currently don't distinguish between SourceConcept, SourceName, and SourceLocation nodes. (And if we ever do need to treat those node types differently, we can do so at query time rather than filtering out unwanted edges when building the graph.)

catalogue_graph/src/transformers/wikidata/names_transformer.py
agnesgaroux: Done! Only for `HAS_FOUNDER` edges though, not `HAS_FIELD_OF_WORK` for name nodes

catalogue_graph/src/transformers/wikidata/concepts_transformer.py
```diff
@@ -36,23 +38,36 @@ def transform_node(self, raw_node: dict) -> SourceConcept | None:
             description=raw_concept.description,
         )
 
-    def extract_edges(self, raw_edge: dict) -> Generator[BaseEdge]:
-        if raw_edge["type"] == "SAME_AS":
+    def extract_edges(self, raw_node: dict) -> Generator[BaseEdge]:
```
kenoir, line 41: Should this still be `raw_edge`?

agnesgaroux: In the base_transformer, extract_edges method:
> """Accepts a raw node from the source dataset and returns a generator of extracted edges as Pydantic models."""
We use the raw "node" to extract edges so I think this is correct

catalogue_graph/src/transformers/wikidata/concepts_transformer.py
```diff
@@ -36,23 +38,36 @@ def transform_node(self, raw_node: dict) -> SourceConcept | None:
...
+                to_id=raw_node["to_id"],
+            )
+        elif (
+            raw_node["type"] == "HAS_INDUSTRY"
+            or raw_node["type"] == "HAS_FIELD_OF_WORK"
+        ):
+            yield SourceConceptHasFieldOfWork(
+                from_id=raw_node["from_id"],
+                to_id=raw_node["to_id"],
             )
```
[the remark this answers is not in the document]
kenoir: This is intentional IIUC

catalogue_graph/src/transformers/wikidata/concepts_transformer.py
StepanBrychta: The base transformer docstring is a bit outdated. Back then we only had MeSH and LoC sources, which only streamed raw nodes, and we always extracted both nodes and edges from them. Wikidata is different — the source class streams raw nodes and raw edges separately. 

So `raw_edge` is a bit more accurate in this case. Sorry for the confusion!

PR #3009, opened 2025-09-09
Add Step Function interface to ID Minter

kenoir:
## What does this change?

This PR introduces a Step Function invocation path for the ID Minter so that batches of source identifiers can be minted deterministically via AWS Step Functions (part of platform issue https://github.com/wellcomecollection/platform/issues/6128). It adds a new generic Step Function lambda framework and wires the ID Minter into it.

Key points:
- New generic Step Function lambda support (`common/lambda/StepFunctionLambdaApp.scala`) with test coverage.
- New ID Minter entrypoint for Step Functions: `StepFunctionMain` + `IdMinterStepFunctionLambda`.
- New configuration model `StepFunctionIdMinterConfiguration` and Step Function request/response models in `StepFunctionModels.scala`.
- Added helper + feature tests (`IdMinterStepFunctionFeatureTest`, `IdMinterStepFunctionLambdaTest`, `StepFunctionMainTest`, `StepFunctionModelsTest`).
- Updated local tooling & compose files to allow invoking the Step Function path locally (updated `docker-compose.yml`, `local.docker-compose.yml`, and Python script `post_to_rie.py`).
- Introduced `JavaMapJsonCodec` to work around Scala type inference restrictions within the AWS Lambda (Java) runtime when decoding Step Function input/output payloads.
- Minor refactors to shared lambda test helpers (`MemoryDownstream`, `SQSMessageHelpers`, etc.).
- Terraform module adjustment to include the new service wiring (`service_id_minter.tf`).

New files of note:
- `common/lambda/JavaMapJsonCodec.scala`
- `common/lambda/StepFunctionLambdaApp.scala`
- `pipeline/id_minter/src/main/scala/.../IdMinterStepFunctionLambda.scala`
- `pipeline/id_minter/src/main/scala/.../StepFunctionMain.scala`
- `pipeline/id_minter/src/main/scala/.../config/models/StepFunctionIdMinterConfiguration.scala`
- `pipeline/id_minter/src/main/scala/.../models/StepFunctionModels.scala`
- Multiple new test suites for Step Function behaviour & models.

### Special note on JavaMapJsonCodec
`JavaMapJsonCodec` is required because Step Function Lambda invocations (through the Java runtime that underlies our Scala code) provide input as a raw Java `Map[String, Object]`. Scala's type inference & erasure make it awkward to apply the usual Circe / JSON decoding directly without either:
1. Re-serialising the map to JSON with lossy / fragile casting; or
2. Performing unsafe casts scattered across the codebase.

The codec centralises safe conversion of the Java map to the strongly typed request model, avoiding brittle reflection or unchecked casts and keeping model decoding in one place. This mitigates runtime ClassCastExceptions and keeps the Step Function entrypoint concise.

## How to test

1. Run the Scala test suite:
   - `sbt "project id_minter" test` (all new Step Function tests should pass).
2. Local invocation (compose):
   - Bring up the local stack with the updated `local.docker-compose.yml`.
   - Use the updated `post_to_rie.py` script to send a minting request (list of source identifier strings) to the Runtime Interface Emulator; verify successful response & that downstream notifications are emitted per invocation.
3. Idempotency check:
   - Re-run the same request 2–3 times; confirm (a) same canonical ID returned, (b) exactly one stored work, (c) multiple downstream notifications referencing the same canonical ID (see `IdMinterStepFunctionFeatureTest`).
4. Invisible works:
   - Confirm that an invisible work still receives a minted identifier and a single downstream notification.

- Deterministic reuse of identifiers for identical source identifiers across repeated Step Function executions.
- No increase in identifier write contention / DynamoDB conditional failures.
- Lambda & Step Function metrics show successful executions with minimal retries.
- Downstream topic/message counts match the number of successful requests (1 message per successful invocation in the feature tests).
- Future: We can add CloudWatch metrics / logs for: successes, failures, decode errors, batch size distribution.

| Risk | Mitigation |
|------|------------|
| Incorrect decoding of Step Function input leading to runtime failures | Centralised `JavaMapJsonCodec` test-covered; unit tests assert round-trip decoding. |
| Increased load / hot partitioning in identifiers table from batch triggering | Step Function requests are currently simple single-work batches; scaling strategy (parallelisation / throttling) to be defined before larger batches. |
| Duplicate downstream notifications if Step Function retries after partial success | Idempotency at ID write layer prevents corrupted data; downstream duplication acceptable short-term, can be deduped later by message body (canonical ID + job ID). |
| Visibility differences (invisible works) causing inconsistent downstream states | Explicit test verifies invisible works still minted; downstream consumers already handle visibility flags. |
| Future schema drift between request model & Java map parsing | Centralised codec + model tests (`StepFunctionModelsTest`) will fail fast. |

## Out of scope / Next steps

- Creating & deploying the actual AWS Step Function state machine definition (this PR only supplies the Lambda entrypoint + infra hook, not the full workflow definition).
- Observability enhancements (structured logs / metrics specific to Step Function path).
- Performance / load testing at scale.

## Related ticket

Partially addresses: https://github.com/wellcomecollection/platform/issues/6128
(Provides Step Function Lambda pathway; remaining work: state machine orchestration, scaling, and metrics.)

## Checklist

- [x] New Lambda entrypoint & configuration added
- [x] Shared Step Function lambda framework added
- [x] JavaMapJsonCodec implemented & unit tested
- [x] Feature tests for idempotency & invisible works
- [x] Local dev tooling updated (compose + script)
- [x] Terraform module updated
- [x] Documentation updated in `pipeline/id_minter/README.md`
- [ ] (Follow-up) Add state machine definition & deployment
- [ ] (Follow-up) Add metrics/alerts for Step Function path


pipeline/id_minter/docker-compose.yml
```diff
@@ -1,13 +1,12 @@
-version: '3.3'
 services:
   mysql:
-    image: "public.ecr.aws/docker/library/mysql:8.0"
```
kenoir, line 4: Should these still be using the `public.ecr.aws` images?

pipeline/id_minter/docker-compose.yml
```diff
@@ -1,13 +1,12 @@
...
+    image: "mysql:8.0"
     ports:
       - "3307:3306"
```
kenoir, line 5: The local.docker-compose.yml is update to use port mapping 3306:3306. Why the inconsistency?

common/lambda/src/main/scala/weco/lambda/JavaMapJsonCodec.scala
```diff
@@ -0,0 +1,72 @@
...
+    case b: java.lang.Boolean => Json.fromBoolean(b.booleanValue())
+    case n: java.lang.Integer => Json.fromInt(n.intValue())
+    case n: java.lang.Long    => Json.fromLong(n.longValue())
+    case n: java.lang.Double  => Json.fromDoubleOrNull(n.doubleValue())
+    case n: java.lang.Float   => Json.fromFloatOrNull(n.floatValue())
+    case n: java.math.BigDecimal =>
+      Json.fromBigDecimal(scala.math.BigDecimal(n))
+    case n: java.lang.Number => Json.fromBigDecimal(BigDecimal(n.toString))
+    case s: String           => Json.fromString(s)
+    case other => Json.fromString(other.toString) // Fallback to toString
```
agnesgaroux, line 32: Would it be useful to log whenever `value` doesn't match any of the explicit cases?

agnesgaroux (approved):
🤝

PR #3010, opened 2025-09-09
Prepare EBSCO adapter output for Step Function consumption

kenoir:
# Prepare EBSCO adapter output for Step Function consumption

## Summary

Modifies the EBSCO adapter transformer to output batch IDs in newline-delimited JSON format suitable for Step Function consumption, enabling integration with the ID Minter state machine.

We are trying to meet the `ItemReader` interface: https://docs.aws.amazon.com/step-functions/latest/dg/input-output-itemreader.html, in order that a distributed map state step can consume the output of this state machine.

My working assumption is that we will create a new state machine that can be triggered by this one, and in the future other adapters, and that state machine will coordinate passing the output to a distributed map state step that starts id_minter instances as required.

Though the `ItemReader` interface accepts JSON files, I have elected to use JSONL / NDJSON (referring to the same format) as that is consumable at scale by any expected consumer due to the ability to stream data from an NDJSON file vs needing to parse a JSON file before reading in most cases.

### Output Format Updates
- **File extension**: Changed from `.ids` to `.ids.ndjson` to signal newline-delimited JSON format
- **Step Function compatibility**: Output format now supported by AWS Step Functions for downstream processing

### Code Changes
- **`transformer.py`**: Updated `_write_batch_file()` to use `.ndjson` extension
- **`step_events.py`**: Enhanced event models for better Step Function integration
- **Test updates**: Updated test fixtures and assertions to match new output format

### Benefits
- **Step Function native**: Output can be directly consumed by Step Functions Map states
- **Clear format signaling**: `.ndjson` extension clearly indicates the file format
- **Downstream compatibility**: Prepares for ID Minter integration in state machine workflow

## Related Issues

- Addresses part of wellcomecollection/platform#6128
- Enables EBSCO adapter → ID Minter state machine integration

## Testing

- Updated test suite to verify new output format
- Maintains backward compatibility with existing processing logic

### DIFF
--- a/ebsco_adapter/ebsco_adapter_iceberg/src/steps/transformer.py
+++ b/ebsco_adapter/ebsco_adapter_iceberg/src/steps/transformer.py
@@ -245,7 +245,9 @@ def handler(
         "works-source",
         (
             # Cascading choice for date in index name
-            event.index_date or config_obj.index_date or config_obj.pipeline_date
+            event.index_date
+            or config_obj.index_date
+            or config_obj.pipeline_date
         ),
     )
 
--- a/ebsco_adapter/ebsco_adapter_iceberg/tests/test_mocks.py
+++ b/ebsco_adapter/ebsco_adapter_iceberg/tests/test_mocks.py
@@ -46,7 +46,9 @@ class MockSmartOpen:
             # Create a temporary file, return a handle and save the location in the file lookup
             # We're ignoring "SIM115 Use a context manager for opening files" as we need to keep
             # the file around for the duration of the test, cleaning it up in reset_mocks.
-            temp_file = tempfile.NamedTemporaryFile(delete=False, mode=mode)  # noqa: SIM115
+            temp_file = tempfile.NamedTemporaryFile(
+                delete=False, mode=mode
+            )  # noqa: SIM115
             # Insert the to_boto3 method to simulate the method provided by smart_open
             # https://github.com/piskvorky/smart_open/blob/develop/howto.md#how-to-access-s3-object-properties
             temp_file.to_boto3 = lambda _: MockBotoS3Object()  # type: ignore[attr-defined]


ebsco_adapter/ebsco_adapter_iceberg/src/models/step_events.py
```diff
@@ -41,12 +41,19 @@ class EbscoAdapterTransformerResult(EbscoAdapterEvent):
     Instead of embedding potentially large batch id lists directly in the
     state machine event, we persist them to S3 and return the location.
 
-    batch_file_location: S3 URI (s3://bucket/prefix/<job_id>.json) containing a
-    JSON array of arrays (each inner list is the ordered IDs for a processed
-    RecordBatch). ``None`` when no work was performed (e.g. idempotent skip).
+    batch_file_location: S3 URI (s3://bucket/prefix/<job_id>.ndjson) containing
```
agnesgaroux, line 44: Given that we have a pattern of one "step A" emitting an "EbscoAdapterStepBEvent" for consumption by the next step in the state machine, is this an "IdMinterEvent"?

ebsco_adapter/ebsco_adapter_iceberg/src/models/step_events.py
```diff
@@ -41,12 +41,19 @@ class EbscoAdapterTransformerResult(EbscoAdapterEvent):
...
+    NDJSON (one JSON object per line) where each line is {"ids": [ ... ]} for a
+    processed RecordBatch. ``None`` when no work was performed (e.g. idempotent
+    skip).
+
```
agnesgaroux, line 48: How does it work with the S3 file and the step function Map State? do we provide a S3 uri and the step uses each line as a item in the Map?

ebsco_adapter/ebsco_adapter_iceberg/src/steps/transformer.py
```diff
@@ -181,13 +186,17 @@ def handler(
                 "Source file previously transformed; skipping transformer work (idempotent short-circuit)."
             )
```
agnesgaroux, line 187: In which scenario would it short-circuit here after not having short-circuited in the loader?

ebsco_adapter/ebsco_adapter_iceberg/src/models/step_events.py
kenoir: I think the adapter doesn't care what the downstream step is, so I would avoid calling it an `IdMinterEvent`.

I do like the notion of it being a generalised `AdapterResult`, but I think we can do that when we have another adapter (and subsequent to adding this to the catalogue-graph folder).

ebsco_adapter/ebsco_adapter_iceberg/src/models/step_events.py
kenoir: I should have clarified in the PR description, we are trying to meet the `ItemReader` interface: https://docs.aws.amazon.com/step-functions/latest/dg/input-output-itemreader.html, in order that a distributed map state step can consume the output of this state machine.

My working assumption is that we will create a new state machine that can be triggered by this one, and in the future other adapters, and that state machine will coordinate passing the output to a distributed map state step that starts id_minter instances as required.

ebsco_adapter/ebsco_adapter_iceberg/tests/conftest.py
```diff
@@ -1,34 +1,127 @@
...
 @pytest.fixture(autouse=True)
 def common_mocks(monkeypatch: MonkeyPatch) -> Generator[None, None, None]:
-    monkeypatch.setattr("boto3.Session", MockBoto3Session)
-    monkeypatch.setattr("boto3.resource", MockBoto3Resource)
-    monkeypatch.setattr("requests.request", MockRequest.request)
-    monkeypatch.setattr("requests.get", MockRequest.get)
-    monkeypatch.setattr("smart_open.open", MockSmartOpen.open)
-    monkeypatch.setattr("elasticsearch.Elasticsearch", MockElasticsearchClient)
-    monkeypatch.setattr("elasticsearch.helpers.bulk", MockElasticsearchClient.bulk)
+    # Ensure modules exist to avoid importing heavy deps under Python 3.13
```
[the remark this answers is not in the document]
kenoir: Good spot. This is LLM-vomit, deposited while trying to fix tests. It's unnecessary, have removed.

agnesgaroux (approved):

PR #3011, opened 2025-09-09
Use daemon threads in MergedWorksSource source

StepanBrychta:
## What does this change?

Threads for extracting documents from the denormalised index now run as daemons so that they automatically exit when the main thread crashes. 

## How to test

Modify the extractor code to throw an exception from the main thread and then run it locally. Does it exit as expected?

The extractor exists (with a non-zero exit code) when the main thread throws an exception.

Risks are minimal.


kenoir:
Maybe add a comment as to why this change has been made and maybe a ref to https://docs.python.org/3/library/threading.html?

PR #3019, opened 2025-09-18
Ingestor incremental mode

StepanBrychta:
## What does this change?

Implements https://github.com/wellcomecollection/platform/issues/6118 and https://github.com/wellcomecollection/platform/issues/6133

* Modify the concepts/works ingestor to use the ES denormalised index as a 'primary' source, supplementing ES documents with data from the catalogue graph.
* Add incremental mode support to the ingestor. This includes several architecture changes — `ingestor_trigger` and `ingestor_loader` have been consolidated into a single service, and all processing now happens in a single instance of `ingestor_loader` and `ingestor_indexer` (as opposed to several parallel instances).
* Split complex Neptune queries into smaller ones, call them independently, and 'stitch' the results together in `GraphConceptsExtractor`. This makes concept queries more maintainable and improves performance. It also fixes https://github.com/wellcomecollection/platform/issues/6104.
* Add a mechanism for parallelising ID-based Neptune queries by splitting large lists of work/concept IDs into smaller chunks.
* Remove `ingestor_reporter` to ensure we don't receive Slack notifications after each incremental ingestor run (every 15 minutes).
* Remove data validation mechanism from `monitor` services. The mechanism assumes that the full dataset is processed on each run, which is not the case when running in incremental mode.

Infrastructure:
* Add a new `catalogue-graph-extractors-incremental` state machine, running the full pipeline (extractor, bulk loader, ingestor), processing both works and concepts. Schedule the state machine to run every 15 minutes.  
* Add a new `catalogue-graph-removers` state machine for removing source concept nodes/edges, decoupling this functionality from the bulk loader. Add this state machine to the monthly pipeline only. 

## How to test

* To test the refactored concept queries, I indexed all concepts into the experimental `concepts-indexed-dev` index in the `2025-08-14` cluster. This index can be tested via a local frontend instance.
* To test the new incremental mode state machine, I applied the terraform changes, connecting the pipeline to non-production `concepts-indexed-dev` and `works-indexed-dev` indexes in the `2025-08-14` cluster. These indexes now get updated every 15 minutes. 

* Works and concepts should be processed incrementally and changes should propagate into final works/concepts indexes every 15 minutes.
* Concept queries should be more reliable and run more quickly. 

* This PR removes various monitoring and safety mechanisms which are not suitable for incremental mode. However, it does not yet replace them with alternative mechanisms, which makes monitoring more difficult. There is a separate ticket to address this: https://github.com/wellcomecollection/platform/issues/6161
* This PR disables the 'graph remover' and 'ingestor deletions' services for works and concepts since existing versions of these services are not suitable for incremental mode. There is a separate ticket to build incremental versions of these services: https://github.com/wellcomecollection/platform/issues/6121

### DIFF
--- a/catalogue_graph/terraform/locals.tf
+++ b/catalogue_graph/terraform/locals.tf
@@ -19,7 +19,7 @@ locals {
 
   # This is a hint that the ingestors might need to be in the pipeline stack!
   pipeline_date       = "2025-10-02"
-  concepts_index_date = "2025-08-21"
+  concepts_index_date = "2025-10-09"
 
   ingestor_types = ["concepts", "works"]
 
--- a/catalogue_graph/terraform/state_machine_schedulers.tf
+++ b/catalogue_graph/terraform/state_machine_schedulers.tf
@@ -29,7 +29,7 @@ resource "aws_scheduler_schedule" "catalogue_graph_pipeline_incremental" {
     input = <<JSON
     {
       "pipeline_date": "${local.pipeline_date}",
-      "index_date": "dev",
+      "index_date": "${local.concepts_index_date}",
       "window": {
         "end_time": "<aws.scheduler.scheduled-time>"
       }
--- a/pipeline/terraform/2025-10-02/main.tf
+++ b/pipeline/terraform/2025-10-02/main.tf
@@ -32,7 +32,7 @@ module "pipeline" {
       concepts = {}
     }
     "2025-10-09" = {
-      works    = {}
+      works    = { indexed = "works_indexed.2024-11-14" }
       images   = {}
       concepts = { indexed = "concepts_indexed.2025-06-17" }
     }


catalogue_graph/src/ingestor/steps/ingestor_loader.py
```diff
@@ -1,79 +1,59 @@
...
 
 
-def get_filename(event: IngestorLoaderLambdaEvent) -> str:
-    return f"{str(event.start_offset).zfill(8)}-{str(event.end_index).zfill(8)}"
-
-
 def handler(
-    event: IngestorLoaderLambdaEvent, config: IngestorLoaderConfig
+    event: IngestorLoaderLambdaEvent,
+    es_mode: ElasticsearchMode = "private",
```
kenoir, line 38: 👌

catalogue_graph/src/pit_opener.py
```diff
@@ -0,0 +1,49 @@
...
+from utils.elasticsearch import ElasticsearchMode, get_client, get_standard_index_name
+
+
+def handler(event: BasePipelineEvent, es_mode: ElasticsearchMode = "private") -> dict:
+    """Create a point in time (PIT) on the denormalised index and return its PIT ID."""
+    es_client = get_client("graph_extractor", event.pipeline_date, es_mode)
+    index_name = get_standard_index_name(
+        config.ES_DENORMALISED_INDEX_NAME, event.pipeline_date
+    )
+    pit = es_client.open_point_in_time(index=index_name, keep_alive="15m")
```
kenoir, line 15: The fixed PIT here means steps using it must complete within the 15m window, are we comfortable this will be true even during the reindex case where there will be many thousands of records to process within a window?

PR #3023, opened 2025-09-29
Force old EBSCO adapter to look only in "ftp/" prefix

kenoir:
## What does this change?

This change updates the prefix that the old EBSCO adapter looks in, in S3 for synced files from FTP.

Currently the old adapter reads files from the new adapter which is using the prefix ftp_v2/

We also make some changes to the old run local script in order to ensure it continues working.

## How to test

- [ ] Make this change and run locally.

The EBSCO adapter continues to sync changes.

Risks should be minimal, we are returning things to a working state.

### DIFF
--- a/ebsco_adapter/ebsco_adapter/src/main.py
+++ b/ebsco_adapter/ebsco_adapter/src/main.py
@@ -25,7 +25,7 @@ reindex_topic_arn = os.environ.get("REINDEX_TOPIC_ARN")
 s3_bucket = os.environ.get("S3_BUCKET", "wellcomecollection-platform-ebsco-adapter")
 s3_prefix = os.environ.get("S3_PREFIX", "dev")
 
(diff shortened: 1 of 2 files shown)


ebsco_adapter/ebsco_adapter/src/main.py
```diff
@@ -25,7 +25,7 @@
 s3_bucket = os.environ.get("S3_BUCKET", "wellcomecollection-platform-ebsco-adapter")
 s3_prefix = os.environ.get("S3_PREFIX", "dev")
 
-ftp_s3_prefix = os.path.join(s3_prefix, "ftp")
+ftp_s3_prefix = os.path.join(s3_prefix, "ftp/")
```
agnesgaroux, line 28: 😱

agnesgaroux (approved):

PR #3025, opened 2025-10-01
 Self-contained Miro Links Script

kenoir:
## What does this change?

This PR restructures the standalone “Miro links” helper script into a self‑contained, reproducible mini-project under `scripts/miro_links/` and ships its current output.

Key points:
- Moves the former `scripts/get_miro_links.py` to `scripts/miro_links/main.py` (clean rename + slight tidy; logic unchanged).
- Adds a `pyproject.toml` and `uv.lock` so the script has pinned, reproducible dependencies (requests / pandas / tqdm) without touching the broader repo dependency surface.

## How to test

Local (requires `uv` installed — fast, no virtualenv boilerplate):
1. From repo root:
   ```bash
   cd scripts/miro_links
   uv run python main.py
   ```
2. Observe progress bar (“Processing pages”).
3. Confirm a new or updated `mirolinks.csv` is written.

- No stale local dependency assumptions.
- Reproducibility: Hash stability of `uv.lock` ensures identical dependency resolution (can diff regenerated lockfile; unchanged = success).
- 

Minimal risk, this is an offline, developer‑run script

### DIFF
--- a/scripts/miro_links/README.md
+++ b/scripts/miro_links/README.md
@@ -1,0 +1,12 @@
+# Miro Links Script
+
+This script is used to find links between Miro images that have been
+individually re-catalogued in Calm and their source Sierra work.
+
+The works from Calm contain a related material note that links to the
+work page for the Sierra work. This script looks up the associated B
+number for that work and creates a CSV file that contains work IDs,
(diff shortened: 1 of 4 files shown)


agnesgaroux (approved):
Does what it says on the tin 👍

PR #3029, opened 2025-10-06
extend concept_override to handle  display images

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6132
This extends the existing concept_override to enable adding display images to concepts 
This is part of the new Collection landing page work and will be used in the "Browse by Theme" component
We will follow with an [RFC](https://github.com/wellcomecollection/platform/issues/6157) describing how we might add a new Wellcome Collection authority source to the graph 

## How to test

Run tests, do they pass? 

❓Given that the concept API returns the `display` object which is not typed, I don't think we need to do anything on the concept-api side?
❓How do we update the specific concepts to add their images in the incremental mode pipeline? 

Concept API returns the additional property in the `display` object
EDITED to reflect the change to using DisplayDigitalLocation type
```
"display": {
            "id": "wev4whps",
            "identifiers": [
              {
                "value": "sh85106740",
                "type": "Identifier",
                "identifierType": {
                  "id": "lc-subjects",
                  "type": "IdentifierType",
                  "label": "Library of Congress Subject Headings (LCSH)"
                }
              }
            ],
            "label": "Printed ephemera",
            "displayLabel": "Printed ephemera",
            "alternativeLabels": [ ],
            "type": "Subject",
 ---->      "displayImages": [
               {
                  "locationType": {
                     "id": "iiif-image", 
                     "label": "IIIF Image API",
                     "type": "LocationType"
                  },
                 "url": "https://iiif.wellcomecollection.org/image/L0029207/info.json",
                 "accessConditions": [], // <- always empty in this case but part of the Location model
                 "type": "DigitalLocation"
               }
            ],
            "relatedConcepts": {
              "relatedTo": [...],
              "fieldsOfWork": [],
              "narrowerThan": [],
              "broaderThan": [...],
              "people": [],
              "frequentCollaborators": [],
              "relatedTopics": [...],
              "foundedBy": []
            },
            "sameAs": [
              "wev4whps"
            ]
          }
```

No overrides have been set yet. The "Browse by theme" component" is behind a toggle until everything is ready and working


agnesgaroux:
> Looks good!
> 
> My understanding is that even though `displayImages` is a list, the current implementation allows specifying at most one display image in the CSV. It might be worth making a small tweak to add support for a separator (e.g. `||`, which we use in other places) to allow assigning multiple images. In the CSV, we could then specify two images like this: `www.image1.info.json||www.image2.info.json`

Excellent suggestion, now implemented

StepanBrychta (approved):

PR #3030, opened 2025-10-06
Move EBSCO Adapter into catalogue-graph folder

kenoir:
## What does this change?

This change moves the new EBSCO adapter code into the catalogue-graph folder, and consolidates the build. It is not yet multiple pipeline aware in deployment so is hard-coded to deploy only to the new pipeline 2025-10-02, a future PR will implement multi-pipeline deployment (https://github.com/wellcomecollection/platform/issues/6168).

It also does not yet refactor to remove duplication, which is part of: https://github.com/wellcomecollection/platform/issues/6166

## How to test

- [x] Do the tests pass?
- [x] Does the new pipeline run (including the new adapter/transformer pair for EBSCO).

Consolidated code and deployment across "pipeline" python projects.

This could impact catalogue-graph deployment if we change shared code.


catalogue_graph/pyproject.toml
```diff
@@ -19,7 +21,14 @@ dependencies = [
     "python-dateutil",
     "types-python-dateutil",
     "pyarrow~=20.0",
-    "pyarrow-stubs"
+    "pyarrow-stubs",
+    "lxml>=6.0.0",
+    "mypy>=1.16.1",
```
paul-butcher, line 26: This is already in the dev dependency group. I don't think it's needed at runtime?

PR #3035, opened 2025-10-09
Add incremental graph remover

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6121

Add a new `graph_remover_incremental` service for removing works, concepts, and work path identifiers from the catalogue graph.

Removal logic works as follows:
* Work nodes: Extract all non-visible works from the merged index and remove corresponding nodes from the catalogue graph (if present).
* Work edges: For each visible work, extract all of its HAS_CONCEPT relationships from the merged index and from the catalogue graph. Compare the two sets of relationships and remove those which exist in the graph but not in the merged index.
* Concept nodes: Remove all nodes which do not have any HAS_CONCEPT edges.
* Path identifier nodes: Similar to concept nodes.
* Path identifier edges: Similar to work edges.

Terraform changes have been applied and the incremental remover now runs as part of the incremental pipeline.

## How to test

Run the incremental remover locally. For example, to remove all unused concept nodes:

```
AWS_PROFILE=platform-developer uv run graph_remover_incremental.py --pipeline-date=2025-08-14 --remover-type=catalogue_concepts --entity-type=nodes
```

To remove all non-visible work nodes which were indexed within the 2025-10-14T10:45 to 2025-10-14T11:00 time window:

```
AWS_PROFILE=platform-developer uv run graph_remover_incremental.py --pipeline-date=2025-08-14 --remover-type=catalogue_concepts --entity-type=nodes --window-end=2025-10-14T11:00
```

The following entities should be regularly removed from the graph:
* Non-visible work nodes
* Concept nodes with no HAS_CONCEPT edges
* Path identifier nodes with no HAS_PATH_IDENTIFIER edges
* HAS_CONCEPT edges which can no longer be extracted from the merged index
* HAS_PATH_IDENTIFIER edges which can no longer be extracted from the merged index

To mitigate the risk of accidentally removing nodes/edges, the incremental graph remover includes a similar safety mechanism to the full graph remover.

### DIFF
--- a/scripts/mimsy_dump/mimsy_dump.py
+++ b/scripts/mimsy_dump/mimsy_dump.py
@@ -143,6 +143,7 @@ def main():
             "VW_DISPOSAL_SOURCES",
             "VW_EXHIBITION_ITEMS",
             "VW_EXHIBITION_ITEMS_STATUS",
+            "VW_HAZARD",
             "VW_EXHIBITIONS",
             "VW_INSTRUCTIONS",
             "VW_ITEM_DESCRIPTIONS",


catalogue_graph/src/clients/base_neptune_client.py
```diff
@@ -192,48 +196,90 @@ def delete_all_nodes_with_label(self, label: str) -> None:
...
 
-        previous_node_count = self.get_total_node_count()
+        return list(self.run_parallel_query(ids, query).keys())
 
-        for batch in batched(ids, ID_DELETE_BATCH_SIZE):
-            self.run_open_cypher_query(delete_query, {"nodeIds": list(batch)})
-            print(f"Deleted a batch of nodes. (Batch size: {ID_DELETE_BATCH_SIZE})")
+    def delete_entities_by_id(self, ids: list[str], entity_type: EntityType) -> None:
+        if entity_type == "nodes":
+            query = "MATCH (n) WHERE id(n) IN $ids DETACH DELETE n"
```
kenoir: Given that we always delete the edges associated with a node using `DETACH DELETE`, does that impact the necessity to remove edges separately? e.g. in the case of deleting path identifier nodes, do we also need to delete the edges or will they already have been deleted?

catalogue_graph/src/converters/cypher/bulk_load_converter.py
```diff
@@ -45,7 +45,7 @@ def _node_to_bulk_cypher(self, model: BaseNode) -> dict:
     def _edge_to_bulk_cypher(self, model: BaseEdge) -> dict:
         bulk_edge = {
             # We need to give the edge a unique ID so that the Neptune bulk loader recognises duplicates
```
kenoir: Should we move this comment to where `edge_id` is generated?

catalogue_graph/src/ingestor/extractors/base_extractor.py
```diff
@@ -92,20 +89,16 @@ def make_neptune_query(
...
-        def _run_query(chunk: Iterable[str]) -> list[dict]:
-            return self.neptune_client.run_open_cypher_query(
-                NEPTUNE_QUERIES[query_type],
-                self.neptune_params | {"ids": sorted(chunk)},
-            )
-
         start = time.time()
-        raw_results = process_stream_in_parallel(
-            ids, _run_query, chunk_size, QUERY_THREAD_COUNT
+        results = self.neptune_client.run_parallel_query(
```
kenoir: I like that logic from `process_stream_in_parallel` has been pushed into `neptune_client` as `run_parallel_query`.

catalogue_graph/src/removers/base_graph_remover_incremental.py
```diff
@@ -0,0 +1,88 @@
...
+from utils.aws import get_neptune_client
+from utils.safety import validate_fractional_change
+from utils.types import EntityType
+
+ES_QUERY_NON_VISIBLE_WORKS = {"bool": {"must_not": {"match": {"type": "Visible"}}}}
+
+BATCH_SIZE = 40_000
+
+
+class BaseGraphRemoverIncremental:
```
kenoir: I wonder if we should split this into a base edge & base node remover class, and/or use multiple inheritance (it looks like re-use batching and validation of fractional change are why we handle edge and node removal in the same class). 

Splitting the class might help to keep it clear what the purpose of each class is.

kenoir (approved):
Some comments to respond to, can either be addressed here or in a future PR depending on justification.

Not blocking for merge.

catalogue_graph/src/clients/base_neptune_client.py
StepanBrychta: I think there are several possible scenarios:
1. A work gets removed (or changes status from visible to non-visible): In this case, we remove the corresponding work node (along with all of its edges) when running the remover with `transformer_type=catalogue_works` and `entity_type=nodes`. This might result in a disconnected path identifier, which we remove as part of the same pipeline run using the `transformer_type=catalogue_work_identifiers` and `entity_type=nodes` remover.
2. A work's collection path gets removed (or changed to a different value). In this case, we remove the edge linking the node to its current path identifier when running the remover with `transformer_type=catalogue_works` and `entity_type=edges`. This will result in a disconnected path identifier, which we remove in the same way as in scenario 1. 
3. The collection paths of two works are swapped (e.g. as part of a cataloguing error correction). In this case, we do not remove any nodes. We only remove the two edges linking each work to its old path identifier.

I think that to cover scenarios 2 and 3, we need a separate edge remover for path identifiers. We could probably handle these scenarios through some custom logic (using the fact that a work can be linked to at most one path identifier), but I think that the solution in this PR is simpler since it uses the same general approach for both `HAS_CONCEPT` edges and `HAS_PATH_IDENTIFIER` edges.

catalogue_graph/src/sources/catalogue/concepts_source.py
```diff
@@ -1,65 +1,90 @@
...
-            event, query=query, fields=fields, es_mode=es_mode
+            event, query=ES_QUERY, fields=ES_FIELDS, es_mode=es_mode
         )
 
-    def stream_raw(self) -> Generator[tuple[dict, WorkConceptKey]]:
+    def stream_raw(self) -> Generator[ExtractedWorkConcept]:
         """Streams raw concept nodes from a work's subjects, genres, and contributors."""
         for work in self.es_source.stream_raw():
-            yield from extract_concepts_from_work(work["data"])
+            work_data = MergedWorkConceptsData.model_validate(work["data"])
```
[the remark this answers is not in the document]
StepanBrychta: Originally I created `MergedWorkConceptsData` model to make it more explicit which fields are expected by the `extract_concepts` method (since we often only load partial data from ES to save bandwidth), but I changed my mind on this and have now replaced the model with `WorkData`.

PR #3039, opened 2025-10-14
Consolidate WorkData between graph & adapter projects

kenoir:
## What does this change?

This change removes duplication of `WorkData` between the new EBSCO adapter and the existing catalogue graph project.

We have not yet aligned `SourceWork` & `MergedWork` to incorporate generalising `WorkState` on a common Work base model. That change will be made in a future PR.

## How to test

Reduced code duplication, easier to work with pipeline services.

This could break existing graph services as it changes services that are in use, we should mitigate by taking care that testing is thorough before release.


paul-butcher (approved):
🔀

PR #3044, opened 2025-10-17
Extract base Work & WorkState

kenoir:
## What does this change?

This change extracts the base work & work state, so it can be used in the new EBSCO adapter.  

## How to test

- [ ] Run the tests, does they pass?

No disruption to the existing ingestor.

This change updates subclasses of `MergedWork` to use multiple inheritance. We should make sure the ingestor continues to function as expected with this change.


catalogue_graph/src/ingestor/models/merged/work.py
```diff
@@ -1,61 +1,42 @@
...
-    source_identifier: SourceIdentifier
+class MergedWorkState(WorkState):
     canonical_id: str
     merged_time: datetime
-    source_modified_time: datetime
     availabilities: list[Id]
     merge_candidates: list[MergeCandidate]
-    relations: WorkRelations
+
+    @model_validator(mode="before")
```
StepanBrychta: I think it would be simpler to use a [computed field](https://docs.pydantic.dev/2.0/usage/computed_fields/) rather than manipulating the raw dictionary. Something like this should work:

```
    @computed_field
    @property
    def modified_time(self) -> datetime:
        return self.merged_time
```

kenoir: I don't think this is possible: https://github.com/pydantic/pydantic/issues/7250

PR #3057, opened 2025-10-24
Remove status datetimes in python, use strings for now

kenoir:
## What does this change?

Updates the EBSCO adapter and related models to use string timestamps instead of datetime objects for `source_modified_time` and `modified_time` in `WorkState`. This ensures consistent ISO format timestamps with timezone information.

This PR is part of https://github.com/wellcomecollection/platform/issues/6150

## How to test

- Run the EBSCO adapter tests: `uv run pytest -q ebsco_adapter/ebsco_adapter_iceberg/tests`
- Run the catalogue graph tests: `pytest catalogue_graph/tests/`
- Verify that the transformer produces valid `VisibleSourceWork` objects with correct timestamp formats

- All tests pass without regressions
- EBSCO records are transformed correctly with proper timestamp strings
- Location models serialize correctly with the type discriminator

- Changes to core pipeline models (`WorkState`) could affect downstream consumers.
- Timestamp parsing in the version calculation uses `dateutil.parser` which may have edge cases
- Ensures backward compatibility in serialization formats

### DIFF
--- a/catalogue_graph/src/utils/timezone.py
+++ b/catalogue_graph/src/utils/timezone.py
@@ -1,0 +1,7 @@
+from datetime import UTC, datetime
+
+
+def convert_datetime_to_utc_iso(dt: datetime) -> str:
+    """Convert a datetime object to an ISO 8601 string in UTC timezone."""
+    # strip +00:00 and replace with Z
+    return dt.astimezone(UTC).isoformat().replace("+00:00", "Z")


catalogue_graph/src/adapters/ebsco/transformers/ebsco_to_weco.py
```diff
@@ -1,5 +1,6 @@
 from datetime import datetime
```
paul-butcher: Include for suggestion on l35
```suggestion
from datetime import datetime, UTC
```

catalogue_graph/src/ingestor/models/debug/work.py
paul-butcher:
```suggestion
from datetime import datetime, UTC
```

catalogue_graph/tests/adapters/ebsco/test_transformer.py
```diff
@@ -456,6 +456,8 @@ def fake_bulk(client, actions, raise_on_error, stats_only):  # type: ignore[no-u
     monkeypatch.setattr("elasticsearch.helpers.bulk", fake_bulk)
 
     now = datetime.now()
+    now_iso = now.isoformat() + "Z"
```
paul-butcher: There are suggestions in application code files about UTC.  As this is freezing a time to create some test data, it probably doesn't matter much here, but we might want to add it for consistency (and just in case we change something else and the tests start mysteriously failing during BST)

paul-butcher (approved):

PR #3059, opened 2025-10-24
Add empty list VisibleSourceWork fields needed for Scala deserialisation

kenoir:
## What does this change?

Adds `internal_work_stubs` and `merge_candidates` fields to `SourceWorkState` in the Python models, defaulting to empty lists. This allows deserialization in the Scala pipeline, which expects these fields.

Updates the test fixture for `VisibleSourceWork` to include the state and renames it from `visible_maximal.json` to `visible_source_maximal.json`.

Modifies the test `test_visible_source_work_maximal_fixtures` to validate the `VisibleSourceWork` serialization against the updated fixture.

Part of https://github.com/wellcomecollection/platform/issues/6167

## How to test

Run the tests in `catalogue_graph/tests/models/test_work_serialisation.py`, particularly `test_visible_source_work_maximal_fixtures`.

Ensure the serialization and deserialization work correctly with the new fields.

- The test passes, confirming that the JSON serialization matches the fixture.
- No regressions in existing work serialization tests.
- Successful integration with the Scala pipeline for deserialization.

The changes add optional fields with default empty lists, so existing code should continue to work. The risk of breaking deserialization in Scala is mitigated by ensuring the fields are present in the serialized output.

### DIFF
--- a/catalogue_graph/tests/models/test_work_serialisation.py
+++ b/catalogue_graph/tests/models/test_work_serialisation.py
@@ -1,4 +1,5 @@
 import json
+import pathlib
 from typing import Any
 
 import pytest
@@ -338,8 +339,6 @@ def test_visible_source_work_maximal_fixtures() -> None:
 
     Combines previous separate tests into one for reduced duplication.
     """
-    import json
-    import pathlib
 
     # VisibleSourceWork fixture (includes state)
     vsw_state = SourceWorkState(


kenoir:
This is not strictly required for parsing at the id_minter, but i'm unsure if it will cause upstream issues and is noise in the index diff, we may want to eventually remove these fields as they may not be required in python-world.

PR #3070, opened 2025-10-28
Transform ebsco relations

StepanBrychta:
## What does this change?

Transform EBSCO Series relations for parity with the Scala transformer. This is done using a new `get_parents` function, which mirrors the logic of the [MarcParents](https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/transformer/transformer_marc_common/src/main/scala/weco/pipeline/transformer/marc_common/transformers/MarcParents.scala) Scala class.

## How to test

Run a full reindex followed by the ES index diff tool.

Unit tests were copied from Scala and expanded on.

All EBSCO works should now have correct Series relationships.

Risks are minimal, this is non-production code at the moment.

### DIFF
--- a/scripts/mimsy_dump/mimsy_dump.py
+++ b/scripts/mimsy_dump/mimsy_dump.py
@@ -140,6 +140,7 @@ def main():
             "VW_CONDITION_DESCRIPTORS",
             "VW_CONSERVATION",
             "VW_CONSERVATION_STATUS",
+            "VW_DATE_COLLECTED",
             "VW_DISPOSAL_SOURCES",
             "VW_EXHIBITION_ITEMS",
             "VW_EXHIBITION_ITEMS_STATUS",


catalogue_graph/src/models/pipeline/work_state.py
```diff
@@ -19,9 +19,18 @@ class WorkRelations(SerialisableModel):
...
         # TODO: This is a temporary 'Series' filter which won't be needed once we remove the relation embedder service
-        return [a for a in raw_ancestors if a["numChildren"] == 0]
+        filtered = []
+        for ancestor in raw_ancestors:
+            if isinstance(ancestor, WorkAncestor):
+                filtered.append(ancestor.model_dump())
+            elif ancestor["numChildren"] == 0:
+                filtered.append(ancestor)
+
+        return filtered
```
StepanBrychta: This is not a great solution but it is a temporary one — we won't need this method once we remove the relation embedder.

kenoir (approved):

PR #3075, opened 2025-10-29
Ebsco transformer date parsing improvements

StepanBrychta:
## What does this change?

Tweak how we parse dates from MARC fields to address edge cases and allow records `ebs11899102e` and `ebs195742e` to be successfully processed.

* If all four digits of a date defined in field 008 are unknown (`uuuu`), do not attempt to parse the date.
* Improve how we parse fields which define a range with more than two dates. For example, given the value `©1928, ©1929-1936`, we should extract `from_date=1928` and `to_date=1936`.

## How to test

Unit tests were expanded to cover the two test cases above.

All EBSCO records are successfully transformed.

This change is low-risk.

### DIFF
--- a/catalogue_graph/src/adapters/ebsco/transformers/parsers/period.py
+++ b/catalogue_graph/src/adapters/ebsco/transformers/parsers/period.py
@@ -168,9 +168,6 @@ def fill_year_prefix(from_year: str, to_year: str) -> str:
     >>> fill_year_prefix("1964", "72")
     '1972'
 
-    >>> fill_year_prefix("1964", "72")
-    '1972'
-
     >>> fill_year_prefix("650", "55")
     '655'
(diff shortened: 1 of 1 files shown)


catalogue_graph/src/adapters/ebsco/transformers/parsers/period.py
```diff
@@ -62,10 +62,6 @@ def preprocess(period: str) -> str:
 
      - Square brackets
     >>> preprocess("[2016]-[2020]")
-    '2016 - 2020'
-
-     - Square brackets
-    >>> preprocess("[2016]-[2020]")
```
StepanBrychta: Remove duplicate examples

kenoir (approved):

PR #3077, opened 2025-10-29
filter-subjects

paul-butcher:
## What does this change?

See https://github.com/wellcomecollection/catalogue-pipeline/issues/3076

The new EBSCO transformer was not filtering out the extra subjects

## How to test
Deploy and run the transformer. Check with es_index_comparison 

Reduce discrepancy between the two indices - remove differences like these:

A French subject
> - **scalar_change** @ `data.subjects[4].id.sourceIdentifier.value`
>   - works-source-2025-10-02: `biology`
>   - works-source-2025-10-06: `biomathematiques - periodiques`

More subjects in the new index than the old

> - **list_length** @ `data.subjects`
>   - works-source-2025-10-02: `7`
>   - works-source-2025-10-06: `16`

### DIFF
--- a/catalogue_graph/src/adapters/ebsco/transformers/subjects.py
+++ b/catalogue_graph/src/adapters/ebsco/transformers/subjects.py
@@ -99,14 +99,19 @@ SUBDIVISION_TRANSFORMS = {
 
 def extract_subjects(record: Record) -> list[Subject]:
     return non_empty(
-        extract_subject(field) for field in record.get_fields(*SUBJECT_FIELDS) if is_subject_to_keep(field)
+        extract_subject(field)
+        for field in record.get_fields(*SUBJECT_FIELDS)
+        if is_subject_to_keep(field)
     )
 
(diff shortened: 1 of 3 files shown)


catalogue_graph/src/adapters/ebsco/transformers/subjects.py
```diff
@@ -99,7 +99,19 @@ def subdivision_concepts_648_650_651(field: Field) -> Generator[Concept]:
...
+        if is_subject_to_keep(field)
+    )
+
+
+def is_subject_to_keep(field: Field) -> bool:
+    return field.indicators is not None and (
+        field.indicators.second in ["0", "2"]
+        or (
+            field.indicators.second == "7"
+            and field.get("2") in ["local", "homoit", "indig", "enslv"]
```
kenoir: Maybe link or reference the Scala implementation of this filter - i've seen it hanging about somewhere!

PR #3089, opened 2025-11-03
Works ingestor tweaks, some pipeline clean-up

kenoir:
## What does this change?

Towards: https://github.com/wellcomecollection/platform/issues/6089

Some tweaks for running the new works ingestor successfully:

- Increase the error threshold to 20%, this allows room for churn during reindexes while still keeping an acceptable error threshold for large unexpected changes.
- Give the bulk load poller some more memory, it was struggling at 128mb and failed intermittently.

See https://wellcome.slack.com/archives/C02ANCYL90E/p1762163724937939?thread_ts=1761912552.903319&cid=C02ANCYL90E

Some other clean-ups
- Remove the id_minter scale var, it was unused.
- Chuck the notebook outputs, they are not needed.

## How to test

- [ ] See testing in Slack thread!

Working towards using the new works ingestor.

Should be minimal, this is all in the new non-prod pipeline.

### DIFF
--- a/catalogue_graph/tests/test_graph_remover_incremental.py
+++ b/catalogue_graph/tests/test_graph_remover_incremental.py
@@ -210,7 +210,7 @@ def test_graph_remover_catalogue_failure() -> None:
 
 def test_graph_remover_safety_mechanism() -> None:
     disconnected_ids = ["byzuqyr5", "vjfb76xy"]
-    mock_neptune_get_total_node_count("Concept", 20)
+    mock_neptune_get_total_node_count("Concept", 9)
     mock_neptune_get_disconnected_concept_nodes(disconnected_ids)
     mock_neptune_get_existing_nodes_response(disconnected_ids)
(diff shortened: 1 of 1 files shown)


pipeline/terraform/modules/pipeline/graph_pipeline/lambda_bulk_load_poller.tf
```diff
@@ -14,8 +14,8 @@ module "bulk_load_poller_lambda" {
     command = ["bulk_load_poller.lambda_handler"]
   }
 
-  memory_size = 128
-  timeout     = 120 // 120 seconds
+  memory_size = 256
+  timeout     = 60 // 60 seconds
```
StepanBrychta: I've seen timeout errors on the bulk loader Lambda too (`bulk_loader_lambda` resource), so I think we should increase its memory in the same way.

StepanBrychta (approved):

PR #3096, opened 2025-11-04
Move ingestor-loader & ingestor-indexer to ECS tasks

kenoir:
## What does this change?

This change moves the `ingestor-loader` and `ingestor-indexer` steps in the graph pipeline ingestor state machine from AWS Lambda functions to ECS tasks. This addresses timeout issues encountered during reindexing operations where Lambda execution limits were exceeded.

Key changes include:
- Added ECS task definitions for `ingestor-loader` and `ingestor-indexer` with appropriate CPU/memory allocations (4096 CPU, 16384 MB memory) and shared infrastructure for Step Functions task token handling.
- Introduced a shared `StepFunctionOutput` helper (with comprehensive tests) and updated `run_ecs_handler` so ECS handlers uniformly report success/failure, including when handlers return `None`.
- Refactored the loader and indexer Python steps to route through `run_ecs_handler`, discover S3 outputs when needed, and emit pipeline reports/CloudWatch metrics directly; the dedicated monitor Lambdas have been removed as a result.
- Updated documentation and CLI entry points to reflect the new ECS execution paths and the retirement of the monitor Lambdas.
- Extended the mock/test suite (S3, Step Functions, loader/indexer flows) to cover the new behaviour, including report generation and Step Functions interactions.
- Added a unified ECS cluster and ECR repository (`unified_pipeline_task`) so the extractor, loader, and indexer can share the same container image, and aligned the extractor state machine to run the consolidated image.
- Increased the extractor works batch size to 40,000 to better utilise the additional ECS capacity.
- Refactored IAM/Terraform modules to support `ecs:runTask.waitForTaskToken`, attach per-task invoke policies, and simplify state machine role attachments.

> [!Note]
> This change is already applied.

This resolves https://github.com/wellcomecollection/platform/issues/6193

Follow-up tasks, these will be covered in a future PR (not done here to allow working deployment to continue with infra changes):
- [ ] Ensure the unified pipeline task image is published with an environment-specific tag (currently referencing the `dev` tag).
- [ ] Update all Lambda deployments that still reference the `prod` image tag to use the appropriate `env.{pipeline_date}` tag.

## How to test

- Deploy the changes to a test environment.
- Run a graph pipeline ingestor job and verify that the loader and indexer steps execute successfully as ECS tasks.
- Check Step Functions execution logs to ensure task tokens are handled correctly and that success/failure is reported back via `StepFunctionOutput`.
- Verify that the output from ECS tasks (including pipeline reports and metrics) is properly written to S3 and CloudWatch.
- Test both CLI and ECS execution modes of the modified scripts, ensuring loader/indexer reports are generated in both modes.

- Ingestor jobs complete without timing out during large reindexing operations.
- Step Functions state machine executions show successful ECS task runs with correct success/failure callbacks.
- Pipeline reports and CloudWatch metrics continue to be produced for loader/indexer runs without the separate monitor Lambdas.
- No regression in functionality compared to Lambda-based execution.
- Monitoring shows ECS tasks have sufficient resources for the workload.

- **Resource allocation**: ECS tasks have higher resource limits than Lambdas, but we've allocated conservative amounts (4 vCPU, 16 GB RAM) based on existing Lambda configurations.
- **Cost**: ECS tasks may incur slightly higher costs than Lambdas.
- **Execution time**: ECS tasks have longer startup times than Lambdas, but overall job completion should be faster due to the absence of Lambda timeout limits.
- **Backward compatibility**: The scripts maintain CLI execution support and continue to work when invoked without a Step Functions task token.
- **IAM permissions**: New policies for ECS task execution, Step Functions integration, and S3 discovery have been carefully scoped.

### DIFF
--- a/catalogue_graph/src/utils/steps.py
+++ b/catalogue_graph/src/utils/steps.py
@@ -25,10 +25,7 @@ class StepFunctionOutput:
         self, task_token: str | None, stepfunctions_client: StepFunctionClient | None
     ) -> None:
         self.task_token = task_token
-        self.stepfunctions_client = stepfunctions_client or boto3.client(
-            "stepfunctions"
-        )
-        self.can_report = task_token is not None and stepfunctions_client is not None
+        self.stepfunctions_client = stepfunctions_client
 
     def _dump_result(self, result: ResultModel | None) -> str:
         if result is not None:
@@ -38,11 +35,9 @@ class StepFunctionOutput:
     def send_success(self, result: ResultModel | None) -> None:
         output = self._dump_result(result)
 
-        if self.can_report:
-            assert self.task_token is not None
-            stepfunctions_client = boto3.client("stepfunctions")
+        if self.stepfunctions_client is not None and self.task_token is not None:
             print("Sending task success to Step Functions.")
-            stepfunctions_client.send_task_success(
+            self.stepfunctions_client.send_task_success(
                 taskToken=self.task_token,
                 output=output,
             )
@@ -58,8 +53,7 @@ class StepFunctionOutput:
             }
         )
 
-        if self.can_report:
-            assert self.task_token is not None
+        if self.stepfunctions_client is not None and self.task_token is not None:
             print(f"Sending task failure to Step Functions: {error_output}")
             self.stepfunctions_client.send_task_failure(
                 taskToken=self.task_token,


catalogue_graph/src/ingestor/steps/ingestor_loader.py
```diff
@@ -129,4 +191,15 @@ def local_handler() -> None:
 
 
 if __name__ == "__main__":
-    local_handler()
+    parser: ArgumentParser = ArgumentParser()
+    parser.add_argument(
+        "--use-cli",
+        action="store_true",
+        help="Whether to invoke the local CLI handler instead of the ECS handler.",
+    )
```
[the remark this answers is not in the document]
kenoir: I think my preference would be to keep it this way, we may want to feed in an event running locally, or indeed discover that is how we most often want to run it to re-run a failed event locally.

catalogue_graph/src/utils/steps.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class StepFunctionOutput:
+    def __init__(
+        self, task_token: str | None, stepfunctions_client: StepFunctionClient | None
+    ) -> None:
+        self.task_token = task_token
+        self.stepfunctions_client = stepfunctions_client or boto3.client(
+            "stepfunctions"
+        )
```
StepanBrychta: Why not instantiate the step functions client in the constructor instead of using dependency injection? That would simplify things (removing the need for the `StepFunctionClient` class) and it would be more consistent with the approach we use in other places. Unit tests should still work due to monkey patching.

catalogue_graph/src/utils/steps.py
```diff
@@ -0,0 +1,109 @@
...
+        error_output = json.dumps(
+            {
+                "message": str(error),
+                "type": type(error).__name__,
+                "stack": str(error.__traceback__),
+            }
+        )
+
+        if self.can_report:
+            assert self.task_token is not None
```
StepanBrychta: It might be simpler to directly check if the task token is defined instead of checking indirectly via the `self.can_report` boolean and then separately asserting than the task token is not None.

catalogue_graph/src/utils/steps.py
```diff
@@ -0,0 +1,109 @@
...
+        if result is not None:
+            return result.model_dump_json()
+        return "Result is None"
+
+    def send_success(self, result: ResultModel | None) -> None:
+        output = self._dump_result(result)
+
+        if self.can_report:
+            assert self.task_token is not None
+            stepfunctions_client = boto3.client("stepfunctions")
```
StepanBrychta: I think we can remove this and use the client defined on the object (`self.stepfunctions_client`).

catalogue_graph/src/utils/steps.py
kenoir: these asserts are to please mypy, but it's just as much code to do the check twice, so I will remove these

PR #3098, opened 2025-11-05
Graph pipeline metrics

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6161

* Introduce a shared way of publishing CloudWatch metrics corresponding to pipeline events. Metrics to be published are declared on individual `PipelineReport` models and published along with the S3 report as part of the `publish` method.
* Increase the number of CloudWatch metrics we collect as part of each pipeline run:
  * For bulk loader events, collect the number of processed records, the number of duplicate records, and the number of errors (parsing, insert, and data type errors)
  * For incremental graph remover events, collect the number of deleted nodes/edges
  * For ingestor events, collect parquet file sizes and the number of indexed records
  * For ingestor deletion events, collect the number of deleted records
* Improve bulk load poller typing and add unit tests

## How to test

Run a relevant service (bulk load poller, incremental graph remover, etc.) locally and check that metrics are published to CloudWatch.

* Reports and relevant metrics should be published by all of the following services:
  * Bulk load poller (only once the bulk load completes)
  * Incremental graph remover
  * Ingestor loader
  * Ingestor indexer
  * Ingestor deletions

This PR refactors/consolidates how we create S3 URIs, which risks breaking existing functionality. Unit tests were expanded to mitigate this.


kenoir (approved):
This is great.

I wonder whether we should think about surfacing reports in Grafana using something like: https://grafana.com/grafana/plugins/simpod-json-datasource/

PR #3099, opened 2025-11-05
Replace print statements with logging library

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6135

Set up structlog to be used in python projects 

## How to test

```
AWS_PROFILE=platform-developer uv run extractor.py --is-local --transformer-type mesh_locations --entity-type nodes --stream-destination local --sample-size 5
```

Better logging 🤷‍♂️

### DIFF
--- a/pipeline/terraform/modules/pipeline/graph/ecs_task_extractor.tf
+++ b/pipeline/terraform/modules/pipeline/graph/ecs_task_extractor.tf
@@ -8,7 +8,7 @@ module "extractor_ecs_task" {
   environment = {
     CATALOGUE_GRAPH_S3_BUCKET   = data.aws_s3_bucket.catalogue_graph_bucket.bucket
     GRAPH_QUERIES_SNS_TOPIC_ARN = module.catalogue_graph_queries_topic.arn
-    LOG_LEVEL                  = "INFO" # Set log level to INFO for testing, WARNING later
+    LOG_LEVEL                   = "INFO" # Set log level to INFO for testing, WARNING later
   }
 
   cpu    = 4096


catalogue_graph/src/extractor.py
```diff
@@ -22,47 +29,90 @@ def raw_event(raw_input: str) -> ExtractorEvent:
...
+                pass
+
+        else:
+            logger.error(
+                "Unsupported stream destination",
+                stream_destination=event.stream_destination,
+            )
             raise ValueError(
-                "To stream to SNS, the GRAPH_QUERIES_SNS_TOPIC_ARN environment variable must be defined."
+                f"Unsupported stream destination: {event.stream_destination}"
```
StepanBrychta: I don't think we need to explicitly log an error just before raising it — raised errors are automatically sent to stderr, so I think this would result in the same error being logged twice.

catalogue_graph/src/utils/logger.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Args:
+        trace_id: Unique execution identifier provided by caller
+        is_local: Whether this is running in local development mode
+
+    Returns:
+        Logger with execution context bound
+    """
+    # Set default log level based on environment
+    setup_structlog(log_level="INFO" if is_local else "WARNING")
```
kenoir: Can this use some external config (like an env var) rather than having to pass a param in?

agnesgaroux: That must have been a leftover, the level is set from env on l.27. Removed

catalogue_graph/src/utils/logger.py
```diff
@@ -0,0 +1,125 @@
...
+        name: Optional logger name (defaults to calling module)
+
+    Returns:
+        Logger that automatically includes bound context
+    """
+    return cast(structlog.stdlib.BoundLogger, structlog.get_logger(name))
+
+
+def setup_logging(
+    context: ExecutionContext, is_local: bool = False
```
kenoir: is_local is not used

catalogue_graph/src/utils/logger.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Get a structlog logger with automatic context from contextvars.
+
+    Args:
+        name: Optional logger name (defaults to calling module)
+
+    Returns:
+        Logger that automatically includes bound context
+    """
+    return cast(structlog.stdlib.BoundLogger, structlog.get_logger(name))
```
kenoir: i'm not sure this needs its own function? Why not do this at the top of the file where logging is used?

catalogue_graph/src/extractor.py
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env python
 
+"""
```
kenoir: Should this comment be here?

kenoir:
In this first pass i'd not to try generalise the use of the new logging approach too much and just try and get it working in a single service - I would test via manual deployment first and then add the results of that test to this PR so we can catch any issues before we merge to main, or adopt a particular way to use structlog.

PR #3119, opened 2025-11-13
Update EBSCO adapter to tolerate missing 001 field, look in 035

kenoir:
## What does this change?

- Fixes the EBSCO loader crash observed on 12 Nov 2025 when records lacked a MARC 001 controlfield by deriving the identifier from alternative fields.
- Introduces `extract_record_id`, which prefers controlfield 001, falls back to datafield 035 subfield `a`, strips the parenthetical OCLC prefix, and raises a clearer error if nothing usable is found.
- Adds targeted unit tests around the new identifier extraction behaviour to guard against regressions.

## How to test

- From `/Users/kennyr/workspace/catalogue-pipeline/catalogue_graph`, run `uv run pytest -q tests/adapters/ebsco/test_loader.py`.

- The EBSCO adapter loader should no longer abort when 001 is missing; monitor the next scheduled run for the absence of the previous exception.
- Record counts emitted to Iceberg should remain steady; discrepancies would indicate unexpected identifier substitutions.
- CloudWatch alarms for the loader Lambda should remain green after deployment.

- Some records may have multiple 035 values; we currently take the first usable entry, which could select a suboptimal identifier. If that appears in the logs we may need to refine the selection logic.
- If neither field carries a suitable identifier we still raise, so the deployment must retain observability to catch new failure modes quickly.

### DIFF
--- a/catalogue_graph/tests/adapters/ebsco/test_loader.py
+++ b/catalogue_graph/tests/adapters/ebsco/test_loader.py
@@ -59,8 +59,6 @@ class TestRecordIdExtraction:
         assert extract_record_id(node) == "9999999"
 
     def test_raises_when_no_identifier_present(self) -> None:
-        node = etree.fromstring("<record></record>")
-
         with pytest.raises(
             Exception, match="Could not find controlfield 001 or usable datafield 035"
         ):
@@ -74,7 +72,7 @@ class TestRecordIdExtraction:
                 </record>
                 """
             )
-        assert extract_record_id(node)
+            extract_record_id(node)
 
     def test_raises_when_no_identifier_fields_present(self) -> None:
         node = etree.fromstring("<record></record>")


catalogue_graph/tests/adapters/ebsco/test_loader.py
```diff
@@ -1,17 +1,59 @@
...
+                <datafield tag="035">
+                    <subfield code="a">(OCoLC)814782</subfield>
+                </datafield>
+            </record>
+            """
+        )
+
+        assert extract_record_id(node) == "814782"
+
+    def test_raises_when_no_identifier_present(self) -> None:
```
paul-butcher: Some other scenarios to exercise - when the fields are present but empty or invalid.

```suggestion
    def test_ignores_controlfield_001_when_empty(self) -> None:
        node = etree.fromstring(
            """
            <record>
                <controlfield tag="001"></controlfield>
                <datafield tag="035">
                    <subfield code="a">(OCoLC)9999999</subfield>
                </datafield>
            </record>
            """
        )
        assert extract_record_id(node) == "9999999"

    def test_raises_when_no_identifier_present(self) -> None:

        with pytest.raises(
                Exception, match="Could not find controlfield 001 or usable datafield 035"
        ):
            node = etree.fromstring(
                """
                <record>
                    <controlfield tag="001">   </controlfield>
                    <datafield tag="035">
                        <subfield code="a"> (OCoLC) </subfield>
                    </datafield>
                </record>
                """
            )
        assert extract_record_id(node)

    def test_raises_when_no_identifier_fields_present(self) -> None:
```

PR #3126, opened 2025-11-17
Skeleton Axiell Collections adapter 

kenoir:
## What does this change?

- Adds the complete Axiell OAI-PMH adapter, including trigger, loader, and transformer steps, plus the supporting config, clients, table helpers, and state models needed for end-to-end harvesting.
- Extracts reusable Iceberg, window-harvesting, and schemata utilities so both the Axiell and EBSCO adapters share the same tested implementations; updates the existing EBSCO steps to use the new helpers.
- **Test Refactoring**: Replaced stubbed Iceberg table implementations in tests with real, temporary local Iceberg tables (via `temporary_table` fixture). This ensures tests validate actual Iceberg interactions, including incremental updates and snapshot management.
- **Deleted Records**: Implemented and tested handling for OAI-PMH deleted records. Records with missing metadata are now correctly processed as soft deletes (content set to `null`) in the Iceberg table.
- Introduces targeted unit tests (trigger/loader/transformer and shared utils), refreshed developer docs, and exploratory notebooks to exercise the adapter locally, resolving wellcomecollection/platform#6199.

[Axiell Collection adapter README](https://github.com/wellcomecollection/catalogue-pipeline/blob/rk/add-axc-adapter/catalogue_graph/src/adapters/axiell/README.md)

## Future work

- Infrastructure wiring (Lambda definitions, EventBridge rules, etc.)
- Use a common framework for new style python transformers
- The transformer should write success / failure manifests to S3

## How to test

```bash
cd catalogue_graph
uv run pytest -q tests/adapters/axiell
uv run pytest -q tests/adapters/utils
uv run pytest -q tests/adapters/ebsco
```

These suites cover the new adapter pipeline as well as the refactored shared utilities and EBSCO regression surface.
The Axiell tests now run against real local Iceberg tables, verifying the full data lifecycle including soft deletes.

- Axiell trigger → loader → transformer runs can be simulated locally with the provided scripts/notebooks, demonstrating that changesets move from Iceberg into Elasticsearch without manual intervention.
- Shared utilities back both Axiell and EBSCO adapters without regressions, giving us a single implementation for Iceberg/window bookkeeping and reducing the maintenance surface.
- Tests and documentation provide a clear runbook for enabling the adapter once the infrastructure layer lands.

- Refactoring the shared harvesting/Iceberg helpers could introduce regressions in the long-running EBSCO adapter; targeted regression tests mitigate this risk but we should monitor its next deployment closely.
- The new adapter assumes the presence of specific Iceberg catalogs and Elasticsearch indices; until the infrastructure change lands, running these Lambdas in AWS will fail fast, so this branch should not be deployed as-is.
- Elasticsearch indexing remains a dummy transform; downstream consumers must be aware that this data is solely for smoke-testing until the real mapping is implemented.

### DIFF
--- a/catalogue_graph/src/adapters/ebsco/steps/trigger.py
+++ b/catalogue_graph/src/adapters/ebsco/steps/trigger.py
@@ -9,7 +9,7 @@ import argparse
 import re
 import tempfile
 from datetime import datetime
-from typing import Any, cast
+from typing import Any
 
 import boto3
 import smart_open
@@ -26,17 +26,11 @@ from adapters.ebsco.models.step_events import (
     EbscoAdapterTriggerEvent,
 )
 from models.events import EventBridgeScheduledEvent
+from utils.aws import get_ssm_parameter
 
 
 class EbscoAdapterTriggerConfig(BaseModel):
     is_local: bool = False
-
-
-def _get_ssm_parameter(parameter_name: str) -> str:
-    """Returns an AWS SSM parameter string associated with a given name."""
-    ssm_client = boto3.Session().client("ssm")
-    response = ssm_client.get_parameter(Name=parameter_name, WithDecryption=True)
-    return cast(str, response["Parameter"]["Value"])
 
 
 def _list_s3_keys(bucket: str, prefix: str) -> list[str]:
@@ -134,10 +128,10 @@ def handler(
 
     job_id = event.job_id
 
-    ftp_server = _get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_server")
-    ftp_username = _get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_username")
-    ftp_password = _get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_password")
-    ftp_remote_dir = _get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_remote_dir")
+    ftp_server = get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_server")
+    ftp_username = get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_username")
+    ftp_password = get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_password")
+    ftp_remote_dir = get_ssm_parameter(f"{SSM_PARAM_PREFIX}/ftp_remote_dir")
 
     with (
         EbscoFtp(ftp_server, ftp_username, ftp_password, ftp_remote_dir) as ebsco_ftp,


catalogue_graph/tests/adapters/axiell/test_trigger.py
```diff
@@ -0,0 +1,167 @@
...
+    monkeypatch.setattr(config, "WINDOW_LOOKBACK_DAYS", 1)
+    store = _populate_store(temporary_window_status_table, [])
+
+    request = trigger.build_window_request(store=store, now=now)
+
+    assert request.window_start == now - timedelta(days=1)
+    assert request.window_end == now
+    assert request.set_spec == config.OAI_SET_SPEC
+    assert request.metadata_prefix == config.OAI_METADATA_PREFIX
+    assert request.job_id
```
paul-butcher: Is it worth asserting that job_id looks as expected, or is it sufficient that  it is something?

catalogue_graph/tests/adapters/axiell/test_trigger.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def test_build_window_request_respects_last_success(
+    monkeypatch: MonkeyPatch,
+    temporary_window_status_table: IcebergTable,
+) -> None:
+    now = datetime(2025, 11, 17, 12, 15, tzinfo=UTC)
+    last_success_end = now - timedelta(minutes=15)
```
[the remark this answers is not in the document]
kenoir: I've had a think about this and I think we should change the behaviour so the boundaries are important and make it generate windows that align to consistent boundaries within the hour, i.e. 0,15,30,45. 

If we do so we will get the benefit that those windows become retryable as part of a larger window later. At present this will not be the case.

catalogue_graph/tests/adapters/axiell/test_loader.py
```diff
@@ -0,0 +1,334 @@
...
+        assert request.window_key == req.window_key
+        harvester = StubHarvester([summary])
+        harvester_holder["instance"] = harvester
+        return harvester
+
+    harvester_holder: dict[str, StubHarvester | None] = {"instance": None}
+    monkeypatch.setattr(loader, "_build_harvester", fake_build_harvester)
+
+    response = loader.execute_loader(req, runtime=runtime)
+
```
paul-butcher: This is how it might look using patch.object

```suggestion
   with patch.object(loader.WindowHarvestManager, 'harvest_recent',
                      return_value=[summary]) as patched_harvest_manager:
        response = loader.execute_loader(req, runtime=runtime)
        patched_harvest_manager.assert_called_once_with(start_time=req.window_start, end_time=req.window_end,
                                                        max_windows=req.max_windows,
                                                        reprocess_successful_windows=False)
```

PR #3131, opened 2025-11-24
Extract transform mets createdate

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/wellcomecollection.org/issues/12464
This extends the MetsXml trait to extract `CREATEDATE` from the mets' header

To minimise confusion, the record's date (used to assert whether the record is the most recent iteration) has been changed from `createdDate` to `modifiedDate`. The METS file `CREATEDATE` uses `createdDate` in its stead. 

`CREATEDATE` is not in a consistent date format, some formatting is performed before instantiating the `InvisibleMetsData` 

## How to test

~~Commit 36856f605957de8efa78fd6563829103f4d28c23 will be reverted before merging and contains files to run the transformer locally:~~
~~- start with `docker compose -f docker-compose.local.yml up` and aws sso login~~
~~- ./setup_elasticsearch.sh~~
~~- ./run_mets_transformer.sh~~
~~- ./send_test_messages.sh~~

From the root of the project:
```
eval $(AWS_PROFILE=storage-developer aws configure export-credentials --format env) && scripts/run_local.sh transformer/transformer_mets dev
```
then from pipeline/transformer/transformer_mets/run_local
```
./send_test_messages.sh
```
You can try with different source records, but you need to ensure they exist in `wellcomecollection-storage-staging` 

Inspect the documents at http://localhost:5601/

`works-source` documents include `createdDate` in the items' DigitalLocation if they're v1. 
The rest of the document is unchanged
v2 METS do not include `createdDate`

This is a new property on the document, which will not be used to perform any pipeline operation (eg. versioning, merging...) 
The items's Location makes it through the pipeline untouched, being simply transferred onto the visible work at the Merger stage.

### DIFF
--- a/pipeline/transformer/transformer_mets/src/main/scala/weco/pipeline/transformer/mets/transformer/MetsData.scala
+++ b/pipeline/transformer/transformer_mets/src/main/scala/weco/pipeline/transformer/mets/transformer/MetsData.scala
@@ -147,17 +147,23 @@ object InvisibleMetsData {
     }
 
     val createdDate = if (version == 1) {
-      root.createdDate.map { dateStr =>
-        // Parse and reformat to ensure consistent format
-        // Try parsing with timezone first, if that fails assume UTC
-        val zonedDateTime = scala.util.Try {
-          ZonedDateTime.parse(dateStr, DateTimeFormatter.ISO_DATE_TIME)
-        }.getOrElse {
-          // No timezone info, parse as LocalDateTime and assume UTC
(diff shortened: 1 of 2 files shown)


agnesgaroux:
> `./scripts/run_local.sh` defines a pattern for running scala services locally, this PR adds some duplication of that functionality and does it a bit differently. Can we reuse the top level run local pattern?

Looking at this but the run_local script was designed for containerised lambdas so slightly different for the transformers

PR #3142, opened 2025-12-02
Add helper deployment scripts for local development

kenoir:
## What does this change?

Adds two helper shell scripts to the `catalogue_graph/scripts/` directory to streamline local development and testing workflows:

- **deploy_lambda.sh**: Builds and deploys the Lambda Docker image for testing
- **deploy_task.sh**: Builds and deploys the Task Docker image for testing

## Why are we making this change?

These scripts have been useful during local development for quick iteration and testing. By adding them to the repository, they can benefit other developers who need to:

- Quickly build and deploy Lambda functions for testing without going through CI/CD
- Test Task container changes rapidly during development
- Iterate on changes without waiting for full pipeline deployments

Both scripts:
- Use the `platform-developer` AWS profile
- Build for `linux/amd64` platform
- Tag images with `:dev` suffix to distinguish from production deployments
- Handle ECR authentication for both public and private registries
- Read Python version from `.python-version` for consistency

## How are we making this change?

Added two new bash scripts:

1. [catalogue_graph/scripts/deploy_lambda.sh](catalogue_graph/scripts/deploy_lambda.sh) - takes a Lambda function name as an argument and updates that specific Lambda with a freshly built image
2. [catalogue_graph/scripts/deploy_task.sh](catalogue_graph/scripts/deploy_task.sh) - builds and pushes the Task image to ECR with the `:dev` tag

Both scripts:
- Set proper bash error handling (`set -e`, `set -o pipefail`, `set -u`)
- Navigate to the `catalogue_graph` directory automatically
- Use `docker buildx` for consistent multi-platform builds

### DIFF
--- a/.github/agents/PR.agent.md
+++ b/.github/agents/PR.agent.md
@@ -11,8 +11,11 @@ Follow this template: https://github.com/wellcomecollection/.github/blob/main/PU
 
 If the user specifies any PRs that this one resolves or is related to retrieve them for context, include them in the PR description.
 
+Set `GH_PROMPT_DISABLED` to `true` in the environment to disable prompt injection when using the GitHub CLI (gh) to create, view or update PRs.
+
 Create or update the PR using the GitHub Copilot Pull Request extension, or the GitHub CLI if the extension is not available.
 
-You will need to create the PR description in markdown format, in a temporary file, and then use that file to create or update the PR.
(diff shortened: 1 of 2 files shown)


paul-butcher (approved):
Handy.

Is it worth making them more DRY? Apart from unified_pipeline_task vs unified_pipeline_lambda (which could be parameterised), they are identical before the call to update-function-code.

Probably not for now, but I can see them being reusable in different projects.

PR #3147, opened 2025-12-02
Fix terraform churn issue

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6225

Depends on https://github.com/wellcomecollection/terraform-aws-secrets/pull/9.

Fix an issue which caused terraform to display a large diff whenever we made a small change in the `elastic` module in the `pipeline` stack.

This issue was caused by the following lines in the `transformers` module, which caused _all_ resources created as part of the module to be recreated whenever _any_ change was made in the `elastic` module.

```
 depends_on = [
    # This updates the API key.  If the transformer is not restarted, then the
    # running container will continue using the old, defunct key, and will fail.
    module.elastic
  ]
```

The lines above were replaced with a different mechanism for redeploying ECS tasks whenever the `transformer` API key is updated, which involves adding the version of the Secrets Manager secret as an environment variable.

## How to test

* Make a change in the elastic module (e.g. modify an index mapping or change the permissions on a non-transformer API key). Run terraform plan. Does it only list the resources you changed?
* Change the permissions of the `transformer`  API key in the elastic module. Run terraform plan. Does it list all transformer ECS tasks?

Terraform diff should only list resources with changes.

This change is low risk.


kenoir:
Can you use `replace_triggered_by` in a lifecycle argument instead: https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#replace_triggered_by?

PR #3161, opened 2025-12-09
Add reporting and align adapter events for AxC

kenoir:
## What does this change?

- Align adapter step events on a shared `BaseAdapterEvent`, including AxC/EBSCO, and switch the AxC loader event to carry an `IncrementalWindow` rather than start/end fields.
- Update the AxC loader/trigger/reloader flows to use the new window shape, surface changed-record totals in the response, and publish an AxC loader report after harvests (supporting the telemetry/reporting work for platform#6216).
- Relax `PipelineReport` so reports can include a window and optionally skip S3 publication; adjust ingestor indexer expectations accordingly.
- Comment out the SNS publish in `sierra_adapter/build_missing_windows.py` while still printing missing windows for inspection.

## How to test

- `uv run pytest . --no-cov` (from `catalogue_graph`).

- Loader/reporting outputs now include consistent job/window context, enabling clearer metrics and downstream alerting for AxC harvests.
- Adapter step events across AxC and EBSCO share the same schema, reducing duplication and easing future changes.

- Event schema changes could break any external callers still emitting `window_start`/`window_end`; verify Step Function inputs are updated in lockstep.
- New loader reporting introduces another dependency point; misconfiguration could prevent report publication or metrics emission.

### DIFF
--- a/catalogue_graph/src/adapters/axiell/reporting.py
+++ b/catalogue_graph/src/adapters/axiell/reporting.py
@@ -3,14 +3,10 @@ from __future__ import annotations
 from typing import ClassVar
 
 from adapters.axiell.models.step_events import AxiellAdapterLoaderEvent, LoaderResponse
-from models.events import IncrementalWindow
 from utils.reporting import PipelineMetric, PipelineReport
 
 
 class AxiellReport(PipelineReport):
-    window: IncrementalWindow
-    publish_to_s3: bool = False
-
     @property
     def metric_dimensions(self) -> dict:
         return {


catalogue_graph/infra/adapters/axiell/iam_policies.tf
```diff
@@ -95,6 +95,23 @@ data "aws_iam_policy_document" "ssm_read" {
...
+    actions = [
+      "cloudwatch:PutMetricData"
+    ]
+
+    resources = ["*"]
+
+    condition {
+      test     = "StringEquals"
+      variable = "cloudwatch:namespace"
+      values   = ["catalogue_graph_pipeline"]
```
StepanBrychta: Maybe we could make the CloudWatch namespace into a configurable property of the reporting classes so that the adapters can publish their metrics into a different namespace? (Non-blocking, we can do this later.)

PR #3162, opened 2025-12-09
Fix EBSCO adapter running

kenoir:
This change fixes the EBSCO adapter, which had been inadvertently disabled, hiding another permissions issue.

Tested & deployed in the current pipeline.

### DIFF
--- a/catalogue_graph/src/adapters/axiell/helpers.py
+++ b/catalogue_graph/src/adapters/axiell/helpers.py
@@ -11,12 +11,15 @@ from adapters.utils.window_store import (
 )
 
 
-def build_adapter_table(use_rest_api_table: bool) -> IcebergTable:
+def build_adapter_table(
+    use_rest_api_table: bool, create_if_not_exists: bool = True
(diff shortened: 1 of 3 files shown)


catalogue_graph/src/adapters/ebsco/steps/transformer.py
```diff
@@ -310,7 +311,10 @@ def handler(
     # Determine index date from config override or fallback to pipeline date
     index_date = config_obj.index_date or config_obj.pipeline_date
 
-    table = helpers.build_adapter_table(config_obj.use_rest_api_table)
+    table = helpers.build_adapter_table(
+        use_rest_api_table=config_obj.use_rest_api_table,
+        create_if_not_exists=False,
```
agnesgaroux: Is it set to False here because in production we never want to create if not exist?

kenoir: The transformer running in production shouldn't (and doesn't have permissions to) create the table, yes.

PR #3178, opened 2026-01-12
Deploy axiell transformer

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6248

Deploy the Axiell transformer, configured to write to an experimental index and to _not_ send records to the ID minter.

* Create a new works source index (`works-source-2026-01-12`) to store works processed by the Axiell transformer.
* Create a new Elasticsearch API key (called `transformer_axiell`), which only has the permission to write to the new index. Configure the Axiell transformer to use this API key.
* Configure the transformer state machine to skip the ID minter map step unless the transformer type is `ebsco`. 
* Give the transformer Lambda the necessary IAM permissions to access the Axiell S3 bucket/table, mirroring EBSCO permissions.
* Create state machine triggers to run the Axiell transformer, mirroring EBSCO. 

Once we are ready to start writing to the production works source index, we should:
* Remove the `transformer_axiell` API key.
* Change Axiell config to:
```python
INDEX_DATE = os.getenv("INDEX_DATE")
ES_API_KEY_NAME = os.getenv("ES_API_KEY_NAME", "transformer")
```

## How to test

* Run the transformer locally. When running with `--transformer-type=ebsco`, the index name should resolve to `works-source-<PIPELINE_DATE>`. When running with `--transformer-type=axiell`, it should resolve to a hardcoded value of `works-source-2026-01-12`. `ES_API_KEY_NAME` should resolve to `transformer_axiell`.
* Apply the terraform changes and deploy the transformer Lambda. Then:
     * Check that the transformer state machine includes the `ShouldRunIdMinter` step. 
     * Run the transformer state machine on an EBSCO changeset, verifying that the ID minter is triggered. 
     * Run the Axiell adapter, verifying that the transformer is triggered and that the ID minter is _not_ triggered.

Items processed via the Axiell transformer should be written to the non-production `works-source-2026-01-12` index. The ID minter should not be invoked. 

* The Axiell transformer writes records to the production `works-source` index: The API key used by the Axiell transformer (defined under `ES_API_KEY_NAME`) doesn't have the permission to do this.
* The Axiell transformer sends records to the ID minter: The `IdMinterMap` step is skipped unless `transformer_type=ebsco`.

### DIFF
--- a/pipeline/terraform/modules/pipeline/iam.tf
+++ b/pipeline/terraform/modules/pipeline/iam.tf
@@ -54,7 +54,7 @@ data "aws_iam_policy_document" "adapter_s3tables_read" {
 
 data "aws_iam_policy_document" "read_ebsco_transformer_pipeline_storage_secrets" {
   statement {
-    actions   = ["secretsmanager:GetSecretValue"]
+    actions = ["secretsmanager:GetSecretValue"]
     resources = [
       "arn:aws:secretsmanager:eu-west-1:760097843905:secret:elasticsearch/pipeline_storage_${var.pipeline_date}/private_host*",
       "arn:aws:secretsmanager:eu-west-1:760097843905:secret:elasticsearch/pipeline_storage_${var.pipeline_date}/port*",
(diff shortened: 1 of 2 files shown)


kenoir (approved):
Looks good.

PR #3185, opened 2026-01-15
Graph integration tests

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6103

Add catalogue graph integration tests, running production queries against the graph and verifying that the returned results are as expected. Integration tests are part of the `graph_pipeline` project, reusing its setup and utility functions. Unlike the project's unit tests, they are located in a separate 'integration' folder and are excluded from regular test runs using a pytest marker.

Each test is associated with a fixture which stores expected results for 20 randomly selected IDs. For example, the fixture associated with the `WORK_ANCESTORS_QUERY` test maps the IDs of 20 random works to their expected ancestor IDs. The contents of this fixture are then compared against results returned from the production graph. If the number of matches is below `MIN_MATCH_RATIO` (currently set to 0.9), the test fails. The `MIN_MATCH_RATIO` is not set to 1 to account for the fact that the graph is not static (works/concepts can be deleted), and so over time a small divergence is expected.

Outstanding work (to be completed in a future PR):
* Deploy integration tests as a GitHub action
* Set up a notifications system reporting integration tests failures

## How to test

* Run `uv run pytest`. Check that integration tests are *not* included in the run.
* Run `AWS_PROFILE=platform-developer uv run pytest -m "integration"`. Check that *only* integration tests are included in the run.
* Make a small change to one of the fixture files (e.g. removing an expected related concept). Run the integration tests again, verifying that they succeed but that a warning is shown. 
* Make a larger change to one of the fixture files (e.g. changing expected related concepts for 3 or more concepts). Run the integration tests again, verifying that they fail.

These tests have two goals:
* Detect when something goes wrong with upstream graph pipeline services (extractors, bulk loaders), causing incorrect relationships to be stored in the catalogue graph.
* Act as regression tests for our production Cypher queries.

For convenience, I included a `generate_fixtures.py` script, which regenerates all integration test fixtures using production Cypher queries. This comes with the assumption that the queries are correct. When a test fails, we might be tempted to regenerate the fixtures instead of investigating. To mitigate this risk, the script shows a warning message and requires confirmation.

### DIFF
--- a/catalogue_graph/integration/graph/generate_fixtures.py
+++ b/catalogue_graph/integration/graph/generate_fixtures.py
@@ -8,9 +8,9 @@ from __future__ import annotations
 
 import json
 import random
-from datetime import datetime, timezone
-from getpass import getuser
 from collections.abc import Callable
+from datetime import UTC, datetime
+from getpass import getuser
 from pathlib import Path
 from typing import Any
 
@@ -45,7 +45,7 @@ def append_regeneration_log(*, reason: str) -> None:
     fixtures_dir = Path(__file__).parent / "fixtures"
     path = fixtures_dir / REGENERATION_LOG_NAME
 
-    timestamp = datetime.now(timezone.utc).isoformat(timespec="seconds")
+    timestamp = datetime.now(UTC).isoformat(timespec="seconds")
     username = getuser()
 
     fixtures_dir.mkdir(parents=True, exist_ok=True)


agnesgaroux (approved):
Nice confidence boost when working on graphy query things! 
Would it be desirable to have a changelog of sorts, for whenever the fixtures are updated to explain and keep track of changes?

PR #3190, opened 2026-01-19
Deploy catalogue graph integration tests

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6103

Deploy catalogue graph integration tests as a GitHub action:
* Tests only run when changes are pushed to the main branch
* A notification is sent to our Slack alerts channel when the tests fail (using a webhook URL retrieved from AWS Secrets Manager)

Here is an example alert:
image

## How to test

I triggered a few test runs on this branch and the action seems to behave as expected, sending a Slack message in case of failure.

Catalogue graph integration tests should run whenever we merge a PR to main. If they fail, we should get a Slack notification.

Test failures should be rare and likely indicative of a serious issue, so the risk of the alerts channel being spammed with low-relevance messages is low.

### DIFF
--- a/catalogue_graph/src/adapters/axiell/steps/loader.py
+++ b/catalogue_graph/src/adapters/axiell/steps/loader.py
@@ -6,10 +6,10 @@ Iceberg, and emits a changeset identifier for the transformer step.
 
 import argparse
 import json
-import logging
 from datetime import datetime
 from typing import Any
 
+import structlog
 from oai_pmh_client.client import OAIClient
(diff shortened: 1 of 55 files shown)


agnesgaroux (commented):
I was thinking about the drift that will occur due to the evolving nature of the graph: this will be cumulative right? drifting little by little over time until we eventually/inevitably reach the threshold? 
Since the github workflow only runs on push to main I'm worried this might put us in a situation where we're blocked from deploying main until we figure out why the integration tests fail

PR #3197, opened 2026-01-22
Weco concept source

paul-butcher:
## What does this change?
https://github.com/wellcomecollection/platform/issues/6157

This promotes the former override facility into a source in its own right, with the highest priority for display values.

## How to test

```
uv run python src/extractor.py \
  --transformer-type=weco_concepts \
  --entity-type=nodes \
  --stream-destination=local \
   --is-local 
```
The resulting csv should contain all the data from what was the overrides csv.

A bit more control over the homonymous concepts.  A more consistent treatment of Wellcome-originated concepts.

Overriding was pretty obvious and explicit. The record would have fields overridden if they were in the list.

However, if the records being "overridden" don't match up, then we might not get the same results.


kenoir (approved):
One comment, related to modifying the source data to remove the need to do ID fiddling.

PR #3223, opened 2026-02-09
Provision identifiers-v2 RDS cluster for migration testing

kenoir:
## What does this change?

Provisions a new `identifiers-v2-serverless` Aurora MySQL Serverless v2 cluster, as part of the work to create a new ID minter database for migration testing ([wellcomecollection/platform#6269](https://github.com/wellcomecollection/platform/issues/6269)).

The new cluster has the same spec as the existing `identifiers-serverless` cluster:
- Aurora MySQL `8.0.mysql_aurora.3.08.2`
- Serverless v2 scaling: 0.5–16 ACUs
- Same VPC, subnet group, security groups, and credentials

Key differences from the existing cluster:
- **Starts empty** — no snapshot restore. Migration data will be loaded separately.
- `snapshot_identifier` is now optional in the `rds-serverless` module (defaulting to `null`), so new clusters can be created without restoring from a snapshot.

New Terraform outputs are added for the v2 cluster (`rds_v2_serverless_cluster_id`, `rds_v2_serverless_cluster_arn`) so downstream stacks can reference it when needed.

Backup and export infrastructure for the v2 database will follow in a later PR.

Follows #3222.

## How to test

```
cd infrastructure/critical
terraform init
terraform plan
```

Expect **8 new resources**: RDS cluster, instance, and 3 Secrets Manager entries (secret + version each). No changes to existing resources.

- `terraform apply` completes successfully
- The `identifiers-v2-serverless` cluster appears in the AWS Console under RDS with the correct engine version and scaling config
- Secrets Manager contains `rds/identifiers-v2-serverless/endpoint`, `rds/identifiers-v2-serverless/reader_endpoint`, and `rds/identifiers-v2-serverless/port`

- **No risk to existing cluster** — this adds a new, independent cluster with no changes to the existing `identifiers-serverless` resources
- The `snapshot_identifier` variable default change is backwards-compatible — the existing module call still passes the value explicitly
- The v2 cluster has `deletion_protection = true` (set in the module), preventing accidental deletion
- Cost is minimal — serverless v2 scales to 0.5 ACU when idle

### DIFF
--- a/builds/deploy_catalogue_pipeline.sh
+++ b/builds/deploy_catalogue_pipeline.sh
@@ -63,7 +63,7 @@ ENV_TAG="env.$PIPELINE_DATE" "$ROOT/builds/update_ecr_image_tag.sh" \
   uk.ac.wellcome/transformer_sierra \
   uk.ac.wellcome/transformer_tei \
   uk.ac.wellcome/unified_pipeline_lambda
-
+  
 if [[ "$TASK" == "tag_images_and_deploy_services" ]]
 then
   echo "Deploying ECS pipeline services to catalogue-$PIPELINE_DATE"
--- a/infrastructure/critical/RDS_BACKUP.md
+++ b/infrastructure/critical/RDS_BACKUP.md
@@ -44,12 +44,12 @@ Each export is encrypted with the `alias/id-minter-rds-export` KMS key.
 
 ## Manual trigger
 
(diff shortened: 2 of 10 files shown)


paul-butcher (approved):
I see this has already been applied  (tf plan says the infrastructure matches)

PR #3224, opened 2026-02-10
Move all transformers to use SQS-to-Step-Function bridge for ID minter

kenoir:
## What does this change?

This follows #3220, which introduced the SQS-to-Step-Function bridge for the ID minter but only wired it up for the Miro transformer.

This PR completes the migration by routing **all** transformer output through the bridge to the ID minter Step Function, instead of triggering the ID minter Lambda directly from SNS.

Specifically:
- **`service_transformers.tf`**: Updates `sqs_id_minter_bridge` to subscribe to all transformer output topics (`local.transformer_output_topic_arns`) instead of only Miro's.
- **`service_id_minter.tf`**: Removes the now-unused direct-invocation `id_minter_lambda` module and its associated `id_minter_output_topic` and `id_minter_environment_variables`, since all traffic now flows through the Step Function path.
- **`service_matcher.tf`**: Removes the reference to the deleted `id_minter_output_topic` from the matcher's input topic list, keeping only the Step Function output topic.
- **`sqs_id_minter_bridge/main.tf`**: Fixes a bug where a single-message batch caused `sourceIdentifiers` to be a string instead of an array (JSONata's `$map` returns a scalar for single items). Wrapped in `$append([], ...)` to guarantee array output.

### Added monitoring

As part of this change, CloudWatch alarms (aborted, failed, timed_out) have been added for:
- The **SQS-to-ID-minter bridge** state machine (`sqs_id_minter_bridge_state_machine_alarms`)
- The **transformer** state machine (`transformer_state_machine_alarms`) — which was previously missing alarms

Both publish to the chatbot topic, matching the pattern used by the graph pipeline state machines.

## How to test

1. Deploy to staging and run a reindex for a few sources (e.g. Sierra, CALM, TEI).
2. Verify that works flow through the pipeline end-to-end: transformer → SQS bridge → ID minter Step Function → matcher → merger → …
3. Confirm no messages are stuck in DLQs / no errors in CloudWatch logs for the ID minter or bridge.

- All transformer outputs are processed successfully through the Step Function bridge with no increase in errors or DLQ messages.
- The old direct-invocation ID minter Lambda is no longer receiving traffic (and can be confirmed removed in the AWS console after deploy).
- State machine failures are now visible via chatbot alerts.

- The SQS-to-Step-Function bridge was already validated with Miro traffic in #3220, so the mechanism is proven. This change extends it to all transformers.
- If issues arise, the commit can be reverted to restore the direct Lambda invocation path.
- During deploy, there may be a brief period where in-flight messages on the old SNS→Lambda path complete while the new bridge takes over. This is expected and safe since both paths ultimately feed the same downstream pipeline.

### DIFF
--- a/pipeline/terraform/modules/pipeline/state_machine_transformers.tf
+++ b/pipeline/terraform/modules/pipeline/state_machine_transformers.tf
@@ -187,6 +187,18 @@ module "transformer_state_machine" {
   }
 }
 
+module "transformer_state_machine_alarms" {
+  source = "../state_machine_alarms"
+
+  state_machine_arn = module.transformer_state_machine.state_machine_arn
+  alarm_name_prefix = "transformer-state-machine"
+  alarm_name_suffix = "-${var.pipeline_date}"
+
+  default_alarm_configuration = {
+    alarm_actions = [local.monitoring_infra["chatbot_topic_arn"]]
+  }
+}
+
 # Trigger State Machine on adapter completed events
 module "adapter_transformer_trigger" {
   for_each = local.transformer_types


kenoir:
This seems to have been working.

PR #3229, opened 2026-02-11
Id pregeneration

paul-butcher:
## What does this change?

https://github.com/wellcomecollection/catalogue-pipeline/issues/3218

This introduces the id pregenerator for the new (still unwritten) id minter.  It is a code-only change operating on an in-memory sqlite database for unit testing.

The interface/scheduling/infrastructure part of it will come later.

## How to test

There is nothing plumbed in.  Examine the unit tests, do they cover all the bases?

This shifts id generation away from the critical path of assigning ids to source ids.  This should allow us to write the id assignation part without the slightly confusing error handling surrounding id clashes.  It also means that if we reach a point in our usage of the id space where clashes start to become common, that won't impact the pipeline.

This is entirely unplugged, so the only risk is that we might have some unused code hanging around.

### DIFF
--- a/builds/terraform/github/gha_role_catalogue_graph.tf
+++ b/builds/terraform/github/gha_role_catalogue_graph.tf
@@ -58,6 +58,7 @@ data "aws_iam_policy_document" "gha_catalogue_graph_ci" {
       "arn:aws:lambda:eu-west-1:760097843905:function:catalogue-*",
       "arn:aws:lambda:eu-west-1:760097843905:function:ebsco-adapter-*",
       "arn:aws:lambda:eu-west-1:760097843905:function:axiell-adapter-*",
+      "arn:aws:lambda:eu-west-1:760097843905:function:folio-adapter-*",
     ]
   }
 
--- a/catalogue_graph/infra/adapters/folio/main.tf
+++ b/catalogue_graph/infra/adapters/folio/main.tf
@@ -166,7 +166,7 @@ resource "aws_scheduler_schedule" "folio_adapter_15_minute_run" {
 
   schedule_expression = "rate(15 minutes)"
   # Enable this to turn on regular harvest
-  state = "DISABLED"
+  state = "ENABLED"
 
   target {
     arn      = aws_sfn_state_machine.state_machine.arn


kenoir:
I think that we probably should report a metric of how many IDs got minted, plus if the 2nd try failed to push us over the expected threshold we should probably indicate that - perhaps in the first case, just with log messages (N IDs minted etc).

catalogue_graph/src/id_minter/identifiers.py
```diff
@@ -0,0 +1,40 @@
+"""
+Generates a series of Wellcome Canonical Identifiers.
+
+A canonical identifier is a unique identifier for anything in the Wellcome catalogue.
+
+Canonical IDs are chosen for the following properties:
+
+*   They should be **short**, ideally something that fits on a single post-it note
```
kenoir: This comment omits that the first character must not be a number so as to be XML identifier safe, although the logic below _does_ implement this.

Suggest updating the comment to match that in the RFC & original Scala implementation.

catalogue_graph/src/id_minter/pregenerate.py
```diff
@@ -0,0 +1,100 @@
...
+    assert isinstance(count, int)
+    return count
+
+
+def save_new_ids(conn: DBConnection, new_ids: Iterable[str]) -> None:
+    if new_ids:
+        cursor = conn.cursor()
+        cursor.executemany(
+            """
+            INSERT OR IGNORE INTO canonical_ids (CanonicalId, Status) VALUES (?, 'free')
```
kenoir: I don't think this will work with the MySQL database, this syntax only works for SQLite. I think perhaps we should be using the MySQL docker container to test at this point so we can avoid this problem.

catalogue_graph/src/id_minter/pregenerate.py
```diff
@@ -0,0 +1,100 @@
...
+    def executemany(self, q: str, args: list[tuple]) -> None: ...
+
+
+class DBConnection[T: DBCursor](Protocol):
+    def cursor(self) -> T: ...
+
+    def commit(self) -> None: ...
+
+
+def top_up_ids(conn: DBConnection, desired_count: int) -> None:
```
kenoir: Suggest you could do something like this to reduce the number of calls: 

```
def top_up_ids(conn: DBConnection, desired_count: int) -> None:
    shortfall = _get_id_shortfall(conn, desired_count)
    if not shortfall:
        return

    inserted = _add_new_ids(conn, shortfall)
    if inserted == shortfall:
        return  # no clashes, we're done — no re-check needed

    logger.info(f"{shortfall - inserted} id clashes on first attempt, retrying")
    shortfall -= inserted

    inserted = _add_new_ids(conn, shortfall)
    if inserted < shortfall:
        raise ShortfallError("nope")
```

Also updating `save_new_ids` to report the number of rows written (presuming `cursor.rowcount` returns the number of ids actually inserted):

```
def save_new_ids(conn: DBConnection, new_ids: Iterable[str]) -> int:
    cursor = conn.cursor()
    cursor.executemany(
        "INSERT OR IGNORE INTO canonical_ids (CanonicalId, Status) VALUES (?, 'free')",
        [(new_id,) for new_id in new_ids],
    )
    conn.commit()
    return cursor.rowcount
```

catalogue_graph/src/id_minter/identifiers.py
```diff
@@ -0,0 +1,40 @@
...
+
+FORBIDDEN_LETTERS = {"o", "i", "l"}
+
+NUMBER_RANGE = [str(n) for n in range(2, 10)]
+LETTER_RANGE = [c for c in string.ascii_lowercase if c not in FORBIDDEN_LETTERS]
+ALLOWED_CHARACTER_SET = LETTER_RANGE + NUMBER_RANGE
+IDENTIFIER_LENGTH = 8
+
+
+def generate_id() -> str:
```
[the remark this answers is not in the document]
paul-butcher: No reason - I think this was a point where I asked Copilot to translate the Scala and then fiddled a bit instead of writing it myself.

PR #3233, opened 2026-02-12
WeCo authority tweaks

StepanBrychta:
## What does this change?

Make a few tweaks to the WeCo authority extractor and to the concepts ingestor loader to correctly index WeCo authority overrides into the final concepts index.

## How to test

* Run the WeCo authority extractor/bulk loader locally and verify that the correct nodes and edges are added to the graph. SAME_AS edges should be bidirectional. 

For example, the concept `zbus63qt` should have four source concepts — LoC, MeSH, Wikidata, and WeCo:
```py
from utils.aws import get_neptune_client
from ingestor.queries.concept_queries import (
    SOURCE_CONCEPT_QUERY,
)

client = get_neptune_client(True)
result = client.run_open_cypher_query(SOURCE_CONCEPT_QUERY, {"ids": ["zbus63qt"]})
```

* Run the ingestor loader locally on an appropriate time window, verifying that outputted concepts have WeCo overrides for titles, descriptions, and images.

Indexed concepts have correct WeCo authority overrides for titles, descriptions, and images.

This change is relatively low-risk, attempting to fix functionality which is currently broken in production.

### DIFF
--- a/pipeline/terraform/modules/pipeline/service_id_minter.tf
+++ b/pipeline/terraform/modules/pipeline/service_id_minter.tf
@@ -1,10 +1,3 @@
-module "id_minter_output_topic" {
-  source = "../topic"
-
-  name       = "${local.namespace}_id_minter_output"
-  role_names = [module.id_minter_lambda.lambda_role_name]
-}
-
 module "id_minter_step_function_output_topic" {
   source = "../topic"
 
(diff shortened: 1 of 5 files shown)


catalogue_graph/src/transformers/weco_concepts/concepts_transformer.py
```diff
@@ -45,10 +45,15 @@ def extract_edges(self, raw_data: dict) -> Generator[SourceConceptSameAs]:
...
+
+        # lookup the id elsewhere and find the corresponding source concept
+        source_id = self._prefixed_id_of(raw_data)
+        concept_id = str(raw_data["id"].strip())
+        attributes = SourceConceptSameAsAttributes(source="weco-authority")
+
+        yield SourceConceptSameAs(
+            from_id=source_id, to_id=concept_id, attributes=attributes
+        )
         yield SourceConceptSameAs(
```
StepanBrychta: We need to yield both directions to make sure both of these queries work:

```
    MATCH (concept: Concept)-[:SAME_AS]->(direct_match)
```

```
    MATCH (concept: Concept)<-[:SAME_AS]-(direct_match)
```

PR #3245, opened 2026-02-19
Catalogue graph dev cluster

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6261
https://github.com/wellcomecollection/platform/issues/6276

Terraform changes:
* Add a new Neptune cluster (`catalogue-graph-dev`) for testing purposes
* Replace the production catalogue graph NLB with a [Neptune public endpoint](https://docs.aws.amazon.com/neptune/latest/userguide/neptune-public-endpoints.html), allowing us to use the same host no matter where we're accessing the graph from

Python changes:
* Remove LambdaNeptuneClient and LocalNeptuneClient classes, replacing them with a single NeptuneClient class
* Remove the `get_neptune_client` utility function and instantiate NeptuneClient directly
* Add a `--neptune-environment` argument to all relevant graph pipeline services, allowing us to choose between the `dev` cluster and the `prod` cluster when running locally (default is set to `dev`)
* Add new utility functions for creating commonly used ArgumentParser arguments to reduce code duplication
* Switch to using dependency injection for Elasticsearch and Neptune clients for more flexibility

## How to test

I tested all affected services by fully populating the dev graph and a local concepts index using this branch.

* A dev graph cluster is available for experimentation and local testing
* A simplified graph networking setup, with a single host and a single NeptuneClient class
* A refactored argument parsing implementation with less repeated code

* Graph cluster security: This change involves moving the graph cluster from private subnets to public subnets. This is not a significant change in practice, since the cluster is already exposed to the public internet via an NLB. The cluster still uses IAM authentication (which is mandatory for clusters with public access enabled). To improve our security posture further, we could restrict cluster access to GlobalProtect IP addresses.
* Egress data transfer costs: According to [AWS documentation](https://docs.aws.amazon.com/neptune/latest/userguide/neptune-public-endpoints.html), the publicly available cluster host still resolves to a private IP address when accessed from inside of the VPC, so there should be no changes in terms of data transfer costs.

### DIFF
--- a/catalogue_graph/infra/graph/modules/catalogue_graph/security_groups.tf
+++ b/catalogue_graph/infra/graph/modules/catalogue_graph/security_groups.tf
@@ -3,11 +3,13 @@ resource "aws_security_group" "neptune_security_group" {
   vpc_id = data.aws_vpc.vpc.id
 }
 
-# Allow any ingress traffic so that we can reach the cluster via its public endpoint.
+# Allow any ingress traffic on port 8182 so that we can reach the cluster via its public endpoint.
 resource "aws_vpc_security_group_ingress_rule" "neptune_ingress" {
   security_group_id = aws_security_group.neptune_security_group.id
   cidr_ipv4         = "0.0.0.0/0"
-  ip_protocol       = "-1"
+  ip_protocol       = "tcp"
+  from_port         = 8182
+  to_port           = 8182
 }
 
 # Allow any egress traffic. The Neptune cluster needs to be able to reach the bulk loader S3 bucket.


catalogue_graph/lambda.Dockerfile
```diff
@@ -19,6 +19,10 @@ RUN dnf install -y ca-certificates git && dnf clean all
 COPY certs/* /etc/pki/ca-trust/source/anchors/
 RUN update-ca-trust extract
 
+# Install pip-system-certs so Python uses the system CA store (including the Sectigo certs above).
+# This is installed separately from the main package to avoid affecting local development.
+RUN pip install pip-system-certs==5.3
```
kenoir: Nice, this should have been pinned 👌

PR #3312, opened 2026-03-31
Integrate image pipeline into graph pipeline

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6273

* Add images into the catalogue graph, using the augmented index as a source.
* Add a new image ingestor, replicating the existing Scala ingestor

These services are not deployed and are not yet production ready. Outstanding tasks are documented in [this](https://github.com/wellcomecollection/platform/issues/6331) ticket.

## How to test

* Run the extractor/bulk_loader services locally in the dev environment using the `catalogue_images` transformer type. Are images loaded to the graph as expected?
* Run the images ingestor in the dev environment. Does it produce valid image documents?

* Images are available in the graph, enabling new features (e.g. portraits on person theme pages)
* Images ingestor is aligned with works/concepts ingestor, eventually replacing the existing Scala service and reducing the amount of code we need to maintain

At this stage, this change is low risk, since it's not deployed yet. Before deployment, we should verify that the new images ingestor produces identical documents to the existing Scala ingestor.


pipeline/terraform/modules/pipeline/elastic/main.tf
```diff
@@ -200,6 +200,12 @@ locals {
       read  = [for idx in local.works_denormalised_list : idx.name]
       write = [for idx in local.works_indexed_list : idx.name]
     }
+    images_ingestor = {
+      read = concat([for idx in local.works_denormalised_list : idx.name], [
+        for idx in local.images_augmented_list : idx.name
+      ])
+      write = ["images-indexed-dev"]
```
StepanBrychta, line 207: Only allow writing to the dev index for now.

pipeline/terraform/modules/pipeline/elastic/main.tf
```diff
@@ -200,6 +200,12 @@ locals {
...
+    }
```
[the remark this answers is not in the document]
StepanBrychta: This is intentional (and I left a comment above highlighting this change). We don't want the experimental ingestor service to write to the production index.

catalogue_graph/src/ingestor/models/step_events.py
```diff
@@ -53,16 +53,19 @@ def get_s3_uri(self, file_name: str, file_format: str | None = None) -> str:
...
             index_date_concepts = args.index_date
+        elif args.ingestor_type == "images":
+            index_date_images = args.index_date
 
         index_dates = PipelineIndexDates(
             merged=args.index_date_merged,
             works=index_date_works,
             concepts=index_date_concepts,
+            images=index_date_images,
         )
```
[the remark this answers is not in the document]
StepanBrychta: Very strange, the suggested code change doesn't correspond to the comment (and suggests making changes which are already in place!). Anyway, the comment itself is valid and I implemented a fix.

catalogue_graph/src/ingestor/extractors/images/images_extractor.py
```diff
@@ -0,0 +1,65 @@
...
+            for es_image in es_batch:
+                work_id = es_image.source.id.canonical_id
+                if work_id in works_batch:
+                    yield ExtractedImage(image=es_image, work=works_batch[work_id])
+                else:
+                    logger.warning(
+                        "Image does not have a corresponding visible work",
+                        image_id=es_image.state.canonical_id,
+                    )
+
```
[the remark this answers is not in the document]
StepanBrychta: Yes, I forgot about unit tests. Fixed.

catalogue_graph/src/ingestor/models/display/image.py
```diff
@@ -0,0 +1,69 @@
...
+                return location
+
+        raise ValueError(f"No iiif-image (thumbnail) location found on image {self.id}")
+
+    @classmethod
+    def from_extracted_image(cls, extracted: ExtractedImage) -> "DisplayImage":
+        image = extracted.image
+        inferred_data = image.state.inferred_data
+
+        return DisplayImage(
```
kenoir, line 60: `DisplayImage.id` is being set to `image.source.id.canonical_id` (the parent work's canonical id), but the existing Scala ingestor sets it to the image's own canonical id (`image.id` → `state.canonicalId`, see `common/display_model/.../image/DisplayImage.scala`). 

With the bundled fixture (`tests/fixtures/ingestor/single_augmented_image.json`) this produces `display.id == "a24esypq"` instead of `"swms7qdx"`, breaking parity with the existing index. The new unit test in `test_images_transformer.py` (`test_display_image` asserts `result.id == MOCK_WORK_ID`) locks in the wrong value.

```suggestion
            id=image.state.canonical_id,
```
(and the matching test should assert `MOCK_IMAGE_ID`).

catalogue_graph/src/ingestor/extractors/works/image_works_extractor.py
```diff
@@ -0,0 +1,47 @@
...
+            event=self.event.copy(update={"window": None, "ids": None}),
+            query=get_works_by_id_query(list(work_ids)),
+            es_client=self.es_client,
+        )
+
+        works_stream = (MergedWork.from_raw_document(w) for w in es_source.stream_raw())
+        extracted_works = self.process_es_works(works_stream)
+        visible_extracted_works = [
+            w for w in extracted_works if isinstance(w, VisibleExtractedWork)
+        ]
```
kenoir, line 50: `ImageWorksExtractor` reuses `GraphBaseWorksExtractor.process_es_works`, which calls `_get_work_ancestors`, `_get_work_children` and `_get_work_concepts` for every batch of parent works. None of the image-side `from_extracted_image` factories (`DisplayImage`, `DisplayImageSource`, `ImageFilterableValues`, `ImageAggregatableValues`, `ImageVectorValues`) appears to read `hierarchy` or `concepts` — they only need `WorkData`. 

If that's right, every image batch is doing a pair of Neptune hierarchy queries (and a concepts extraction) we then throw away. 

Is it worth either skipping those calls for the images path or adding a comment on why we still need them?

kenoir (commented):
Mostly LGTM — one blocker (`DisplayImage.id` regression vs. the Scala ingestor) plus one efficiency question on `ImageWorksExtractor`. Nice refactor splitting `works_extractor` into `base/index/image` modules and adding the unit tests.

catalogue_graph/src/ingestor/extractors/works/image_works_extractor.py
StepanBrychta: I think we need both concepts and ancestors when constructing final image documents. We need the resolved concepts to get consistent concept labels in `ImageFilterableValues` (which I think is one of the main benefits of rewriting the ingestor in Python), and we need the ancestors to recreate the full collection path in `QueryWork` (which is a field on the `QueryImage` model).

catalogue_graph/src/ingestor/models/display/image.py
StepanBrychta: Yes, well spotted.

catalogue_graph/src/ingestor/extractors/works/image_works_extractor.py
StepanBrychta: I added a comment clarifying this.

kenoir (approved):

PR #3328, opened 2026-04-20
scan for source and optional predecessor to mint

agnesgaroux:
https://github.com/wellcomecollection/platform/issues/6326

## What does this change?

The id minter's embedder previously required predecessor identifiers to be passed in as an external `dict[SourceIdentifierKey, SourceIdentifierKey]` parameter to `process_work()`. This was a placeholder — the mechanism for deriving predecessors from the work JSON had not been designed.

This change implements predecessor discovery by scanning the document itself. Any node that has a `sourceIdentifier` can now optionally have a sibling `predecessorIdentifier`, and both are extracted together as a pair.

Key changes in `embedder.py`:
- `extract_source_identifiers` now returns `list[tuple[SourceIdentifierKey, SourceIdentifierKey | None]]` — each source identifier paired with its optional predecessor
- Removed the `predecessors` parameter from `process_work()` — predecessors are discovered from the document
- Uses `SourceIdentifier.model_validate()` from the pipeline models instead of manual dict access for the state-level source identifier
- Removed the dead `if not keys` early-return (state always has a `sourceIdentifier`)
- Simplified logging to use `SourceIdentifier.__str__()` via structlog

Test changes in `test_embedder.py`:
- Updated existing tests to match the new `(key, predecessor | None)` return type
- Added scan tests: node with predecessor, mixed nodes with/without predecessors
- Added process_work tests: nested item with predecessor, state with predecessor, multiple predecessors across document
- Test docs now include `state.sourceIdentifier` to mirror real data shape

## How to test

```bash
cd catalogue_graph
uv run pytest tests/id_minter/test_embedder.py -v
```

All 21 tests should pass.

When work documents with `predecessorIdentifier` fields (e.g. Axiell records replacing Sierra records) are processed, the predecessor's canonical ID should be inherited rather than a new one minted. This can be verified by checking the minting resolver logs for `method=inherited` entries.

- **Backward compatible**: Documents without `predecessorIdentifier` fields behave exactly as before (predecessor is `None`).
- **Type boundary**: `SourceIdentifierKey` (NamedTuple) is structurally compatible with `SourceId` (plain tuple) used in the DB layer. A `type: ignore[arg-type]` comment documents this at the boundary.
- **No new external dependencies**: Uses the existing `SourceIdentifier` pydantic model already in the codebase.

### DIFF
--- a/catalogue_graph/src/id_minter/embedder.py
+++ b/catalogue_graph/src/id_minter/embedder.py
@@ -129,8 +129,7 @@ def process_work(
         predecessor_count=predecessor_count,
     )
 
-    found = resolver.mint_ids(mint_requests)  # type: ignore[arg-type]  # SourceIdentifierKey is a tuple[str, str, str]
-
+    found = resolver.mint_ids(mint_requests)  # type: ignore[arg-type]  # list invariance: SourceIdentifierKey is a NamedTuple subtype of SourceId
     id_map: dict[SourceIdentifierKey, str] = {
         SourceIdentifierKey(*k): v for k, v in found.items()
     }
--- a/catalogue_graph/src/id_minter/embedder.py
+++ b/catalogue_graph/src/id_minter/embedder.py
@@ -124,7 +124,7 @@ def process_work(
 
     logger.info(
         "Embedding canonical IDs",
-        source_identifier=source_id,
+        source_identifier=str(source_id),
         source_identifier_count=len(mint_requests),
         predecessor_count=predecessor_count,
     )
@@ -139,7 +139,7 @@ def process_work(
 
     logger.info(
         "Finished embedding canonical IDs",
-        source_identifier=source_id,
+        source_identifier=str(source_id),
         ids_embedded=len(id_map),
     )
 
--- a/catalogue_graph/src/id_minter/embedder.py
+++ b/catalogue_graph/src/id_minter/embedder.py
@@ -14,9 +14,9 @@ import structlog
 from id_minter.models.identifier import (
     TYPES_NORMALIZED_TO_CONCEPT,
     IdResolver,
-    SourceId,
     SourceIdentifierKey,
 )
+from models.pipeline.identifier import SourceIdentifier
 
 logger = structlog.get_logger(__name__)
 
@@ -64,10 +64,17 @@ def make_key(source_identifier: dict) -> SourceIdentifierKey:
     )
 
 
-def extract_source_identifiers(work_json: dict) -> list[SourceIdentifierKey]:
+def extract_source_identifiers(
+    work_json: dict,
+) -> list[tuple[SourceIdentifierKey, SourceIdentifierKey | None]]:
     return [
-        make_key(node["sourceIdentifier"])
-        for node in scan(work_json, lambda d: "sourceIdentifier" in d)
+        (
+            make_key(node["sourceIdentifier"]),
+            make_key(node["predecessorIdentifier"])
+            if "predecessorIdentifier" in node
+            else None,
+        )
+        for node in scan(work_json, lambda node: "sourceIdentifier" in node)
     ]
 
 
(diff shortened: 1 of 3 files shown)


catalogue_graph/src/id_minter/embedder.py
```diff
@@ -64,10 +64,17 @@ def make_key(source_identifier: dict) -> SourceIdentifierKey:
...
-        make_key(node["sourceIdentifier"])
-        for node in scan(work_json, lambda d: "sourceIdentifier" in d)
+        (
+            make_key(node["sourceIdentifier"]),
+            make_key(node["predecessorIdentifier"])
+            if "predecessorIdentifier" in node
+            else None,
+        )
+        for node in scan(work_json, lambda node: "sourceIdentifier" in node)
     ]
```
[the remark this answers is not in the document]
agnesgaroux: that shouldn't ever be the case? 🤔

kenoir, line 78: If the same `sourceIdentifier` appears on more than one node in a document (e.g. it's mirrored on `state` and at the top level), we'll emit it once per node. Downstream, `_mint_ids_with_rollback` collapses these via `predecessors = {req[0]: req[1] for req in requests if req[1] is not None}` ([minting_resolver.py L170](https://github.com/wellcomecollection/catalogue-pipeline/blob/handle-predecessor-ids/catalogue_graph/src/id_minter/resolvers/minting_resolver.py#L170)), so if duplicated nodes ever disagreed on `predecessorIdentifier` (or one had `None`), the last non-`None` would silently win.

I don't think this can happen with our current shapes (the document carries a single canonical `(source, predecessor)` pair, mirrored on `state` but a short comment here noting that invariant would be good.

catalogue_graph/src/id_minter/embedder.py
```diff
@@ -109,43 +116,21 @@ def _add_canonical_id(node: dict) -> dict:
...
         predecessor_count=predecessor_count,
     )
 
-    if not keys:
-        logger.info(
-            "No source identifiers found, skipping",
-            source_identifier=source_id_str,
-        )
-        return work_json
+    found = resolver.mint_ids(mint_requests)  # type: ignore[arg-type]  # SourceIdentifierKey is a tuple[str, str, str]
```
kenoir, line 132: The comment is a little misleading — `SourceIdentifierKey` is a `NamedTuple` so it _is_ structurally a `tuple[str, str, str]`. The reason mypy complains is `list` invariance (`list[tuple[SourceIdentifierKey, ...]]` isn't a subtype of `list[tuple[SourceId, ...]]`). Suggest updating the comment so the next reader doesn't go looking for a nominal mismatch:

```suggestion
    found = resolver.mint_ids(mint_requests)  # type: ignore[arg-type]  # list invariance: SourceIdentifierKey is a NamedTuple subtype of SourceId
```

Alternative (no follow-up needed here): widen `IdResolver.mint_ids`'s parameter to `Sequence[tuple[SourceId, SourceId | None]]` in `id_minter/models/identifier.py` — that drops the ignore entirely. Up to you.

catalogue_graph/tests/id_minter/test_embedder.py
```diff
@@ -363,24 +389,141 @@ def test_with_predecessor_inheritance(self) -> None:
...
         resolver = FakeResolver(ids={(k_new[0], k_new[1], k_new[2]): cid})
 
-        result = process_work(doc, resolver, predecessors={k_new: k_pred})
+        result = process_work(doc, resolver)
 
-        assert result["canonicalId"] == cid
+        assert result["state"]["canonicalId"] == cid
         assert len(resolver.mint_calls) == 1
-        assert resolver.mint_calls[0][0][1] == (k_pred[0], k_pred[1], k_pred[2])
+        assert (k_new, k_pred) in resolver.mint_calls[0]
```
kenoir, line 404: [nit] `(k_new, k_pred) in resolver.mint_calls[0]` will silently pass if an unexpected extra request also slipped in. `set(resolver.mint_calls[0]) == {(k_new, k_pred)}` (and similarly for the other multi-request tests on L460-461 / L528-529) would be stricter. Not blocking.

kenoir (approved):
Nice, the new shape (sibling `predecessorIdentifier` discovered during scan) is much cleaner. Tests are thorough and pass locally; mypy is clean. A few small comments below, none blocking.

catalogue_graph/src/id_minter/embedder.py
agnesgaroux: I'd not considered the possibility but now that I think of it, is it not something we want to defend against? there should be a 1:1 mapping of source and predecessor. If we ever have nodes with the same sourceId but different predecessors, or different sourceIds with the same predecessor, that means something has gone Really Wrong, right? Is it actually a transformer concern, making sure that different sourceIds can't have the same predecessor? Could it happen across separate records, in which case, how do we find out?

catalogue_graph/tests/id_minter/test_embedder.py
agnesgaroux: Ironically, this exposed an example of your above comment 

> If the same sourceIdentifier appears on more than one node in a document

catalogue_graph/src/id_minter/embedder.py
kenoir: We could add a guard - I don't think it's blocking for this PR. Yes it would be best practise to raise an exception here if we see data that should be impossible.

catalogue_graph/src/id_minter/embedder.py
agnesgaroux: Went for the updated comment here, to avoid having to amend the type over many files

catalogue_graph/tests/id_minter/test_embedder.py
agnesgaroux: For the record, the shape of the test data was not something that would occur so I modified it to match real data

PR #3331, opened 2026-04-21
Adapter refactoring

StepanBrychta:
## What does this change?

Refactor adapter classes (mostly `WindowHarvestManager` and `WindowStore`), simplifying logic and reducing code duplication.

* `WindowStore` changes:
    * Consolidate duplicate PyArrow and Iceberg schemas
    * Consolidate duplicate logic in `list_in_range` and `load_status_map` methods
    * Replace custom `_normalize_row` utility with Arrow's `maps_as_pydicts` feature
    * Use `pa.Table.from_pylist` instead of `pa.Table.from_pydict`
    * Remove the `WindowStatusRecord` dataclass, use `WindowSummary` instead 

* `WindowHarvestManager` changes:
    * Make `client` and `record_callback` required in the constructor, since they are necessary for the class to function
    * Remove the `WindowKey` model and absorb its functionality into the `IncrementalWindow` model
    * Integrate the class with the `IncrementalWindow` model. Move UTC conversion logic to `IncrementalWindow` to avoid repeatedly calling `ensure_datetime_utc`.

## How to test

Run adapter services (loader, reloader) and verify that they behave as expected.

Simpler implementation, no changes in behaviour.

The code is covered by extensive unit tests, so risk is low.

### DIFF
--- a/catalogue_graph/src/adapters/utils/window_generator.py
+++ b/catalogue_graph/src/adapters/utils/window_generator.py
@@ -31,7 +31,9 @@ class WindowGenerator:
         self.window_minutes = window_minutes or DEFAULT_WINDOW_MINUTES
         self.allow_partial_final_window = allow_partial_final_window
 
-    def generate_windows(self, window: IncrementalWindow) -> list[IncrementalWindow]:
+    def generate_windows(
+        self, time_range: IncrementalWindow
+    ) -> list[IncrementalWindow]:
         """Generate aligned time windows between start_time and end_time.
 
         Windows are aligned to fixed boundaries based on ALIGNMENT_EPOCH and
@@ -39,7 +41,7 @@ class WindowGenerator:
         at :00, :15, :30, :45 of each hour.
 
         Args:
-            window: The incremental window defining the time range.
+            time_range: The incremental window defining the time range.
 
         Returns:
             List of IncrementalWindow objects.
@@ -47,11 +49,8 @@ class WindowGenerator:
         Raises:
             ValueError: If start_time >= end_time.
         """
-        start_time = window.start_time_utc
-        end_time = window.end_time_utc
-
-        if start_time >= end_time:
-            raise ValueError("start_time must be earlier than end_time")
+        start_time = time_range.start_time_utc
+        end_time = time_range.end_time_utc
 
         delta = timedelta(minutes=self.window_minutes)
         effective_end_time = end_time
--- a/catalogue_graph/src/adapters/utils/window_generator.py
+++ b/catalogue_graph/src/adapters/utils/window_generator.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from datetime import UTC, datetime, timedelta
 
-from utils.timezone import ensure_datetime_utc
+from models.incremental_window import IncrementalWindow
 
 logger = logging.getLogger(__name__)
 
@@ -31,9 +31,7 @@ class WindowGenerator:
         self.window_minutes = window_minutes or DEFAULT_WINDOW_MINUTES
         self.allow_partial_final_window = allow_partial_final_window
 
-    def generate_windows(
-        self, start_time: datetime, end_time: datetime
-    ) -> list[tuple[datetime, datetime]]:
+    def generate_windows(self, window: IncrementalWindow) -> list[IncrementalWindow]:
         """Generate aligned time windows between start_time and end_time.
 
         Windows are aligned to fixed boundaries based on ALIGNMENT_EPOCH and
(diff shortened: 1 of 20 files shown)


catalogue_graph/src/adapters/utils/window_harvester.py
```diff
@@ -135,98 +102,57 @@ def harvest_range(
...
+                set_spec=self.set_spec,
             )
-            if not callback and records_in_window:
-                raise RuntimeError(
-                    "A record callback must be supplied via record_callback to persist harvested records."
-                )
 
-            if callback:
-                records_with_ids: list[tuple[str, Record]] = []
-                for idx, record in enumerate(records_in_window, 1):
```
StepanBrychta, line 203: Was there a reason for using 1-based indexing? Python is 0-based, so I think keeping the error message 0-based is more intuitive.

catalogue_graph/src/adapters/utils/window_harvester.py
```diff
@@ -135,98 +102,57 @@ def harvest_range(
...
-        record_callback: WindowCallback | None = None,
-    ) -> WindowSummary:
-        start = ensure_datetime_utc(start)
-        end = ensure_datetime_utc(end)
-        key = WindowKey.from_dates(start, end)
+    def process_window(self, window: IncrementalWindow) -> WindowSummary:
+        start = window.start_time_utc
+        end = window.end_time_utc
+        key = window.to_iso_string()
         attempts = 1
```
StepanBrychta, line 119: I noticed that `attempts` is always `1`. Was this supposed to be part of a retry mechanism?

catalogue_graph/src/adapters/utils/window_generator.py
```diff
@@ -32,29 +32,25 @@ def __init__(
...
         Raises:
             ValueError: If start_time >= end_time.
         """
-        start_time = ensure_datetime_utc(start_time)
-        end_time = ensure_datetime_utc(end_time)
-
-        if start_time >= end_time:
-            raise ValueError("start_time must be earlier than end_time")
+        start_time = time_range.start_time_utc
+        end_time = time_range.end_time_utc
```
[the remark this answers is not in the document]
StepanBrychta: IncrementalWindow already validates that `start_time > end_time`, but this check is slightly stricter, so I added it back.

catalogue_graph/src/adapters/utils/window_harvester.py
kenoir: I imagine not, 0 based is fine.

catalogue_graph/src/adapters/utils/window_summary.py
```diff
@@ -4,36 +4,11 @@
 from typing import Any
 
 from pydantic import BaseModel, computed_field, field_validator
-from pydantic_core import core_schema
 
+from models.events import IncrementalWindow
```
kenoir: nit: the rest of the PR imports `IncrementalWindow` from `models.incremental_window` directly. Routing through `models.events` here drags in the whole pipeline-event module (`SourceScope`, `config`, transformer types) just to build a window-key string, and is one rename away from a circular import. Could this use the direct import for consistency?

catalogue_graph/src/adapters/utils/window_harvester.py
```diff
@@ -277,9 +199,4 @@ def _record_identifier(
         )
 
     def _merge_tags(self, custom_tags: dict[str, str] | None) -> dict[str, str] | None:
```
kenoir: nit: the only caller now passes `custom_tags` as a `dict[str, str]` initialised to `{}`, never `None`. Could tighten the parameter to `dict[str, str]` to make the contract clearer.

catalogue_graph/src/models/incremental_window.py
```diff
@@ -62,6 +74,17 @@ def from_argparser(cls, args: argparse.Namespace) -> Self | None:
 
         return window
 
+    def to_iso_string(self) -> str:
+        return f"{self.start_time_utc.isoformat()}_{self.end_time_utc.isoformat()}"
+
+    @classmethod
+    def from_iso_string(cls, s: str) -> IncrementalWindow:
+        start_str, end_str = s.split("_")
```
kenoir: nit, fine to skip: `s.split("_", 1)` would be marginally more defensive and document that we expect exactly one separator.

kenoir (approved):
Nice cleanup — the consolidation of `WindowStatusRecord` into `WindowSummary` and the `IncrementalWindow`-based plumbing make the harvester noticeably easier to follow. A few small nits inline; none blocking.

PR #3332, opened 2026-04-21
Extract id-minter RDS into a reusable module and add a test instance

### LINKED ISSUE #6325: Provision identifiers-v3 cluster
infrastructure/critical/rds_id_minter.tf
- Add v3 cluster with in/out security groups
- set up migration tooling to load v2 into v3 when we start transforming axiell records with predecessor ids
NOTE: v2 is exported every day -> s3://wellcomecollection-platform-id-minter/exports/identifiers-v2-serverless/

kenoir:
Resolves https://github.com/wellcomecollection/platform/issues/6325. Required by https://github.com/wellcomecollection/platform/issues/6326 / wellcomecollection/catalogue-pipeline#3329.

## What does this change?

Refactors the id-minter Aurora cluster into a reusable module so multiple instances can co-exist, and stands up a `test` instance restored from a recent AWS Backup snapshot of production.

**Module extraction (`infrastructure/critical/modules/id-minter-rds/`):**
- Wraps the existing `rds-serverless` sub-module along with the subnet group and the two security groups that surround it.
- Exposes the previously-flat outputs (`rds_cluster_id`, `rds_cluster_arn`, `master_user_secret_arn`, `ingress_security_group_id`, `subnet_group_name`).
- `moved` blocks were used in the initial extraction commit and have since been removed — the existing prod cluster is unchanged in state.

**New variables:**
- `name_suffix` (default `""`) — appended to the cluster identifier, security group names, and the access SG `Name` tag so multiple instances don't collide on AWS-side names. The empty default keeps the existing prod resources byte-identical.
- `snapshot_identifier` (default `null`) — passed through to `aws_rds_cluster.snapshot_identifier` so a new instance can be restored from an existing snapshot. Only consulted on initial create; once provisioned, do **not** edit or remove this value (it would force replacement).

**New `test` instance (`rds_id_minter.tf`):**
- `module.id_minter_rds_test` with `name_suffix = "test"` and `snapshot_identifier` set to the AWS Backup recovery point taken on 2026-04-21 04:18 UTC+01:00.
- Produces `identifiers-v2-serverless-test` (cluster), `identifiers_v2_database_sg_test` and `pipeline_rds_v2_ingress_security_group_test` (security groups).

**Outputs (`outputs.tf`):**
- New map output `id_minter_rds` keyed by suffix (`prod`, `test`) — each value is an object exposing the same fields as before.
- Legacy flat outputs (`rds_v2_serverless_cluster_id`, etc.) are retained pointing at the prod instance, so existing downstream stacks that read this state via `terraform_remote_state` continue to work unchanged.

## Consuming the new outputs from #3329

PR #3329 currently has the RDS test database wiring commented out pending this change. To plug it in, that PR can read from the new map output via the existing `terraform_remote_state` pointer to this stack (the catalogue-pipeline pipeline stack already reads `data.terraform_remote_state.infra_critical.outputs.*`). Suggested pattern:

```hcl
locals {
  id_minter_rds_instances = data.terraform_remote_state.infra_critical.outputs.id_minter_rds

  rds_v2_test_config = {
    cluster_arn               = local.id_minter_rds_instances["test"].cluster_arn
    cluster_id                = local.id_minter_rds_instances["test"].cluster_id
    ingress_security_group_id = local.id_minter_rds_instances["test"].ingress_security_group_id
    master_user_secret_arn    = local.id_minter_rds_instances["test"].master_user_secret_arn
    subnet_group_name         = local.id_minter_rds_instances["test"].subnet_group_name
  }
}
```

The test Lambda then takes its DB endpoint / secret ARN from `local.rds_v2_test_config` in the same shape as the existing `local.rds_v2_config` (which can also be migrated to read from `id_minter_rds["prod"]` later, in a follow-up).

The remote-state output reference name on the consuming side will be whatever data source already exposes the `infrastructure/critical` outputs — the existing flat outputs continue to work, so #3329 can be merged in either order.

## How to test

- `terraform plan` in `infrastructure/critical/` against `main`: shows only the **additions** for the new `module.id_minter_rds_test.*` resources and the new `id_minter_rds` map output. The existing prod cluster, security groups, subnet group, and legacy outputs show **no diff**.
- After apply, confirm the new cluster is `available`:
  ```bash
  aws rds describe-db-clusters \
    --db-cluster-identifier identifiers-v2-serverless-test \
    --query 'DBClusters[0].{Status:Status,Endpoint:Endpoint}'
  ```
- Confirm the new outputs from this stack:
  ```bash
  cd infrastructure/critical && terraform output -json id_minter_rds | jq
  ```

The new `identifiers-v2-serverless-test` cluster comes up restored from the chosen snapshot, with the same schema/data as prod at that point in time. The pipeline stack in #3329 can then point its test id_minter at it without touching any prod infra.

- **Non-destructive to prod**: every change to the existing module call defaults to its previous behaviour (`name_suffix = ""`, `snapshot_identifier = null`). The plan shows zero changes against the existing prod cluster.
- **Snapshot value is sticky**: `snapshot_identifier` is read only at create time but tracked in state. Removing or changing the value on the `test` instance would force replacement of the cluster. The comment in the test module call calls this out; if we want to detach later we can either leave the literal in place forever or add `lifecycle { ignore_changes = [snapshot_identifier] }` inside `modules/rds-serverless/main.tf` (deferred — would touch a shared sub-module).
- **Cost**: the test cluster is Aurora Serverless v2 with the same scaling config as prod (min 0.5 ACU, max 16 ACU). It will scale low when idle but is not free.
- **Output compatibility**: legacy flat outputs are retained, so no downstream stack breaks. The new `id_minter_rds` map can be adopted incrementally; the legacy outputs can be removed in a later cleanup PR once consumers migrate.

### DIFF
--- a/infrastructure/critical/backup_id_minter.tf
+++ b/infrastructure/critical/backup_id_minter.tf
@@ -44,6 +44,6 @@ resource "aws_backup_selection" "id_minter" {
   plan_id      = aws_backup_plan.id_minter_daily.id
 
   resources = [
-    module.identifiers_v2_serverless_rds_cluster.rds_cluster_arn
+    module.id_minter_rds.rds_cluster_arn
   ]
 }
--- a/infrastructure/critical/modules/id-minter-rds/main.tf
+++ b/infrastructure/critical/modules/id-minter-rds/main.tf
@@ -1,3 +1,8 @@
+locals {
+  hyphen_suffix     = var.name_suffix == "" ? "" : "-${var.name_suffix}"
+  underscore_suffix = var.name_suffix == "" ? "" : "_${var.name_suffix}"
+}
+
 resource "aws_db_subnet_group" "default" {
   subnet_ids = var.private_subnet_ids
 }
@@ -5,7 +10,7 @@ resource "aws_db_subnet_group" "default" {
 resource "aws_security_group" "database_v2_sg" {
   description = "Allows connection to identifiers-v2 RDS instance via TCP and egress to the world"
   vpc_id      = var.vpc_id
-  name        = "identifiers_v2_database_sg"
+  name        = "identifiers_v2_database_sg${local.underscore_suffix}"
 
   ingress {
     protocol  = "tcp"
@@ -36,10 +41,12 @@ resource "aws_security_group" "database_v2_sg" {
 module "identifiers_v2_serverless_rds_cluster" {
   source = "../rds-serverless"
 
-  cluster_identifier = "identifiers-v2-serverless"
+  cluster_identifier = "identifiers-v2-serverless${local.hyphen_suffix}"
   database_name      = "identifiers"
   master_username    = var.master_username
   master_password    = null
+
+  snapshot_identifier = var.snapshot_identifier
 
   manage_master_user_password = true
 
@@ -52,7 +59,7 @@ module "identifiers_v2_serverless_rds_cluster" {
 }
 
 resource "aws_security_group" "rds_v2_ingress_security_group" {
-  name        = "pipeline_rds_v2_ingress_security_group"
+  name        = "pipeline_rds_v2_ingress_security_group${local.underscore_suffix}"
   description = "Allow traffic to identifiers-v2 rds database"
   vpc_id      = var.vpc_id
 
@@ -64,6 +71,6 @@ resource "aws_security_group" "rds_v2_ingress_security_group" {
   }
 
   tags = {
-    Name = "pipeline-rds-v2-access"
+    Name = "pipeline-rds-v2-access${local.hyphen_suffix}"
   }
 }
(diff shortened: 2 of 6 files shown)


agnesgaroux (approved):
There's this change when I plan 
` enable_http_endpoint                  = false -> true` 
on `"identifiers-v2-serverless-test"`
I think that's what we want for the data api to work but wondering why it's set to false atm

kenoir:
> There's this change when I plan ` enable_http_endpoint = false -> true` on `"identifiers-v2-serverless-test"` I think that's what we want for the data api to work but wondering why it's set to false atm

Also not sure - I think it might be related to the way terraform provisions the database, and this option is only available after the db has been created, but the initial apply doesn't realise that?

PR #3338, opened 2026-04-23
get bag from S3 on 307 from storage bag api

agnesgaroux:
## What does this change?

The METS adapter's `BagRetriever` was failing with `java.lang.Exception: Received error from storage service: 307 Temporary Redirect` when the storage service returned a 307 redirect for large bags. This happens when the bag response is too large for a standard HTTP response — the storage service uploads it to S3 and returns a presigned URL via a `307 Temporary Redirect` with a `Location` header.

This PR adds handling for the `307 TemporaryRedirect` case in `HttpBagRetriever.handleResponse`:

- Reads the `Location` header from the 307 response and follows the redirect to fetch the bag JSON from the presigned S3 URL.
- Uses a separate `redirectClient: HttpClient` (a plain Pekko HTTP client, no OAuth) to follow the redirect. This avoids leaking OAuth credentials to the S3 presigned URL.
- A single `PekkoHttpClient` instance is shared between the OAuth client and the redirect client (in `Main.scala`) to avoid creating duplicate connection pools.
- **Redirect URL validation**: the redirect is only followed if the URL starts with `https://wellcomecollection-storage-prod-large-response-cache.s3.eu-west-1.amazonaws.com/responses/` — any other redirect target is rejected with a clear error.
- **Presigned URL redaction**: query parameters (which contain presigned URL credentials) are stripped from all log messages and error messages.
- **Connection pool hygiene**: response entities are discarded on all non-OK status paths to prevent Pekko HTTP connection pool leaks.
- Fails with a clear error if the 307 response is missing a `Location` header.

See Slack thread for context: https://wellcome.slack.com/archives/C02ANCYL90E/p1776938470259489

## How to test

Unit tests in `BagRetrieverTest` cover the new behaviour:

- **Successful redirect**: a 307 redirect is followed to fetch the bag from the presigned URL.
- **Missing Location header**: a 307 without a `Location` header produces a clear error.
- **Unexpected redirect URL**: a 307 pointing to an unexpected host is rejected.

All `mets_adapter` tests pass:

```
sbt "project mets_adapter" "test"
```

Large bags that previously caused `307 Temporary Redirect` errors in the METS adapter will now be fetched successfully. We should see the related errors disappear from logs after deployment.

- The redirect client is a plain HTTP client with no authentication — this is intentional since the redirect target is a presigned S3 URL that already contains its own auth. Sending an OAuth token to S3 would be unnecessary and a credential leak risk.
- Redirect URLs are validated against an expected S3 bucket prefix before following, so the adapter won't chase arbitrary redirects.
- The redirect is only followed for `307 TemporaryRedirect`; other redirect status codes continue to be treated as errors.
- Sharing a single `PekkoHttpClient` between the OAuth wrapper and the redirect client avoids resource duplication without changing behaviour — `PekkoHttpClient` is stateless and thread-safe.

### DIFF
--- a/mets_adapter/mets_adapter/src/test/scala/weco/pipeline/mets_adapter/services/BagRetrieverTest.scala
+++ b/mets_adapter/mets_adapter/src/test/scala/weco/pipeline/mets_adapter/services/BagRetrieverTest.scala
@@ -209,7 +209,7 @@ class BagRetrieverTest
   }
 
   it("follows a 307 redirect to fetch the bag from S3") {
-    val redirectUri = Uri("https://s3.example.com/large-bag-response.json")
+    val redirectUri = Uri("https://wellcomecollection-storage-prod-large-response-cache.s3.eu-west-1.amazonaws.com/responses/digitised/b30414726/v1")
     val bagJson =
       """
         |{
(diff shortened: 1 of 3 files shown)


mets_adapter/mets_adapter/src/test/scala/weco/pipeline/mets_adapter/services/BagRetrieverTest.scala
```diff
@@ -207,15 +208,164 @@ class BagRetrieverTest
     }
```
kenoir, line 208: The new tests cover the happy path, the unexpected-URL path, and the missing-`Location` path — nice. The one branch that isn't exercised is the `Received error following redirect to …: $status` failure inside `followRedirect` (e.g. S3 returns 403 because the presigned URL has expired). Worth adding a small test for that so the redacted error message is also locked in by the test suite?

kenoir (approved):
LGTM — nice clean fix. One optional follow-up below for test coverage of the redirect-error branch; not blocking.

PR #3341, opened 2026-04-27
id_minter: batch mint_ids across works and coalesce per-row INSERTs

### LINKED ISSUE #6328: Address ID minter performance issues
The new ID minter Lambda tends to time out when processing a window containing a large number of identifiers (100,000+). See https://wellcome.slack.com/archives/CQ720BG02/p1776652587007929 for more info.

Possible fixes include (but are not limited to):
* Parallelising the ID minting Python code (at the moment, the code is single threaded, processing one work document at a time)
* Reducing the size of each window 
* Switching to an ECS task to remove the 15 minute limit

kenoir:
Resolves https://github.com/wellcomecollection/platform/issues/6328.

## What does this change?

Two layered batching changes to the ID minter to reduce database round trips, which were the dominant cost when processing large windows (100,000+ identifiers) and were causing the Lambda to time out:

1. **Batch `resolver.mint_ids` across works in a chunk** (`id_minting_transformer.py`). The transformer now groups works into chunks (default 500, controlled by `mint_batch_size`) and issues a single `resolver.mint_ids` call — one DB transaction — for the combined set of source identifiers, instead of one transaction per work. Each work's embedding step still runs individually using a shared `id_map` of the minted canonical IDs.
   - If the batched mint call raises (e.g. one work has a missing predecessor), the transformer logs a warning and falls back to per-work minting for that chunk, so a single bad work cannot poison its neighbours.
   - Refactored `embedder` so that `extract_source_identifiers` and `embed_canonical_ids` are usable independently of the existing per-work `process_work` helper, which is retained for the fallback path.

2. **Coalesce per-row INSERTs into multi-row INSERTs** (`resolvers/minting_resolver.py`). Inside `_mint_ids_with_rollback`, the predecessor-inheritance branch (Step 3) and the new-ID branch (Step 5) each used to execute one `INSERT ... ON DUPLICATE KEY UPDATE` per source identifier. Each is now a single multi-row `INSERT ... VALUES (...), (...), ...` statement — same number of rows written, but only one network round trip per branch.
   - Per-row semantics are preserved: `ON DUPLICATE KEY UPDATE CanonicalId = CanonicalId` keeps the statement idempotent under concurrency, and the existing `FOR SHARE` re-read in Step 6 still detects lost races so only canonical IDs we actually won are marked `assigned` in Step 7.

No behaviour changes are intended for callers; `mint_batch_size` defaults preserve current semantics for small inputs, and the structured-log error contract on individual works is unchanged.

## How to test

From `catalogue_graph/`:

```bash
uv run pytest tests/id_minter/test_id_minting_transformer.py tests/id_minter/test_minting_resolver.py
uv run mypy src/id_minter tests/id_minter
uv run ruff format src/id_minter tests/id_minter
uv run ruff check --fix src/id_minter tests/id_minter
```

New tests cover:

- `test_id_minting_transformer.py`: that a chunk of works results in a single `mint_ids` call, that the resulting `id_map` is correctly distributed to each work's embedding, and that a failing batch falls back to per-work minting without dropping the surviving works.
- `test_minting_resolver.py` (`TestRoundTripBatching`): wraps `cursor.execute` to count `INSERT` statements and asserts a single round trip in the new-only, inheritance-only, and mixed cases, plus that the resulting rows are actually persisted in the `identifiers` table.

End-to-end, deploy to a staging stack and replay a large reindex window (100k+ identifiers). The job should complete inside the Lambda's 15 minute limit; check the CloudWatch metric `success_count` matches the expected work count and `failure_count` stays at zero.

- ID minter Lambda no longer times out on 100k+ identifier windows (compare against the timeouts referenced in #6328 and the linked Slack thread).
- Significant reduction in `mint_ids` wall time per batch — visible in the existing structured `Batch minted canonical IDs` log line and in CloudWatch Lambda duration for the id-minter step.
- `failure_count` metric remains zero across reindexes; no change in alarm rate.

- **Larger transactions.** A failed batched mint rolls back up to `mint_batch_size` works' worth of inserts. The fallback path then reprocesses those works one-by-one, so no data is lost; the only cost is the extra round trips for that chunk. The default of 500 was chosen to keep the transaction footprint modest while still amortising the per-call overhead.
- **Concurrency / lost-update races.** The multi-row `INSERT ... ON DUPLICATE KEY UPDATE` keeps the same idempotent, race-tolerant semantics as the previous per-row inserts, and Step 6's `FOR SHARE` re-read is unchanged, so race detection in Step 7 still applies.
- **Error attribution.** Per-work errors still flow through `_add_error` with the original source identifier as `row_id`, so structured logs and the `failure_count` metric continue to attribute failures to the correct work.
- **Memory.** Holding a chunk of raw works plus the combined `mint_requests` list in memory is bounded by `mint_batch_size`; for the default of 500 works this is well within the Lambda's memory envelope.

## Benchmark

A standalone script, [`scripts/bench_id_minter.py`](scripts/bench_id_minter.py), benchmarks the id_minter transformer + resolver across multiple git refs against a local Docker MySQL and prints a markdown table. It uses `git worktree` so the active branch is never touched, and re-invokes itself inside each worktree's `uv` venv so the timed code is loaded from that ref's `src/`. See the script docstring for full usage.

Reproduce these results:

```bash
python scripts/bench_id_minter.py \
    --refs main 3c2de53cb 4e2a5ca1f \
    --iterations 5 --include-noop
```

Latest run (5 iterations/cell, Python 3.13.5 on arm64, MySQL 8 in Docker on localhost):

| Workload | main `e735671fb` (med ms) | + commit 1 `3c2de53cb` (med ms) | Δ% | + commit 2 `4e2a5ca1f` (med ms) | Δ% | 
|---|---|---|---|---|---|
| all-new | 7349.5 | 2822.4 | -61.6% | 425.6 | **-94.2%** | 
| all-existing | 1405.4 | 219.9 | -84.4% | 216.7 | **-84.6%** | 
| mixed | 6798.9 | 2610.6 | -61.6% | 394.8 | **-94.2%** | 
| large-fanout | 5590.3 | 4538.7 | -18.8% | 661.2 | **-88.2%** | 

Production caveats: this benchmark excludes ES indexing (timed only `transform()`, not `stream_to_index`), and loopback MySQL has lower RTT than RDS in-VPC. Net Lambda-level speedup will likely be smaller in % terms than this table suggests, but the absolute round-trip savings should be larger.


catalogue_graph/src/id_minter/id_minting_transformer.py
```diff
@@ -44,12 +77,67 @@ def transform(self, raw_nodes: Iterable[Any]) -> Generator[tuple[str, dict]]:
...
+            return
+
+        combined_requests: list[
+            tuple[SourceIdentifierKey, SourceIdentifierKey | None]
+        ] = []
+        for _, _, mint_requests in works_in_batch:
+            combined_requests.extend(mint_requests)
+
+        try:
+            found = self.resolver.mint_ids(combined_requests)  # type: ignore[arg-type]  # SourceIdentifierKey is a NamedTuple subtype of SourceId
```
kenoir: `combined_requests` flattens identifiers from up to `mint_batch_size` works into one list. Inside `_mint_ids_with_rollback`, `predecessors = {req[0]: req[1] for req in requests if req[1] is not None}` is keyed by `source_id`, so if two works in the same chunk happen to mention the same source identifier with *different* `predecessorIdentifier` values (or one with a predecessor and one without), the dict-comprehension will silently keep just one of them.

The pre-existing comment in `_mint_ids_with_rollback` ("upstream guarantees a 1:1 mapping between sourceIdentifier and predecessorIdentifier, so this is safe") used to apply *within* a single work's transaction; this change extends that invariant to "across all works in a chunk". Is that already guaranteed by the source data (e.g. a sub-identifier shared by two works always carries the same predecessor info), or is it a new assumption worth a defensive check / log line?

catalogue_graph/src/id_minter/steps/id_minter.py
```diff
@@ -179,6 +179,7 @@ def handler(
             success_count=success_count,
             failure_count=failure_count,
         ).publish()
+
```
kenoir: Looks like a stray whitespace addition unrelated to the batching change — easy to drop if you are rebasing anyway.

kenoir (commented):
Reviewing my own change pre-emptively here.

catalogue_graph/src/id_minter/id_minting_transformer.py
```diff
@@ -3,38 +3,71 @@
...
     ):
         super().__init__()
         self.source = minting_source
         self.resolver = resolver
+        self.mint_batch_size = mint_batch_size
 
     def transform(self, raw_nodes: Iterable[Any]) -> Generator[tuple[str, dict]]:
-        for raw_doc in raw_nodes:
+        for chunk in batched(raw_nodes, self.mint_batch_size):
+            yield from self._transform_chunk(chunk)
```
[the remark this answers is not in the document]
kenoir: Leaving as-is: `itertools.batched` already raises a clear `ValueError` the first time `transform` runs, so a misconfigured value fails loudly without per-value extra validation. Per the repo's `AGENTS.md` ("Don't add error handling for scenarios that can't happen. Only validate at system boundaries"), the constructor isn't really an external boundary here — config validation happens upstream where the value enters the system.

catalogue_graph/src/id_minter/id_minting_transformer.py
```diff
@@ -3,38 +3,71 @@
...
+
+
+# Per-work data carried through the batched mint pipeline:
+# (row_id, raw_doc, mint_requests) where mint_requests is the list returned by
+# extract_source_identifiers(raw_doc).
+_WorkEntry = tuple[
+    str,
+    dict,
+    list[tuple[SourceIdentifierKey, SourceIdentifierKey | None]],
+]
```
StepanBrychta: Minor/non-blocking: This type (`tuple[SourceIdentifierKey, SourceIdentifierKey | None]`) is used in several places across the ID minter. It might be good to give it a name (e.g. `IdentifierWithPredecessor` or `IdentifierPredecessorPair`, or similar) so that it's easier to parse.

catalogue_graph/src/id_minter/id_minting_transformer.py
```diff
@@ -44,12 +77,89 @@ def transform(self, raw_nodes: Iterable[Any]) -> Generator[tuple[str, dict]]:
...
+                    " falling back to per-work minting",
+                    source_id=str(key),
+                    works_in_batch=len(works_in_batch),
+                )
+                yield from self._transform_per_work(works_in_batch)
+                return
+            seen_predecessors[key] = predecessor
+
+        try:
+            found = self.resolver.mint_ids(combined_requests)  # type: ignore[arg-type]  # SourceIdentifierKey is a NamedTuple subtype of SourceId
```
StepanBrychta: Minor and not necessarily in scope for this PR, but is there a good reason for having both a `SourceId` tuple and a `SourceIdentifierKey` named tuple? It looks like it might be cleaner to only have `SourceIdentifierKey`. It would also us to remove the `# type: ignore[arg-type]`.

catalogue_graph/src/id_minter/id_minting_transformer.py
```diff
@@ -44,12 +77,89 @@ def transform(self, raw_nodes: Iterable[Any]) -> Generator[tuple[str, dict]]:
...
+            if key in seen_predecessors and seen_predecessors[key] != predecessor:
+                logger.warning(
+                    "Conflicting predecessors for source id across works in chunk;"
+                    " falling back to per-work minting",
+                    source_id=str(key),
+                    works_in_batch=len(works_in_batch),
+                )
+                yield from self._transform_per_work(works_in_batch)
+                return
+            seen_predecessors[key] = predecessor
```
StepanBrychta: Since it's the resolver making the assumption about the 1:1 mapping, would it be better for the resolver to enforce it? 

The resolver could check that the mapping is 1:1 and throw an exception if it's not. Then, the transformer would automatically switch to processing works one by one due to the general error handling on line 120 below.

catalogue_graph/src/id_minter/id_minting_transformer.py
```diff
@@ -44,12 +77,89 @@ def transform(self, raw_nodes: Iterable[Any]) -> Generator[tuple[str, dict]]:
...
+                self._add_error(e, "embed", row_id)
+                continue
+            yield row_id, embedded
+
+    def _transform_per_work(
+        self, works_in_batch: list[_WorkEntry]
+    ) -> Generator[tuple[str, dict]]:
+        for row_id, raw_doc, _ in works_in_batch:
+            try:
+                embedded = process_work(raw_doc, self.resolver)
```
StepanBrychta: At the moment, there is a lot of duplication between the 'batch flow' (`_transform_chunk` method) and the 'single work flow' (`process_work` utility function). Both flows have a similar structure:
* extract source identifiers and build minting requests
* mint IDs via the resolver
* collect mapping between minted IDs and source IDs
* embed canonical IDs into work documents

This is a bit confusing, partially because in the batch flow, the `resolver.mint_ids` call is being made from this class, whereas in the single work flow, it's being made from the utility function. 

I think that with a bit of refactoring, it should be possible to remove most of the duplication between the two flows and have a single place which makes the `resolver.mint_ids` call.

catalogue_graph/src/id_minter/id_minting_transformer.py
kenoir: Good call — added a `MintRequest` type alias in `id_minter/models/identifier.py` and used it in `extract_source_identifiers`, the transformer's `_WorkEntry`, and `combined_requests` (788478804).

catalogue_graph/src/id_minter/id_minting_transformer.py
kenoir: Done in 2fe18e397 — dropped `SourceId` and unified everything (resolvers, transformer, embedder, tests) on `SourceIdentifierKey`. Both `# type: ignore[arg-type]` comments are gone.

catalogue_graph/src/id_minter/id_minting_transformer.py
kenoir: Done in af4612f8a — the 1:1 invariant is now enforced inside `MintingResolver._mint_ids_with_rollback` (raises `ValueError("Conflicting predecessors …")`). The transformer's existing batch-failure `except` turns that into per-work minting, so I dropped the dedicated pre-check loop. Added resolver-level tests for the new error path.

catalogue_graph/src/id_minter/id_minting_transformer.py
kenoir: Done in 0a61848dc. Both the batch path and the per-work fallback now go through a single `_mint_and_embed(works)` helper, so `resolver.mint_ids` is called from exactly one place in the transformer. The fallback just runs the same helper one work at a time. `embedder.process_work` (the standalone single-work flavour of the same logic) is gone — the embedder module is now just the primitives `extract_source_identifiers` / `embed_canonical_ids` / `scan` / `make_key`. The unique coverage process_work had (predecessor reaching the resolver as a `(sid, pred)` tuple) moved to a new transformer test.

StepanBrychta (approved):
Looks good, the new combined batch/single work minting flow is great

PR #3350, opened 2026-04-30
Elasticsearch bulk indexing tweaks

StepanBrychta:
## What does this change?

* Fix https://wellcome.slack.com/archives/CQ720BG02/p1777571115196479
* If some documents fail to index, do not raise an exception immediately. Wait until the end of the run so that other documents in the same window are processed.

## How to test

Run the ingestor_indexer service locally on a window which contains works whose source modified date is `1970-01-01T00:00:00Z`. For example:

```
AWS_PROFILE=platform-developer uv run ingestor/steps/ingestor_indexer.py --ingestor-type=works --pipeline-date=2025-10-02 --use-cli --index-date=2026-03-03 --window-end=2026-04-30T17:30 --es-mode=public --environment=prod
```

This issue is addressed: https://wellcome.slack.com/archives/CQ720BG02/p1777571115196479

This change should be low-risk.

### DIFF
--- a/catalogue_graph/src/core/transformer.py
+++ b/catalogue_graph/src/core/transformer.py
@@ -1,13 +1,15 @@
 from collections.abc import Generator, Iterable
 from itertools import batched
-from typing import Any, TypeVar, cast
+from typing import Any, TypeVar
 
-import elasticsearch.helpers
 import structlog
 from elasticsearch import Elasticsearch
 from pydantic import BaseModel
 
 from core.source import BaseSource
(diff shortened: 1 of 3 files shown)


kenoir (approved):
Merging this for now, as it will prevent errors. I think we need to go back and re-assess https://github.com/wellcomecollection/catalogue-pipeline/pull/3347, see https://wellcome.slack.com/archives/CQ720BG02/p1777571115196479

PR #3357, opened 2026-05-07
Consolidate OAI-PMH adapter entrypoints and restructure adapters

kenoir:
## What does this change?

Implements wellcomecollection/platform#6311 — consolidates adapter entrypoints for the Axiell and FOLIO OAI-PMH adapters.

Previously each adapter (`axiell/`, `folio/`, `ebsco/`) lived as a flat peer under `src/adapters/` with its own `steps/` module for Lambda entrypoints. The Axiell and FOLIO step files were near-identical thin wrappers that only differed in which config singleton they imported. This PR:

1. **Consolidates OAI-PMH entrypoints** — the trigger, loader, and reloader now use an `adapter_type` field from the Lambda event to resolve the correct `OAIPMHRuntimeConfig` via a registry, eliminating the per-adapter `steps/` wrappers.

2. **Restructures `src/adapters/` into three clear layers:**
   - `extractors/` — source-specific extraction logic (OAI-PMH framework, EBSCO FTP)
   - `transformers/` — MARC/EBSCO/Axiell transformation logic
   - `steps/` — all Lambda/CLI entrypoints, namespaced by adapter type (`steps/oai_pmh/`, `steps/ebsco/`, `steps/transformer.py`)

3. **Updates Terraform** to point Lambda handler commands at the new `adapters.steps.*` module paths. Adds a `steps_namespace` variable so axiell and folio modules both resolve to `adapters.steps.oai_pmh`.

4. **Fixes `WindowNotifier`** — the remediation CLI command was hardcoded to `--adapter-type axiell`; it now accepts `adapter_name` and uses it dynamically.

5. **Updates documentation** — READMEs, Copilot instructions, and CLI examples all reference the new paths.

184 files changed (mostly renames), net reduction of ~234 lines.

There will be another infrastructure PR that follows this change to move to using one state machine for both OAI-PMH adapters.

## How to test

1. Check out this branch and run the test suite:
   ```
   cd catalogue_graph && uv run pytest tests/adapters/ -q --tb=short
   ```
   All 559 adapter tests pass. mypy and ruff are clean (2 pre-existing mypy errors in `marcxml_transformer.py` only).

2. Verify Terraform plans cleanly for both adapter and pipeline infra:
   ```
   cd catalogue_graph/infra/adapters && terraform plan
   cd pipeline/terraform && terraform plan
   ```

3. After merge, deploy and verify that both the axiell and folio adapters trigger, load, and transform correctly via their scheduled runs.

- All adapter Lambda invocations continue to succeed after deploy (CloudWatch metrics / logs unchanged).
- No increase in error rate for axiell or folio adapters.
- Future OAI-PMH adapters only need a config + runtime module — no boilerplate `steps/` wrapper.

- **Terraform apply coordination required:** after merging, we will need to co-ordinate an adapter and pipeline terraform apply — the Lambda handler paths change in both `catalogue_graph/infra/adapters/` and `pipeline/terraform/`. These should be applied together to avoid handler-not-found errors.
- **No runtime behaviour change:** the consolidated entrypoints call the exact same handler logic with the same config objects. The only difference is _how_ the config is resolved (registry lookup from event vs. import-time singleton).
- **EBSCO adapter unchanged:** EBSCO's FTP-based trigger and file-based loader are fundamentally different from OAI-PMH, so they keep their own step files under `steps/ebsco/`.


catalogue_graph/tests/adapters/transformers/axiell/test_reconciler_transformer.py
```diff
@@ -45,7 +45,7 @@ def _run_reconciler(
         {"axiell_reconciler": lambda **kwargs: adapter_table},
     )
     monkeypatch.setattr(
-        "adapters.transformers.transformer.axiell_helpers.build_reconciler_table",
+        "adapters.transformers.transformer.AXIELL_CONFIG.build_reconciler_table",
         lambda **kwargs: reconciler_table,
     )
```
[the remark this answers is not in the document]
kenoir: Thanks for flagging — this is actually working as intended. `monkeypatch.setattr("a.b.c.OBJ.attr", val)` splits on the last dot, so it resolves `adapters.steps.transformer.AXIELL_CONFIG` via `getattr` on the module (not as a module import), then patches `build_reconciler_table` on that object. All 7 tests in this file pass. Resolving as no change needed.

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
```diff
@@ -241,21 +246,29 @@ def build_runtime(
...
-        Serialized OAIPMHLoaderEvent for the next step.
+    The ``time`` field is optional: when present (e.g. from an EventBridge
+    scheduled event) it is used as the reference timestamp; otherwise
+    ``datetime.now(UTC)`` is used.
     """
-    scheduled_event = EventBridgeScheduledEvent.model_validate(event)
-    event_time = datetime.fromisoformat(scheduled_event.time.replace("Z", "+00:00"))
+    adapter_type = event.get("adapter_type")
+    if adapter_type is None:
+        raise ValueError("Event must contain 'adapter_type'")
```
StepanBrychta, line 263: Optional: Maybe we could add `adapter_type` as a field to the `BaseAdapterEvent` class to validate this automatically. (And storing the adapter type on the Pydantic model representing the event might be useful in general.)

catalogue_graph/src/adapters/steps/oai_pmh/loader.py
```diff
@@ -285,3 +287,34 @@ def run_cli(
...
+    config = get_config(pre_args.adapter_type)
+    parser = build_cli_parser(config)
+    parser.add_argument(
+        "--adapter-type",
+        required=True,
+        choices=typing.get_args(AdapterType),
+        help="Which adapter to load",
+    )
+    args = parser.parse_args()
+    run_cli(config, args)
```
StepanBrychta: Do we need this two-stage parsing? The `pre_parser` is only used so that the adapter type can be included in the description of the main parser, which seems like a small benefit given the added complexity. 

(Same for the reloader service.)

catalogue_graph/src/adapters/extractors/oai_pmh/axiell/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Environment variables
+
+| Variable                      | Default       | Description                                |
+| ----------------------------- | ------------- | ------------------------------------------ |
+| `AXIELL_WINDOW_MINUTES`       | 15            | Harvesting window duration                 |
+| `AXIELL_WINDOW_LOOKBACK_DAYS` | 7             | Fallback range when no history             |
+| `AXIELL_MAX_LAG_MINUTES`      | 360           | Maximum allowed lag before circuit breaker |
+| `AXIELL_OAI_SET_SPEC`         | `collect`     | OAI set specification                      |
+| `AXIELL_OAI_METADATA_PREFIX`  | `oai_marcxml` | OAI metadata prefix                        |
```
StepanBrychta: I can't find these environment variables. The variables defined in `axiell/config.py` do not have the `AXIELL_` prefix.

Optional: I wonder if this table and the table in the `Configuration` section above could be joined into a single table, since they largely describe the same configuration options.

catalogue_graph/src/adapters/extractors/oai_pmh/registry.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_config(adapter_type: str) -> OAIPMHRuntimeConfig:
+    """Resolve an adapter_type string to its OAIPMHRuntimeConfig.
+
+    Raises:
+        ValueError: If the adapter_type is not recognised.
+    """
+    global _CONFIGS
+    if _CONFIGS is None:
+        _CONFIGS = _load_configs()
```
StepanBrychta: I think we can achieve the same result with the `@cache` decorator instead of keeping a global variable. We can keep the lazy import. Maybe something like:

```py
@cache
def get_config(adapter_type: AdapterType) -> OAIPMHRuntimeConfig:
    """Resolve an adapter_type string to its OAIPMHRuntimeConfig.

    Raises:
        ValueError: If the adapter_type is not recognised.
    """

    # Lazy imports to avoid heavyweight module-level side effects
    # (SSM lookups, HTTP client construction) when not needed.
    if adapter_type == "axiell":
        from adapters.extractors.oai_pmh.axiell.runtime import AXIELL_CONFIG
        return AXIELL_CONFIG
    if adapter_type == "folio":
        from adapters.extractors.oai_pmh.folio.runtime import FOLIO_CONFIG
        return FOLIO_CONFIG

    raise ValueError(
        f"Unknown adapter_type: {adapter_type!r}. "
        f"Expected one of: {get_args(AdapterType)}"
    )
```

StepanBrychta (approved):

catalogue_graph/src/adapters/extractors/oai_pmh/registry.py
StepanBrychta: Actually, do we need the lazy imports? I'm not seeing any expensive side effects at construct time. The SSM lookups only happen when we construct the client.

PR #3372, opened 2026-05-12
Predecessor validation tweaks

StepanBrychta:
## What does this change?

* **Transformer tweak**: When indexing transformed documents to the `works-source` index, drop nulls for parity with the Scala pipeline and with the Python ingestor service. Keeping fields will null values causes downstream issues in the ID minter.
* **ID minter tweak**: Temporarily deploy an ID generator Lambda for the test ID minter database (`identifiers-v2-serverless-test`). During my [Folio predecessor testing](https://github.com/wellcomecollection/platform/issues/6349), the ID pool in the testing database ran out of free IDs, and adding this Lambda allows for easy top ups.

### DIFF
--- a/catalogue_graph/src/core/transformer.py
+++ b/catalogue_graph/src/core/transformer.py
@@ -1,4 +1,3 @@
-import json
 from collections.abc import Generator, Iterable
 from itertools import batched
 from typing import Any, TypeVar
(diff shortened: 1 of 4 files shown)


agnesgaroux (approved):

catalogue_graph/src/core/transformer.py
```diff
@@ -1,3 +1,4 @@
+import json
```
agnesgaroux: Stray import

PR #3398, opened 2026-05-28
Wikidata source refactoring

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6379

Due to https://github.com/wellcomecollection/catalogue-pipeline/pull/3297, the Wikidata extractor service now filters out Wikidata nodes which do not have an English label. This causes an issue in the downstream Wikidata edge extractor, which assumes that no filtering occurs and outputs edges whose start nodes do not exist in the graph. 

The fix involves checking that the start node of each edge returned by the Wikidata API exists in the graph (by reading the bulk load file outputted by the Wikidata node extractor).

This PR also improves unit test coverage and refactors surrounding code to make it easier to understand:
* Split the logic in the `WikidataLinkedOntologySource` class into three classes, with separate subclasses dedicated to extracting nodes and edges
* Add a `WikidataBaseTransformer` superclass, selecting the correct source class at construction time based on the selected `entity_type`. 
* Move the `_parallelise_sparql_requests` function to the `WikidataSparqlClient` class

## How to test

Run the Wikidata edge/node extractor locally and compare outputs with existing production outputs. Outputted files should largely be the same, but the file outputted by the new edge extractor shouldn't include edges whose node IDs aren't outputted by the node extractor. 

The Wikidata extractor does not emit edges with missing IDs and is easier to maintain.

Since the Wikidata edge extractor now checks IDs outputted by the node extractor, it is dependent on it. This shouldn't be an issue, since the deployed state machine always runs the node extractor first.

### DIFF
--- a/catalogue_graph/tests/graph/sources/test_wikidata_concepts_source.py
+++ b/catalogue_graph/tests/graph/sources/test_wikidata_concepts_source.py
@@ -2,11 +2,16 @@ import json
 from typing import Literal
 
 from config import WIKIDATA_SPARQL_URL
-from graph.sources.wikidata.linked_ontology_source import WikidataLinkedOntologySource
+from graph.sources.wikidata.linked_ontology_edge_source import (
+    WikidataLinkedOntologyEdgeSource,
+)
+from graph.sources.wikidata.linked_ontology_node_source import (
+    WikidataLinkedOntologyNodeSource,
+)
 from models.events import ExtractorEvent
 from tests.mocks import MockRequest
 from tests.test_utils import add_mock_transformer_outputs_for_ontologies, load_fixture
-from utils.ontology import get_extracted_ids, is_id_in_ontology
+from utils.ontology import get_ids_for_transformer, is_id_extracted_for_ontology
 
 
 def _add_mock_wikidata_requests(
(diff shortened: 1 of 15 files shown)


catalogue_graph/tests/graph/sources/test_wikidata_concepts_source.py
```diff
@@ -55,37 +60,35 @@ def test_wikidata_concepts_source_edges() -> None:
...
 
     assert len(same_as_edges) == 2
     assert ("Q1", "sh00000001") in same_as_edges
     assert ("Q2", "sh00000001") in same_as_edges
 
-    assert len(has_parent_edges) == 3
-    assert ("Q1", "Q4") in has_parent_edges
+    assert len(has_parent_edges) == 1
     assert ("Q2", "Q1") in has_parent_edges
-    assert ("Q2", "Q3") in has_parent_edges
```
StepanBrychta, line 88: Two of the HAS_PARENT edges are no longer outputted since Q3 and Q4 are not included in the `wikidata_linked_mesh_concepts__nodes` fixture file, and so the edges get filtered out.

agnesgaroux (approved):
Ran 1000 edges on `main`: 4 missing START_ID
Ran again on this branch: all checks out 
💪

PR #3407, opened 2026-06-09
Minimal Axiell transformer

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6303

Refactor transformer classes:
* Delegate the transformation of individual records to a dedicated `SourceWorkBuilder` class, with subclasses for each transformer. This keeps transformer classes more focused and makes it easier to write unit tests without boilerplate.
* Add a `MarcXmlWorkBuilder` class (subclassing from `SourceWorkBuilder`), responsible for turning individual MARC fields into source work fields, allowing for easy code reuse across transformers. Each field has a dedicated method containing a default MARC transformation, which can be overridden in subclasses.

Modify the Axiell transformer:
* Replace the existing Axiell transformer stub with a minimal transformer, emitting visible works with predecessor identifiers.

## How to test

This change is covered by unit tests. The Axiell transformer predecessor ID logic cannot be tested on real data yet, since the predecessor ID is not being sent from the source system. 

* Better transformer code structure, with more focused classes and a framework for building reusable MARC transformation code.
* A minimal Axiell transformer we can build on top of. 

This change is low risk.


catalogue_graph/tests/adapters/transformers/axiell/bdd/features/minimal_record.feature
```diff
@@ -6,8 +6,10 @@ Feature: The bare minimum MARC record for an item
   Scenario: A minimal MARC record from Axiell
```
agnesgaroux, line 6: Do we still want to test for mimsy/invisible v all the rest/visible?

agnesgaroux:
Good-ish point

> extract_predecessor_id() deduplicates the raw 907$a values before normalising them with lstrip('.'). This means values like ".<uuid>" and "<uuid>" will be treated as two distinct IDs and raise, even though they normalise to the same identifier. Also, the UUID regex only accepts lowercase hex, so valid uppercase UUIDs would be rejected.

For the sierra predecessor identifier it'd be more robust to normalise then deduplicate (not your doing, mine) so that we  ignore identical values if the only difference is the `.` we don't even care about

catalogue_graph/src/adapters/transformers/builders/marc_xml_work_builder.py
```diff
@@ -0,0 +1,206 @@
...
+        return super().work_state.model_copy(
+            update={
+                "predecessor_identifier": self.predecessor_identifier,
+                "relations": WorkRelations(ancestors=self.ancestors),
+            }
+        )
+
+    @property
+    def visible_work(self) -> VisibleSourceWork:
+        return VisibleSourceWork(
```
agnesgaroux, line 202: Noted in DDB tests too but are we removing InvisibleWork transform?

catalogue_graph/src/adapters/transformers/builders/marc_xml_work_builder.py
```diff
@@ -0,0 +1,206 @@
...
+    def designation(self) -> list[str]:
+        return []
+
+    @property
+    def ancestors(self) -> list[WorkAncestor]:
+        return get_parents(self.record)
+
+    @property
+    def work_data(self) -> WorkData:
+        return WorkData(
```
agnesgaroux, line 161: Nice having everything set up and implement as we go! 👏

catalogue_graph/src/adapters/transformers/builders/marc_xml_work_builder.py
StepanBrychta: Yes, we agreed with @kenoir to remove the Mimsy/invisible work logic for now. Mimsy data is returned from a different OAI set (`exhibition` instead of `collect`) and will be structured differently (e.g. no predecessor identifiers), and so we should probably create a dedicated adapter and transformer for Mimsy.

agnesgaroux: Makes sense!

catalogue_graph/tests/adapters/transformers/axiell/bdd/features/minimal_record.feature
agnesgaroux: Addressed below 👍

catalogue_graph/src/adapters/transformers/builders/folio_work_builder.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @property
+    def predecessor_identifier(self) -> WorkSourceIdentifier | None:
+        if (value := extract_sierra_predecessor_id(self.record)) is not None:
+            return WorkSourceIdentifier(
+                identifier_type=Id(id="sierra-system-number"),
+                value=value,
+            )
+
+        return None
```
agnesgaroux, line 21: Should we have `source_modified_time` here as well? it's in the 005 field as well I think

agnesgaroux: Discussed and no because we can't have it consistent across ebsco axiell and folio

agnesgaroux (approved):
Ve 👏 ry 👏 nice 👏

PR #3411, opened 2026-06-11
Transform MARC notes

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6396

* Extract notes from 5xx MARC fields. This replaces an existing implementation, which was created as part of the Axiell transformer for exhibitions data.
* Move note extraction from the shared `MarcXmlWorkBuilder` class to `AxiellWorkBuilder`. At the moment, we do not extract any notes from EBSCO records, and this ensures that we do not accidentally start doing so.

This is a port of the Scala implementation (see `MarcNotes.scala`), minus the custom logic for removing references to Codebreakers, which may no longer be needed (to be confirmed later). 

Most notes are extracted using a standard handler, which concatenates all subfields of a given 5xx subfield (except for the suppressed subfield $5 - 'Institution to which field applies'). There are two special handlers:
* 535 notes are categorised as either location of duplicates or location of original, depending on ind1
* 561 notes are hidden if marked as private

## How to test

This change is covered by unit tests ported from Scala, plus BDD tests.

The new note extraction logic should allow us to reproduce the `notes` field outputted by the CALM transformer in the Axiell transformer.

This change does not affect production and is low risk.

### DIFF
--- a/catalogue_graph/src/adapters/transformers/builders/axiell_work_builder.py
+++ b/catalogue_graph/src/adapters/transformers/builders/axiell_work_builder.py
@@ -2,10 +2,12 @@ from adapters.transformers.builders.marc_xml_work_builder import MarcXmlWorkBuil
 from adapters.transformers.marc.last_transaction_time import (
     extract_last_transaction_time_to_datetime,
 )
+from adapters.transformers.marc.notes import extract_notes
 from adapters.transformers.marc.predecessor_identifier import (
     extract_calm_predecessor_id,
 )
 from models.pipeline.identifier import Id, WorkSourceIdentifier
+from models.pipeline.note import Note
 from utils.timezone import convert_datetime_to_utc_iso
 
 
@@ -31,3 +33,7 @@ class AxiellWorkBuilder(MarcXmlWorkBuilder):
         """Reads from MARC 005 (last transaction time) rather than the adapter row's last_modified."""
         last_modified = extract_last_transaction_time_to_datetime(self.record)
         return convert_datetime_to_utc_iso(last_modified)
+
+    @property
+    def notes(self) -> list[Note]:
+        return extract_notes(self.record)
--- a/catalogue_graph/src/adapters/transformers/marc/notes.py
+++ b/catalogue_graph/src/adapters/transformers/marc/notes.py
@@ -48,6 +48,11 @@ EXHIBITIONS_NOTE = IdLabel(id="exhibitions-note", label="Exhibitions note")
 AWARDS_NOTE = IdLabel(id="awards-note", label="Awards note")
 
 
+# TODO: The Scala notes transformer (MarcNotes.scala) had custom logic removing all references to Codebreakers.
+# Do we need to reimplement this logic here? To be confirmed once we start validating Python transformed source works
+# against Scala transformed source works.
+
+
 def _create_note_from_contents(
     field: Field,
     note_type: IdLabel,
--- a/catalogue_graph/src/adapters/transformers/marc/notes.py
+++ b/catalogue_graph/src/adapters/transformers/marc/notes.py
@@ -64,14 +64,14 @@ def _create_note_from_contents(
         if subfield_tag in suppressed:
             continue
 
-        # Subfield ǂu is rendered as an HTML link when it contains a valid URL.
+        # Subfield $u is rendered as an HTML link when it contains a valid URL.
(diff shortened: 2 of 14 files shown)


catalogue_graph/src/adapters/transformers/marc/notes.py
```diff
@@ -1,20 +1,149 @@
...
+        if subfield_tag == "u":
+            trimmed = value.strip()
+            if _is_url(trimmed):
+                parts.append(f'<a href="{trimmed}">{trimmed}</a>')
+            else:
+                logger.warning(
+                    "Subfield $u which doesn't look like a URL",
+                    contents=value,
+                )
+                parts.append(value)
```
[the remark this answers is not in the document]
StepanBrychta: The Scala implementation does not trim the value if it's not a valid URL. This might be on purpose, so I think I will keep the Scala logic for now.

catalogue_graph/src/adapters/transformers/utils/html.py
```diff
@@ -17,3 +21,13 @@ def is_url(maybe_url: str) -> bool:
     # So we need to do a bit of extra checking to see if it really
     # looks like a URL.
     return bool(url.scheme in ["http", "https"] and url.netloc)
+
+
+def format_as_html_link(maybe_url: str) -> str:
```
kenoir, line 26: Can we have a unit test for this function?

catalogue_graph/src/adapters/transformers/builders/axiell_work_builder.py
```diff
@@ -2,10 +2,12 @@
 from adapters.transformers.marc.last_transaction_time import (
     extract_last_transaction_time_to_datetime,
 )
+from adapters.transformers.marc.notes import extract_notes
```
kenoir, line 5: I wonder whether the Axiell / Marc terminology should be reversed here. The EBSCO source is the one that deviates from what you might expect interpreting plain Marc. 

Perhaps it doesn't matter and we'll end up with a work builder per Marc source, picking and choosing how to transform specific fields, but the names we choose here should reflect that choice.

catalogue_graph/src/adapters/transformers/marc/notes.py
```diff
@@ -1,20 +1,136 @@
...
+    if field.indicator1 == "1":
+        return _create_note_from_contents(field, OWNERSHIP_NOTE)
+    return None
+
+
+def _create_location_of_note(field: Field) -> Note | None:
+    # Use indicator 1 to distinguish location of originals vs duplicates.
+    if field.indicator1 == "2":
+        return _create_note_from_contents(field, LOCATION_OF_DUPLICATES_NOTE)
+    return _create_note_from_contents(field, LOCATION_OF_ORIGINAL_NOTE)
```
kenoir, line 87: The Scala `MarcNotes` transformer had special-case logic here that stripped five hardcoded "Codebreakers" sentences from 535 location-of-originals notes, and that logic is not carried over in this port. The PR description flags it as "may no longer be needed (TBC)". Is it tracked anywhere (a ticket or an inline comment) so the decision is not silently lost? 535 is the exact field that logic targeted, so a breadcrumb here would make it easy to revisit if the Codebreakers records turn up again.

Not blocking.

kenoir (approved):

catalogue_graph/src/adapters/transformers/builders/axiell_work_builder.py
StepanBrychta: We already have separate builders for each source - there is an `EbscoWorkBuilder`, a `FolioWorkBuilder`, and an `AxiellWorkBuilder`, all subclassing from the base `MarcXmlWorkBuilder` class. The idea is that `MarcXmlWorkBuilder` will contain default transformations, which can be overridden by the subclasses.

catalogue_graph/src/adapters/transformers/marc/notes.py
StepanBrychta: Good point, I added a TODO comment so that this doesn't get lost.

For context, I checked a subset of the CALM data and the Codebreaker references seem to be gone. The idea is that when we start comparing source works transformed by the Python transformer against existing source works transformed by Scala, any remaining Codebreaker references would be surfaced and we would then know if we have to reimplement the Scala logic in Python.

PR #3413, opened 2026-06-15
Scheduled image-inference state machine + Python inference manager

kenoir:
## What does this change?

This moves image inference off the always-on, SQS-driven ECS service and onto a **scheduled Step Functions state machine**, and reimplements the Scala `inference_manager` as a **Python** service in `catalogue_graph`.

Compute stays on the existing **EC2 capacity provider** (GPU is on the near-term roadmap, and GPU on ECS is EC2-only).

This is **additive**: the old SQS-driven `module.image_inferrer` service stays in place until cutover, and the new EventBridge schedule ships **disabled** behind a feature flag.

By default the new pipeline reads and writes only **shadow indices** (date `2026-06-15`): a source `images-initial-2026-06-15` whose mapping indexes `modifiedTime`, and an output `images-augmented-2026-06-15`. It never touches the live `images-initial-2025-10-02` or the prod `images-augmented-2026-04-29` that the graph pipeline reads, so its output can be validated against the Scala inferrer before any cutover. See *Rollout strategy* below.

### Python (`catalogue_graph/src/inferrer/`)

- `steps/inference_manager.py` — the ECS-task step (task-token harness). It mgets `Image[Initial]`, downloads the IIIF thumbnail, calls the 3 inferrer sidecars over localhost, assembles `InferredData`, and bulk-writes `Image[Augmented]` with `state.augmentedTime` using `external_gte` versioning.
  - **Strictly all-or-nothing**: empty/wrong-length features, an empty palette, or a missing aspect ratio / average colour cause the task to **FAIL** (so the state machine retries transient infra) rather than writing a "poisoned" augmented doc. This fixes a known defect where the Scala manager could silently index poisoned docs.
- `steps/find_work.py` — work-discovery Lambda: a time-window range query over `images-initial.modifiedTime`, partitioned for the Map fan-out. This is why the shadow source index must index `modifiedTime` (see below).
- `adapters.py` — base64 little-endian float32 decode.
- `image_downloader.py` — IIIF `info.json` -> `full/!400,400/0/default.jpg`, download plus cleanup.
- `models.py` — data models for the above.
- Small shared edits: `config.py` (+`ES_IMAGES_INITIAL_INDEX_NAME`), `utils/elasticsearch.py` (+`get_images_initial_index_name`), `tests/mocks.py` (`MockRequest` now accepts/passes through `timeout`).
- Tests: new `tests/inferrer/` suite (23 tests).

### Terraform (additive)

- New `modules/inferrer_ecs_task/` — EC2, multi-container task definition (Python manager + 3 inferrer sidecars + a shared `/data` volume). The existing `ecs_task` module is Fargate-only / single-container, so it can't be reused here.
- `modules/pipeline/state_machine_image_inferrer.tf` — `ConstructEvent` (build the time window from the schedule) -> `FindWork` (Lambda) -> `InferImages` (Map: one EC2 inference task per partition via `ecs:runTask.waitForTaskToken`). Plus an EventBridge schedule shipped **disabled** behind a new `enable_image_inferrer_schedule` variable (default `false`), plus alarms.
- `modules/pipeline/ecs_task_inference_manager.tf` — task instance + IAM (task-token, pipeline-storage secret read).
- `modules/pipeline/lambda_inference_find_work.tf` — the find-work Lambda.
- The state machine reads/writes **two shadow indices** (both date `2026-06-15`), selected by two new variables so the new path never touches live data:
  - `image_inferrer_initial_index_date` (default `2026-06-15`) selects the **source** index. It defaults to a shadow `images-initial-2026-06-15` whose mapping **indexes `modifiedTime`** (a new `index_config/mappings.images_initial.2026-06-15.json` with `dynamic: false` + a `modifiedTime` `date` field). This is needed because the live `images-initial` uses the `empty`/`dynamic:false` mapping, so `modifiedTime` is stored in `_source` but **not indexed** and the find-work window query returns 0 hits. The shadow source mirrors how `works-denormalised` selectively maps `state.mergedTime`. `find_work` and the manager are otherwise unchanged. At cutover, set this to `var.pipeline_date` once the live `images-initial` indexes `modifiedTime`.
  - `image_inferrer_augmented_index_date` (default `2026-06-15`) selects the **output** index. It defaults to a shadow `images-augmented-2026-06-15` (reusing the `images_augmented.2026-04-29` mapping) so output can be compared against the Scala inferrer's prod output. At cutover, set this to `graph_index_dates.augmented`.
  - Both are backed by a new date-only `index_config` entry (`2026-06-15`) in the `2025-10-02` root.

### Parity tooling

- `scripts/es_index_comparison/configs/image_inferrer_parity.yaml` — config to diff the shadow augmented index against the prod augmented index for the same image ids, ignoring `state.augmentedTime` (a per-run timestamp that legitimately differs). Used in Phase 3 below.

## How to test

Local validation already run on this branch:

- `ruff` clean.
- `mypy` clean (13 files).
- 23 new tests pass; 61 existing tests pass (no regression).
- `terraform fmt` clean and `terraform validate` passes.

To re-run locally from `catalogue_graph/`:

- `ruff check src tests`
- `mypy src`
- the `tests/inferrer/` suite

The following need an **AWS-capable** check (they cannot be validated by `terraform validate`):

- Confirm the Map `runTask` `CapacityProviderStrategy` key casing against a real execution.
- Confirm the state-machine role's `iam:PassRole` (task + execution roles) is sufficient for an EC2 capacity-provider `runTask`.
- Confirm the secret prefix `elasticsearch/pipeline_storage_<pipeline_date>/*` matches the actual secret names.
- After the Phase 2 reindex, `GET images-initial-2026-06-15/_mapping` to confirm `modifiedTime` is `date`-typed and that a window range query over it returns hits (the whole point of the shadow source index).

- The scheduled state machine, once enabled, augments the same set of images the SQS-driven service would have, with no "poisoned" augmented docs being indexed (the all-or-nothing behaviour above).
- Parity is proven: a shadow-vs-prod comparison (`scripts/es_index_comparison`) shows `state.augmentedTime` as the only differing field.
- No always-on inference service idling between work — inference runs on a schedule rather than a continuously-running ECS service.
- No regression in the existing `catalogue_graph` test suite (61 existing tests still pass).

- The change is additive and the schedule ships **disabled** (`enable_image_inferrer_schedule` default `false`), so merging does not change runtime behaviour. The old SQS-driven service remains the live path until cutover.
- Until cutover the state machine reads/writes only the **shadow** indices, so even a manual/parity run cannot corrupt the live `images-initial` or the prod augmented index the graph pipeline reads.
- The Python manager is strictly all-or-nothing on inference output, which is safer than the Scala manager (no silent poisoned docs) but means a partial/bad inference result fails the task and relies on state-machine retries for transient infra issues.
- Several IAM / capacity-provider / secret-naming details (listed under *How to test*) can only be confirmed against a real AWS execution and should be verified before enabling the schedule.

## Rollout strategy

The migration is staged so the new Python pipeline is **non-destructive until we deliberately cut over**. Until cutover it reads/writes only shadow indices (date `2026-06-15`): source `images-initial-2026-06-15` (mapping indexes `modifiedTime`) and output `images-augmented-2026-06-15`. The live `images-initial-2025-10-02`, the prod `images-augmented-2026-04-29`, and the Scala SQS service are untouched; the schedule ships disabled (`enable_image_inferrer_schedule = false`).

- **Phase 0 — Merge (safe by default):** merging changes nothing operationally. The Scala SQS inferrer keeps owning prod; the new state machine is disabled and points at the shadow source/output.
- **Phase 1 — Deploy (additive):** `terraform apply` creates the two shadow indices, the state machine, the find-work lambda, and the task definition, and grants the inferrer key read/write on the shadow indices. No existing resource changes behaviour.
- **Phase 2 — Populate the shadow source:** one-off ES `_reindex` `images-initial-2025-10-02` -> `images-initial-2026-06-15` (reindex re-indexes through the new mapping, so `modifiedTime` becomes queryable). Point-in-time snapshot — fine for validation.
- **Phase 3 — Parity validation:** run the state machine over a sample (manual execution / brief schedule). It windows the shadow source on `modifiedTime` and writes the shadow output; this also exercises the real `runTask` on the EC2 capacity provider, confirming `CapacityProviderStrategy` casing and `PassRole` sufficiency. Compare shadow-augmented vs prod-augmented with `scripts/es_index_comparison` (ignore `state.augmentedTime`); iterate until only `augmentedTime` differs.
- **Phase 4 — Production cutover (migrate images-initial, NOT a full pipeline reindex):**
  1. Provision a production images-initial index with the `modifiedTime`-indexed mapping.
  2. ES `_reindex` the current images-initial -> the new index (backfill).
  3. Switch where the merger writes: repoint `es_images_initial_index` (`modules/pipeline/locals.tf`, consumed in `service_merger.tf`) to the new index and redeploy the merger; reindex the delta (docs merged between the snapshot and the switch) to close the gap. No merger code change — it already sets `modifiedTime` and its key already writes all images-initial indices.
  4. Point the inferrer at prod: set `image_inferrer_augmented_index_date = graph_index_dates.augmented` and `image_inferrer_initial_index_date = <new images-initial date>`, enable the schedule, and stop/scale down the Scala service. Overlap is safe (idempotent `external_gte` writes). Then remove the merger->SQS subscription, the SQS queue, and queue-depth autoscaling.
  - *Lighter alternative:* `put_mapping` `modifiedTime` onto the existing images-initial + `update_by_query` to backfill, leaving the merger untouched (mutates the live index but avoids the write-switch gap).
- **Phase 5 — Decommission:** remove the Scala `inference_manager` (sbt project, CI step, image) and the SQS service; drop the shadow indices when no longer needed.
- **Rollback:** pre-cutover, disable the schedule (Scala still owns prod). Post-cutover, re-point the merger to the old images-initial, re-enable the Scala SQS subscription, and disable the schedule; everything is additive until the explicit teardown.

---

## Validation & resilience hardening (deployed + tested against real data)

This PR was deployed pre-merge to the **2025-10-02** pipeline and exercised end-to-end against real prod data, using a `dev`-tagged image so the shared `env.2025-10-02` tag (used by the live id-minter / transformer / graph services) was never touched. The branch was first merged up to `main` (it was behind, which otherwise made the plan want to destroy the live `transformer_folio` ES key).

### Phase 3 parity — passed
A manual execution over a 200-image sample wrote `images-augmented-2026-06-15` and was compared against the prod `images-augmented-2026-04-29` for the same ids. This also confirmed the three AWS unknowns: **`CapacityProviderStrategy` casing, the state-machine `PassRole` for an EC2 capacity-provider `runTask`, and the `elasticsearch/pipeline_storage_2025-10-02/*` secret prefix** all work.

- ✅ **Core inference matches**: `features` to float32 precision (max abs diff 1.5e-8), `aspectRatio` and `averageColorHex` identical — proving the base64 little-endian float32 decode is correct.
- ✅ `paletteEmbedding` differs (~0.0006/run) **by design** — the palette encoder adds *unseeded* Gaussian noise per pixel (`palette_inferrer/app/palette_encoder.py`), so it is non-deterministic; neither value is "more correct".
- ✅ `type`-tag differences on the copied `source`/`locations` blocks are **cosmetic** — the image ingestor re-derives them on read (lenient pydantic parsing), so the indexed output is identical (proven by the live API running off the Scala-shaped prod index).

The parity config (`scripts/es_index_comparison/configs/image_inferrer_parity.yaml`) was updated to ignore these expected-to-differ fields, so the comparison now surfaces only meaningful diffs.

### Scale test (10k images) — surfaced robustness gaps, now fixed
Running 10k images through the state machine revealed that the all-or-nothing partitions plus the Map's fail-fast default mean **any single transient error aborts the entire run**. The volume test hit three distinct transient classes, plus a capacity/concurrency mismatch:

1. IIIF thumbnail `502/503/504` (gateway/overload) were not retried.
2. ECS task placement raced on the EC2 capacity provider (`ECS.AmazonECSException: Insufficient CPU available` — each task needs a whole instance and the Map fans out `runTask` concurrently); this error was not in the retry list and **more instances reduces but can't eliminate the race**.
3. Elasticsearch bulk writes hit transient connection drops (`Remote end closed connection without response`) that were not retried.

**Resolved in this PR** (commit `Make image inference resilient to transient failures at scale`):
- `image_downloader`: retry transient HTTP statuses + connection/timeout errors with backoff; fail fast on permanent statuses (e.g. 404).
- `index_es_batch`: retry transient ES `ConnectionError`/`ConnectionTimeout`.
- State machine: retry `ECS.AmazonECSException` (capacity/placement) generously, and add a `Catch` on the inference task so a partition that still fails is **recorded and tolerated** rather than aborting the whole Map (its images are picked up by a later window via idempotent `external_gte` writes).
- Added the missing `index_config/analysis.images_initial.2026-06-15.json` so the shadow source index builds.

After these, the 10k run **completes**: `34/34` partitions, 15 transient task failures all retried or tolerated, **0 aborts**, in ~15.5 min (6-wide on the EC2 capacity provider). New unit tests cover the download and ES-write retries.

> Note: the scale test rewrote the shadow source `modifiedTime` into a single window to force one execution to sweep the whole sample, and temporarily bumped the inferrer ASG — both reverted afterwards. The deployed test infra (state machine, shadow indices) remains; production cutover still follows the Rollout strategy above.

---

## Concurrency guardrail + large-window scaling (follow-up)

Two follow-ups surfaced while running larger volumes through the new path:

### Map concurrency now matches the ASG capacity
The Map hard-coded `MaxConcurrency = 10` while the inferrer ASG was capped at 2 instances
(each task fills a whole `c5.xlarge`), so the Map fanned out more tasks than the capacity
provider could place. Both are now driven by a single `image_inferrer_max_concurrency`
variable (default 10) via `local.inference_max_concurrency`, used for the ASG `max_instances`
and the Map `MaxConcurrency` — the Map can never over-subscribe the ASG. The ASG scales to 0
when idle, so this is only a ceiling.

### Work-list handed off via S3 (removes the 256 KB payload cap)
`find_work` returned every partition's ids **inline**, and the Map iterated that array, so a
large window's result exceeded the Step Functions **256 KB** state-payload limit
(`States.DataLimitExceeded`) — capping a single execution at **~17k images**. It now writes
each partition to S3 and returns small refs (`{s3_uri, image_count}`); the inline Map iterates
the refs; each task resolves its ref back to the full event from S3 (the same pass-by-reference
idiom the graph extractor/bulk-loader use, reusing `utils.aws.pydantic_to/from_s3_json`). The
Map payload is now just refs (137k images → ~460 refs ≈ 70 KB), so a single execution can cover
the whole index. New IAM: `s3:PutObject` for the find_work lambda, `s3:GetObject` for the task,
on the `catalogue-graph/inferrer/*` prefix.

**Validated end-to-end:** a 26k-image window that previously failed with `DataLimitExceeded`
now succeeds — `find_work` writes 88 partition objects to S3, returns 88 refs, and the Map fans
out 10-wide with each task loading its partition from S3.


catalogue_graph/src/inferrer/steps/find_work.py
```diff
@@ -0,0 +1,140 @@
...
+                break
+
+            ids.extend(hit["_id"] for hit in hits)
+            search_after = hits[-1]["sort"]
+            if result.get("pit_id"):
+                pit_id = result["pit_id"]
+    finally:
+        es_client.close_point_in_time(body={"id": pit_id})
+
+    return ids
```
StepanBrychta: Can we use the established 'source' pattern that's used in other places? It should be possible to subclass from `ElasticSource` (like we do with the `AugmentedImagesSource` class) instead of duplicating the ES fetching logic here. The ElasticSource class comes with some nice features, like retries and query parallelisation.

catalogue_graph/src/inferrer/steps/find_work.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def partition(ids: list[str], size: int) -> list[list[str]]:
+    return [ids[i : i + size] for i in range(0, len(ids), size)]
```
StepanBrychta: Minor: I think we can use `itertools.batched` instead of a custom utility.

catalogue_graph/src/inferrer/steps/inference_manager.py
```diff
@@ -0,0 +1,253 @@
...
+    if inferred.average_color_hex is None:
+        problems.append("average colour is missing")
+    if inferred.aspect_ratio is None:
+        problems.append("aspect ratio is missing")
+
+    if problems:
+        raise PoisonedImageError(
+            f"Refusing to index poisoned augmented doc for image {image_id}: "
+            + "; ".join(problems)
+        )
```
StepanBrychta, line 90: Minor: It might be cleaner to do this validation directly on the `InferredData` model `@model_validator(mode="after")`. This would guarantee that the object always has the correct data,  regardless of where it's instantiated.

catalogue_graph/src/inferrer/steps/inference_manager.py
```diff
@@ -0,0 +1,253 @@
...
+        operations.append(
+            {
+                "_index": index_name,
+                "_id": image.state.id(),
+                "_source": source,
+                "_version": version,
+                "_version_type": "external_gte",
+            }
+        )
+    return operations
```
StepanBrychta: Minor: This function is nearly identical to the one defined in `ingestor_indexer.py`. I wonder if we could reuse the same function in both places. This should be possible by making the `AugmentedImage` model subclass from `IndexableRecord`.

catalogue_graph/tests/mocks.py
```diff
@@ -363,6 +365,8 @@ def get(
         data: dict | None = None,
         headers: dict | None = None,
         params: dict | None = None,
+        timeout: float | None = None,
+        **kwargs: Any,
     ) -> MockResponse:
         return MockRequest.request("GET", url, stream, data, headers, params)
```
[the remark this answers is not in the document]
kenoir: The `timeout`/`**kwargs` params exist only for signature parity with the real `requests` API, so call sites that pass a timeout don't break against the mock. The matcher deliberately keys on method/url/params/data only, and nothing currently asserts timeout behaviour, so wiring timeout into the matcher would be speculative test infrastructure. Declining for now.

catalogue_graph/src/inferrer/models.py
```diff
@@ -0,0 +1,92 @@
...
+    # window queries.
+    augmented_time: str
+
+
+class AugmentedImageToIndex(SerialisableModel):
+    state: AugmentedImageStateToIndex
+    source: ParentWork
+    locations: list[DigitalLocation]
+    version: int
+    modified_time: str
```
StepanBrychta: I don't think we need these models. We should be able to use the ones defined in `ingestor/models/augmented/image.py` (though we might want to move them elsewhere).

The `augmented_time` field can be added to the `AugmentedImageState` model directly (but it should be made optional, since older images don't have it), and then `AugmentedImageToIndex` will become a duplicate of `AugmentedImage`.

catalogue_graph/tests/mocks.py
```diff
@@ -373,6 +377,8 @@ def post(
         data: dict | None = None,
         headers: dict | None = None,
         params: dict | None = None,
+        timeout: float | None = None,
+        **kwargs: Any,
     ) -> MockResponse:
         return MockRequest.request("POST", url, stream, data, headers, params)
```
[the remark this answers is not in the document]
kenoir: The `timeout`/`**kwargs` params exist only for signature parity with the real `requests` API, so call sites that pass a timeout don't break against the mock. The matcher deliberately keys on method/url/params/data only, and nothing currently asserts timeout behaviour, so threading timeout through would be speculative test infrastructure. Declining for parity with the sibling get() thread.

catalogue_graph/src/inferrer/adapters.py
```diff
@@ -0,0 +1,98 @@
...
+
+class InferrerError(Exception):
+    pass
+
+
+def decode_base64_floats(encoded: str) -> list[float]:
+    """Decode a base64 little-endian float32 buffer into a list of floats."""
+    raw = base64.b64decode(encoded)
+    count = len(raw) // 4
+    return list(struct.unpack(f"<{count}f", raw[: count * 4]))
```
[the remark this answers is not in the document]
kenoir: Good catch. Fixed in b1d3e4b9e: `decode_base64_floats()` now raises `InferrerError` when the decoded buffer length isn't a multiple of 4, rather than silently truncating. Added a unit test covering a non-multiple-of-4 buffer.

catalogue_graph/src/inferrer/steps/inference_manager.py
```diff
@@ -0,0 +1,253 @@
...
+
+def _build_augmented(
+    image: InitialImage, inferred: InferredData
+) -> AugmentedImageToIndex:
+    state = AugmentedImageStateToIndex(
+        canonical_id=image.state.canonical_id,
+        source_identifier=image.state.source_identifier,
+        inferred_data=inferred,
+        augmented_time=datetime.now(UTC).isoformat(),
+    )
```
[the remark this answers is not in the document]
kenoir: Fixed in b1d3e4b9e: `augmented_time` now uses `convert_datetime_to_utc_iso` from `utils/timezone.py`, so it serialises with a `Z` suffix in line with the rest of the repo.

catalogue_graph/src/inferrer/steps/inference_manager.py
```diff
@@ -0,0 +1,254 @@
...
+        },
+    )
+    args = parser.parse_args()
+    validate_es_mode_for_writes(parser, args)
+    event = InferenceManagerEvent.from_argparser(args)
+    handler(event, es_mode=args.es_mode)
+
+
+if __name__ == "__main__":
+    from utils.steps import ecs_handler
```
StepanBrychta: Minor: Can we move this import to the top of the file?

catalogue_graph/src/inferrer/models.py
StepanBrychta: Also, I wonder if `AugmentedImage` should subclass from `InitialImage` to capture the fact that it builds on top of the InitialImage model.

StepanBrychta (commented):
This is great. I left a few non-critical comments (mostly about code organisation/duplication).

catalogue_graph/src/inferrer/steps/find_work.py
kenoir: Good call. Done in 3e55f5e: added `inferrer/source.py::ImagesInitialSource` subclassing `ElasticSource` (PIT + search_after, retries, slice parallelism), and `find_work` now does `list(ImagesInitialSource(event, es_client).stream_raw())` instead of the hand-rolled `scan_ids`. Only `_id` is needed so `fields=[]` (`_source` disabled) and `worker_target` yields the doc id.

catalogue_graph/src/inferrer/steps/find_work.py
kenoir: Agreed. Done in 3e55f5e: removed the custom `partition()` helper and now build partitions with `itertools.batched(ids, event.partition_size)` (using `ids=list(chunk)`). Dropped the now-redundant `test_partition_chunks_ids`.

catalogue_graph/src/inferrer/steps/inference_manager.py
kenoir: Done in 3e55f5e: `from utils.steps import ecs_handler` moved out of the `if __name__ == "__main__"` block to the top-level imports. No circular import (`utils.steps` only pulls boto3/structlog/utils.logger).

catalogue_graph/src/inferrer/steps/inference_manager.py
kenoir: I agree with centralising invariants in principle, but I'd prefer to keep this as a write-side guard here rather than a `@model_validator(mode="after")` on `InferredData`. `InferredData` lives in `ingestor/models/augmented/image.py` and is shared with the ingestor READ path (it parses existing augmented docs out of ES), so a strict after-validator would reject legacy docs with empty palettes / null colour — a read-side behaviour change well beyond the scope of this PR. The dimension check also depends on `FEATURE_VECTOR_SIZE`, which is an inferrer-layer constant. Keeping `validate_inferred` as a write-side poison guard preserves the all-or-nothing intent without constraining reads. Declining for now.

catalogue_graph/src/inferrer/steps/inference_manager.py
kenoir: Done in 3e55f5e: `generate_operations` is now shared. I moved it to `utils.elasticsearch` (typed as `Sequence[IndexableRecord]`), made `AugmentedImage` subclass `IndexableRecord` (implementing `get_id`/`get_modified_time`), and both `ingestor_indexer` and `inference_manager` import the single shared function. The ingestor and inferrer `generate_operations` tests both still pass against it.

catalogue_graph/src/inferrer/models.py
kenoir: Done in 3e55f5e: deleted `AugmentedImageStateToIndex`/`AugmentedImageToIndex` and the write path now reuses `ingestor.models.augmented.image.AugmentedImage`. I added `augmented_time: str | None = None` to `AugmentedImageState` (optional, so older images without it still parse), as you suggested. On the `AugmentedImage(InitialImage)` idea: I've left that out for now — `AugmentedImage` already needed to subclass `IndexableRecord` for the shared `generate_operations`, and the two models don't currently share a field shape (`InitialImage` here is the inferrer read model with a plain `ImageState`), so layering them would be a larger refactor than this PR warrants. Happy to revisit separately.

StepanBrychta (approved):

catalogue_graph/src/inferrer/image_downloader.py
```diff
@@ -0,0 +1,123 @@
...
+                )
+            last_error = f"status {response.status_code}"
+
+        if attempt < MAX_DOWNLOAD_ATTEMPTS:
+            time.sleep(DOWNLOAD_BACKOFF_SECONDS * 2 ** (attempt - 1))
+
+    raise ImageDownloadError(
+        f"Image request for {url} failed after {MAX_DOWNLOAD_ATTEMPTS} "
+        f"attempts ({last_error})"
+    )
```
StepanBrychta: In other places we use the `backoff` decorator to handle retries automatically. It can be configured to retry a fixed number of times using exponential backoff. I would prefer if we used the same approach here instead of mixing the fetching logic with custom retry logic.

catalogue_graph/src/utils/elasticsearch.py
```diff
@@ -80,12 +105,30 @@ def get_client(
...
+                "Transient Elasticsearch error during bulk write; retrying",
+                attempt=attempt,
+                max_attempts=ES_BULK_MAX_ATTEMPTS,
+                error=repr(exc),
+            )
+            if attempt < ES_BULK_MAX_ATTEMPTS:
+                time.sleep(ES_BULK_BACKOFF_SECONDS * 2 ** (attempt - 1))
+    else:
+        assert last_error is not None
+        raise last_error
```
StepanBrychta: Same as above - can we use the `backoff` decorator here?

catalogue_graph/src/inferrer/steps/find_work.py
```diff
@@ -62,12 +70,51 @@ def handler(
     return FindWorkResult(partitions=partitions)
 
 
+def _partition_s3_uri(event: FindWorkEvent, run_id: str, index: int) -> str:
+    bucket = config.CATALOGUE_GRAPH_S3_BUCKETS[event.environment]
+    return (
+        f"s3://{bucket}/{config.INFERRER_S3_PREFIX}/{event.pipeline_date}"
+        f"/find_work/{run_id}/partition-{index}.json"
+    )
```
StepanBrychta: In other parts of the pipeline, the standard S3 path corresponding to a given event is constructed on the event itself (e.g. see `file_path_parts` or `get_s3_uri` on the `BasePipelineEvent` class) and uses the time window in the path prefix instead of the run ID. For example, the extractor uses:

`s3://<BUCKET>/<SERVICE_NAME>/<PIPELINE_DATE>/<WINDOW_START>-<WINDOW_END>/...`

I think using the window is slightly preferable, since it keeps things organised in S3.

This is minor and non-blocking.

catalogue_graph/src/inferrer/image_downloader.py
kenoir: Good call — switched to the `backoff` decorator (`@backoff.on_exception(backoff.expo, …)`) to match the convention used in `neptune_client`/`core.source`/`sparql_client`, so the retry policy is declarative and no longer interleaved with the fetch logic. Behaviour (transient-only retries, 404 fail-fast, env-tunable attempts) is preserved and covered by the existing tests. Done in a5d2f3efc.

catalogue_graph/src/utils/elasticsearch.py
kenoir: Good call — switched to the `backoff` decorator here too (`@backoff.on_exception(backoff.expo, …)`), matching the `neptune_client`/`core.source`/`sparql_client` convention. The per-retry warning is preserved via an `on_backoff` handler, and transient-only retries + env-tunable attempts are unchanged and still covered by the existing tests. Done in a5d2f3efc.

catalogue_graph/src/inferrer/steps/find_work.py
kenoir: I would prefer it align with the other services, I will resolve here.

kenoir: Good call — moved the path construction onto the event (`FindWorkEvent.find_work_s3_prefix` / `partition_s3_uri`) and keyed it by the run scope (window → `windows/<start>-<end>`, ids → `by_id/<segment>`, else `full`), mirroring the `file_path_parts`/`get_s3_uri` layout used elsewhere, instead of the Lambda request id. A given window's partitions now land in a stable, predictable S3 location. Done in b2f98a9f4.

PR #3428, opened 2026-06-24
Bump pipeline transitive dependencies (urllib3, form-data)

kenoir:
## What does this change?

This bundles five Dependabot transitive dependency bumps in the pipeline area into one PR. All bumps are transitive only. No source files or version constraints changed, only lockfiles.

- `urllib3` 2.6.3 -> 2.7.0 across the four image inferrer uv projects: `pipeline/inferrer/common`, `pipeline/inferrer/feature_inferrer`, `pipeline/inferrer/aspect_ratio_inferrer`, and `pipeline/inferrer/palette_inferrer`.
- `form-data` 3.0.4 -> 3.0.5 in `pipeline/matcher_merger/matcher/scripts` (yarn.lock), pulled in transitively via the existing `form-data@^3.0.0` range.

Closes #3391
Closes #3384
Closes #3383
Closes #3379
Closes #3415

## How to test

Quality checks run locally:

- `aspect_ratio_inferrer`: `uv run pytest` passes (4 passed).
- `palette_inferrer`: `uv run pytest` passes (5 passed).
- `matcher_merger/matcher/scripts`: `yarn install --frozen-lockfile` succeeds, confirming the lockfile is consistent.
- `feature_inferrer`: pytest cannot collect on this machine due to a pre-existing `RuntimeError: MKL-DNN build is disabled`. This is a torch/hardware limitation in the local environment, not a result of this change. The same error reproduces on `main` without this change, so it is unrelated to the urllib3 bump. CI will run these tests.

CI passes on this branch and the inferrer and matcher services continue to build and run as before. There is no behavioural change to measure.

Low risk. These are patch and minor transitive bumps with no changes to source or to declared version constraints. The version ranges already permitted these versions; the lockfiles are simply being moved forward. The one gap is that `feature_inferrer` tests could not run locally because of an unrelated environment issue, so we rely on CI to exercise that project.


kenoir:
Checked this against the buildkite issue from #3386, where a urllib3 bump to 2.7.0 broke CI because 2.7.0 needs Python >=3.10 and the lambda publish step pip-installs on a Python 3.9 runner.

This PR is safe on two counts:

1. The inferrer images are not built by buildkite. Nothing in `.buildkite/` calls `build_inferrer_image.sh` or `publish_inferrer_image_to_ecr.sh`. The pipeline deploy step only re-tags existing ECR images via `update_ecr_image_tag.sh`, so these `uv.lock` files are never consumed by a buildkite build or pip install.

2. Where the inferrers are built, it is Python 3.10 via uv, not the path that broke #3386. The Dockerfile builds with `--build-arg pythonversion="$(cat .python-version)"` (3.10) and installs with `uv sync --frozen` on a `python:3.10-slim` base. All four inferrer pyproject.toml files declare `requires-python >=3.10`, so urllib3 2.7.0 resolves fine.

The #3386 failure is specific to `builds/publish_lambda_zip.py` running `pip3 install` on the Python 3.9 buildkite agent, which is used only by these lambdas: `calm_window_generator`, `calm_deletion_check_initiator`, `common/window_generator`, `tei_updater`. This PR does not touch any of their published requirements.txt.

The yarn.lock change is JS and unrelated to the Python runner.

agnesgaroux (approved):

PR #3433, opened 2026-06-26
Retire image-inferrer cutover variables (augmented index + schedule gate)

kenoir:
Stacked on #3432. Base this PR on `remove-old-scala-inferrer`, review and merge it after #3432, and apply its terraform after #3432's terraform.

This is the variable-collapse follow-up flagged in #3426 and again in #3432. With the old Scala image-inferrer gone, the variables that existed only to gate the cutover have outlived their purpose. This PR retires that cutover plumbing in two parts: the augmented-index overrides, and the schedule enable gate.

### Augmented-index variables

Now that the old Scala image-inferrer is removed (#3432), the two cutover override variables that decoupled the augmented index during the migration are redundant. The new Python inferrer's write target and the graph read-path source can both resolve from a single `graph_index_dates.augmented`.

In `pipeline/terraform`:

- `2025-10-02/main.tf`: set `graph_index_dates.augmented = "2026-06-15"` (was `"2026-04-29"`), and remove the `image_inferrer_augmented_index_date = "2026-06-15"` and `graph_images_augmented_index_date = "2026-06-15"` overrides. Kept `image_inferrer_initial_index_date = "2026-06-15"`. Refreshed the cutover comment and the two `index_config` comments that referenced the removed variable.
- `modules/pipeline/variables.tf`: removed the `image_inferrer_augmented_index_date` and `graph_images_augmented_index_date` variable definitions. Kept `image_inferrer_initial_index_date` and `image_inferrer_max_concurrency`.
- `modules/pipeline/locals.tf`: removed the two resolving locals. Kept `image_inferrer_initial_index_date`, which still falls back to `pipeline_date`.
- `modules/pipeline/subsystem_graph.tf`: the graph module's `index_dates.augmented` now reads `var.graph_index_dates.augmented` directly (was `local.graph_images_augmented_index_date`).
- `modules/pipeline/state_machine_image_inferrer.tf`: the inferrer's ConstructEvent `index_dates.augmented` now reads `var.graph_index_dates.augmented` directly (was `local.image_inferrer_augmented_index_date`).

This is a net-zero change. Before this, both overrides were set to `"2026-06-15"`, so the read-path source and the inferrer write target already resolved to `images-augmented-2026-06-15`. Setting `graph_index_dates.augmented = "2026-06-15"` and dropping the overrides resolves to the same value through the default fallback. No index routing changes.

### Schedule enable gate

`enable_image_inferrer_schedule` was shipped disabled to gate the cutover from the old SQS-driven service. That service is gone and the scheduled Python inferrer is now the sole inferrer, so the steady state is enabled. This keeps the variable but changes what it means.

- `modules/pipeline/variables.tf`: flip the `enable_image_inferrer_schedule` default from `false` to `true`, and reword its description. It is no longer a cutover gate; it is now a kill-switch to pause scheduled inference, for example during an incident or a large reindex.
- `2025-10-02/main.tf`: remove the now-redundant `enable_image_inferrer_schedule = true` override, since the default now carries that value.

This is also net-zero. The override was already `true`, so the schedule resolves to `ENABLED` either way. Dropping the override and defaulting the variable to `true` lands on the same `state = "ENABLED"` for the inferrer's scheduling rule.

### Out of scope

This does not delete the orphaned `images-augmented-2026-04-29` index. Every index is an `elasticstack_elasticsearch_index` with `deletion_protection = !allow_delete_indices`, and `allow_delete_indices = false` on this stack, so removing the `index_config` entry would make `terraform apply` fail on the blocked destroy. Deleting that index is a separate operational step: flip `allow_delete_indices`, apply, flip back, or delete out of band. Its `index_config` entry is left in place on purpose, with a comment in `main.tf` noting this.

## How to test

This PR is stacked on #3432. Apply its terraform from the branch before merging, after #3432 has been applied and merged. Same ordering discipline as #3432.

1. Run `./run_terraform.sh plan` in `pipeline/terraform/2025-10-02/` and confirm it shows no functional changes. This is a net-zero variable collapse, so the variable and local plumbing simplification is plan-invisible. The augmented index keeps resolving to `images-augmented-2026-06-15` and the inferrer schedule rule stays `ENABLED`.
2. Apply, then merge.

Verification performed on this branch:

- `terraform fmt -check` is clean and `terraform validate` succeeds on the `2025-10-02` stack.
- A grep sweep confirms no remaining references to the two removed augmented-index variables or locals, and that the only remaining `enable_image_inferrer_schedule` references are the variable definition and its single use in the inferrer schedule rule.
- Not run: a live `terraform plan` against real state, which needs AWS credentials.

`terraform plan` shows no functional resource changes. The image read-path and the inferrer write target continue to resolve to `images-augmented-2026-06-15`, with one variable now driving both instead of three, and the scheduled inferrer rule stays `ENABLED` with the schedule controlled by a single defaulted variable rather than a per-stack override.

Both changes are net-zero. The two removed augmented-index overrides were already set to `"2026-06-15"`, the same value `graph_index_dates.augmented` now carries, so routing is unchanged. The `enable_image_inferrer_schedule` override was already `true`, the same value the variable now defaults to, so the schedule stays `ENABLED`. The plan should confirm no functional resource changes before apply.

The orphaned `images-augmented-2026-04-29` index is deliberately left under terraform management. Removing its `index_config` entry while `allow_delete_indices = false` would break apply, so that cleanup is held back as a separate operational step.

Related: #3426, #3432.

### DIFF
--- a/pipeline/terraform/modules/pipeline/subsystem_graph.tf
+++ b/pipeline/terraform/modules/pipeline/subsystem_graph.tf
@@ -2,12 +2,9 @@ module "graph_pipeline" {
   source = "./graph"
 
   pipeline_date = var.pipeline_date
-  # The graph subsystem's augmented source can be decoupled from graph_index_dates.augmented (which
-  # the old Scala inferrer still writes) via graph_images_augmented_index_date — see that variable and
-  # local.graph_images_augmented_index_date. All other dates pass through unchanged.
   index_dates = {
     merged    = var.graph_index_dates.merged
-    augmented = local.graph_images_augmented_index_date
+    augmented = var.graph_index_dates.augmented
     works     = var.graph_index_dates.works
     concepts  = var.graph_index_dates.concepts
     images    = var.graph_index_dates.images
--- a/pipeline/terraform/2025-10-02/main.tf
+++ b/pipeline/terraform/2025-10-02/main.tf
@@ -111,11 +111,9 @@ module "pipeline" {
   allow_delete_indices = false
 
   # Image-inferrer. The scheduled Python inferrer is the sole inferrer: it reads images-initial-2026-06-15
-  # and writes images-augmented-2026-06-15, which the graph READ-path (extractor + ingestor + remover)
-  # also reads. graph_index_dates.augmented = "2026-06-15" is the single source for both the write target
-  # and the read source. image_inferrer_initial_index_date stays overridden because this in-place pipeline's
-  # live images-initial used the old "empty"/dynamic:false mapping, so the merger and inferrer were moved
-  # onto the modifiedTime-mapped images-initial-2026-06-15 (it falls back to pipeline_date otherwise).
+  # and writes images-augmented-2026-06-15, which the graph read-path also reads.
+  # graph_index_dates.augmented = "2026-06-15" is the single source for both. image_inferrer_initial_index_date
+  # is overridden because it otherwise falls back to pipeline_date.
   image_inferrer_initial_index_date = "2026-06-15"
 
   # Base AMI for ECS instances
--- a/pipeline/terraform/2025-10-02/main.tf
+++ b/pipeline/terraform/2025-10-02/main.tf
@@ -116,7 +116,6 @@ module "pipeline" {
   # and the read source. image_inferrer_initial_index_date stays overridden because this in-place pipeline's
   # live images-initial used the old "empty"/dynamic:false mapping, so the merger and inferrer were moved
   # onto the modifiedTime-mapped images-initial-2026-06-15 (it falls back to pipeline_date otherwise).
-  enable_image_inferrer_schedule    = true
   image_inferrer_initial_index_date = "2026-06-15"
 
   # Base AMI for ECS instances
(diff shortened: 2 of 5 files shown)


pipeline/terraform/modules/pipeline/subsystem_graph.tf
```diff
@@ -2,12 +2,9 @@ module "graph_pipeline" {
...
-  # The graph subsystem's augmented source can be decoupled from graph_index_dates.augmented (which
-  # the old Scala inferrer still writes) via graph_images_augmented_index_date — see that variable and
-  # local.graph_images_augmented_index_date. All other dates pass through unchanged.
   index_dates = {
     merged    = var.graph_index_dates.merged
-    augmented = local.graph_images_augmented_index_date
+    augmented = var.graph_index_dates.augmented
     works     = var.graph_index_dates.works
     concepts  = var.graph_index_dates.concepts
     images    = var.graph_index_dates.images
```
StepanBrychta, line 10: Minor: I think this can be simplified back to `index_dates = var.graph_index_dates`.

StepanBrychta (approved):

PR #3435, opened 2026-06-29
Extract Axiell languages

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6409

Transform languages and language notes in the Axiell transformer. Implementation ported from Scala `CalmLanguages.scala`, with some minor simplifications which shouldn't affect behaviour. 

## How to test

All unit tests were ported from Scala to show that the ported code behaves the same as the Scala code.

Relevant fields on Axiell source works (`languages` and `notes`) should have the same values as corresponding CALM source works.

This change does not touch production. The Axiell transformer will be fully validated against the CALM transformer at a later stage.


catalogue_graph/src/adapters/transformers/axiell/languages.py
```diff
@@ -0,0 +1,129 @@
...
+        _match_after_corrections,
+        _match_after_stripping_tags,
+    ):
+        result = matcher(lang_field)
+        if result is not None:
+            return result
+    return None
+
+
+def _match_exact(lang_field: str) -> list[Language] | None:
```
kenoir, line 98: The Scala version tried ExactLanguageMatch (matching the whole field against a single language name) before MultiLanguageMatch (split on separators). Folding both into a split-first matcher means a field value that is itself a language name containing one of the separator chars no longer matches. 123 names in the MARC list contain a separator, e.g. the inverted variant forms "Circassian, West" and "Spanish, Aragonese". For "Circassian, West" the Scala code returned [Language(ady)] with no note, whereas this returns ([], [LanguageNote(...)]).

Probably rare in real 546 data, but it's a behaviour difference from CALM rather than a no-op simplification. Worth a quick exact-match-first pass, or at least a note that it's a known, accepted divergence?

(Verified by running both lookups against the bundled languages.xml.)

kenoir (commented):
Reviewed the port against CalmLanguages.scala and MarcLanguageCodeList.scala, and ran the new tests (37 passing) plus the lookup against the bundled languages.xml.

The test port is thorough and faithful, the TODO comments documenting the source-system fixes are clear, and excluding 546 from the base note path so it isn't double-handled is the right call.

One behaviour difference from CALM noted inline. I've also confirmed Copilot's obsolete-code thread is a genuine parity gap (details on that thread). Neither is a blocker for a change that's off the production path.

catalogue_graph/src/lookups/languages.py
```diff
@@ -30,22 +30,29 @@ def load_language_code_to_name_map() -> dict[str, str]:
...
+    uf_tag = f"{{{CODELIST_NS}}}uf"
     for language_element in doc.findall("c:languages/c:language", namespaces=ns_decl):
         code = language_element.findtext(code_tag)
         name = language_element.findtext(name_tag)
         yield code, name
 
+        if include_name_variants:
+            for uf in language_element.findall(uf_tag):
+                yield code, uf.findtext(name_tag)
+
```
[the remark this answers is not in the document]
kenoir: Confirming this is a real parity gap. Running the name->code map against the bundled languages.xml: 31 names resolve to more than one code, so from_name logs a spurious "Multiple language codes" warning for each (e.g. Croatian -> ['hrv', 'scr']), and obsolete-only names like Aljamía -> ajm now resolve where the Scala MarcLanguageCodeList returned None. Ivatan also comes back as ['phi', 'phi'], so duplicate codes are not deduped. The chosen id was still the authorized one in every case I sampled (it appears first in document order), so I do not think it produces a wrong id today, but filtering status="obsolete" as the Scala code did would remove both the warning noise and the obsolete-only resolution.

catalogue_graph/src/adapters/transformers/axiell/languages.py
StepanBrychta: Yes, I added a `_match_exact_single` matcher. In the real data, there are no cases where this makes a difference at the moment, but I agree it's a more robust approach.

kenoir (approved):

PR #3436, opened 2026-06-29
Transform Axiell production

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6337

Tweak Axiell `production` extraction logic, bringing it closer to parity with the Scala pipeline. Previously, we were extracting Axiell production using the EBSCO `extract_production` function, which parses period start/end dates from the production label free text field. However, this frequently produced incorrect results, since the Python production parser is not as elaborate as the Scala one. 

The new approach uses parsed start and end dates from Axiell where available, with custom parsing only used as fallback.

This PR does not make any improvements to the Python date parser, relying on parsed dates from the source system, whose quality is at least as high as the output of the Scala parser. At some point, we will likely need to bring the parser closer to the Scala one to parse Folio dates.

## How to test

This change is covered by unit tests.

I tested this change on the full dataset, producing a 94% match between Axiell production dates and CALM production dates. Most of the remaining discrepancies are due to differences in how approximate dates are interpreted in the custom Scala parser versus the Axiell parser:
* In Scala, approximate dates like `c. 1930` are parsed as a range from `1920-01-01` to `1939-12-31` (versus `1920-01-01` to `1940-12-31` in Axiell).
* In Scala, approximate dates like `c.1960s` are parsed as a range from `1950-01-01` to `1970-12-31` (versus `1950-01-01` to `1979-12-31` in Axiell).

The above differences are minor and both variants can be considered correct. When ignored, there is a 99% match between Axiell and CALM.

The `production` field on Axiell source works should be more closely aligned with its CALM counterpart. 

At the moment, we do not know how the parsed Axiell dates are populated. The approach in this PR assumes that they will keep getting populated in the future, removing the need for a custom date parser in the Axiell transformer. This unknown is documented on a Notion page and will be raised with the relevant team.


catalogue_graph/src/lookups/languages.py
```diff
@@ -60,12 +60,13 @@ def _iter_languages(
...
         name = language_element.findtext(name_tag)
         code = code_element.text
-        yield code, name
 
         if include_name_variants:
             # Find all descendant instances of `<name>` to make sure we yield all variant names
             for name_element in language_element.findall(f".//{name_tag}"):
                 yield code, name_element.text
+        else:
+            yield code, name
```
StepanBrychta, line 69: Ensure that the main (code, name) pair is not yielded twice.

catalogue_graph/src/adapters/transformers/axiell/production.py
```diff
@@ -0,0 +1,68 @@
...
+    if not production_labels:
+        return []
+
+    # Unlike CALM, Axiell usually returns parsed start and end production dates (stored in MARC 046).
+    # If available, and if there is exactly one production label, use the provided dates instead
+    # of trying to parse the period ourselves from the production label.
+    if len(production_labels) == 1 and start_date is not None and end_date is not None:
+        periods = [_period_from_dates(production_labels[0], start_date, end_date)]
+    else:
+        periods = [parse_period(label) for label in production_labels]
```
kenoir: `parse_period` raises ValueError for century labels that don't exactly match the "nth century" pattern: "early 20th century", "mid 19th century" and "19th-20th century" all raise "Invalid century format". Nothing in the builder catches this, so the whole record fails to transform. The Scala side never fails here: `ParsedPeriod` delegates to `PeriodParser`, which returns an Option, so an unparseable date just gives a Period with no range (ParsedPeriod.scala:14).

This isn't new in this PR (the old EBSCO path also ran `parse_period` on 264 $c), but this fallback keeps it live for records without 046 dates. Did the full-dataset run hit any of these? If it did, would it be worth catching ValueError here and emitting a range-less Period to match Scala?

catalogue_graph/src/adapters/transformers/axiell/dates.py
```diff
@@ -11,20 +11,29 @@
 def _try_parse_date(s: str) -> date | None:
     """Try to parse date in yyyy-M-d format (e.g. 2039-01-01). Return `None` if not valid date."""
     try:
-        y, m, d = s.split("-")
-        parsed = date(int(y), int(m), int(d))
-        return parsed
+        return datetime.strptime(s, "%Y-%m-%d").date()
```
kenoir, line 14: `strptime` with `%Y` is stricter about years than the old split/int parser: "850-01-01" now returns None (with a warning) where the old code returned `date(850, 1, 1)`. Zero-padded "0850-01-01" is fine. Do we know whether Axiell zero-pads years in 046 for early material? If it does, this is a non-issue. The failure mode is soft either way (production falls back to parsing the label).

kenoir (commented):
The event shape matches `CalmTransformer.production` exactly (labels joined with a space, one period per label, empty places/agents, no function), and using the parsed 046 dates instead of free-text parsing is a sensible improvement. A couple of things I checked while reviewing: the `languages.py` dedupe removes 485 duplicate yields with nothing lost or gained, and the `Period` type default fixes a real serialisation bug (Periods built without an explicit type came out as "Concept"). All 239 Axiell and MARC transformer tests pass at head.

Two questions inline, neither blocking.

catalogue_graph/src/adapters/transformers/axiell/production.py
kenoir: Answered this by scanning the full adapter store (228,506 records, scanned 2026-07-03): 12 records hit this. 8 have real century labels ("Early 20th century", "Late 17th century", "c.17th century" and similar) and 4 are cataloguing test records with XSS/SQLi strings in 264 $c. All 12 failed to transform and were dropped from the index.

Fixed in a1f8692: the fallback now catches the ValueError and emits a range-less Period, matching the Scala PeriodParser. Re-running the transform over the live store now gives zero failures. Most of the century-labelled records also carry bare-year 046 dates, so with the bare-year parsing added in the same commit they get proper ranges from 046 rather than a range-less period.

catalogue_graph/src/adapters/transformers/axiell/dates.py
kenoir: Checked the live 046 data: there are no 3-digit years anywhere, so the zero-padding concern is moot. The scan found something bigger though: 383 of the 387 non-empty 046 $k/$l values in the store are bare years like "1900", which neither the old nor the new parser accepted. Only 2 records carry full ISO dates (e.g. collect:4608 pairs 046 $k1855-07-21 with 264 $c "July 21st, 1855"), and even those fell back because their first 046 field is bare-year. So this branch fired for exactly zero records on the current store.

Axiell appears to write $k/$l at the precision it knows: full dates for exact dates, bare years for approximate ones. Fixed in a1f8692: production dates now also accept bare years, widened to start of year for $k and end of year for $l. With that, 151 of the 215 records with a production label use the Axiell-parsed dates.

@StepanBrychta worth double-checking which dataset the 94% comparison ran against. The examples in the PR description (c. 1930 giving 1920-01-01 to 1940-12-31) look like year-precision values, which matches what the store holds, but the code as merged from this branch previously required full ISO dates.

catalogue_graph/src/lookups/languages.py
kenoir: Verified: the main <name> element is a direct child of the language element, so findall(".//name") includes it and the variants branch yields it exactly once. load_language_name_to_codes_map no longer collects duplicate codes for the primary name.

kenoir (approved):

catalogue_graph/src/adapters/transformers/axiell/dates.py
StepanBrychta: I ran the comparison against the full, up-to-date Axiell dataset that I downloaded locally for testing purposes. All works in that dataset store values like `1900-01-01` in the 046 field. The adapter store is out of date and includes Mimsy works, which are out of scope for the transformer.

The year-precision value from the PR description (`c. 1930 giving 1920-01-01 to 1940-12-31`) is the raw production label stored in $264 $c. 

Modifying the parser to also accept bare years is definitely an improvement, but I don't think it's strictly needed.

PR #3438, opened 2026-06-30
Enrich FOLIO items in the OAI-PMH adapter

kenoir:
> [!Note] 
> This implements [the enricher option](https://github.com/wellcomecollection/location-movement-control-docs/pull/20) from the FOLIO OAI-PMH item-enrichment discovery doc. 

## What does this change?

The FOLIO OAI-PMH feed (`marc21_withholdings`) carries item data in MARC 952 but no item UUID, so FOLIO works currently have no stable per-item id. This implements Option C: enrich each harvested instance with item UUIDs from mod-inventory-storage's `oai-pmh-view/enrichedInstances`, store them in a second Iceberg table, and join that onto the bib store at transform time. That keeps the transformer a pure function of the stores, so a full reindex never calls FOLIO.

The `enrichedInstances` contract has been verified read-only against the live FOLIO system, and the code matches it. The endpoint sits behind OKAPI, so auth uses `x-okapi-token` and `x-okapi-tenant` rather than the OAI feed's `Authorization` token. The client logs in to OKAPI itself (POST `/authn/login`) and refreshes the token on a 401, so no static token is stored. It streams concatenated JSON objects (`{...}{...}`, no array brackets or newlines) with lowercase keys; each record is `{instanceId, itemsAndHoldingsFields: {items: [...]}}`; items carry nested `callNumber` and `location` objects; and an instance with no items sends `itemsAndHoldingsFields: ""`. A full harvest, enrichment, and transform run against the live system produced works with real item UUIDs in a local Elasticsearch.

Enrichment ingest (new, isolated, under `adapters/extractors/oai_pmh/folio/enrichment/`):
- `models.py`: payload models matching the live contract (concatenated-stream friendly, nested `callNumber`/`location` flattened). Parsing fails loud: a record missing `instanceId`, or an item missing `id`, raises rather than being coerced to the string `"None"`.
- `inventory_client.py`: read-only `enrichedInstances` client (batched; parses the concatenated-JSON-object stream the live endpoint returns, and still tolerates an array, a wrapper object, or NDJSON). An unparseable record fails the batch rather than being silently dropped.
- `enricher.py` and `runtime.py`: query by the bare instance UUID, then re-key each result back to the bib store's full OAI identifier so the transform join lines up. Items are upserted into a `folio-items` namespace store (reusing `AdapterStore`), idempotent via `incremental_update`. A returned instance that matches no requested id raises rather than being stored as an orphaned, unjoinable row.
- `steps/oai_pmh/folio_enrich.py`: the "Run enrichment" step. It reads the changed bib store ids from the bib changeset, fetches their items, upserts them, and passes both changesets forward. The ECS entrypoint uses the shared `ecs_handler` (`waitForTaskToken`), matching the loader.

Transform-time join:
- `AdapterStoreSource` optionally joins an items store onto bib rows by id. Adapters with no items store (EBSCO/Axiell) are unaffected.
- `FolioWorkBuilder.items` emits items carrying a `folio-item` source identifier with the inventory UUID. It falls back to no items when an instance is not yet enriched, and never guesses from MARC 952.
- `Item.id` is widened to include the source-stage `Identifiable`, matching the existing `models/pipeline/concept.py` union, so item ids are mintable.

Terraform (behind a new `enable_item_enrichment` module variable, default `false`, set `true` only on the FOLIO module so EBSCO/Axiell are unchanged):
- `modules/adapter/ecs_enrichment.tf`: the enrichment ECS task (gated by `count`) with the same IAM shape as the loader, plus a state-machine policy to run it.
- `modules/adapter/state_machine.tf`: a "Run enrichment" state between Run loader and Publish event, carrying `items_changeset_ids` for provenance. It is gated so the generated definition is byte-identical for non-enriching adapters.
- `ssm_parameters.tf`: `inventory_api_url`, `inventory_username`, `inventory_password` and `inventory_tenant` (placeholder values with `ignore_changes`, matching the existing pattern; the password is a `SecureString`).

## How to test

Quality gate, run inside `catalogue_graph/` (mirrors the `python_check` CI action):

```
uv run ruff check .
uv run ruff format --check .
uv run mypy .
uv run pytest        # or scope: uv run pytest tests/adapters/extractors/oai_pmh/folio/enrichment tests/adapters/transformers/folio
```

Results on this branch: ruff clean, `mypy .` clean (569 files), 1367 tests pass (27 integration deselected, 0 failures).

Tests:
- `tests/adapters/extractors/oai_pmh/folio/enrichment/`: payload models against the live shape (concatenated stream, lowercase keys, empty items, nested `callNumber`/`location`), client batching and parsing, enricher idempotency and UUID re-keying, and the enrichment step.
- Fail-loud regression tests: a missing item id raises, a missing `instanceId` raises, and an enriched instance that matches no requested id raises in the enricher.
- `tests/adapters/transformers/folio/test_item_enrichment.py`: `FolioWorkBuilder` items, an end-to-end transform join producing works with `folio-item` item identifiers, and the no-enrichment fallback.

Terraform (no apply): inside `catalogue_graph/infra/adapters/`, `terraform validate` passes and `terraform plan` is clean, reporting 15 to add, 4 to change, 0 to destroy. The adds are the enrichment task, its IAM, and the two SSM params; the only meaningful change is the FOLIO state machine. The 3 `trigger_lambda` updates are pre-existing package-hash drift.

## Dev test deployment (validated end to end)

Deployed to dev and validated end to end. The FOLIO adapter runs its loader and enrichment tasks on a `:dev` image (ebsco/axiell stay on `:prod`), with `enable_item_enrichment = true` and the OKAPI inventory credentials populated in SSM.

A real item change (a `952$n` edit on an instance that has items) flowed through the whole state machine:
- the loader harvested the one changed record and produced a bib changeset,
- `Should enrich?` routed to `Run enrichment`,
- the enrichment task ran on Fargate: its task role read the SSM credentials, the client logged in to OKAPI, called `enrichedInstances`, and wrote the items store (container exit code 0),
- the execution published `folio.adapter.completed` and finished `SUCCEEDED`.

Also confirmed along the way: the enrichment client authenticates itself via OKAPI login and refreshes on 401 (no static token); the enrichment task role has the SSM read, KMS decrypt and s3tables write it needs; and the adapter security group allows egress to OKAPI.

The change was then followed downstream through the transformer and id-minter. The transformer joins the enriched items store onto works; this PR also fixes it to open that store read-only (`create_if_not_exists=False`), since only the enrichment step should create it. With that fix the enriched work reached `works-source` carrying its `folio-item` items, and the id-minter minted stable canonical IDs for the work and all six items in `works-identified`. That is the point of this change: the OAI-PMH feed carries no item identifier, so items had no stable public ID; the enrichment supplies the item UUIDs and the id-minter turns them into durable canonical IDs.

Note on memory: the transformer join tips a single-record transform over the previous 4096 MB (the bib store is sorted by id, so a changeset read cannot prune and materialises most of the table). This PR raises the adapter transformer Lambda to 10240 MB as an interim mitigation; #3444 tracks the real fix (read the changeset by record id).

Once enabled and deployed (not in this PR):
- FOLIO source works carry items with stable `folio-item` source identifiers (and, after the minter, canonical ids), so the public catalogue can address individual items.
- The items store stays current incrementally. An item or holdings change re-advances the instance's OAI datestamp into the bib window (verified read-only against the dev feed: 29 of 46 changed records over 30 days were driven by non-bib changes), so no separate item-change tracking is needed.
- A full reindex still replays from the stores with no live FOLIO calls, so transformer purity is preserved.

- **Shared-model change:** widening `Item.id` to include `Identifiable`. The ingestor consumes post-mint `Identified`, so this is additive; the union order matches the existing `concept.py` pattern, and the full models and adapters suites pass.
- **Sequencing:** apply the SSM params (real OKAPI inventory URL, username, password and tenant) and seed the items store before enabling the enrichment state, so the first publish after cutover has item rows to join. Until enrichment runs, works are still valid, with no items rather than wrong items.
- **Scale:** the items-side load is addressed (the join now fetches items per bounded batch with an id-scoped `In("id", ...)` lookup instead of reading the whole items namespace). The analogous bib-side full-reindex materialisation, which predates this PR and affects all adapters, is tracked in #3439 (and its sibling #3444, the un-prunable changeset read the dev test surfaced).


catalogue_graph/src/adapters/transformers/adapter_store_source.py
```diff
@@ -1,36 +1,81 @@
...
+        self.items_store = items_store
 
     def stream_raw(self) -> Generator[dict[str, Any]]:
         if self.changeset_ids:
             for changeset_id in self.changeset_ids:
                 table = self.adapter_store.get_records_by_changeset(
                     changeset_id, self.snapshot_id
                 )
-                yield from table.to_pylist()
+                yield from self._with_enrichment(table.to_pylist())
```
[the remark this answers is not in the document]
kenoir: Thanks. This is tracked scale work rather than something this PR introduces. The changeset-read memory cost comes from get_records_by_changeset scanning the whole bib table on an un-prunable changeset filter, captured in #3444, where the real fix is reading by id. Resolving as out of scope for this PR.

catalogue_graph/src/adapters/transformers/adapter_store_source.py
```diff
@@ -1,36 +1,81 @@
...
         else:
             logger.info("No changeset_id provided; performing full reindex of records.")
 
             # During a full reindex we are writing into an empty index,
             # so no need to include deleted rows to overwrite documents.
             table = self.adapter_store.get_active_namespace_records(self.snapshot_id)
-            yield from table.to_pylist()
+            yield from self._with_enrichment(table.to_pylist())
```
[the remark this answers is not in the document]
kenoir: Thanks. This is the full-reindex case already tracked in #3439, where a full reindex materialises the entire namespace. It predates this PR and is out of scope here, so resolving against that issue.

catalogue_graph/src/adapters/extractors/oai_pmh/folio/enrichment/okapi_auth.py
```diff
@@ -0,0 +1,76 @@
...
+    ) -> Generator[httpx.Request, httpx.Response, None]:
+        if self._token is None:
+            self._store_token((yield self._login_request()))
+        self._apply(request)
+        response = yield request
+        if response.status_code == 401:
+            # Token has likely expired: re-login once and retry the request.
+            self._store_token((yield self._login_request()))
+            self._apply(request)
+            yield request
```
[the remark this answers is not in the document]
kenoir: Checked this against httpx's auth flow: the client reads and closes each intermediate response before sending the next request from auth_flow (_send_handling_auth calls response.read() / aread(), and closes on error), on both the sync and async paths. The login and retry responses here are consumed by httpx, so there is no leaked connection to close manually. Resolving as not applicable to httpx.

catalogue_graph/src/adapters/extractors/oai_pmh/folio/enrichment/models.py
```diff
@@ -0,0 +1,153 @@
...
+        nested = (
+            _first_present(data, "itemsAndHoldingsFields", "itemsandholdingsfields")
+            or {}
+        )
+        raw_items = (
+            _first_present(data, "items")
+            or (nested.get("items") if isinstance(nested, dict) else None)
+            or []
+        )
+        instance_id = _first_present(data, "instanceId", "instanceid", "instance_id")
```
StepanBrychta: On this line, and in other places in this file, we check for several field name variants. Why do we do this? Does the Folio API inconsistently/arbitrarily return all of these variants? If so, why does this line check for three variants (instanceId/instanceid/instance_id), but line 133 only checks for two (instanceId/instanceid)? 

Also, I think that checking for both `instanceId` and `instanceid` is redundant, since `_first_present` already lowercases both key sets.

kenoir: This was a pre-emptive measure that I think can be trimmed as the API output was not originally clear, I think this can be removed now.

kenoir: Trimmed in 2c6f181619. The alternates (instance_id, itemId, materialTypeId, callNumberString, effectiveLocation) were speculative and are gone, along with the case-duplicate keys that _first_present already covers. The case-insensitive lookup itself stays because the live stream lowercases its wrapper keys (instanceid, itemsandholdingsfields) while item fields are camelCase; there is now a docstring note explaining that.

catalogue_graph/infra/adapters/modules/adapter/state_machine.tf
StepanBrychta, line 1: All the `enable_item_enrichment` checks make me wonder whether merging all the adapters into a single Terraform stack was a mistake. I think we didn't anticipate that the state machines would have a different structure for different data sources. (Just a thought, nothing to address in this PR. It might be possible to do some refactoring in the future to at least reduce the number of repeated `var.enable_item_enrichment ? 1 : 0` checks.)

catalogue_graph/src/adapters/extractors/oai_pmh/folio/enrichment/inventory_client.py
```diff
@@ -0,0 +1,134 @@
...
+from adapters.extractors.oai_pmh.folio.enrichment.models import FolioEnrichedInstance
+
+logger = structlog.get_logger(__name__)
+
+ENRICHED_INSTANCES_PATH = "oai-pmh-view/enrichedInstances"
+
+
+def _chunk(ids: list[str], size: int) -> Iterator[list[str]]:
+    for start in range(0, len(ids), size):
+        yield ids[start : start + size]
```
StepanBrychta: Minor: I think we can use `itertools.batched` here instead of a custom helper.

catalogue_graph/src/adapters/extractors/oai_pmh/folio/enrichment/inventory_client.py
```diff
@@ -0,0 +1,134 @@
...
+    try:
+        payload = response.json()
+    except json.JSONDecodeError:
+        # Streamed/concatenated (or newline-delimited) JSON objects.
+        return _decode_json_stream(text)
+
+    if isinstance(payload, list):
+        return cast("list[dict]", payload)
+    if isinstance(payload, dict):
+        for key in ("enrichedInstances", "instances", "records"):
```
StepanBrychta: Do we need to check all of these keys?

catalogue_graph/src/adapters/transformers/adapter_store_source.py
```diff
@@ -1,32 +1,43 @@
...
 from core.source import BaseSource
 
 logger = structlog.get_logger(__name__)
 
+# Bib rows are enriched in batches of this size, so the per-batch item lookup (and the
+# items it holds in memory) stays bounded regardless of how many rows are transformed.
+ITEM_ENRICHMENT_BATCH_SIZE = 10_000
+
 
 class AdapterStoreSource(BaseSource):
```
StepanBrychta, line 12: Minor: The added logic is Folio-specific, and I wonder if it would be possible to isolate it into a Folio-specific subclass (e.g. `FolioStoreSource`). We use inheritance in other places where different sources require modifications to the shared logic.

catalogue_graph/src/adapters/steps/transformer.py
```diff
@@ -119,7 +121,26 @@ def build_transformer(
     if event.transformer_type == "ebsco":
         return EbscoTransformer(adapter_store, event.changeset_ids, snapshot_id)
     if event.transformer_type == "folio":
-        return FolioTransformer(adapter_store, event.changeset_ids, snapshot_id)
+        # Join the enriched items store so FOLIO works carry items with stable
+        # `folio-item` identifiers.
+        # The transformer only reads the items store; the enrichment step creates it.
+        # Before enrichment has ever run the items table does not exist yet, so treat
+        # a missing table as "no enrichment" and transform without items (works still
+        # emit no items rather than failing the whole transform).
```
StepanBrychta: Is it preferable to emit works without items rather than failing the transformer if the items store is not found? Doesn't this risk silently emitting works without items if, for example, someone were to accidentally remove the Folio items store?

kenoir:
> [!Note]
> We've rebased onto the this branch [and addressed.](https://github.com/wellcomecollection/catalogue-pipeline/pull/3438/commits/55674ac70a30503cde85ddc9ebf7b93360c60560)

Heads up from #3447 (changeset reads in the transformer): PipelineStore now has get_records_by_changesets, which bounds the content read by the changesets' minimum last_modified so file stats can prune. On the FOLIO store a fresh single-record changeset read drops from about 2.3 GB and 39 seconds to under 300 MB and 2 seconds.

Two places in this branch will want it after rebasing:

- steps/oai_pmh/folio_enrich.py reads the bib changeset with bib_store.get_records_by_changeset(changeset_id), which is the plain unbounded scan. Switching to bib_store.get_records_by_changesets([changeset_id]) gets the pruned read with identical results.
- adapter_store_source.py's changeset branch has changed on main: it now calls get_records_by_changesets once for all changeset ids and converts one record batch at a time. When resolving the conflict, wrap both the per-batch yield and the full-reindex branch in _with_enrichment as planned; batched() accepts any iterable so the enrichment join composes unchanged.

StepanBrychta (commented):
I think there is an edge case we might want to think about:

1. The adapter loader runs, processing some window W.
2. It processes N records, storing them in the adapter store, marking the window as processed in the window store, and returning some changeset ID C.
3. The enricher starts but fails repeatedly, causing the whole state machine to fail.

At this stage, if we fix whatever caused the enricher to fail and redrive the state machine from failure, it will safely process all N records with the changeset ID C.

However, if we start a new state machine execution, the adapter would not produce a changeset ID, since the window W is already marked as processed. The state machine run would be a no-op, and item processing would silently be skipped.

Is this something we want to prevent? Should we track enrichment status in the window status table too? (I don't have a solution.)

catalogue_graph/infra/adapters/modules/adapter/state_machine.tf
kenoir: Agreed, the conditionals are noisy. I'd rather not restructure the stack in this PR, but a follow-up that pulls the enrichment states into their own module (or splits the stacks) seems worth it.

catalogue_graph/src/adapters/extractors/oai_pmh/folio/enrichment/inventory_client.py
kenoir: Done in e2477cc12d.

catalogue_graph/src/adapters/extractors/oai_pmh/folio/enrichment/inventory_client.py
kenoir: No, they were speculative. The live endpoint streams concatenated JSON objects, so I've dropped the wrapper-key checks and kept list / single-object / stream handling. Done in a96e40734f.

catalogue_graph/src/adapters/transformers/adapter_store_source.py
kenoir: Good idea. Done in 8b27ecd853: the join now lives in a FolioStoreSource subclass, and the base class just exposes a _process_rows hook.

catalogue_graph/src/adapters/steps/transformer.py
kenoir: Fair point, the fallback made table loss too quiet. Done in d86cac0be3: a missing items table now fails the transform. The one operational consequence is that a fresh environment needs one enrichment pass before a full reindex, since the enrichment step is what creates the table.

kenoir:
> However, if we start a new state machine execution, the adapter would not produce a changeset ID, since the window W is already marked as processed.

Good catch. Worth noting it's broader than enrichment: in that scenario the publish never fires either, so the transformer also never runs for changeset C. The bibs sit in the store untransformed as well as unenriched. That makes it the same class of failure as the existing stranded-window problem, and it predates this PR; enrichment just widens the gap between marking the window and publishing.

I'd suggest:

- Treat redrive-from-failure as the recovery path and document it. The execution fails and alarms, so the only silent case is starting a fresh execution instead of redriving.
- Follow-up ticket for a reconciliation sweep: an enrichment mode that scans for active bib ids with no items row and enriches them. That self-heals any stranding regardless of cause, same pattern as the inferrer's index scan.

I'd rather not add enrichment status to the window table in this PR. It only covers half the symptom (publish is skipped too), and re-running the loader on an already loaded window yields an empty changeset anyway, so it wouldn't recover without reloader-style forcing.

StepanBrychta:
Maybe we could also make the loader return the changeset ID corresponding to the earlier successful run. That would make the loader output truly idempotent: If we input window W into the loader but the window was already processed, return the changeset ID produced by the earlier run, so that downstream services can use it.

StepanBrychta (approved):

kenoir:
> Maybe we could also make the loader return the changeset ID corresponding to the earlier successful run.

Good news: the loader already does this. Window summaries store their changeset ids in tags, harvest_range returns reused summaries for windows already marked success instead of re-harvesting them, and OAIPMHLoaderResponse.from_summaries folds changeset ids from all summaries into the response, reused ones included. So re-feeding window W returns changeset C.

The gap is in the trigger it starts the next range at last_success_end, so a fresh execution never re-covers W. Recovery is a new execution with an explicit range covering W, no redrive needed. Making it fully automatic would mean advancing the trigger on published rather than loaded windows, which I'd rather scope as the follow-up along with the reconciliation sweep.

I've added a test pinning the reuse behaviour (changeset ids from reused summaries reach the loader response) in 2d66e0aeb1.

Issue #3444, opened 2026-07-01
OAI-PMH transformer: get_records_by_changeset scans the whole bib table (un-prunable changeset filter)


kenoir:
## Problem

Reading a changeset's records with `get_records_by_changeset` on the large bib store is memory-heavy. The adapter store table is sorted by `id` (`ADAPTER_STORE_SORT_ORDER`), so a filter on `changeset` cannot prune, and the scan touches nearly every data file even to return a single record.

## Evidence

Profiling the FOLIO transformer on one changed record (a changeset with 1 row) from the 1.7M-record bib store:

- Peak RSS jumps from ~190 MB to ~1.8 GB at `get_records_by_changeset(...)`, for **1 row** returned.
- The scan touches **19 of 23 data files** (the changeset filter does not prune; the table is sorted by `id`).
- The item-enrichment join adds ~0 MB (it is id-scoped and prunable).
- In the deployed transformer Lambda this exceeded 4 GB and OOM'd (`Runtime.OutOfMemory`) transforming a single record.

## Impact

Every transform reads its changeset this way, so memory scales with the whole bib table rather than the changeset size. It surfaced while validating FOLIO item enrichment (the join tipped a marginal 4 GB transform into OOM), but the cost is in the shared changeset read and affects all OAI-PMH adapters.

## Proposed fix

Read the changeset's bib rows by their record ids (`In("id", record_ids)`), which is prunable on the id-sorted table, instead of the un-prunable `get_records_by_changeset` scan. The changed record ids are already tracked by the loader (`record_ids_changed` in the window-status tags / loader response), so they can be threaded to the transformer. This mirrors the id-scoped pattern already used for the items lookup.

Alternatives: partition or add a sort field on `changeset`; or (interim) raise the transformer Lambda memory.

## Related

Sibling of #3439 (full reindex materialises all rows). Found while validating PR #3438.

kenoir:
Data point from the dev validation: live-bumping the transformer Lambda (`catalogue-2025-10-02-transformer`) from 4096 to 10240 MB was enough to get a single-record FOLIO enrichment transform through (it OOM'd at 4096). This confirms the cost is the changeset read itself, not a leak. The bump is currently a live edit (config drift); if we keep it as the interim mitigation it should be set in terraform rather than live.

PR #3445, opened 2026-07-02
Stream the full-reindex read from the adapter store

### LINKED ISSUE #3439: OAI-PMH adapter: full reindex materialises the entire bib namespace in memory
## Problem

`AdapterStoreSource.stream_raw` (`catalogue_graph/src/adapters/transformers/adapter_store_source.py`), on the full-reindex path (no changeset), calls `self.adapter_store.get_active_namespace_records(self.snapshot_id)` and then `table.to_pylist()`, materialising all active bib rows for the adapter into a single Python list before any processing. At catalogue scale (for example millions of FOLIO bib records) the whole namespace is held in memory at once.

## Scope

This is shared by every adapter that uses `AdapterStoreSource` (Axiell, EBSCO, FOLIO). It predates and is independent of item enrichment. The incremental/changeset path also calls `to_pylist()`, but per changeset (one harvest window), so that path is bounded and fine. Only the full-reindex path is unbounded.

## Relation to PR #3438

PR #3438 fixed the analogous items-side load: the enrichment join now fetches items per bounded batch with `In("id", ids)` instead of reading the whole items namespace. The bib-side materialisation was called out as out of scope there; this issue tracks it.

## Suggested direction

Stream the bib table in bounded batches (iterate PyArrow record batches, or slice the table) instead of one `to_pylist()`, so memory stays flat on a full reindex. Keep the downstream batching in `catalogue_graph/src/core/transformer.py` (`_transform_batches`, `batched(..., ES_BULK_INDEX_BATCH_SIZE)`) and the per-changeset path behaving as they do today. The enrichment join in `_with_enrichment` already batches, so it composes with a batched bib stream.

kenoir:
Closes #3439.

## What does this change?

On a full reindex (no changeset ids), `AdapterStoreSource.stream_raw` loaded the whole adapter store namespace into memory twice: once as an Arrow table via `get_active_namespace_records`, and again as a Python list via `to_pylist()`. For the 1.7M-record FOLIO bib store that peaks at about 13 GB, which cannot fit in the transformer Lambda. The same path is shared by Axiell and EBSCO.

This change streams the read as Arrow record batches instead:

- `PipelineStore.stream_namespace_records`: same namespace filter and snapshot pinning as `get_namespace_records`, but returns `scan.to_arrow_batch_reader()` (cast to the store schema) instead of a materialised table.
- `AdapterStore.stream_active_namespace_records`: applies the same non-deleted filter as `get_active_namespace_records`.
- `AdapterStoreSource.stream_raw`: the full-reindex branch iterates batches and converts one batch at a time. The changeset branch is untouched.

Downstream needs no changes: `ElasticBaseTransformer._transform_batches` already consumes `stream_raw()` lazily in 10k chunks.

### Measurements

Read-only runs against the production FOLIO bib S3 Table (1,706,137 active rows, 23 data files), each in a fresh process, reporting `ru_maxrss`:

| Approach | Peak RSS | Elapsed |
| --- | --- | --- |
| Current eager read (`get_active_namespace_records().to_pylist()`) | 12,995 MB | 117s |
| This change (`to_arrow_batch_reader`) | 4,530 and 4,885 MB (two runs) | 45s |
| Per-file `ArrowScan` streaming (considered, not chosen) | 2,746 MB | 80s |

All three returned identical row counts.

### Choice of API

pyiceberg's `to_arrow_batch_reader` prefetches data files on a thread pool regardless of consumption rate (each file scan task materialises its whole file, and all tasks are submitted up front), so its peak memory is a large constant-factor improvement over the eager read rather than a hard per-batch bound. This is a known upstream issue (apache/iceberg-python#2407) with an open improvement PR (#1995).

A per-file read via `ArrowScan.to_table([task])` gives a genuinely bounded profile (2.7 GB, flat as the table grows) and was implemented in an earlier iteration of this branch (commit `3cf3ebb72`), but `ArrowScan` sits below pyiceberg's public API and the library is pre-1.0, so its signature can change in a minor release. The public reader was chosen deliberately, in discussion with the repo owner: it is the supported surface, it is simpler and faster, upstream memory improvements land for free, and the measured 4.5 to 4.9 GB peak fits comfortably in the transformer Lambda's 10240 MB. The per-file pattern is preserved in this branch's history if a hard bound is ever needed.

## How to test

Inside `catalogue_graph/`:

```
uv run ruff check .
uv run ruff format --check .
uv run mypy .
uv run pytest
```

On this branch: ruff clean, `mypy .` clean (558 files), 1350 tests pass (27 deselected). New tests:

- `tests/adapters/utils/test_adapter_store.py` (5): streamed rows match the eager read, deleted rows are excluded, snapshot pinning is honoured, an empty table streams no rows, and batches carry the store schema.
- `tests/adapters/transformers/test_adapter_store_source.py` (4): the full-reindex path yields all active rows, the changeset path is unchanged, consuming one row pulls only one batch and never calls the eager read (laziness), and an empty store yields nothing.

On a full reindex, the transformer reads the FOLIO bib namespace without running out of memory. Peak RSS drops from about 13 GB to about 4.5 GB, well inside the transformer Lambda's 10240 MB, at current scale, and the read is faster (about 45s against 117s). Row counts match the eager read exactly, so the streamed read produces the same works.

- Peak memory scales with table size and with how far the reader's prefetch runs ahead of the consumer, rather than being bounded by one batch. At current scale it is 4.5 GB against a 10 GB Lambda. If the stores grow substantially, options are: cap `PYICEBERG_MAX_WORKERS` on the Lambda (upstream reports a large memory drop with unchanged duration), pick up the upstream reader improvements, or restore the per-file read from commit `3cf3ebb72`.
- The manifest bookkeeping in `stream_to_index` still accumulates all successful ids and an id mapping for the run. That is O(n) in short id strings, small next to record content, and out of scope here.

### Related work

- #3444 is the sibling issue about the changeset read scanning the whole table with an un-prunable filter. This work does not touch `get_records_by_changeset` or the changeset branch of `stream_raw`, so the two fixes are independent.
- PR #3438 wraps both `stream_raw` branches in `_with_enrichment(rows)`. The only textual conflict is the tail of the full-reindex branch. To resolve when both are in: build the row generator (`row for batch in batches for row in batch.to_pylist()`), pass it to `_with_enrichment`, and loosen that method's parameter type from `list` to `Iterable`. Its body uses `itertools.batched`, which accepts any iterable, so the enrichment join stays bounded and the stream stays lazy end to end.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/catalogue_graph/src/adapters/transformers/adapter_store_source.py
+++ b/catalogue_graph/src/adapters/transformers/adapter_store_source.py
@@ -32,5 +32,9 @@ class AdapterStoreSource(BaseSource):
 
             # During a full reindex we are writing into an empty index,
             # so no need to include deleted rows to overwrite documents.
-            table = self.adapter_store.get_active_namespace_records(self.snapshot_id)
-            yield from table.to_pylist()
+            # Stream record batches so the full table is never held in memory at once.
+            batches = self.adapter_store.stream_active_namespace_records(
+                self.snapshot_id
+            )
+            for batch in batches:
+                yield from batch.to_pylist()
--- a/catalogue_graph/src/adapters/transformers/adapter_store_source.py
+++ b/catalogue_graph/src/adapters/transformers/adapter_store_source.py
@@ -32,9 +32,14 @@ class AdapterStoreSource(BaseSource):
 
             # During a full reindex we are writing into an empty index,
             # so no need to include deleted rows to overwrite documents.
-            # Stream record batches so the full table is never held in memory at once.
+            # Stream record batches so the full table need not be materialised
+            # at once. Close the reader on exit so an abandoned stream (e.g. a
+            # consumer error mid-reindex) does not leave prefetch reads running.
             batches = self.adapter_store.stream_active_namespace_records(
                 self.snapshot_id
             )
-            for batch in batches:
-                yield from batch.to_pylist()
+            try:
+                for batch in batches:
+                    yield from batch.to_pylist()
+            finally:
+                batches.close()
--- a/catalogue_graph/src/adapters/utils/adapter_store.py
+++ b/catalogue_graph/src/adapters/utils/adapter_store.py
@@ -1,4 +1,3 @@
-from collections.abc import Generator
 from datetime import UTC, datetime
 
 import pyarrow as pa
@@ -72,7 +71,7 @@ class AdapterStore(PipelineStore):
 
     def stream_active_namespace_records(
         self, snapshot_id: int | None = None
-    ) -> Generator[pa.RecordBatch]:
+    ) -> pa.RecordBatchReader:
         """Stream non-deleted records in the store namespace as record batches."""
         non_deleted_filter = Or(EqualTo("deleted", False), IsNull("deleted"))
         return self.stream_namespace_records(non_deleted_filter, snapshot_id)
--- a/catalogue_graph/src/adapters/utils/adapter_store.py
+++ b/catalogue_graph/src/adapters/utils/adapter_store.py
@@ -1,3 +1,4 @@
+from collections.abc import Generator
 from datetime import UTC, datetime
 
 import pyarrow as pa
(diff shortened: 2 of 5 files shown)


catalogue_graph/src/adapters/transformers/adapter_store_source.py
```diff
@@ -32,5 +32,9 @@ def stream_raw(self) -> Generator[dict[str, Any]]:
 
             # During a full reindex we are writing into an empty index,
             # so no need to include deleted rows to overwrite documents.
-            table = self.adapter_store.get_active_namespace_records(self.snapshot_id)
-            yield from table.to_pylist()
```
StepanBrychta, line 36: The `get_active_namespace_records` function is now unused in production code. Maybe we could remove it to make it impossible for someone to accidentally use it in the future without understanding the memory implications?

StepanBrychta (approved):

catalogue_graph/src/adapters/transformers/adapter_store_source.py
kenoir: Good spot, it is unused in production after this change. I have left it in place deliberately: the open FOLIO enrichment PR (#3438) changes its signature and adds a new production call site (reading item records from the items store), so removing it here would conflict with that work. It is also still the verification helper in a fair few tests. If #3438 does not land we can remove it then.

PR #3446, opened 2026-07-02
Axiell normalise HTML text

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6337

Port HTML normalisation (sanitisation) logic from the CALM transformer (`NormaliseText.scala`) to the Axiell transformer. Sanitisation is done using the [nh3](https://pypi.org/project/nh3/) package, which is based on [ammonia](https://github.com/rust-ammonia/ammonia).

To mirror the Scala logic, a `rel="nofollow"` is added to all HTML links, but we should also consider adding `noreferrer`, since it's good practice. 

## How to test

Unit tests were ported from Scala. Validation against CALM showed that the relevant fields (description, physical description, notes) now match.

HTML sanitisation parity between CALM transformer and Axiell transformer.

This change is low risk, reproducing functionality of the existing CALM transformer.

### DIFF
--- a/catalogue_graph/pyproject.toml
+++ b/catalogue_graph/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
     "pymysql>=1.0.0",
     "cryptography>=48.0.1",
     "yoyo-migrations>=8.2.0",
+    "nh3"
 ]
 
 [tool.uv]
--- a/catalogue_graph/pyproject.toml
+++ b/catalogue_graph/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
     "pymysql>=1.0.0",
     "cryptography>=48.0.1",
     "yoyo-migrations>=8.2.0",
-    "nh3"
+    "nh3>=0.3"
 ]
 
 [tool.uv]
--- a/catalogue_graph/src/adapters/transformers/axiell/description.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/description.py
@@ -1,0 +1,9 @@
+from pymarc.record import Record
+
+from adapters.transformers.marc.common import non_empty_subfields
+from adapters.transformers.utils.html import normalise_text
+
+
+def extract_description(record: Record) -> str | None:
+    description = " ".join(non_empty_subfields("520", "a", record))
+    return normalise_text(description) or None
--- a/catalogue_graph/src/adapters/transformers/axiell/notes.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/notes.py
@@ -11,6 +11,7 @@ from adapters.transformers.marc.notes import (
     _create_note,
 )
 from adapters.transformers.marc.notes import extract_notes as base_extract_notes
+from adapters.transformers.utils.html import normalise_text
 from models.pipeline.id_label import IdLabel
 from models.pipeline.note import Note
 
@@ -40,5 +41,8 @@ def extract_notes(record: Record) -> list[Note]:
         notes.append(Note(contents=terms_of_use, note_type=TERMS_OF_USE))
 
     notes += extract_languages(record)[1]
+
+    for note in notes:
+        note.contents = normalise_text(note.contents)
 
     return notes
--- a/catalogue_graph/src/adapters/transformers/axiell/notes.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/notes.py
@@ -45,4 +45,5 @@ def extract_notes(record: Record) -> list[Note]:
     for note in notes:
         note.contents = normalise_text(note.contents)
 
-    return notes
+    # Filter out any notes whose contents were fully removed by `normalise_text`
+    return [n for n in notes if n.contents]
--- a/catalogue_graph/src/adapters/transformers/axiell/physical_description.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/physical_description.py
@@ -1,0 +1,14 @@
+from pymarc.field import Field
+from pymarc.record import Record
+
+from adapters.transformers.marc.common import non_empty_subfields
+from adapters.transformers.utils.html import normalise_text
+
+
+def _field_content(field: Field) -> str:
+    return " ".join(field.get_subfields("a", "b", "c", "e")).strip()
+
+
+def extract_physical_description(record: Record) -> str | None:
+    description = " ".join(non_empty_subfields("300", "a", record))
+    return normalise_text(description) or None
--- a/catalogue_graph/src/adapters/transformers/axiell/physical_description.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/physical_description.py
@@ -1,12 +1,7 @@
-from pymarc.field import Field
 from pymarc.record import Record
 
 from adapters.transformers.marc.common import non_empty_subfields
 from adapters.transformers.utils.html import normalise_text
-
-
-def _field_content(field: Field) -> str:
-    return " ".join(field.get_subfields("a", "b", "c", "e")).strip()
 
 
 def extract_physical_description(record: Record) -> str | None:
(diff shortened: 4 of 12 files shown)


catalogue_graph/tests/adapters/transformers/axiell/bdd/features/physical_description.feature
```diff
@@ -16,19 +16,9 @@ Feature: Physical description extraction from Axiell MARC records
...
-
-  Scenario: Single 300 field with multiple subfields joined by space
-    Given the MARC record has a 300 field with subfields:
-      | code | value              |
-      | a    | 1 folder           |
-      | b    | ill.               |
-      | c    | 30 cm              |
-      | e    | + 1 map            |
-    When I transform the MARC record
-    Then the work's physical_description is "1 folder ill. 30 cm + 1 map"
```
StepanBrychta, line 28: Removing this test, since Axiell MARC records always store physical description in 300 $a.

catalogue_graph/src/adapters/transformers/utils/html.py
```diff
@@ -1,10 +1,69 @@
...
+    "none": set(),
+    "italics_only": {"i"},
+}
+
+
+def normalise_text(s: str, allowed_tags: AllowedTags = "basic") -> str:
+    """Sanitise an HTML string, retaining only the tags permitted by allowed_tags."""
+
+    # TODO: `rel="nofollow"` added to mirror Scala. Do we also want to add `noreferrer`?
+    cleaned = nh3.clean(s, tags=_TAGS[allowed_tags], link_rel="nofollow")
```
kenoir: The tag list matches jsoup's `Safelist.basic()`, but nh3's default attribute and URL-scheme allowances are wider than jsoup's. I ran a quick probe against this branch:

- `<em title="x" lang="en">hi</em>` keeps `title` and `lang` (jsoup basic only allows `a[href]`, `blockquote[cite]`, `q[cite]`)
- `<a href="tel:+44123">` keeps the `tel:` href (jsoup basic restricts `a[href]` to http, https, ftp, mailto)
- `hreflang` is kept on `a`

If strict parity with CALM output matters for the validation you did, it might be worth passing explicit `attributes={"a": {"href"}, "blockquote": {"cite"}, "q": {"cite"}}` and `url_schemes={"http", "https", "ftp", "mailto"}` to `nh3.clean`. If the wider defaults are fine in practice, happy to leave as is.

catalogue_graph/src/adapters/transformers/axiell/physical_description.py
```diff
@@ -0,0 +1,14 @@
+from pymarc.field import Field
+from pymarc.record import Record
+
+from adapters.transformers.marc.common import non_empty_subfields
+from adapters.transformers.utils.html import normalise_text
+
+
+def _field_content(field: Field) -> str:
```
kenoir: `_field_content` (and the `Field` import) look like leftovers from copying `marc/physical_description.py`; nothing calls it here. Worth removing?

catalogue_graph/src/adapters/transformers/axiell/notes.py
```diff
@@ -41,4 +42,7 @@ def extract_notes(record: Record) -> list[Note]:
 
     notes += extract_languages(record)[1]
 
+    for note in notes:
```
kenoir, line 45: A note whose contents normalise to an empty string will stay in the list as an empty note, whereas `extract_description` and `extract_physical_description` drop empties with `or None`. CalmNotes has the same behaviour so this is parity, just checking it's deliberate rather than filtering with something like `[n for n in notes if n.contents]`.

catalogue_graph/pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     "pymysql>=1.0.0",
     "cryptography>=48.0.1",
     "yoyo-migrations>=8.2.0",
+    "nh3"
```
kenoir: `nh3` is the only dependency without a version specifier. Worth adding `>=0.3` or similar to match the rest of the list?

catalogue_graph/src/adapters/transformers/axiell/physical_description.py
```diff
@@ -0,0 +1,14 @@
...
+    return " ".join(field.get_subfields("a", "b", "c", "e")).strip()
+
+
+def extract_physical_description(record: Record) -> str | None:
+    description = " ".join(non_empty_subfields("300", "a", record))
+    return normalise_text(description) or None
```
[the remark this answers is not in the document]
StepanBrychta: Yes, extracting the physical description only from 300 $a is intentional. The Axiell MARC stylesheet never uses other subfields. I removed the now redundant `_field_content` function.

catalogue_graph/src/adapters/transformers/axiell/physical_description.py
StepanBrychta: Yes, see https://github.com/wellcomecollection/catalogue-pipeline/pull/3446#discussion_r3514005163.

catalogue_graph/pyproject.toml
StepanBrychta: Yes, done

catalogue_graph/tests/adapters/transformers/axiell/bdd/features/physical_description.feature
```diff
@@ -16,19 +16,9 @@ Feature: Physical description extraction from Axiell MARC records
...
-
-  Scenario: Multiple 300 fields are joined with <br/>
+    
+  Scenario: Multiple 300 fields are joined by space
     Given the MARC record has a 300 field with subfield "a" value "42 leaves"
     And the MARC record has another 300 field with subfield "a" value "1 map"
     When I transform the MARC record
-    Then the work's physical_description is "42 leaves<br/>1 map"
+    Then the work's physical_description is "42 leaves 1 map"
```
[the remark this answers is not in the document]
StepanBrychta: I updated the outdated file description. The scenario was removed on purpose, see https://github.com/wellcomecollection/catalogue-pipeline/pull/3446#discussion_r3513794192.

catalogue_graph/src/adapters/transformers/axiell/notes.py
StepanBrychta: Well spotted, I think we should filter out any empty notes (correctness is more important than parity). I added the filtering logic and a unit test covering this edge case.

catalogue_graph/src/adapters/transformers/axiell/description.py
```diff
@@ -0,0 +1,9 @@
+from pymarc.record import Record
+
+from adapters.transformers.marc.common import non_empty_subfields
+from adapters.transformers.utils.html import normalise_text
+
+
+def extract_description(record: Record) -> str | None:
+    description = " ".join(non_empty_subfields("520", "a", record))
+    return normalise_text(description) or None
```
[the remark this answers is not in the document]
StepanBrychta: I added a few BDD tests covering this field.

catalogue_graph/src/adapters/transformers/utils/html.py
StepanBrychta: Good point. I don't think this matters much from a Scala parity perspective, since I haven't seen this in the real data, but I wonder if we should restrict allowed attributes and URL schemes anyway (and not necessarily in a way with matches jsoup's defaults). For example, does it make sense to support URL schemes other than `https`? Do we want to support `ftp`?

For now, I added the jsoup defaults you suggested, and maybe we can revisit this later.

catalogue_graph/src/adapters/transformers/utils/html.py
```diff
@@ -1,10 +1,69 @@
...
+    result: list[str] = []
+    for line in lines:
+        # Strip consecutive blank lines
+        if not result and line == "":
+            continue
+        if result and result[-1] == "" and line == "":
+            continue
+        result.append(line)
+
+    return html.unescape("\n".join(result)).strip()
```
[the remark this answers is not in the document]
StepanBrychta: Yes, I think this is a real vulnerability ported from the Scala implementation. I don't know why the Scala pipeline unescapes the output, and I can't think of a good reason to port the `unescape` call to Python. The frontend uses `dangerouslySetInnerHTML` to render the HTML content, so I think that the content should _not_ be unescaped. If we want to unescape any HTML tags or character entities in the raw text, maybe we can do it _before_ the `clean` call rather than after.

kenoir (approved):

PR #3449, opened 2026-07-03
Advance the OAI-PMH trigger on published windows

kenoir:
## What's changing

Follow-up to the recovery discussion on #3438. A state machine execution that dies after the loader commits but before the publish event strands its changesets: windows are marked success at harvest time, the trigger resumes from last_success_end, so the next execution starts past the stranded windows and their changesets are never enriched or published.

The adapter state machine now ends in a mark-published step, so the trigger can resume from the last window that made it all the way through:

```mermaid
flowchart TD
    trigger["Run trigger (lambda): compute window, resume from last published window"]
    loader["Run loader (ECS): harvest pending windows, mark success, name the covered windows and re-emit their changesets"]
    enrichQ{"Should enrich? (folio only)"}
    enrich["Run enrichment (ECS): item enrichment"]
    publishQ{"Should publish event?"}
    publish["Publish event (EventBridge putEvents, ns.adapter.completed)"]
    mark["Mark published (lambda): stamp the covered windows with published_at"]
    done(["Success"])

    trigger --> loader
    loader --> enrichQ
    enrichQ -->|changesets exist| enrich
    enrichQ -->|no changesets| publishQ
    enrich --> publishQ
    publishQ -->|changesets exist| publish
    publishQ -->|quiet, no changesets| mark
    publish --> mark
    mark --> done
```

The loader response names the success windows whose changesets it carries (covered_window_keys), and mark-published stamps exactly those rows, so windows written by a concurrently running execution, or rows whose changesets the loader never re-emitted, are never marked published. "Should enrich?" and "Run enrichment" are folio only; "Mark published" only renders when published tracking is enabled, otherwise both publish paths go straight to Success. The self-healing property is that if an execution dies between "Run loader" and "Mark published" the published cursor does not advance, so the next scheduled run re-covers the range without re-harvesting, re-emits the changesets, and re-publishes.

This makes recovery automatic:

- A new mark-published lambda runs at the end of the adapter state machine and stamps the windows named by the loader response with a published_at tag. It runs on both the publish path and the quiet no-publish path, so the cursor keeps moving when there is nothing to publish.
- The trigger resumes from the last published window, cascading to last_success_end while no window carries a stamp yet. Loaded-but-unpublished windows stay inside the next harvest range: the loader skips re-harvesting them (no OAI calls) but re-emits their changeset ids, which was pinned by test in #3438, so the next scheduled run re-publishes anything stranded. Enrichment and item writes are idempotent, so replays are safe.
- The trigger only prefers the published cursor when PUBLISHED_CURSOR_ENABLED is set, which terraform drives from the same enable_published_tracking flag that wires the stamping step. Turning the feature off therefore reverts cleanly to last_success_end rather than pinning the cursor on stale stamps.
- The lag circuit breaker measures harvest liveness (last_success_end), not the published cursor: a stale stamp, whether from a stalled mark-published step or a disable and re-enable cycle, cannot halt harvesting, and the next successful run re-covers and re-stamps the stale range. Stamping failures still alarm through the failed execution.
- The loader response carries the covered window keys itself; a JSONata Output merge threads them past the enrichment state only. Everything is gated behind a new enable_published_tracking module flag, enabled for folio and axiell; the ebsco state machine renders unchanged.
- Cursor computation reads only the published_at key rather than parsing the JSON-bearing tag values, so one corrupt tag on a historical row cannot fail the trigger. A published_at value that is not a valid timestamp is treated as unstamped and re-stamped.

Behaviour notes:

- Deploying this code before the terraform is applied changes nothing: the env flag is absent, so the cursor stays on last_success_end, which is today's behaviour.
- Failed or partial windows are never stamped, so the cursor cannot pass unfinished work; gap alerts and the reloader remain the remediation for failed windows, exactly as today.
- Re-publishing is at-least-once. The downstream transformer must stay idempotent on changeset ids.

## How to test

- New tests cover the tag round-trip, the batched window-store write and list_by_keys, the reporter's last_published_end including corrupt and garbage tag values, the mark-published handler (stamps only covered rows, skips already-stamped, ignores uncovered and unknown keys, idempotent retries), the payload contracts the state machine feeds mark-published on both adapter paths, and the trigger's cursor behaviour with the published cursor enabled and disabled.
- Full adapters suite: 824 passed. mypy and ruff clean on the changed files.
- Terraform validate passes; the state machine changes need a terraform apply after this merges, followed by a small CI PR adding the two new lambdas to the deploy matrix.


catalogue_graph/src/adapters/utils/window_harvester.py
```diff
@@ -26,6 +26,7 @@
 class WindowSummaryTags(BaseModel):
     changeset_ids: list[str] = []
     upserted_record_count: int = 0
+    published_at: str | None = None
     other_tags: dict[str, str] = {}
```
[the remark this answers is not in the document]
kenoir: WindowSummaryTags is a pydantic BaseModel, and pydantic v2 deep-copies mutable defaults per instance, so cross-instance mutation cannot occur here. Leaving the plain defaults as they are.

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
```diff
@@ -66,27 +66,27 @@ def generate_job_id(timestamp: datetime) -> str:
...
+    if cursor_end is None:
         return
-    lag_minutes = (now - last_success_end).total_seconds() / 60
+    lag_minutes = (now - cursor_end).total_seconds() / 60
     if lag_minutes > max_lag_minutes:
         raise RuntimeError(
-            f"{adapter_name.title()} adapter is too far behind: last successful window "
+            f"{adapter_name.title()} adapter is too far behind: last published window "
             f"ended {lag_minutes:.1f} minutes ago (limit={max_lag_minutes})."
         )
```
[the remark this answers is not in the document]
kenoir: Good catch, the cursor can be last_success_end before anything is stamped. Reworded in terms of the cursor in f04abe8386.

catalogue_graph/src/adapters/steps/oai_pmh/mark_published.py
```diff
@@ -0,0 +1,206 @@
...
+        set(event.covered_window_keys) - {row.window_key for row in rows}
+    )
+    if missing_keys:
+        logger.warning(
+            "Covered windows missing from the window store",
+            adapter=runtime.adapter_name,
+            job_id=event.job_id,
+            missing_window_keys=missing_keys,
+        )
+
```
StepanBrychta, line 92: Is a warning sufficient here? It sounds like if there are missing keys, something is definitely wrong, so maybe it would make sense to fail?

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
```diff
@@ -56,6 +57,13 @@ class TriggerRuntime(BaseModel):
     oai_metadata_prefix: str
     oai_set_spec: str | None = None
     adapter_name: str
+    use_published_cursor: bool = False
+    """Resume from the last published window rather than the last loaded one.
+
+    Driven by the PUBLISHED_CURSOR_ENABLED environment variable, which
+    terraform sets from the same enable_published_tracking flag that wires
+    the mark-published step, so the trigger never prefers stamps that
+    nothing is maintaining."""
```
StepanBrychta: Does this need to be a configurable parameter? Folio and Axiell both use the same window store, which handles the published stamps.

In general, I think configurability is a good thing, but I also think that with every configurable option we add, the system grows in complexity and maintainability goes down. It also seems like this toggle has an element of speculative generality, preparing for a hypothetical future OAI-PMH adapter which cannot handle published stamps.

StepanBrychta (approved):
I think this is a nice way of ensuring automatic recovery. It does add a bit of complexity, but I think that's a reasonable trade-off for the added safety.

catalogue_graph/src/adapters/steps/oai_pmh/mark_published.py
kenoir: Agreed, changed to raise in aa270dc962. Failing is safe here: nothing gets stamped, so the next run re-covers and re-stamps the same windows.

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
kenoir: Fair point, removed the toggle in e9126ad23c. The one real distinction is ebsco, which shares the terraform module but has no OAI-PMH window store, and the module already captures that via steps_namespace, so the mark-published resources are now gated on that instead. The env var and the use_published_cursor field are gone: the trigger always prefers the published cursor and already falls back to last_success_end when no stamps exist, so an adapter that never stamps behaves the same.

PR #3451, opened 2026-07-03
Decouple Elasticsearch cluster lifecycle from pipeline stacks

agnesgaroux:
## Add shared infrastructure for the new pipeline stack

Adds the long-lived resources that the new `pipeline_new` module (and its `2026-07-03` root) will depend on. These live outside the pipeline stack so their lifecycle is independent.

### What's added

**ES cluster** (`infrastructure/critical/`)
- New `es-cluster` module that creates an Elastic Cloud deployment + read-only user
- `es_cluster_2026_07_03` instance with outputs (endpoint, credentials, host, port, protocol)

**RDS for id-minter** (`infrastructure/critical/`)
- New Aurora Serverless v2 cluster `identifiers-2026-07-03` for the id-minter
- Exposed via the `id_minter_rds` map output (keyed by date)

**Neptune graph DB** (`catalogue_graph/infra/graph/`)
- New `2026-07-03` Neptune cluster

### Other changes

- `pipeline/README.md` — updated architecture diagram
- `infrastructure/critical/ecr.tf` — removed stale ECR repos for retired services
- `infrastructure/critical/run_terraform.sh` — added Elastic Cloud API key retrieval

### What's NOT in this PR

The `pipeline_new` module and the `2026-07-03` pipeline root will be in a follow-up PR.


infrastructure/critical/es_cluster.tf
```diff
@@ -0,0 +1,16 @@
+module "es_cluster_2026-07-03" {
```
agnesgaroux: That won't work because of the secret names :( 
Ditto below

pipeline/terraform/modules/pipeline/elastic/variables.tf
```diff
@@ -1,4 +1,8 @@
 variable "pipeline_date" { type = string }
+variable "es_cluster_date" {
+  type    = string
+  default = ""
```
agnesgaroux: To ensure the existing secrets' keys are preserved as `elasticsearch/pipeline_storage_2025-10-02`
New secrets will be like `elasticsearch/es-cluster-2026-07-03/service-name-pipeline-date/api_key`

infrastructure/critical/es_cluster.tf
agnesgaroux: Found a workaround

pipeline/terraform/modules/pipeline/elastic_indices/main.tf
```diff
@@ -0,0 +1,211 @@
...
+
+  api_key_versions = {
+    for k, v in local.service_index_permissions : k => module.pipeline_services[k].version
+  }
+}
+
+module "pipeline_services" {
+  for_each            = local.service_index_permissions
+  source              = "../../pipeline_es_api_key"
+  name                = "${each.key}-${var.pipeline_date}"
```
agnesgaroux: Use in secret path, ie cluster-date/service_name-date

pipeline/terraform/modules/pipeline/service_matcher_test.tf
```diff
@@ -2,6 +2,7 @@
 # .buildkite/pipeline.yml -> matcher_test
 
 module "matcher_test" {
+  count        = contains(var.enabled_services, "matcher_test") ? 1 : 0
   source       = "./matcher"
```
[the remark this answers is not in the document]
agnesgaroux: The count guard is temporary, to ensure the existing pipeline is untouched *and* we created a new pipeline without these duplicated services

pipeline/terraform/modules/pipeline/service_merger_test.tf
```diff
@@ -2,6 +2,7 @@
 # .buildkite/pipeline.yml -> merger_test
 
 module "merger_test" {
+  count        = contains(var.enabled_services, "merger_test") ? 1 : 0
   source       = "./merger"
```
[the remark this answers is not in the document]
agnesgaroux: The count guard is temporary, to ensure the existing pipeline is untouched *and* we created a new pipeline without these duplicated services

pipeline/terraform/modules/pipeline_es_api_key/main.tf
```diff
@@ -55,7 +60,7 @@ module "pipeline_service_api_key_secrets" {
   deletion_mode = "IMMEDIATE"
 
   key_value_map = {
-    "elasticsearch/pipeline_storage_${var.pipeline_date}/${var.name}/api_key" = elasticstack_elasticsearch_security_api_key.pipeline_service.encoded
+    "${local.secret_prefix}/${var.name}-${var.pipeline_date}/api_key" = elasticstack_elasticsearch_security_api_key.pipeline_service.encoded
```
kenoir: This `-${var.pipeline_date}` suffix changes the platform secret path for the existing 2025-10-02 pipeline, and the read side doesn't follow.

On `main` all three paths agree: this module writes `.../pipeline_storage_2025-10-02/<service>/api_key` in both accounts, and `elastic/main.tf:243` gives each service `es_apikey = .../pipeline_storage_2025-10-02/<service>/api_key`.

For 2025-10-02, `es_cluster_date` is empty, so `secret_prefix` stays the legacy `.../pipeline_storage_2025-10-02`, but this line now appends `-2025-10-02` to the service name. `elastic/main.tf:243` still reads `.../<service>/api_key` and isn't updated. The secrets module keys each secret on the map key, so the old secret is destroyed and recreated at the new path, and the live platform services (matcher, merger, id_minter, ingestors, graph) end up reading a path that no longer exists.

This also means the 2025-10-02 plan won't be 0/0/0 as the description says. I think that claim predates the secret-path commits, so worth re-running.

Simplest fix: keep the legacy branch of `secret_prefix` writing the old path with no suffix, and only add the `-<date>` disambiguation on the new shared-cluster path.

pipeline/terraform/modules/pipeline/subsystem_graph.tf
```diff
@@ -12,14 +12,14 @@ module "graph_pipeline" {
 
   ecs_cluster_arn = aws_ecs_cluster.cluster.arn
 
-  es_cluster_host     = module.elastic.pipeline_storage_private_host
-  es_cluster_port     = module.elastic.pipeline_storage_port
-  es_cluster_protocol = module.elastic.pipeline_storage_protocol
+  es_cluster_host     = var.elastic_outputs.pipeline_storage_private_host
```
kenoir: For the new 2026-07-03 stack these outputs point at the shared cluster, but I don't think the graph runtime reads secrets from where they now live.

`get_pipeline_config` in `catalogue_graph/src/utils/elasticsearch.py:111` builds its own secret paths from `pipeline_date`:

```
secret_prefix = f"elasticsearch/pipeline_storage_{pipeline_date}"
# {secret_prefix}/private_host, /port, /protocol, /{api_key_name}/api_key
```

For `pipeline_date = 2026-07-03` that resolves to `elasticsearch/pipeline_storage_2026-07-03/...`, and nothing writes that prefix now. The es-cluster module writes host/port/protocol under `elasticsearch/es-cluster-2026-07-03/...`, and the api key lands at `.../es-cluster-2026-07-03/graph_extractor-2026-07-03/api_key`. So graph entrypoints that go through `get_pipeline_config` (pit_opener is one, see `tests/test_environment_settings.py`) resolve secrets that don't exist on the new stack.

The `es_apikey` path passed here via `es_secrets` is correct, so if every graph entrypoint reads its path from env this is fine. But the code that reconstructs the legacy layout needs teaching about the `es-cluster-<date>` prefix and the `<service>-<date>` key name. Worth confirming which path the graph services use at runtime.

Same class of issue for the transformer IAM grants (`modules/pipeline/iam.tf:86,98,110`, which allow `.../pipeline_storage_<date>/transformer/api_key*`) and the ops scripts (`scripts/_common.py:56,74`) once the platform key name gains the `-<date>` suffix noted on `pipeline_es_api_key/main.tf`.

pipeline/terraform/modules/pipeline_es_api_key/main.tf
```diff
@@ -70,6 +75,6 @@ module "pipeline_catalogue_service_api_key_secrets" {
 
   deletion_mode = "IMMEDIATE"
   key_value_map = {
-    "elasticsearch/pipeline_storage_${var.pipeline_date}/${var.name}/api_key" = elasticstack_elasticsearch_security_api_key.pipeline_service.encoded
+    "${local.secret_prefix}/${var.name}/api_key" = elasticstack_elasticsearch_security_api_key.pipeline_service.encoded
```
kenoir: Small thing: this catalogue path keeps `${var.name}` with no `-${var.pipeline_date}`, whereas the platform path just above (line 63) now has it, so the same api key ends up under two different names.

One thing worth thinking about for the shared-cluster model: the `-<date>` suffix is what disambiguates two pipelines sharing a cluster. Since the catalogue path drops it, if a later stack pointed at the same `es_cluster_date` and also exposed `catalogue_api` / `snapshot_generator` / `concepts_api`, both would write `elasticsearch/es-cluster-<cluster_date>/catalogue_api/api_key`. Not an issue today with a single stack, but might be worth aligning the two sides now.

If the catalogue-account consumers are meant to resolve the key without the pipeline_date, that would explain the difference. Could you confirm which way it's intended?

pipeline/terraform/modules/pipeline/elastic_indices/main.tf
```diff
@@ -0,0 +1,211 @@
...
+  pipeline_date       = var.pipeline_date
+  es_cluster_date     = var.es_cluster_date
+  expose_to_catalogue = contains(var.catalogue_account_services, each.key)
+  providers = {
+    aws.catalogue = aws.catalogue
+  }
+}
+
+resource "elasticstack_elasticsearch_ingest_pipeline" "set_indexed_at" {
+  name        = "set-indexed-at"
```
kenoir: Since this module runs against a shared cluster, the fixed name `set-indexed-at` is cluster-global rather than namespaced per stack (ingest pipelines aren't scoped to an index the way the dated index names above are).

If a second stack ever runs `elastic_indices` against the same cluster, both terraform states would manage a pipeline of this name. The PUT is idempotent so applies wouldn't obviously fail, but both states would then own the same object, and tearing one stack down would delete `set-indexed-at` out from under the other. The `works-source-*` indices set `default_pipeline = set-indexed-at` (line 14), so they'd start rejecting writes once it disappeared.

Not an issue with a single stack on the cluster today. Might be worth namespacing it (e.g. `set-indexed-at-${var.pipeline_date}`) so the shared-cluster path stays clean.

pipeline/terraform/modules/pipeline/variables.tf
```diff
@@ -1,7 +1,24 @@
-variable "pipeline_date" {
+variable "pipeline_date" { // nammespace for the pipeline services
   type = string
 }
 
+variable "es_cluster_date" { // the es cluster that the pipeline services use, eg. "es-cluster-2026-07-03"
+  type = string
+}
+
+variable "enabled_services" {
```
kenoir: Reading this alongside the `count = contains(var.enabled_services, "<service>_test")` guards, the variable's surface doesn't quite match its behaviour.

The only `contains()` checks in the module are for `matcher_test`, `merger_test`, and `id_minter_test`. None of the main services in this default (transformers, id_minter, matcher, merger, image_inferrer, graph_pipeline) are guarded, so listing them has no effect, and the description "Omit a service to skip it entirely" reads as if any of them can be toggled.

It works for both stacks as written: 2025-10-02 sets the three `_test` names, and 2026-07-03 keeps the default (which contains none of the `_test` names, so they're skipped). Two things that might be worth tidying:
- the default listing main-service names that are never read is a bit of a trap, since it looks like the authoritative on/off list
- if the main services later gain the same guard to match the description, 2025-10-02 (whose value omits them all) would tear its whole pipeline down

Might be clearer to name this `enabled_test_services` and default it to the tests you want on, or wire the guard for the main services too.

infrastructure/critical/modules/es-cluster/main.tf
```diff
@@ -0,0 +1,88 @@
+locals {
+  # deployment_name = replace(var.cluster_date, "_", "-")
```
kenoir: Couple of tiny things, non-blocking:

- This line is a leftover commented-out `deployment_name` local. Safe to drop.
- The elasticstack/ec provider versions are split across the two sides that now manage the same cluster. This stack (`infrastructure/critical`) pins `elasticstack 0.16.1` / `ec 0.13.0`, while `elastic_indices` and the 2026-07-03 stack pin `elasticstack 0.7.0`. Both are legal since they're separate root modules, but 0.7.0 is quite old for the brand-new index module. Might be worth a conscious call on whether the new module should track the newer provider rather than inherit 0.7.0 from the old stack.

pipeline/terraform/modules/pipeline/elastic/variables.tf
agnesgaroux: Nope

pipeline/terraform/modules/pipeline/elastic_indices/main.tf
agnesgaroux: Reverted

pipeline/terraform/modules/pipeline/variables.tf
```diff
@@ -1,7 +1,24 @@
...
+  description = "Set of services to create in this stack. Omit a service to skip it entirely."
+  default = [
+    "transformers",
+    "id_minter",
+    "matcher",
+    "merger",
+    "image_inferrer",
+    "graph_pipeline",
+  ]
+}
```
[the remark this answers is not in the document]
agnesgaroux: I considered count guards everywhere to list precisely which services to created but decided against it. Forgot to update the variable!

pipeline/terraform/modules/pipeline/variables.tf
agnesgaroux: Forgot to correct this after deciding against count guarding every module

infrastructure/critical/rds_id_minter.tf
```diff
@@ -25,5 +25,20 @@ module "id_minter_rds_test" {
...
 
+  master_username = data.aws_ssm_parameter.rds_username.value
+}
+
+module "id_minter_rds_test_2026_07_03" {
+  source = "./modules/id-minter-rds"
+
+  name_suffix = "2026-07-03"
+  # Restore from production, use recent snapshot
+  snapshot_identifier = ""
```
agnesgaroux: Add latest snapshot identifier before applying TF

infrastructure/critical/modules/es-cluster/main.tf
agnesgaroux: Upgraded to elasticstack 0.16.1

pipeline/terraform/modules/pipeline_es_api_key/main.tf
agnesgaroux: Reverted to elasticsearch/pipeline_storage_${var.pipeline_date} secret string to keep consumers happy

pipeline/terraform/modules/pipeline/subsystem_graph.tf
agnesgaroux: See above

pipeline/terraform/modules/pipeline_es_api_key/main.tf
agnesgaroux: Secrets now namespaced with pipeline date

infrastructure/critical/modules/es-cluster/main.tf
```diff
@@ -0,0 +1,53 @@
+locals {
+  elastic_id     = ec_deployment.cluster.elasticsearch.resource_id
+  elastic_region = ec_deployment.cluster.elasticsearch.region
+}
```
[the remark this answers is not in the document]
agnesgaroux: The ec provider ≥ 0.6.0 changed elasticsearch from a single-item list to an object (breaking change in that release). We're on v0.13.0, so .elasticsearch.resource_id is correct — no [0] indexing needed. Plan succeeds as-is.
See [https://github.com/elastic/terraform-provider-ec/releases/tag/v0.6.0](https://github.com/elastic/terraform-provider-ec/releases/tag/v0.6.0)

infrastructure/critical/modules/es-cluster/outputs.tf
```diff
@@ -0,0 +1,39 @@
+output "https_endpoint" {
+  value     = ec_deployment.cluster.elasticsearch.https_endpoint
+  sensitive = true
+}
```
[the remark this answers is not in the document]
agnesgaroux: Same as above

infrastructure/critical/modules/es-cluster/provider.tf
```diff
@@ -0,0 +1,25 @@
...
+      source = "hashicorp/random"
+    }
+  }
+}
+
+provider "elasticstack" {
+  elasticsearch {
+    username  = ec_deployment.cluster.elasticsearch_username
+    password  = ec_deployment.cluster.elasticsearch_password
+    endpoints = [ec_deployment.cluster.elasticsearch.https_endpoint]
```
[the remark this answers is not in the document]
agnesgaroux: The ec provider ≥ 0.6.0 changed elasticsearch from a single-item list to an object

infrastructure/critical/modules/es-cluster/outputs.tf
```diff
@@ -0,0 +1,39 @@
...
+  value = "${local.elastic_id}.vpce.${local.elastic_region}.aws.elastic-cloud.com"
+}
+
+output "public_host" {
+  value = "${local.elastic_id}.${local.elastic_region}.aws.found.io"
+}
+
+output "port" {
+  value = reverse(split(":", ec_deployment.cluster.elasticsearch.https_endpoint))[0]
+}
```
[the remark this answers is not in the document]
agnesgaroux: The ec provider ≥ 0.6.0 changed elasticsearch from a single-item list to an object

infrastructure/critical/modules/es-cluster/outputs.tf
```diff
@@ -0,0 +1,39 @@
...
+
+output "protocol" {
+  value = split(":", ec_deployment.cluster.elasticsearch.https_endpoint)[0]
+}
```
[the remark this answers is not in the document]
agnesgaroux: The ec provider ≥ 0.6.0 changed elasticsearch from a single-item list to an object

catalogue_graph/infra/graph/neptune.tf
```diff
@@ -39,6 +39,21 @@ module "catalogue_graph_neptune_cluster_dev" {
   }
 }
 
+module "catalogue_graph_neptune_cluster_2026_07_03" {
+  source = "./modules/catalogue_graph"
+
+  graph_date                 = "2026-07-03"
+  namespace                  = local.namespace
```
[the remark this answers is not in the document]
agnesgaroux: Done

StepanBrychta (approved):

PR #3454, opened 2026-07-06
Emit stub source concept nodes from the incremental concepts transformer

kenoir:
> Do not merge yet. Merging auto-deploys to prod. This PR is up so the team can judge whether the approach is worth the complexity before we commit to it.

## What does this change?

The incremental graph pipeline failed on 2026-07-06 (execution `0904923d-c17a-4a14-894e-be12337f0d1b` of `graph-pipeline-incremental-2025-10-02`). The `catalogue_concepts` transformer label-matches concepts against the LoC/MeSH bulk load CSVs in S3. The monthly pipeline refreshes those CSVs hours before it bulk loads the corresponding nodes into Neptune, and the full run takes about 22h. During that gap an incremental window matched the concept "Laboratory manuals" to a newly minted LCSH heading (`sh2026001150`) that wasn't in Neptune yet. The edge load failed with `FROM_OR_TO_VERTEX_ARE_MISSING`, and one bad edge is enough to exceed the `1e-4` insert error threshold on a small window. This recurs on every monthly run. Relaxing the threshold instead would silently drop the edge forever, because nothing re-loads catalogue edges in full.

The fix: the transformer now also emits a minimal stub node (id, label, source) for every matched source concept into the same window's nodes CSV. The nodes CSV loads before the edges CSV, so edge targets always exist. The monthly full load enriches stubs in place: loads run with `updateSingleCardinalityProperties=TRUE` and empty CSV cells never clear properties.

Stubs carry the same node label (`SourceConcept` / `SourceName` / `SourceLocation`) as the full node that later enriches them, tracked per id in `IdLabelChecker`. `IdLabelChecker` now keeps labels in their original casing (lookups still lowercase at comparison time). Without that, a stub upserted over an already-enriched node would rewrite "United States" as "united states".

Both edges and stubs now derive from a new shared `RawCatalogueConcept.matched_source_concepts()`, so the two sets can't drift. The stub models share `Concept`'s exact field set so `Concept` and stub rows can live in one bulk load CSV (the header is derived from the first row).

Some honest trade-offs worth weighing:

- The nodes pass now builds `IdLabelChecker` too (it previously only ran in the edges pass). That means downloading the multi-GB LoC/MeSH CSVs every 15 minutes in the nodes extractor task as well. It's the same 16GB task envelope the edges pass already uses.
- There's a residual rarer race. The nodes and edges passes build their checkers from S3 independently, so a monthly CSV rewrite landing between the two reads could still produce an edge without a stub.
- Stub-only nodes serve sparse concept docs (no description or alternative labels) until the next monthly enrichment.

## How to test

Run the `catalogue_graph` suite from `catalogue_graph/`:

```
uv run --group dev pytest
uv run --group dev ruff check .
uv run --group dev ruff format --check .
uv run --group dev mypy .
```

The full suite passes (1428 tests) and ruff is clean. mypy reports pre-existing errors in `src/utils/marc.py` only, unrelated to this change.

New tests cover: a stub emitted per matched source id with an original-case label; the stub set equals the edge target set; dedup when two concepts match the same id; mixed `Concept` / `SourceConcept` rows in one CSV under a single header; original-case `get_label`; and per-id transformer tracking in the transformer.

The incremental graph pipeline stops failing with `FROM_OR_TO_VERTEX_ARE_MISSING` around monthly source ontology refreshes. No new `HAS_SOURCE_CONCEPT` edge load errors from missing target vertices, and the previously dropped concept-to-heading edges (e.g. "Laboratory manuals" to `sh2026001150`) land in the window they were matched in rather than being lost.

- Merging auto-deploys to prod, so this stays unmerged until the team is happy with the approach.
- The nodes extractor task now downloads the same multi-GB CSVs every 15 minutes. It reuses the 16GB envelope the edges pass already runs in, but it's extra work and cost on the nodes pass.
- The narrow race between the two independent S3 reads remains. A monthly CSV rewrite landing between the nodes read and the edges read could still produce an edge without a stub. Rarer than the original window-vs-Neptune gap, but not zero.
- Stub-only nodes are sparse until the next monthly enrichment, so anything reading those concepts sees id, label and source but no description or alternative labels in the meantime.
- The label-casing change touches `IdLabelChecker` used across the catalogue graph transformers, not just stubs. Lookups still lowercase at comparison time, and the mixed-case tests guard against regressions there.

### DIFF
--- a/catalogue_graph/src/graph/converters/cypher/bulk_load_converter.py
+++ b/catalogue_graph/src/graph/converters/cypher/bulk_load_converter.py
@@ -33,7 +33,7 @@ class CypherBulkLoadConverter(CypherBaseConverter):
         return f"{field_name}:{field_type}"
 
     def _node_to_bulk_cypher(self, model: BaseNode) -> dict:
-        bulk_node = {":ID": model.id, ":LABEL": type(model).__name__}
+        bulk_node = {":ID": model.id, ":LABEL": type(model).bulk_load_label()}
 
         for field_name, raw_value in model.dict().items():
             column_header = self._get_bulk_loader_column_header(model, field_name)
--- a/catalogue_graph/src/graph/transformers/catalogue/concepts_transformer.py
+++ b/catalogue_graph/src/graph/transformers/catalogue/concepts_transformer.py
@@ -9,11 +9,27 @@ from graph.sources.catalogue.concepts_source import (
 from graph.transformers.graph_transformer import GraphBaseTransformer
 from models.events import BasePipelineEvent
 from models.graph_edge import ConceptHasSourceConcept, ConceptHasSourceConceptAttributes
-from models.graph_node import Concept
+from models.graph_node import (
+    Concept,
+    SourceConceptStub,
+    SourceLocationStub,
+    SourceNameStub,
+)
 from utils.ontology import get_transformers_from_ontology
+from utils.types import TransformerType
 
-from .id_label_checker import IdLabelChecker
+from .id_label_checker import IdLabelChecker, concept_source_from_id
 from .raw_concept import RawCatalogueConcept
+
+# The node label each source ontology transformer bulk loads its nodes under, so that stubs
+# minted here carry the same label as the full node which later enriches them in place.
+STUB_CLASSES_BY_TRANSFORMER: dict[TransformerType, type[SourceConceptStub]] = {
+    "loc_concepts": SourceConceptStub,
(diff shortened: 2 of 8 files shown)


kenoir:
Closing in favour of the simpler threshold override in the follow-up PR. The review found the stub approach doubles the ontology CSV download work in every incremental window and leaves orphaned stub nodes with no cleanup path, which we decided was not worth it for a failure that a threshold override handles. Keeping the branch around for reference.

PR #3456, opened 2026-07-06
Add tflint to the Terraform workflow

anubhavijay:
## What does this change?

Adds [tflint](https://github.com/terraform-linters/tflint) as a Terraform linting step for the repository.

Four files are added or modified:

### `builds/run_tflint.sh` — shared tflint helper (new)

Extracts the Terraform-root discovery and lint loop into a single script so CI and local developer instructions share one source of truth and can't drift independently again.

- Discovers all Terraform roots by looking for files that declare a `backend` block (root modules only, not child modules). Excludes `.terraform/` directories so locally-cached provider modules can't be picked up as false roots.
- Accepts optional positional arguments to target specific roots (e.g. `builds/run_tflint.sh calm_adapter/terraform`) for faster local runs.
- Runs `tflint --init` then lints each root with `--minimum-failure-severity=error` so warnings don't fail CI.
- Skips `reindexer/terraform` (with a comment) because a shared module it references fails static evaluation without provider initialisation.

### `.tflint.hcl` — repo-level tflint configuration

- Module evaluation disabled (`call_module_type = "none"`) so static analysis doesn't follow cross-module references that require provider credentials to resolve.
- `terraform` plugin enabled with the `recommended` preset.
- `aws` ruleset pinned to v0.48.0 for reproducible runs.
- `terraform_naming_convention` enabled with an explicit comment noting it is **advisory**: the rule emits Notice-severity findings, which are logged but do not fail CI under the current `--minimum-failure-severity=error` threshold. (Dated module labels like `sierra-adapter-20200604` are a repo-wide pattern; enforcement would need a cleanup pass first.) The redundant `format = "snake_case"` default is omitted.

### `.github/workflows/tflint.yml` — CI workflow

Triggers on:
- Pull requests that touch any `.tf` file, `.terraform.lock.hcl`, `.tflint.hcl`, or the workflow itself.
- Pushes to `main` on the same paths.
- Manual `workflow_dispatch`.

The workflow now delegates entirely to `builds/run_tflint.sh` — the inline discovery loop has been removed. tflint pinned to v0.64 (updated from v0.54.0).

### `AGENTS.md` — local instructions for agents and developers

- Terraform guidance promoted to its own `## Terraform conventions` section (previously nested under `## Python conventions`).
- Local command simplified to `builds/run_tflint.sh`, matching CI exactly.

## How to test

1. Open a PR that modifies a `.tf` file and observe the **Terraform: tflint** check running.
2. Introduce a deliberate error (e.g. reference an undefined variable) and confirm the check fails.
3. Introduce a tflint warning (not an error) and confirm the check still passes.
4. Locally, from the repo root:
   ```bash
   builds/run_tflint.sh
   ```
   Or target a single root:
   ```bash
   builds/run_tflint.sh calm_adapter/terraform
   ```

- The `tflint` check appears in the GitHub Actions tab for PRs that touch Terraform files.
- The check passes on `main` after merge.
- Future Terraform PRs surface errors that would previously only be caught in `terraform plan`.

- **False positives causing blocked PRs.** Mitigated by `--minimum-failure-severity=error`; only genuine errors block the check, not style warnings.
- **New failures on existing code.** The workflow was validated against all current Terraform roots before this PR was opened; all pass (aside from `reindexer/terraform` which is skipped with a comment).
- **Plugin version drift.** The aws ruleset is pinned to v0.48.0 in `.tflint.hcl` and tflint itself is pinned to v0.64 in the workflow, giving reproducible results until intentionally upgraded.
- **Local module cache false-positives.** The discovery grep excludes `.terraform/` directories, so cached provider modules cannot be picked up as Terraform roots.

### DIFF
--- a/.github/workflows/tflint.yml
+++ b/.github/workflows/tflint.yml
@@ -1,0 +1,60 @@
+name: "Terraform: tflint"
+
+on:
+  pull_request:
+    paths:
+      - "**/*.tf"
+      - "**/.terraform.lock.hcl"
+      - ".tflint.hcl"
+      - ".github/workflows/tflint.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "**/*.tf"
+      - "**/.terraform.lock.hcl"
+      - ".tflint.hcl"
+      - ".github/workflows/tflint.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  tflint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: terraform-linters/setup-tflint@v4
+        with:
+          tflint_version: v0.54.0
+
+      - name: Initialise tflint plugins
+        run: tflint --init
+
+      - name: Run tflint on all Terraform roots
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          config_file="$(pwd)/.tflint.hcl"
+
+          mapfile -t terraform_roots < <(
+            find . -type f -name '.terraform.lock.hcl' -exec dirname {} \; | sort -u
+          )
+
+          if [[ ${#terraform_roots[@]} -eq 0 ]]; then
+            echo "No Terraform roots found"
+            exit 1
+          fi
+
+          for root in "${terraform_roots[@]}"; do
+            if [[ "${root}" == "./reindexer/terraform" ]]; then
+              echo "Skipping ${root}; this root currently fails static evaluation in a shared module"
+              continue
+            fi
+
+            echo "Linting ${root}"
+            tflint --chdir="${root}" --config="${config_file}" --format=compact --minimum-failure-severity=error
+          done
--- a/.github/workflows/tflint.yml
+++ b/.github/workflows/tflint.yml
@@ -40,8 +40,14 @@ jobs:
 
           config_file="$(pwd)/.tflint.hcl"
 
+          # A Terraform root is identified by having a backend configuration.
+          # Child modules (under modules/, etc.) never declare a backend, so this
+          # discovers every root without depending on a committed .terraform.lock.hcl.
           mapfile -t terraform_roots < <(
-            find . -type f -name '.terraform.lock.hcl' -exec dirname {} \; | sort -u
+            grep -rlE '^\s*backend\s+"' --include='*.tf' . \
+              | xargs -r -n1 dirname \
+              | sed 's|^\./||' \
+              | sort -u
           )
 
           if [[ ${#terraform_roots[@]} -eq 0 ]]; then
@@ -50,7 +56,7 @@ jobs:
           fi
 
           for root in "${terraform_roots[@]}"; do
-            if [[ "${root}" == "./reindexer/terraform" ]]; then
+            if [[ "${root}" == "reindexer/terraform" ]]; then
               echo "Skipping ${root}; this root currently fails static evaluation in a shared module"
               continue
             fi
--- a/.github/workflows/tflint.yml
+++ b/.github/workflows/tflint.yml
@@ -28,39 +28,8 @@ jobs:
 
       - uses: terraform-linters/setup-tflint@v4
         with:
-          tflint_version: v0.54.0
-
-      - name: Initialise tflint plugins
-        run: tflint --init
+          tflint_version: v0.64
 
       - name: Run tflint on all Terraform roots
         shell: bash
-        run: |
-          set -euo pipefail
-
-          config_file="$(pwd)/.tflint.hcl"
-
-          # A Terraform root is identified by having a backend configuration.
-          # Child modules (under modules/, etc.) never declare a backend, so this
-          # discovers every root without depending on a committed .terraform.lock.hcl.
-          mapfile -t terraform_roots < <(
-            grep -rlE '^\s*backend\s+"' --include='*.tf' . \
-              | xargs -r -n1 dirname \
-              | sed 's|^\./||' \
-              | sort -u
-          )
-
-          if [[ ${#terraform_roots[@]} -eq 0 ]]; then
-            echo "No Terraform roots found"
-            exit 1
-          fi
-
-          for root in "${terraform_roots[@]}"; do
-            if [[ "${root}" == "reindexer/terraform" ]]; then
-              echo "Skipping ${root}; this root currently fails static evaluation in a shared module"
-              continue
-            fi
-
-            echo "Linting ${root}"
-            tflint --chdir="${root}" --config="${config_file}" --format=compact --minimum-failure-severity=error
-          done
+        run: builds/run_tflint.sh
--- a/.github/workflows/tflint.yml
+++ b/.github/workflows/tflint.yml
@@ -28,7 +28,7 @@ jobs:
 
       - uses: terraform-linters/setup-tflint@v4
         with:
-          tflint_version: v0.64
+          tflint_version: v0.64.0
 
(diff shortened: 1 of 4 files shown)


.tflint.hcl
```diff
@@ -0,0 +1,21 @@
...
+  version = "0.48.0"
+  source  = "github.com/terraform-linters/tflint-ruleset-aws"
+}
+
+# Enforce the recommended Terraform naming convention (snake_case, the
+# rule default) for resources, variables, outputs, etc.
+rule "terraform_naming_convention" {
+  enabled = true
+  format  = "snake_case"
+}
```
kenoir, line 21: concern: I don't think this rule can ever fail the build as configured. `terraform_naming_convention` emits Notice-severity issues (https://github.com/terraform-linters/tflint-ruleset-terraform/blob/main/docs/rules/terraform_naming_convention.md), and the workflow runs with `--minimum-failure-severity=error`, which only gates the exit code at Error. You can see this in the tflint run on this PR: it passed while printing `module name "sierra-adapter-20200604" must match the following format: snake_case`. So right now the rule is advisory output in the logs only. Is that the intent? If so a comment saying it's advisory would help; if you want it enforced, tflint has no per-rule severity override, so the options are lowering the failure threshold (which would also make the existing warnings fail) or accepting it as-is. Note also that dated module labels like `sierra-adapter-20200604` are a repo-wide pattern, so enforcement would need a cleanup pass first. (Minor: `format = "snake_case"` is the rule default, so that line is redundant.)

.github/workflows/tflint.yml
```diff
@@ -0,0 +1,66 @@
...
+
+          for root in "${terraform_roots[@]}"; do
+            if [[ "${root}" == "reindexer/terraform" ]]; then
+              echo "Skipping ${root}; this root currently fails static evaluation in a shared module"
+              continue
+            fi
+
+            echo "Linting ${root}"
+            tflint --chdir="${root}" --config="${config_file}" --format=compact --minimum-failure-severity=error
+          done
```
kenoir: suggestion: The discovery-and-loop logic now lives in two places (this workflow and AGENTS.md), and they already drifted once during this PR (the AGENTS.md snippet used lockfile discovery until the last commit, which is what Copilot flagged). Would it be worth extracting this into a script, e.g. `builds/run_tflint.sh`, and having both CI and the AGENTS.md instructions call it? The repo already keeps helper scripts in `builds/` for exactly this kind of thing.

AGENTS.md
```diff
@@ -74,6 +74,29 @@ uv run ruff check --fix
...
+### Before finalising a Terraform change
+
+Run from the repository root:
+
+```bash
+tflint --init
+
+config_file="$(pwd)/.tflint.hcl"
+
+grep -rlE '^\s*backend\s+"' --include='*.tf' . \
```
kenoir: suggestion: Unlike CI, a local working copy has `.terraform/` module caches after `terraform init` (mine currently holds ~14k cached `.tf` files). None of them declare a backend today so the command happens to work, but any fetched module that ships an example root with a backend block would get picked up as a false root. Adding `--exclude-dir=.terraform` would make this robust (and faster). Same flag would be harmless in the workflow for consistency.

AGENTS.md
```diff
@@ -74,6 +74,29 @@ uv run ruff check --fix
 Scope `pytest`/`mypy` paths to the area you changed when the project is large
 (e.g. `catalogue_graph/`); a full run can be slow.
 
+### Before finalising a Terraform change
```
kenoir, line 80: nit: This `### Before finalising a Terraform change` heading sits inside the `## Python conventions` section. Since it's not Python-related, would it be better as its own `## Terraform conventions` section (there are already parallel `## Python conventions` and `## Scala conventions` sections)?

.github/workflows/tflint.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  tflint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: terraform-linters/setup-tflint@v4
+        with:
+          tflint_version: v0.54.0
```
kenoir: nit: v0.54.0 is from November 2024 and the current release is v0.63.1, while the aws ruleset is pinned to the newest release (0.48.0, June 2026). It clearly works together since the check passed on this PR, so no blocker, but since this is a fresh introduction it might be worth starting from a current tflint pin rather than a 19-month-old one.

kenoir (commented):
Nice addition. The backend-block root discovery is a neat way to find roots without relying on committed lockfiles, and I checked it finds exactly the same 10 roots as lockfile discovery with no false matches. Pinning both tflint and the aws ruleset is the right call, and the check already runs green on this PR.

Main thing to decide before merge: the naming-convention rule cannot fail CI at `--minimum-failure-severity=error` (it emits Notice-level issues), so as written it is advisory log output only. Details in the line comment. The rest is small stuff: deduplicating the loop between CI and AGENTS.md, excluding `.terraform` dirs from the local grep, and the heading placement in AGENTS.md.

One small thing on the PR description itself: the "How to test" step 4 still shows the old lockfile-based discovery command; the workflow and AGENTS.md now use backend-block discovery. Worth updating so future readers copy the right command.

.tflint.hcl
anubhavijay: Kept this advisory for now since so that dated repo-wide pattern will not need to be changed, Would be good to fix the other warnings reported on the tflint so that we dont ignore the naming conventions advisory or further warnings for the newer added infra

.github/workflows/tflint.yml
anubhavijay: implemented

.github/workflows/tflint.yml
anubhavijay: using 0.64.0 newest version now

AGENTS.md
```diff
@@ -74,6 +74,21 @@ uv run ruff check --fix
...
+
+```bash
+tflint --init
+find . -type f -name '.terraform.lock.hcl' -exec dirname {} \; | sort -u | while IFS= read -r root; do
+  if [[ "$root" == "./reindexer/terraform" ]]; then
+    continue
+  fi
+
+  tflint --chdir="$root" --config="$(pwd)/.tflint.hcl" --format=compact --minimum-failure-severity=error
+done
```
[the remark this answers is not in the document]
anubhavijay: outdated now

builds/run_tflint.sh
```diff
@@ -0,0 +1,54 @@
...
+else
+  # A Terraform root is identified by having a backend configuration.
+  # Child modules (under modules/, etc.) never declare a backend, so this
+  # discovers every root without depending on a committed .terraform.lock.hcl.
+  mapfile -t terraform_roots < <(
+    grep -rlE '^\s*backend\s+"' --exclude-dir='.terraform' --include='*.tf' . \
+      | xargs -r -n1 dirname \
+      | sed 's|^\./||' \
+      | sort -u
+  )
```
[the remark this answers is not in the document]
anubhavijay: Fixed

kenoir (approved):
Thanks for the follow-up changes. The extracted `builds/run_tflint.sh` is a nice result: CI and the local instructions now share one code path, the root discovery regex is POSIX-safe, and `.terraform/` caches are excluded. I re-ran the discovery grep locally and it correctly finds all 11 current roots, including the new `pipeline/terraform/2026-07-03` stack that landed after my first review. The advisory note on `terraform_naming_convention` in `.tflint.hcl` makes the intent clear. Approving.

PR #3458, opened 2026-07-07
Replace graph environment with graph_date

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6431

Infrastructure changes:
* Replace graph environments (`prod` or `dev`) with graph dates, functioning similarly to pipeline dates: Each Neptune cluster (and other graph infrastructure) is dated, and any cluster can become the production graph. Graph dates should be in the standard format (e.g. 2026-01-01), except for the special `graph_date=dev` cluster, which is reserved for experimentation and cannot be productionised.
* Remove the `wellcomecollection-catalogue-graph-dev` S3 bucket: The number of dated clusters we can provision is now unbounded, so all of them should use the same bucket, instead of a new bucket being provisioned for each cluster. Data isolation is handled by reserving a prefix for each cluster (`/graph-{graph_date}`) and only allowing Neptune bulk loader to read from the reserved prefix. 
* Remove the `graph_scaler` and `graph_status_poller` services, since we no longer scale the graph as part of each pipeline run.

> [!NOTE]  
> To prevent the current production cluster from being recreated, it intentionally has a blank graph date to preserve its ID. When we switch to a new (dated) production cluster, we should make non-blank graph dates mandatory.

Pipeline (Python) changes:
* Standardise S3 paths across all services. Each pipeline run now reads from and writes to S3 paths like `graph-{graph_date}/pipeline-{pipeline_date}/{service_prefix(es)}/{scope}`, where
    * `graph_date` identifies the catalogue graph (Neptune cluster)
    * `pipeline_date` identifies the Elasticsearch pipeline cluster 
    * `service_prefix` identifies the service (bulk loader, ingestor, etc.) and, optionally, Elasticsearch index dates (e.g. `index-2026-01-01`).
    * `scope` reflects the pipeline run mode (window-based, ID-based, full)

Jupyter notebook changes:
* Add a new notebook for copying monthly bulk load files from one S3 location to another. This will become useful once we start creating new clusters, allowing us to populate them with LoC/Wikidata/MeSH data without running the full monthly pipeline.

## How to test

Run all affected services locally and verify that:
* Artefacts are written to and read from the correct S3 paths
* The correct Neptune cluster is used

We should be able to provision new graph clusters and turn them into production clusters as needed.

## Release plan

To prevent the current production cluster from being recreated, it intentionally has a blank graph date to preserve its ID. When we switch to a new (dated) production cluster, we should make non-blank graph dates mandatory.

To release this change, incremental graph pipeline runs need to be paused. This is because the new S3 paths are not backwards compatible, and if the new code was released in the middle of an active pipeline run, new services would read from and write to S3 locations not compatible with the old services.

Release steps:
1. Pause graph pipeline schedule and wait for any pending runs to complete
2. Apply Terraform changes and merge this PR
3. Manually backfill any missed windows
4. Enable graph pipeline schedule
5. Wait a few days, then remove the old artefacts from the S3 bucket, so that all files are stored under the new path structure


catalogue_graph/src/utils/argparse.py
```diff
@@ -113,8 +112,8 @@ def validate_es_mode_for_writes(
     """
     When running services which write to an Elasticsearch index locally, we disallow combining
     the production Elasticsearch cluster (`es_mode=public` or `es_mode=private`) with the development
-    Neptune cluster (`environment=dev`) so that we cannot accidentally write non-production data to
+    Neptune cluster (`graph_date=dev`) so that we cannot accidentally write non-production data to
     production indexes.
```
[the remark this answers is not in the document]
StepanBrychta: I think the docstring is fine. Yes, `es_mode=private` is not allowed in local runs (since it only works in our VPC), but the doctoring doesn't contradict that. It just says that both `es_mode=public` and `es_mode=private` connect to the production Elasticsearch cluster.

catalogue_graph/notebooks/copy_monthly_bulk_load_files.ipynb
```diff
@@ -0,0 +1,110 @@
+{
```
agnesgaroux, line 1: Can we describe why we have this, ie. load monthly run artefacts into a new graph to avoid running the monthly graph pipeline

catalogue_graph/src/graph/steps/bulk_load_poller.py
```diff
@@ -56,22 +56,28 @@ def print_detailed_bulk_load_errors(payload: BulkLoadStatusResponse) -> None:
...
         r"(?P<transformer_type>[^/]+)__(?P<entity_type>[^/]+)\.csv$"
     )
 
     m = regex.match(s3_uri)
     if not m:
         raise ValueError(f"S3 uri '{s3_uri}' does not match the expected format.")
 
+    uri_graph_date = m.group("graph_date")
+    # This should never happen, but validate just in case
+    if uri_graph_date != (graph_date or "prod"):
```
kenoir, line 76: Following up on Copilot's comment on this check: I think it is actually correct for the legacy prod cluster. The comparison is against `(graph_date or "prod")`, so an empty `graph_date` normalises to `"prod"`, and the S3 path is written under `graph-prod/` via `BasePipelineEvent.s3_prefix_parts`. For `graph_date=""` the URI segment is `prod` and the caller side also resolves to `prod`, so they match and no error is raised. I do not think a change is needed here.

kenoir (commented):
I think its a blocker that the image inferrer state machine does not pass graph_date.

`graph_date` is now a required field on `GraphPipelineScope` with no default, so every pipeline event inherits it as required, including `FindWorkEvent`.

The image inferrer state machine (`pipeline/terraform/modules/pipeline/state_machine_image_inferrer.tf`) is not part of this PR, so its `ConstructEvent` state was not updated. It still emits only `pipeline_date`, `index_dates`, and `window`, and never sets `graph_date`. That output feeds `find_work.lambda_handler`, which builds the event with `FindWorkEvent(**event)` (`find_work.py:102`), so the lambda will raise a pydantic `ValidationError` on every run once this ships. The image inferrer runs on its own schedule, separate from the graph pipeline, so it fails on its own.

The `pipeline` module already receives `graph_date` (default `""`), so the fix is to add one line to the `ConstructEvent` output:

```hcl
"graph_date" : var.graph_date,
```

That produces `graph_date=""`, which matches the legacy prod convention the 2025-10-02 pipeline uses everywhere else. It is probably worth checking whether any other event producers outside the graph state machine files were missed in the same way.

catalogue_graph/src/models/events.py
```diff
@@ -73,6 +56,45 @@ def from_argparser(cls, args: argparse.Namespace) -> Self:
...
+    def s3_prefix_parts(self) -> list[str]:
+        """Build the S3 path prefix for this run's output files.
+
+        All services share the same top-level layout:
+
+            graph-{graph_date}/pipeline-{pipeline_date}/{service_prefix(es)}/{scope}
+
+        where:
+            - ``graph_date`` identifies the Neptune graph cluster (temporarily defaults to ``prod``)
+            - ``pipeline_date`` identifies the Elasticsearch pipeline cluster
```
agnesgaroux: The ES cluster will be persistent going forward, so pipeline_date is the namespace for the pipeline services

catalogue_graph/notebooks/copy_monthly_bulk_load_files.ipynb
StepanBrychta: Good idea, I added a new paragraph to the README file explaining what it does.

catalogue_graph/src/graph/steps/bulk_load_poller.py
StepanBrychta: Copilot's comment was referring to an older version of this code, which didn't have the 'prod' fallback, so I added it.

StepanBrychta:
@kenoir Well spotted, I added the missing `graph_date`. I think all state machines should have it now. But if there is a service which doesn't, Pydantic should protect us from any unexpected behaviour. The service would fail instantly before doing any damage.

catalogue_graph/src/models/events.py
StepanBrychta: I assumed we would keep the Elasticsearch clusters dated (even if we have just one of them most of the time), just like we do with the graph cluster now. If not, happy to update the docstring.

StepanBrychta: Actually, I suppose that dating the clusters would require yet another date in the S3 path. Maybe that's something to discuss later. For now, updating the docstring makes sense.

kenoir (approved):

catalogue_graph/src/models/events.py
agnesgaroux: The service writes/reads to S3 so namespacing by service name should be enough? It's more like 
``pipeline_date`` identifies the service that uses the prefix

StepanBrychta: Yes, I think it's enough for now. I was thinking about a scenario where we have multiple dated Elasticsearch clusters. In that case, it might make sense to include the Elasticsearch cluster date in the S3 path to keep data between clusters isolated. The idea is that every S3 path would be fully qualified, isolating artefacts belonging to the different clusters and indexes. 

That's probably not something we have to worry about now, especially if we're not planning to make the Elasticsearch cluster dated.

PR #3459, opened 2026-07-08
Add a reconciliation audit for OAI-PMH adapters

kenoir:
## What does this change?

Windowed OAI-PMH harvesting cannot detect a bulk source load whose records carry backdated datestamps. Normal harvest lag self-heals, because the trigger resumes from the last successful window and catches recent windows up within the lag window. But records loaded with datestamps that fall inside windows which already ran and were marked successful sit behind the cursor and are never picked up. The Axiell load in early July 2026 was exactly this, and it went unnoticed for days.

This adds a reconciliation audit for the OAI-PMH adapters. It compares a source-of-truth record count against the number of active records in the adapter store and emits a `drift_count` metric (CloudWatch namespace `catalogue_adapters`, dimension `adapter_type`). Positive drift means the source holds records the store is missing.

The audit is run on demand rather than on a schedule. We are comfortable with the adapter falling behind on recent windows for up to the lag window, because that case self-heals. The audit is for the case that does not self-heal, and it emits its metric every time it runs, so it can be attached to a schedule and an alarm later without any code change.

The source count is adapter-specific. The base adapter reads the OAI `completeListSize`. Axiell overrides it with the Axiell WebAPI (`wwwopac.ashx`), which is served by the same application as the OAI endpoint and accepts the same token, and gives a direct database count that is unaffected by faults in the OAI module. When drift is found the audit can log a sample of the specific ids present at the source but missing from the store (Axiell enumerates ids statelessly through the WebAPI). Those ids can be fed to the recover step in the follow-up PR.

The audit is one-directional by design: it detects records missing from the store, not stale records lingering in it.

New files: the WebAPI client (`webapi_client.py`) and the reconcile step (`reconcile.py`). Modified: `runtime.py` adds `source_of_truth_count` and `enumerate_source_ids` with Axiell overrides; `adapter_store.py` adds `count_active_namespace_records` (an id-only scan so a whole-table count stays memory-flat); `reporting.py` adds `OAIPMHReconcileReport`; plus the README.

## How to test?

- `cd catalogue_graph && uv run pytest tests/adapters -q` (all green, including tests for the WebAPI client paging and page cap, the base OAI completeListSize count path, `count_active_namespace_records`, and the reconcile drift and missing-id logic)
- `uv run ruff check src tests` and `uv run mypy src` (clean)
- Run it against Axiell: `uv run python -m adapters.steps.oai_pmh.reconcile --adapter-type axiell --use-rest-api-table --list-missing` reports the current drift.

The reconcile can be run on demand to confirm the adapter store matches the source, and gives the id list needed to close any gap.

This is the first of two PRs from the July Axiell incident. This one is detection; a follow-up adds remediation tooling (recover-by-id and reconciler baseline rebuild).

### DIFF
--- a/catalogue_graph/src/adapters/extractors/oai_pmh/axiell/runtime.py
+++ b/catalogue_graph/src/adapters/extractors/oai_pmh/axiell/runtime.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from collections.abc import Iterator
+
 import httpx
 from oai_pmh_client.client import OAIClient
 
@@ -11,6 +13,9 @@ from adapters.extractors.oai_pmh.axiell.clients import (
 )
 from adapters.extractors.oai_pmh.axiell.clients import (
     build_oai_client as _build_axiell_oai_client,
+)
+from adapters.extractors.oai_pmh.axiell.clients import (
+    build_webapi_client as _build_axiell_webapi_client,
 )
 from adapters.extractors.oai_pmh.axiell.config import AXIELL_ADAPTER_CONFIG
 from adapters.extractors.oai_pmh.runtime import OAIPMHAdapterConfig, OAIPMHRuntimeConfig
@@ -44,6 +49,28 @@ class AxiellRuntimeConfig(OAIPMHRuntimeConfig):
         Overrides base to include Axiell-specific retry configuration.
         """
         return _build_axiell_oai_client(http_client=http_client)
+
+    def _webapi_database(self) -> str:
+        return self.config.oai_set_spec or "collect"
+
+    def source_of_truth_count(self) -> int | None:
+        """Total record count from the Axiell WebAPI (wwwopac.ashx).
+
+        A direct database count, so unaffected by the OAI module's
+        datestamp-query faults.
+        """
+        return _build_axiell_webapi_client(database=self._webapi_database()).count()
+
+    def enumerate_source_ids(self) -> Iterator[str]:
+        """Yield every record id via the Axiell WebAPI (stateless paging).
+
+        Ids are ``<database>:<priref>`` to match the OAI identifiers stored in
+        the adapter table.
+        """
+        database = self._webapi_database()
+        return _build_axiell_webapi_client(database=database).enumerate_ids(
+            namespace=database
+        )
 
     def build_reconciler_table(
         self,
--- a/catalogue_graph/src/adapters/extractors/oai_pmh/axiell/webapi_client.py
+++ b/catalogue_graph/src/adapters/extractors/oai_pmh/axiell/webapi_client.py
@@ -1,0 +1,132 @@
+"""Client for the Axiell WebAPI (``wwwopac.ashx``).
+
+The Axiell OAI-PMH endpoint (``oai.ashx``) is one handler inside a WebAPI
+application that also serves ``wwwopac.ashx`` on the same host with the same
+``Token`` header. Unlike OAI-PMH it supports stateless ``startfrom`` paging with
+no resumption tokens, so it is a reliable source of truth for reconciliation: a
+full id enumeration cannot be derailed by one dropped page.
+
+Only read-only ``search`` requests are used. The URL is derived from the OAI
+endpoint (``oai.ashx`` -> ``wwwopac.ashx``) and reuses the Axiell Token client,
+so no new credentials are needed.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+
+import httpx
+import structlog
+from lxml import etree
+
+logger = structlog.get_logger(__name__)
+
+DEFAULT_DATABASE = "collect"
+DEFAULT_PAGE_SIZE = 1000
+# Absolute page ceiling. At the default page size this is ~1 billion records,
+# far beyond any real database, so it only ever trips on a server that ignores
+# ``startfrom`` and never drains (which would otherwise loop until timeout).
+DEFAULT_MAX_PAGES = 1_000_000
+
+
+def oai_url_to_webapi_url(oai_endpoint: str) -> str:
+    """Derive the wwwopac.ashx URL from the oai.ashx URL (same application)."""
+    base = oai_endpoint.split("?", 1)[0]
+    if "/oai.ashx" not in base:
+        raise ValueError(f"Cannot derive WebAPI URL: expected '/oai.ashx' in {base!r}")
+    return base.replace("/oai.ashx", "/wwwopac.ashx")
+
+
+class AxiellWebApiClient:
+    """Read-only client for Axiell ``wwwopac.ashx`` search requests.
+
+    Args:
+        base_url: The full ``wwwopac.ashx`` URL.
+        client: An ``httpx.Client`` carrying the Axiell ``Token`` header. Injected
+            so it can be mocked in tests.
+        database: The Axiell database to query (default ``collect``).
+        page_size: Records per page when enumerating (default 1000).
+    """
+
+    def __init__(
+        self,
+        *,
+        base_url: str,
+        client: httpx.Client,
+        database: str = DEFAULT_DATABASE,
+        page_size: int = DEFAULT_PAGE_SIZE,
+        max_pages: int = DEFAULT_MAX_PAGES,
+    ) -> None:
+        self._base_url = base_url
+        self._client = client
+        self._database = database
+        self._page_size = page_size
+        self._max_pages = max_pages
+
+    def _search(self, *, fields: str, limit: int, startfrom: int) -> etree._Element:
+        response = self._client.get(
+            self._base_url,
+            params={
+                "database": self._database,
+                "search": "all",
+                "fields": fields,
+                "limit": str(limit),
+                "startfrom": str(startfrom),
+            },
+        )
+        response.raise_for_status()
+        if not response.content:
+            raise ValueError(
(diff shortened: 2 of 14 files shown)


catalogue_graph/src/adapters/steps/oai_pmh/reconcile.py
```diff
@@ -0,0 +1,182 @@
...
+
+MISSING_ID_SAMPLE = 50
+
+
+class ReconcileResponse(BaseModel):
+    adapter_type: str
+    server_count: int | None
+    harvested_count: int
+    drift_count: int
+    missing_id_sample: list[str] = []
```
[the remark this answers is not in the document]
kenoir: Fixed: `missing_id_sample` now uses `Field(default_factory=list)`.

catalogue_graph/src/adapters/steps/oai_pmh/reconcile.py
```diff
@@ -0,0 +1,182 @@
...
+        return []
+
+    have = set(runtime.store.get_active_namespace_records()["id"].to_pylist())
+    missing: list[str] = []
+    for record_id in source_ids:
+        if record_id not in have:
+            missing.append(record_id)
+            if len(missing) >= MISSING_ID_SAMPLE:
+                break
+    return missing
```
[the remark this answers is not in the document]
kenoir: Fixed: `_sample_missing_ids` now streams the store's ids id-only into a set via `stream_active_namespace_records`, so record content is never materialised. (This path only runs with `--list-missing`; the scheduled Lambda has also been dropped, so it is a local diagnostic now.)

catalogue_graph/src/adapters/utils/adapter_store.py
```diff
@@ -88,6 +88,21 @@ def stream_active_namespace_records(
...
+        keeping memory flat for a whole-table count.
+        """
+        non_deleted_filter = Or(EqualTo("deleted", False), IsNull("deleted"))
+        full_filter = And(EqualTo("namespace", self.namespace), non_deleted_filter)
+        scan = self.table.scan(
+            row_filter=full_filter,
+            selected_fields=("id",),
+            snapshot_id=snapshot_id,
+        )
+        return scan.to_arrow().num_rows
```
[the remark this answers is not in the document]
kenoir: Fixed: `count_active_namespace_records` now sums row counts from `to_arrow_batch_reader()` instead of building a full `to_arrow()` table.

catalogue_graph/src/adapters/extractors/oai_pmh/runtime.py
```diff
@@ -177,3 +178,48 @@ def build_oai_client(self, *, http_client: httpx.Client | None = None) -> OAICli
...
+        separator = "&" if "?" in endpoint else "?"
+        params = (
+            f"verb=ListIdentifiers&metadataPrefix={self.config.oai_metadata_prefix}"
+        )
+        if self.config.oai_set_spec:
+            params += f"&set={self.config.oai_set_spec}"
+
+        client = self.build_http_client()
+        response = client.get(endpoint + separator + params)
+        response.raise_for_status()
```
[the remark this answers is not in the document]
kenoir: Fixed: the request now passes a `params=` dict to httpx so the set spec and metadata prefix are URL-encoded.

catalogue_graph/src/adapters/extractors/oai_pmh/axiell/runtime.py
```diff
@@ -45,6 +50,25 @@ def build_oai_client(self, *, http_client: httpx.Client | None = None) -> OAICli
         """
         return _build_axiell_oai_client(http_client=http_client)
 
+    def source_of_truth_count(self) -> int | None:
+        """Total record count from the Axiell WebAPI (wwwopac.ashx).
+
+        More reliable than the OAI ``completeListSize`` default: it is a direct
+        SQL count, unaffected by the OAI module's datestamp-query faults.
+        """
+        return _build_axiell_webapi_client().count()
```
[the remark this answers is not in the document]
kenoir: Fixed: `source_of_truth_count` and `enumerate_source_ids` now share a `_webapi_database()` helper (`oai_set_spec`), so the count and the id enumeration always query the same database.

catalogue_graph/src/adapters/extractors/oai_pmh/axiell/webapi_client.py
```diff
@@ -0,0 +1,132 @@
...
+                    f"draining (seen={seen}, total={total}); startfrom may be "
+                    f"ignored server-side"
+                )
+            root = self._search(
+                fields="priref", limit=self._page_size, startfrom=startfrom
+            )
+            pages += 1
+            if total is None:
+                hits = root.findtext(".//diagnostic/hits")
+                total = int(hits) if hits is not None else None
```
StepanBrychta: This defensive logic guards against situations where the server arbitrarily drops the total record count (`.//diagnostic/hits`) from the response.

Do we have a good reason to think that the Axiell server might be broken in this specific way? If so, it might be helpful to include a comment. If not, do we need this logic? There are many ways in which a server could be broken (e.g. it might send the same record on multiple pages), and I think there is limited value in speculatively defending against some of those ways.

On top of adding complexity, there are situations in which this approach could make things worse. For example, if the server suddenly started returning the total record count in a different field (e.g. `.//diagnostic/total` instead of `.//diagnostic/hits`), we would not notice this change. Conversely, if our logic was written to always expect the `.//diagnostic/hits` field, it would fail loudly on the next run.

The `count` method on this class already makes the `.//diagnostic/hits` field mandatory, so we could call that method from `enumerate_ids` to keep things simple and separate concerns.

catalogue_graph/src/adapters/extractors/oai_pmh/axiell/webapi_client.py
```diff
@@ -0,0 +1,132 @@
...
+        total: int | None = None
+        seen = 0
+        pages = 0
+        while True:
+            if pages >= self._max_pages:
+                raise RuntimeError(
+                    f"WebAPI enumeration exceeded {self._max_pages} pages without "
+                    f"draining (seen={seen}, total={total}); startfrom may be "
+                    f"ignored server-side"
+                )
```
StepanBrychta: Following up on my other comment, do we need this logic?

I think the argument for keeping this backstop is stronger than with the `.//diagnostic/hits` field, but it still feels a little bit arbitrary (unless we've already seen the server ignore `startfrom`). 

Even if we have seen the server do this, the backstop, as configured, would take a long time to kick in. `DEFAULT_MAX_PAGES` is currently set to 1 million, and assuming it takes one second to retrieve a page, it would take ~10 days for the backstop to trigger.

If we do want to keep the backstop, an alternative approach would be to keep a set of all returned IDs and terminate if the server does not return any new IDs not already stored in the set. This would be much faster than retrieving a million instances of the same page, at the cost of O(n) memory (likely small, even with ~1 million IDs). This would also allow us to deduplicate IDs before yielding, streaming a generator of deduplicated IDs.

catalogue_graph/src/adapters/steps/oai_pmh/audit.py
StepanBrychta, line 1: Minor: There is already an Axiell `reconciler` service, so maybe we could choose a different name for this service to avoid confusion.

StepanBrychta (approved):
Looks good!

At the moment, the new service sits between a local-only ad hoc tool and a deployed service. It publishes metrics and has a `lambda_handler`, but it's not actually deployed. Non-blocking, but I think we should decide if we want to deploy it, or if it's fine as a local-only service (in which case we might not want to publish the metrics). We can decide later.

catalogue_graph/src/adapters/extractors/oai_pmh/axiell/webapi_client.py
kenoir: Agreed, no reason to think the server drops it. enumerate_ids now takes its total from count(), so a missing hits field fails on the next run instead of being tolerated.

catalogue_graph/src/adapters/extractors/oai_pmh/axiell/webapi_client.py
kenoir: Taken the set-based approach. Worth noting the backstop was worse than slow: if the server ignored startfrom, seen still advanced by a page each iteration, so seen >= total ended the loop after ~222 pages and max_pages never fired. It yielded one page repeated to the total, which reconciliation would report as a huge fake mismatch.

Yielded ids are now tracked, so a repeated page stops enumeration immediately, and a distinct count short of the total raises. The memory is already paid for anyway, since reconcile.py collects these into a set. max_pages is gone.

catalogue_graph/src/adapters/steps/oai_pmh/audit.py
kenoir: Good catch, renamed to `audit`. The step is now `adapters.steps.oai_pmh.audit`, with `OAIPMHAuditReport` and the report label `adapter_audit`. I also dropped "reconciliation" from the surrounding prose so the word is left to the transformer.

Cheap to do now because the scheduled Lambda came out earlier in this review, so nothing is emitting the old metric label yet.

kenoir:
Closing this. On reflection the added complexity of a separate audit service isn't worth it here.

The recover-by-id work from #3460 is being kept and reworked into an ID mode on the loader, per your review comment there.

PR #3460, opened 2026-07-08
Fold recover-by-id into the OAI-PMH loader as an id mode

kenoir:
## What does this change?

Adds an "id mode" to the OAI-PMH loader for recovering specific records the source still holds but the adapter store is missing (typically records written with datestamps inside windows already harvested, as in the July 2026 Axiell incident). An operator starts an adapter state machine execution with `{adapter_type, ids}`; a `Which mode?` choice routes to the id loader, which fetches each id via OAI `GetRecord` and writes it through the normal adapter store path. Per review it lives in the loader rather than a separate `recover.py` service.

Both modes return changeset ids down the same `loader_next` transition, so an id run composes with the publish and Axiell reconcile chain (from wellcomecollection/platform#6446) for free: recovered records get `adapter.completed`, guid-change detection and deletion facts exactly like a harvested window, with no manual step. The reconciler-rebuild step this PR once carried is gone, because the continuous adapter-side reconcile from #6446 now keeps the mapping baseline and a one-off rebuild no longer has a rationale.

<details><summary>Risk-relevant design choices</summary>

- **Separate state, same task.** `Run id loader` reuses the window loader's task definition but keeps its own state purely for retry policy. Window retries skip already-successful windows, but id mode keeps no progress state so a retry re-fetches every id with a politeness delay. At the per-run ceiling that is hours of duplicated load on a flaky source, so id mode has no automatic retries; failed ids come back in the report for a smaller second run.
- **Removed ids are reported, not tombstoned.** `idDoesNotExist` also fires for a typo or an id from another set, and changesets now publish, so tombstoning would auto-propagate a delete downstream on a weak signal.
- **No window state is written**, so an id run never shifts the trigger's resume cursor onto a range that was never harvested (`covered_window_keys` stays empty).
- **`commit_every` and run size are bounded** (default 10,000, capped at 50,000 ids); each published changeset costs the transformer roughly a full materialisation of the bib store, so commit rarely.

Supporting refactors: accumulator state moved into classes, step events validated with Pydantic, a shared `build_adapter_store_row`, an optional `flush_threshold` on the buffered writer, and the `oai_pmh` README documents both modes.
</details>

## How to test

- From `catalogue_graph/`: `uv run --group dev ruff check .`, `mypy .`, `pytest` (963 adapter tests pass, 28 covering id mode and the record writer). `terraform validate` in `infra/adapters`. Every state machine `Next`/`Default` target was cross-checked for each mode and flag combination.
- Validated against the live Axiell OAI endpoint via CLI (`publish_to_s3: false`), recovering five real records: one changeset minted, `covered_window_keys: []`, window-status table untouched.

The next list-of-missing-ids incident is handled by a supported step instead of an ad hoc script, reporting recovered / removed / unfetchable counts and minting changesets that publish the records downstream unaided.

- **No staging for this path.** `infra/adapters` is production and the loader task is pinned to `:prod`, so an end-to-end run of unmerged code would hit live infra; validated by CLI instead.
- **Unfetchable ids degrade quietly.** A large run reports "some ids unfetchable" and needs re-running on the residue rather than failing loudly; genuine bugs still surface as exceptions.


catalogue_graph/src/adapters/steps/oai_pmh/recover.py
```diff
@@ -0,0 +1,235 @@
...
+                "id": record_id,
+                "content": content,
+                "last_modified": record.header.datestamp,
+                "deleted": content is None,
+            }
+        )
+        if len(buffer) >= commit_every:
+            flush()
+
+    flush()
```
StepanBrychta: Minor: It might be preferable to encapsulate this logic into a class.

catalogue_graph/src/adapters/steps/oai_pmh/recover.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def lambda_handler(event: dict[str, Any], context: Any) -> dict[str, Any]:
+    """Lambda entry point. Expects ``adapter_type`` and an ``ids`` list."""
+    adapter_type = event.get("adapter_type")
+    if adapter_type is None:
+        raise ValueError("Event must contain 'adapter_type'")
+    ids = event.get("ids")
+    if not isinstance(ids, list):
+        raise ValueError("Event must contain an 'ids' list")
```
StepanBrychta: Minor: In other places, we load the raw event dictionary into a Pydantic model so that we don't have to manually validate each field.

catalogue_graph/src/adapters/steps/oai_pmh/rebuild_reconciler.py
```diff
@@ -0,0 +1,205 @@
...
+                {
+                    "namespace": runtime.namespace,
+                    "id": row["id"],
+                    "guid": guid,
+                    "last_modified": row["last_modified"],
+                }
+            )
+            if len(buffer) >= batch_size:
+                flush()
+    flush()
```
StepanBrychta: Same as above, it might be preferable to encapsulate this into a class rather than managing state via closure.

catalogue_graph/src/adapters/extractors/oai_pmh/README.md
```diff
@@ -161,6 +161,49 @@ being deleted at the source), because `drift_count` is a count comparison
...
+`idDoesNotExist`), or unfetchable (neither returned nor reported gone, after the
+client's retries). Unfetchable ids are left absent from the store and listed in
+the report, never backfilled with stale content.
+
+### 7. Rebuild reconciler → reseed the id-to-GUID baseline (Axiell)
+
+The reconciler maps each record id to the source-identifier GUID it produces, so
+a later GUID change can mark the old work deleted. It only runs incrementally,
+per changeset, and cannot run over a full snapshot. After a full reindex (a
+clean-slate rebuild of the adapter store) there is otherwise no supported way to
```
StepanBrychta: This reminds me: I think we don't have a documented/standard way of rebuilding the adapter store from a clean slate. Not something to address in this PR, it just occurred to me we will probably need one so that we can wipe test records and Mimsy data from the Axiell adapter store.

StepanBrychta (dismissed):
Question: After running the `recover` service to process missing records, how will we ensure that those records get propagated downstream in the transformer? It looks like we would have to manually find the corresponding changeset IDs and input them into the transformer, or reprocess all adapter records in full mode. That's probably fine if we're not expecting to do this often.

It would be convenient if the 'process by ID' functionality was integrated into the loader service rather than a separate service outside of the main pipeline. We would then be able to run the adapter either in 'window mode' or in 'ID mode', automatically propagating changeset IDs to the transformer. But this would probably involve a significant refactor, so I'm not saying we should do it now.

catalogue_graph/src/adapters/steps/oai_pmh/recover.py
kenoir: Agreed, done in 7bd4099. The accumulator state is now a RecoveryBatch class with add_recovered/add_removed/add_unfetchable/flush/to_response, with unit tests exercising it directly.

catalogue_graph/src/adapters/steps/oai_pmh/rebuild_reconciler.py
kenoir: Agreed, done in 7bd4099. This is now a ReconcilerBaselineBatch class, so the nonlocal written counter becomes an attribute. Unit tests cover it directly, including that written totals across flushes.

catalogue_graph/src/adapters/steps/oai_pmh/recover.py
kenoir: Agreed, done in 419dfd5. Applied to reconcile and rebuild_reconciler as well so the three ad-hoc steps are consistent.

One wrinkle worth noting: BaseAdapterEvent requires job_id, but none of these three steps run as part of a harvest, so they inherit from a new AdapterRecoveryEvent base that makes job_id optional. Validation errors are now ValidationError rather than ValueError, which subclasses ValueError, so anything catching ValueError still behaves the same.

catalogue_graph/src/adapters/extractors/oai_pmh/README.md
kenoir: Good point, and agreed it is out of scope here. I have raised #3473 to cover it and put it in Next on the Digital platform board. It notes the Axiell motivation, the test records and Mimsy data we want wiped rather than corrected, which the incremental harvest path cannot remove.

catalogue_graph/src/adapters/steps/oai_pmh/rebuild_reconciler.py
```diff
@@ -0,0 +1,241 @@
...
+) -> RebuildReconcilerResponse:
+    """Recompute and write id->GUID mappings for every active record."""
+    setup_logging(execution_context)
+
+    batch = ReconcilerBaselineBatch(runtime, batch_size=batch_size)
+
+    for record_batch in runtime.adapter_store.stream_active_namespace_records():
+        for row in record_batch.to_pylist():
+            batch.add_row(row)
+    batch.flush()
```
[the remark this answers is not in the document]
kenoir: Good catch, fixed in 790dfb0. The stream is now wrapped in try/finally, matching the pattern already used in AdapterStoreSource.stream_raw.

catalogue_graph/src/adapters/extractors/oai_pmh/models/step_events.py
```diff
@@ -54,6 +75,60 @@ class OAIPMHLoaderEvent(BaseAdapterEvent):
...
+        # no-op that reads like a success.
+        if not ids:
+            raise ValueError("No ids supplied; id mode needs at least one record id")
+        if len(ids) > MAX_IDS_PER_RUN:
+            raise ValueError(
+                f"{len(ids)} ids exceeds the {MAX_IDS_PER_RUN} per-run ceiling; "
+                f"split the work across several runs"
+            )
+        return ids
+
```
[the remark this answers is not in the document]
kenoir: Fixed in 0b05979, with a split on the zero case. commit_every is now ge=1, since 0 disables auto-flush and buffers the whole run. polite_delay_seconds is ge=0, since no delay is a reasonable thing to ask for and the id-mode tests rely on it.

kenoir:
Note: wellcomecollection/platform#6446 moves reconciliation into the adapter, which removes the routine need for the rebuild step in this PR. The mapping baseline will be maintained continuously at load time, reindexes don't touch it, and new pipelines share the one store instead of needing a seeded copy. The id-mode loader changes are unaffected.

The rebuild step still has value as a repair tool for baseline drift or loss (e.g. reconciliation disabled while loads continue, or the table having to be recreated), but it drops from required tooling to optional. Could trim it from this PR to keep it focused on id mode, or keep it as-is and just update its docstring once 6446 lands.

kenoir:
Updated and trimmed the reconciler rebuild step as discussed above. The adapter-side reconcile step from wellcomecollection/platform#6446 now maintains the baseline continuously, so the after-reindex reseed this step existed for no longer arises, and a rebuild from the adapter store would absorb in-flight guid changes without emitting their deletions. The PR is now just the loader id mode. Title and description updated to match.

Note: when https://github.com/wellcomecollection/catalogue-pipeline/pull/3479 merges it adds reconcile states to the same adapter state machine locals this PR touches, so expect one small rebase then. 972 adapter tests, ruff, mypy, and terraform validate all pass on the rebased branch.

catalogue_graph/src/adapters/extractors/oai_pmh/models/step_events.py
```diff
@@ -29,12 +44,18 @@ class OAIPMHTriggerEvent(BaseAdapterEvent):
...
 
     The loader step receives this event from the trigger and harvests
     records from the OAI-PMH endpoint within the specified window.
     """
 
+    mode: Literal["window"] = "window"
+    """Discriminator selecting the loader's window mode.
+
+    Defaults to ``window`` so trigger payloads and stored events written before
+    id mode existed stay valid."""
```
StepanBrychta: Do we need this discriminator? The presence of a `window` field is what defines an event as a 'window mode' event. Typing the LoaderEvent as 

```
LoaderEvent = OAIPMHLoaderEvent | OAIPMHIdLoaderEvent
```

should work. Pydantic is smart and will choose the right model based on whether the input dictionary has an `ids` field or a `window` field.

catalogue_graph/src/adapters/extractors/oai_pmh/models/step_events.py
```diff
@@ -54,6 +75,67 @@ class OAIPMHLoaderEvent(BaseAdapterEvent):
...
+
+class AdapterRecoveryEvent(BaseAdapterEvent):
+    """Shared base for the manually invoked recovery steps.
+
+    These steps are run ad hoc against a chosen adapter rather than as a stage
+    of a harvest, so unlike the trigger and loader they belong to no job and
+    ``job_id`` is optional.
+    """
+
+    job_id: str | None = None  # type: ignore[assignment]
```
StepanBrychta: This doesn't seem to be used anywhere.

catalogue_graph/src/adapters/extractors/oai_pmh/models/step_events.py
kenoir: Agreed, done in bf6c450. The mode field is gone from both events, the union is plain and the smart union picks by field presence, so the validate_loader_event shim that injected a default mode is gone too, along with the mode merge in the state machine's Prepare id run state and the CLI's mode key.

catalogue_graph/src/adapters/extractors/oai_pmh/models/step_events.py
kenoir: Right, deleted in 2041f74. It was the base for the ad hoc recovery steps, which left this PR when adapter-side reconciliation made the rebuild step obsolete.

catalogue_graph/src/adapters/extractors/oai_pmh/models/step_events.py
```diff
@@ -121,3 +203,73 @@ def from_summaries(
...
             changed_record_count=upserted_record_count,
         )
+
+
+UNFETCHABLE_SAMPLE_SIZE = 100
+"""Unfetchable ids carried inline on the response.
+
+The publish state passes the whole response onward as its output, so an
+unbounded list risks breaching the Step Functions payload limit after the store
+writes have already succeeded. The full list goes to the report in S3."""
```
StepanBrychta: Is there a good reason for returning a sample of unfetchable IDs in the response? In the transformer and in the ID minter, we only return counts (success count/failure count), and we also return the S3 URI of the full report so that the full lists can be retrieved easily. It would be good to be consistent between services.

Being able to see a sample of unfetchable IDs directly in the response is nice, but every change like this adds a little bit of complexity, which gradually compounds and makes the codebase harder to understand.

kenoir: This grew out of having a set of unfetchable things that I needed to identify, which we hope is now resolved so agree this is vestigial and can be removed.

kenoir: Removed in d677d95. The response now carries counts plus report_s3_uri, matching IdMinterResult, and the full removed and unfetchable lists stay in the report.

catalogue_graph/src/adapters/extractors/oai_pmh/reporting.py
```diff
@@ -120,3 +127,102 @@ def metrics(self) -> list[PipelineMetric]:
...
+    def publish_to_cloudwatch(self) -> bool:
+        return self.emit_metrics
+
+    @property
+    def s3_uri(self) -> str:
+        return (
+            f"s3://{self.report_s3_bucket}/{self.report_s3_prefix}"
+            f"/reports/{self.adapter_type}/{self.label}"
+            f"/{self.job_id}_{self.report_id}.json"
+        )
```
StepanBrychta, line 175: Minor: In other places, we compose the path using `PurePosixPath`.

catalogue_graph/src/adapters/extractors/oai_pmh/record_writer.py
```diff
@@ -20,14 +20,32 @@
...
+    return {
+        "namespace": namespace,
+        "id": identifier,
+        "content": content,
+        "last_modified": record.header.datestamp,
+        "deleted": content is None,
+    }
+
+
 class WindowRecordWriter:
```
StepanBrychta, line 49: I think the `WindowRecordWriter` name (and the `BufferedWindowRecordWriter` name) are now misleading, because they do not always write records as part of a window. Maybe we could change the name and also remove the `window_range` parameter, instead allowing callers to pass an optional dictionary of tags?

StepanBrychta (approved):

PR #3461, opened 2026-07-09
Graph date IAM fixes

StepanBrychta:
## What does this change?

Hotfix following https://github.com/wellcomecollection/catalogue-pipeline/pull/3458, updating S3 paths accessible by deployed services to match the new pattern.

## How to test

These changes were already applied as a hotfix and have been proven to work.


kenoir (approved):
I think the co-pilot comment is correct, but it's a pre-existing issue so your call as to whether to remediate.

PR #3463, opened 2026-07-13
Implement Axiell to  FOLIO sync adapter

anubhavijay:
## What does this change?

Implements the complete Axiell → FOLIO synchronization adapter: a write path that exports changed records from the Axiell Iceberg store to FOLIO Inventory via the OKAPI REST API.

### Implementation breakdown:

**Core adapter step** (`adapters.steps.axiell_folio_sync`):
- Listens to `axiell.adapter.completed` EventBridge events
- Reads changed records from the Axiell adapter table
- Filters records by selection criteria (via `is_selected_for_sync`)
- Maps MARCXML → FOLIO Inventory payloads (Instance → Holdings → Item)
- Upserts records via OKAPI in batches with error recovery
- Writes success manifests to S3 (NDJSON format with configurable retention)
- Supports dry-run mode (default: true) for validation before FOLIO writes

**Transformers** (`adapters.transformers.axiell_folio_sync`):
- `mapping.py` — MARCXML → FOLIO payload translation, selection logic
- `ref_cache.py` — Reference caches for FOLIO Inventory lookups (holdings types, item statuses, etc.); intermediate record lists removed to reduce memory footprint
- `upsert.py` — OKAPI upsert orchestration (PUT Instance, create/update Holdings + Items) with delete-based rollback on failure

**FOLIO client** (`clients.folio_client`):
- Thin OKAPI REST wrapper with configurable retries and SSL context
- Supports both environment variable auth (local dev) and SSM SecureString (Lambda)

**Infrastructure** (`infra/adapters/modules/axiell_folio_sync`):
- Lambda function (runs in shared `unified_pipeline_lambda` image)
- Step Function with retries and error handling
- EventBridge rule (`axiell.adapter.completed` → Step Function)
- S3 bucket for run manifests (with configurable retention)
- IAM roles (read Axiell table, read/write S3 manifests, read SSM secrets, write CloudWatch Logs)
- SSM SecureString for OKAPI credentials (placeholder seeded; real values set out-of-band)

### Key design decisions:

- **Dry-run by default**: OKAPI writes are guarded; flip `dry_run_default=false` in Terraform only after validation
- **Reference caching**: FOLIO Inventory reference objects (statuses, types) are cached in memory per Lambda invocation (survives warm starts)
- **Selection logic**: Not all Axiell records are written to FOLIO; selection is configurable and documented in `is_selected_for_sync()`
- **Manifest storage**: One NDJSON per successful run; expires after `manifest_retention_days` (configurable, default 90 days)
- **Error recovery**: Failed upserts are logged and included in counts; the step continues processing other records (fail-open policy)
- **Tombstone handling**: `deleted=true` records are recorded as advisory tombstones (RFC 090); authoritative deletes come from the reconciler, not this path

### Post-PR fixes included:

- **`folio_delete` callable**: Added a proper `DELETE` path to `_make_folio_callables` so rollback on failed Holdings/Item writes issues real FOLIO DELETE calls (404 responses treated as success)
- **S3 error handling**: `_flush_success_batch` now catches `botocore.exceptions.ClientError` specifically (checking for `NoSuchKey`/`404`/`NotFound`) rather than swallowing all exceptions
- **`RefCache` cleanup**: Removed intermediate `_*_records` instance attributes — reference lookups now use local variables only, reducing per-invocation memory
- **`run_sync` / `handler` wiring**: `folio_delete` threaded through `run_sync` and `handler` call sites
- **Upsert tests**: Added `tests/transformers/axiell_folio_sync/test_upsert.py` covering create/update/rollback paths
- **mypy fixes**: Corrected `httpx.MockTransport` handler type in `test_client.py`; added explicit return annotation in `test_axiell_folio_sync.py`; excluded `rfcs/` from mypy via `pyproject.toml`
- **Ruff lint**: Fixed import ordering in `test_client.py`

### Local validation (before FOLIO writes)

```bash
cd catalogue_graph
source .venv/bin/activate  # or: uv shell
uv run python -m adapters.steps.axiell_folio_sync \
  --use-cli --job-id local-test-1 --use-rest-api-table --sample-limit 5 --dry-run
```

This reads 5 sample records from the Axiell REST API table (DEV/staging schema) and logs the payloads without writing to FOLIO.

### Run the test suite

```bash
cd catalogue_graph
uv run pytest tests/adapters/steps/test_axiell_folio_sync.py -v
uv run pytest tests/transformers/axiell_folio_sync/ -v
uv run pytest tests/clients/folio_client/ -v
uv run mypy .
uv run ruff check
```

### Deploy to dev/staging (Lambda)

```bash
cd catalogue_graph
./scripts/deploy_lambda.sh axiell-folio-sync-adapter-lambda
```

This builds and pushes the shared `unified_pipeline_lambda:dev` image, then updates the Lambda function. The Step Function and EventBridge rule are already deployed (Terraform-managed).

After deployment:
1. Trigger a test run via EventBridge (AWS Console or CLI)
2. Monitor the Lambda execution in CloudWatch Logs
3. Verify the manifest in S3 (bucket name in Terraform outputs)
4. Query FOLIO directly or via OKAPI to spot-check created/updated records

- **Dry-run validation**: Manifest is created with 0 errors (or only expected mapping errors) when `dry_run=true`
- **FOLIO Inventory**: Records appear in FOLIO Inventory (Instances, Holdings, Items) matching the Axiell source data
- **Counts accuracy**: Manifest counts (successful, errors, skipped) match query results from FOLIO API
- **Event delivery**: EventBridge `axiell.adapter.completed` events trigger the Step Function reliably
- **Error handling**: Failed FOLIO writes are logged (in both CloudWatch and the manifest) but don't block the step from processing other records
- **Reference cache performance**: Multiple runs within a warm Lambda invocation should reuse the cache (faster, fewer FOLIO API calls)

- **FOLIO write failures**: Implemented delete-based rollback — if an Item write fails, `folio_delete` is called to clean up; if Holdings write fails, Holdings delete is attempted (404 treated as already-clean). Failed records are logged and included in the manifest but don't stop the step.
- **Authentication**: OKAPI credentials are stored in SSM SecureString (never in Terraform state). Ensure credentials have the minimum required FOLIO permissions (Instance read/write, Holdings read/write, Item read/write).
- **Dry-run default**: The step defaults to `dry_run=true`; ensure `dry_run_default` is set to `false` only after validation. The event can override this.
- **Reference cache staleness**: Reference caches (e.g., item statuses) are loaded once per Lambda invocation; if FOLIO Inventory references change during execution, cached values will be stale until the next invocation.
- **Rate limiting**: The step respects FOLIO OKAPI rate limits via configurable retries (default 3 retries with exponential backoff). If records are large or OKAPI is rate-limited, the step may time out or be throttled.
- **Data consistency**: Records are upserted (read → update or create); concurrent writes from other systems could result in lost updates. This is a known limitation of the upsert pattern.
- **Manifest retention**: S3 manifests are retained for `manifest_retention_days` (default 90 days); older manifests are automatically deleted.

### Testing recommendations

1. **Credentials**: Validate SSM SecureString is correctly populated with valid FOLIO credentials before flipping `dry_run_default=false`
2. **Permissions**: Test with a reduced permission set to ensure the Lambda only has access to required FOLIO Inventory endpoints
3. **Load**: Test with a large batch (e.g., 10,000 records) to validate timeout and rate-limit handling
4. **Concurrency**: If multiple Axiell changesets are processed concurrently, ensure FOLIO can handle the load
5. **Rollback**: Manually test the delete-based rollback by injecting FOLIO write failures and verifying the Lambda recovers cleanly


catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,653 @@
...
+
+
+def _flush_success_batch(bucket: str, job_id: str, batch: list[dict]) -> None:
+    if not batch:
+        return
+    key = f"manifests/{job_id}.ids.ndjson"
+    ndjson_lines = "\n".join(json.dumps(record) for record in batch)
+    _s3().put_object(
+        Bucket=bucket,
+        Key=key,
```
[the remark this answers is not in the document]
anubhavijay: Prevented success manifest data loss by appending batch output to the existing NDJSON object instead of overwriting on each flush

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+    if _ssm_client is None:
+        _ssm_client = boto3.client("ssm", region_name=os.environ["AWS_REGION"])
+    return _ssm_client
+
+
+def _s3() -> Any:
+    global _s3_client
+    if _s3_client is None:
+        _s3_client = boto3.client("s3", region_name=os.environ["AWS_REGION"])
+    return _s3_client
```
StepanBrychta: Minor: I think that using `lru_cache` instead of global variables is more idiomatic and readable:

```python
@lru_cache(max_size=1)
def _s3() -> Any:
    return boto3.client("s3", region_name=os.environ["AWS_REGION"])

```

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+    # Dev/smoke-test fallback: a sample of active records (no changesets given).
+    limit = sample_limit or 10
+    logger.info("adapter_read", mode="sample", limit=limit)
+    return store.get_active_namespace_records().to_pylist()[:limit]
+
+
+# ── CloudWatch metrics ────────────────────────────────────────────────────────
+
+
+def _emit_metrics(counts: dict) -> None:
```
StepanBrychta: Minor suggestion: Throughout the pipeline, we use dedicated `MetricReporter` and `PipelineReport` classes to publish metrics. It might be possible to integrate those classes with this service.

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+            {
+                "MetricName": "RecordsProcessed",
+                "Value": counts.get("total", 0),
+                "Unit": "Count",
+            },
+        ],
+    )
+
+
+# ── S3 NDJSON manifest writing ────────────────────────────────────────────────
```
StepanBrychta: We already have a `ManifestWriter` class for publishing manifests, though using it here might not be feasible. Maybe something to explore?

Either way, I think it would be good to move this logic into its own file. This file is large (nearly 700 lines of code), and splitting it up would make it easier to parse.

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+            raise RuntimeError(f"PUT {path} failed: {status} {str(data)[:200]}")
+        return cast("dict", data)
+
+    def folio_delete(path: str) -> dict:
+        status, data = client.request("DELETE", path, accept="text/plain")
+        if status >= 400 and status != 404:
+            raise RuntimeError(f"DELETE {path} failed: {status} {str(data)[:200]}")
+        return cast("dict", data)
+
+    return folio_get, folio_post, folio_put, folio_delete
```
StepanBrychta: Would it be possible to encapsulate these into a class (either `FolioClient` directly or maybe a subclass if coupling is a concern)? Having to pass all four functions separately to callers is a bit awkward.

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+    store = AdapterStore(table, namespace=AXIELL_CONFIG.config.adapter_namespace)
+
+    if changeset_ids:
+        logger.info("adapter_read", mode="changesets", changeset_ids=changeset_ids)
+        return store.get_records_by_changesets(changeset_ids).to_pylist()
+
+    # Dev/smoke-test fallback: a sample of active records (no changesets given).
+    limit = sample_limit or 10
+    logger.info("adapter_read", mode="sample", limit=limit)
+    return store.get_active_namespace_records().to_pylist()[:limit]
```
StepanBrychta: This duplicates some of the logic from the `AdapterStoreSource` class. Could we use the class here?

catalogue_graph/src/adapters/transformers/axiell_folio_sync/ref_cache.py
```diff
@@ -0,0 +1,133 @@
...
+    holdings_source_uuid = cache.resolve_holdings_source("MARC")
+"""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable
+from typing import Any, cast
+
+logger = logging.getLogger(__name__)
```
StepanBrychta: Minor: In other places, we usually use `structlog` (though there are a few places where we still use `logging`).

catalogue_graph/src/adapters/transformers/axiell_folio_sync/mapper.py
```diff
@@ -0,0 +1,92 @@
...
+        nodes = root.findall(
+            f".//marc:datafield[@tag='{tag}']/marc:subfield[@code='{code}']", MARC_NS
+        )
+    else:
+        nodes = root.findall(f".//datafield[@tag='{tag}']/subfield[@code='{code}']")
+    for node in nodes:
+        value = (node.text or "").strip()
+        if value:
+            return value
+    return None
```
StepanBrychta: In other places, we use `pymarc` to parse MARC21 records. Could we use it here to simplify this code?

We also have existing helper functions which might be relevant here. For example, see the `first_non_empty_subfield` function, which could probably be used in place of the `_first_subfield` helper.

catalogue_graph/src/clients/folio_client/client.py
```diff
@@ -0,0 +1,113 @@
...
+
+class FolioClient:
+    def __init__(
+        self,
+        base_url: str,
+        tenant: str,
+        *,
+        username: str | None = None,
+        password: str | None = None,
+        credentials_provider: CredentialsProvider | None = None,
```
StepanBrychta, line 48: Minor: This provides two ways of passing credentials (direct `username/password` vs `credentials_provider`), but only one of them is currently used by callers. Do we want to keep both?

catalogue_graph/src/adapters/transformers/axiell_folio_sync/mapping.py
```diff
@@ -0,0 +1,347 @@
...
+    "loan_type_code": "949$l",
+    "copy_number": "876$p",
+    "volume": "876$t",
+    "electronic_access_uri": "856$u",
+}
+
+
+def _safe_segment(value: str) -> str:
+    """Lowercase slug from an arbitrary string — used in composite hrids."""
+    return re.sub(r"[^a-z0-9]+", "-", value.strip().lower()).strip("-") or "unknown"
```
StepanBrychta: Minor: This function doesn't seem to be called anywhere.

catalogue_graph/src/adapters/transformers/axiell_folio_sync/mapping.py
```diff
@@ -0,0 +1,347 @@
...
+        "item": item.model_dump(exclude_none=True),
+        "meta": {
+            "source_id": rec.source_id,
+            "instance_hrid": _instance_hrid(rec.source_id),
+            "holdings_hrid": _holdings_hrid(rec.source_id),
+            "item_hrid": _item_hrid(rec.source_id),
+            "mapping_version": VERSION,
+            "deleted": deleted,
+        },
+    }
```
StepanBrychta: Would it make sense to turn this return value into a Pydantic model, instead of an untyped dictionary? `upsert_from_payloads` could then take the typed model, allowing us to simplify/remove some of the defensive coding in that function (e.g. `meta.get`).

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+        "suppressed": 0,
+        "skipped": 0,
+        "tombstone": 0,
+        "failed": 0,
+        "total": 0,
+    }
+
+    for row in rows:
+        source_id: str = row.get("id", "unknown")
+        changeset_id: str = row.get("changeset", "unknown")
```
StepanBrychta: Is there a good reason for falling back to 'unknown' for rows which don't have an ID? It's a required field, so should its absence be silently tolerated?

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+            errors_list.append(
+                {
+                    "jobId": event.job_id,
+                    "sourceId": source_id,
+                    "changesetId": changeset_id,
+                    "stage": "mapping",
+                    "error": str(exc),
+                    "timestamp": _utc_now_iso(),
+                }
+            )
```
StepanBrychta: There are four repeated blocks like this, which all do the same thing: increment  `counts["failed"]`, increment `total_errors`, and add an error to `error_list`. To make this a bit less verbose, we could introduce a closure inside of the `for row in rows` loop to capture the shared fields (`job_id`, `source_id`, `changeset_id`)

```python
def add_error(stage: str, error: str) -> None:
    counts["failed"] += 1
    errors_list.append(
        {
            "jobId": event.job_id,
            "sourceId": source_id,
            "changesetId": changeset_id,
            "stage": stage,
            "error": error,
            "timestamp": _utc_now_iso(),
        }
    )
```

and then use it like this:

```python
add_error(stage="scan", error="empty_content")
```

I think we can remove the `total_errors` counter, since the same info is already captured in `counts["failed"]`.

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
```diff
@@ -0,0 +1,687 @@
...
+
+        # Loader tombstones are advisory only (RFC 090): the loader's deleted=true is
+        # unreliable, so we record and metric the signal but do NOT suppress/remove.
+        # Authoritative deletes come from the reconciler, not this path.
+        if row.get("deleted"):
+            counts["tombstone"] += 1
+            logger.info(
+                "tombstone_advisory", source_id=source_id, changeset_id=changeset_id
+            )
+            continue
```
StepanBrychta: Are we planning to handle deletions later, or is that not a requirement? The reconciler communicates deletions to downstream pipeline services, but not to Folio.

catalogue_graph/infra/adapters/modules/axiell_folio_sync/s3_manifest.tf
```diff
@@ -0,0 +1,27 @@
+# S3 bucket for NDJSON manifest files produced by sync operations.
+
+data "aws_caller_identity" "current" {}
+
+data "aws_region" "current" {}
+
+resource "aws_s3_bucket" "axiell_folio_sync_manifests" {
+  bucket = var.manifest_bucket_name != "" ? var.manifest_bucket_name : "${var.namespace}-adapter-manifests-${data.aws_caller_identity.current.account_id}-${data.aws_region.current.region}"
+}
```
StepanBrychta, line 9: Minor: We don't usually include account_id/region in S3 bucket names. Most of our buckets are named like `wellcomecollection-*` (e.g. `wellcomecollection-catalogue-graph`, `wellcomecollection-platform-axiell-adapter`).

I also think it might be more intuitive and explicit to make the `manifest_bucket_name` variable required instead of providing a default.

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
anubhavijay: this is fixed in the refactor

catalogue_graph/src/adapters/steps/axiell_folio_sync.py
anubhavijay: yes as a separate ticket as we need to understand the architecture around it

catalogue_graph/src/adapters/transformers/axiell_folio_sync/ref_cache.py
```diff
@@ -0,0 +1,134 @@
...
+    def load(self) -> RefCache:
+        """Fetch all reference data from FOLIO. Returns self for chaining."""
+        location_records = self._fetch_records("/locations", "locations")
+        self._locations = self._build_map(location_records, key="code")
+        self._location_names = self._build_map(location_records, key="name")
+        logger.info("RefCache: %d locations", len(self._locations))
+
+        material_type_records = self._fetch_records("/material-types", "mtypes")
+        self._material_types = self._build_map(material_type_records, key="name")
+        logger.info("RefCache: %d material types", len(self._material_types))
```
StepanBrychta, line 61: We usually prefer using structured logs. Something like:

```python
logger.info("RefCache: material types loaded", count=len(self._material_types))
```

Same comment applies to other logs throughout this PR.

StepanBrychta (approved):
Looks good!

PR #3468, opened 2026-07-15
Refactor adapter transformer manifests/reports

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6444

**Isolate transformer manifests from different stacks**
At the moment, adapter transformer manifest files are written to S3 paths like `/prod/transformer/batches/<changeset_id>.<job_id>.ids.ndjson`, where `job_id` is in the format `YYYYMMDDTHHMM`. This means that two transformers from different stacks (pipeline dates) working on the same changes ID and starting at the same time (which would give them the same `job_id`) can overwrite each other's manifests. 

This PR changes manifest S3 paths to `/pipeline-<pipeline_date>/<transformer_type>/prod/<changeset_id>__<job_id>.json` to keep S3 paths isolated and to follow a convention used in the graph pipeline.

**Simplify and consolidate manifest/report writing across services**
At the moment, the Python codebase defines two related classes for publishing run artefacts to S3: `PipelineReport` (which publishes CloudWatch metrics and S3 reports) and `ManifestWriter` (which publishes S3 manifests). Most pipeline services use `PipelineReport` to publish both metrics and reports, but the adapter transformer and the ID minter use both `PipelineReport` (to publish metrics) and `ManifestWriter` (to publish S3 run artefacts). 

This PR consolidates these classes, removing `ManifestWriter` and publishing all run artefacts using `PipelineReport`.

## How to test

These changes are covered by unit tests. To test manually, run the transformer locally and verify that S3 reports are written to the expected location in the right format. 

* S3 artefact isolation between different pipeline stacks
* Simpler and consolidated report writing across the pipeline

The manifest files written by the `ManifestWriter` were in a different format than those written by `PipelineReport`: Successes and failures were in separate files, and successful IDs were grouped into batches of 5,000, with each batch written to a separate line. This is because the manifests used to be consumed by downstream step functions, which required limiting the size of each batch.

However, manifest consumers are now gone: The ID minter (downstream of the transformer) runs in 'window' mode (reading from an ES index to determine which works to process), and the matcher/merger (downstream of the ID minter), subscribes to an SNS topic. Manifests are now used purely for debugging/observability, so we can simplify how they're written.


catalogue_graph/src/id_minter/steps/id_minter.py
```diff
@@ -158,36 +153,36 @@ def handler(
...
+
+    report = IdMinterReport(
+        pipeline_date=runtime.config.pipeline_date,
+        job_id=event.job_id,
+        successful_ids=successful_ids,
+        errors=errors,
+        s3_bucket=runtime.config.s3_bucket,
+        s3_prefix=runtime.config.s3_prefix,
+    )
+    report.publish()
```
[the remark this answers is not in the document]
StepanBrychta: This is intentional. See the first comment on [this PR](https://github.com/wellcomecollection/catalogue-pipeline/pull/3464#discussion_r3571096611) for full reasoning.

pipeline/terraform/modules/pipeline/id_minter/main.tf
```diff
@@ -80,7 +80,7 @@ data "aws_iam_policy_document" "id_minter_s3_write" {
   statement {
     actions = ["s3:PutObject"]
     resources = [
-      "arn:aws:s3:::${var.env_vars.S3_BUCKET}/${var.env_vars.S3_PREFIX}/id_minter/*",
+      "arn:aws:s3:::${var.env_vars.S3_BUCKET}/pipeline-${var.pipeline_date}/id_minter/${var.env_vars.S3_PREFIX}/*",
```
kenoir: I think this grant may not cover the namespaced instance of the module. The path segment is hardcoded to `id_minter` here, but the code derives it from `IdMinterReport.label`, which is `os.environ.get("PIPELINE_STEP", "id_minter")` (`src/id_minter/reporting.py:14`), and line 34 of this file sets `PIPELINE_STEP = "id_minter${local.underscore_namespace}"`. `state_machine_id_minter_test.tf:83` passes `namespace = "test"`, so that lambda's label is `id_minter_test`.

I ran the report model with that env set to check:

```
label: id_minter_test
uri  : s3://wellcomecollection-platform-id-minter/pipeline-2025-10-02/id_minter_test/test/20260720T1200.json
```

which falls outside `.../pipeline-<date>/id_minter/test/*`. Combined with `publish_to_s3` no longer being pinned to `False` and `publish()` no longer being wrapped, an AccessDenied there would fail the whole run, and that lambda is on a 15 minute schedule.

The old policy matched because `batch_s3_prefix` was `<S3_PREFIX>/id_minter` with `label="id_minter"` passed explicitly, so this looks like it only becomes a mismatch with the new path scheme. Would using `id_minter${local.underscore_namespace}` here line the two back up?

catalogue_graph/src/adapters/transformers/reporting.py
```diff
@@ -10,11 +14,15 @@ class TransformerReport(PipelineReport):
...
-    failure_count: int
-    # S3 publishing is disabled because the manifest writer already writes
-    # successes and failures to S3, from which these counts can be inferred.
-    publish_to_s3: Literal[False] = False
+    changeset_ids: list[str]
+    job_id: str
+    snapshot_id: int | None = None
+
+    successful_ids: list[str]
+    errors: list[TransformationError]
```
kenoir, line 22: `pydantic_to_s3_json` writes `model.model_dump_json()` in one go, so the whole ID list gets built as a single string in memory, where the old writer streamed it in 5,000 line batches. For changeset runs that's obviously fine. Do we know this will work with a no-changeset reindex?

kenoir (commented):
Read through this one carefully. The consolidation reads well, and returning `report_s3_uri` on the result is a nice improvement over the old manifest shape for tracing an artefact. I checked for remaining consumers of the old manifest format in the state machine definitions and in Scala and found none, which backs up the risk section of the description.

Two comments inline: one on the id minter IAM path, which I think is a real mismatch for the namespaced instance, and one question about report size on a reindex.

One thing outside the diff: `catalogue_graph/notebooks/id_minter_rerun_failures.ipynb` reads `result["successes"]["count"]` and `result.get("failures").get("count")` from the lambda response. With `IdMinterResult` those keys are gone, so the rerun notebook will KeyError. Worth updating in this PR or a follow up, since it's the documented recovery path when minting fails.

catalogue_graph/src/adapters/transformers/reporting.py
StepanBrychta: Assuming ~100 bytes per ID and 1 million IDs, this would consume ~100 MB of memory. I think that's small compared to how much memory we consume holding the Iceberg data. (Also notice that the `ElasticBaseTransformer` class already holds the full list of IDs in `self.successful_ids`, so we already do this in other places.)

More broadly, I'm actually not sure we need to list all the IDs in the report. I included them to mirror the `ManifestWriter`, but maybe we don't need them. Other reports only write ID counts, and the transformed records themselves can be found in the works-source index. I think that the main reason why the `ManifestWriter` listed all IDs in the manifest files (and why it batched them) was because the IDs were consumed by downstream services, but nothing consumes these files anymore.

pipeline/terraform/modules/pipeline/id_minter/main.tf
StepanBrychta: Yes, fixed

kenoir (approved):

PR #3469, opened 2026-07-16
Sierra transformer: Fallback to MARC 099 if 001 undefined

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6436

Fallback to MARC 099 in Sierra transformer if MARC 001 not specified. 

Some (but not all) Sierra records will have their Sierra unique reference numbers moved from 001 to 099. We should continue to extract these moved reference numbers.

See https://wellcome.slack.com/archives/C8X9YKM5X/p1777629042762979 for more info.

## How to test

This change is covered by unit tests.

The Sierra transformer should continue extracting Sierra collection paths, falling back to MARC 099 if no value is found under MARC 001.

MARC 099 is only used as fallback, so existing records with a MARC 001 should be processed in the same way as before. Any values migrated from 001 to 099 will be removed from 001, so we do not have to worry about reading stale data from MARC 001. 

There might be existing records which do *not* have a MARC 001 but store some unrelated data in MARC 099, which would now get interpreted as MARC 001. This risk is small: I downloaded a large sample of ~300k Sierra works and verified that none of them currently include a MARC 099 field in their bib data.


pipeline/transformer/transformer_sierra/src/main/scala/weco/pipeline/transformer/sierra/data/BibDataAsMarcRecord.scala
```diff
@@ -74,6 +74,15 @@ class BibDataAsMarcRecord(bibData: SierraBibData)
...
       .map(SierraMarcDataConversions.varFieldToMarcControlField)
 
+  override def controlField(tag: String): Option[MarcControlField] =
+    super.controlField(tag).orElse {
+      // Fall back to 099 if the value from 001 has been moved there. See https://wellcome.slack.com/archives/C8X9YKM5X/p1777629042762979
+      if (tag == "001")
+        controlFieldsWithTags("099").headOption
+          .map(f => MarcControlField("001", f.content))
+      else None
+    }
```
kenoir, line 90: The fallback doesn't fire on the 099 fields that exist in data indicated by the reporting cluster. The reporting cluster (sierra_varfields) currently has 1,594 bibs with a 099 field, and those records store the value in a $a subfield with no top-level content, e.g. b14589321:

```json
{"fieldTag":"l","marcTag":"099","ind1":"0","ind2":"0",
 "subfields":[{"tag":"a","content":"HM3001"}]}
```

This matches how Sierra returns MARC data fields generally: only subfield-less fields get a content string (https://techdocs.iii.com/sierraapi/Content/zObjects/varFieldsArray.htm). All three fallback sites in this PR read varField.content, so for these records they return nothing, and the new tests only cover the bare-content shape, which doesn't appear in the live data.

This site also introduces a small regression for exactly these records: varFieldToMarcControlField maps missing content to "", so the fallback returns Some(MarcControlField("001", "")) instead of None, and MarcCollectionPath will build a CollectionPath with an empty path for root records where today it logs a warning and returns None.

Suggested fixes: read the value from $a (falling back to content if you want to keep that path), add tests using the subfielded shape, and filter out empty content so we never synthesise an empty 001. One thing worth checking with Nicola: the sampled values (HM3001, HM3003, and a few literal "tag 099$a" test records) look like call numbers rather than moved reference numbers, so it may be worth confirming what the migrated values will look like before wiring the fallback to $a.

StepanBrychta: You're right, the value is always stored in the $a subfield. I forgot that the reporting cluster exists, otherwise I would have checked. This issue should now be fixed. The value is always read from 099 $a, and I made a few other tweaks (e.g. not falling back to 099 if multiple instances of 001 exist to continue treating 001 as a non-repeatable field).

The sampled values you listed (`HM3001`, `tag 099$a`, etc.) all belong to works which are either suppressed or deleted. I checked this by retrieving the latest versions of all 1,592 records which contain a 099 field from the VHS store. Out of the 1,592 records:
* 62 are deleted or suppressed
* 180 contain both a 099 field and a 001 field (i.e. no fallback)
* 1,350 contain a 099 field and do not contain a 001 field (i.e. the fallback logic applies)

All of the 1,350 records for which the fallback logic applies store values like `A123` or `Z123` in 099 $a, which is the expected format according to [this Slack message](https://wellcome.slack.com/archives/C8X9YKM5X/p1777629042762979).

kenoir (approved):

PR #3476, opened 2026-07-20
Add a deploy_all_pipelines toggle for deploying every dated pipeline

kenoir:
## What does this change?

Adds a `deploy_all_pipelines` toggle so a merge to main can update every active pipeline instead of just the newest one.

The toggle lives in a new file, `pipeline/terraform/deploy_settings.json`, and defaults to off:

```json
{
  "deploy_all_pipelines": false
}
```

When it is off, behaviour is unchanged: both CI paths deploy only the dated pipeline directory with the most recent date under `pipeline/terraform/`. When it is on, both paths deploy every dated pipeline.

There are two deploy paths and each reads the same setting:

- Buildkite runs `.buildkite/scripts/deploy_latest_pipeline.py`. It now reads `deploy_settings.json` and, when the flag is on, deploys to every dated pipeline. It attempts all of them even if one fails, and exits non-zero at the end if any failed, so one broken pipeline does not skip the rest.
- GitHub Actions runs `catalogue-graph-deploy.yml`. It gains a `discover-pipeline-dates` job that emits a JSON array of the dates to deploy to (all dates when the flag is on, just the newest when off). That array feeds a `pipeline_date` matrix dimension on the dated image-retag and lambda-deploy jobs.

The Actions change also tidies up how discovery and retagging are wired:

- The `retag-dated-pipeline-image` composite action is deleted. Its date-discovery script moves to a new `discover-pipeline-dates` action, which also honours `deploy_settings.json`.
- `deploy-dated-pipeline-lambda` now takes an explicit `pipeline_date` input instead of re-discovering the date and re-retagging the image inside each of its ~12 matrix jobs. Retagging the dated lambda image happens once, up front, in a dedicated `retag-dated-unified-pipeline-lambda-image` job that uses the shared `retag-docker-image` action.

Adds `docs/pipeline/deployment.md` describing how deployment works: dated pipelines, the ECR tag scheme, the two CI paths, and the toggle. It is linked from `docs/SUMMARY.md` and the root `SUMMARY.md`, with a cross-reference from `REINDEXING.md`.

Addresses https://github.com/wellcomecollection/platform/issues/6443.

## How to test

The default is off, so the existing single-pipeline behaviour is unchanged and nothing needs to change to keep the current setup working.

To exercise the toggle:

- Run the selection logic in `deploy_latest_pipeline.py` and `discover.sh` against `pipeline/terraform` with `deploy_all_pipelines` off and confirm only the newest dated directory is chosen; flip it on and confirm every dated directory is chosen. Both were run locally in both states with the expected results.
- Check the workflow wiring in `catalogue-graph-deploy.yml`: `discover-pipeline-dates` output feeds the `pipeline_date` matrix on the dated retag and lambda-deploy jobs via `fromJSON`, and the dependent jobs list it under `needs`. The YAML was parsed to confirm the needs/matrix/fromJSON references line up.

While `deploy_all_pipelines` is on, a merge to main should redeploy every dated pipeline rather than only the newest, which is what platform#6443 asks for. Success is that all active pipelines pick up new code from main without a manual per-pipeline deploy.

The flag defaults to off and the off path is the same code path as before, so merging this does not change how deployments behave until someone turns it on in a follow-up pull request.

With the flag on, deploying to more pipelines means more work per merge and more chances for a single pipeline deploy to fail. The Buildkite script attempts every pipeline and only fails the build at the end, so one failure does not silently skip the others. Turning the flag on is the wrong choice when an older pipeline is deliberately pinned (for example when a change is incompatible with its index mappings); the deployment docs call this out and note that flipping the flag is itself a pull request against `deploy_settings.json`.


.buildkite/scripts/deploy_latest_pipeline.py
```diff
@@ -28,6 +30,19 @@ def deploy_to(root, pipeline_date):
...
+    Reads pipeline/terraform/deploy_settings.json, which controls whether we
+    deploy to every dated pipeline or just the most recent one.
+    """
+    settings_path = f"{root}/pipeline/terraform/deploy_settings.json"
+    try:
+        with open(settings_path) as f:
+            return json.load(f)
+    except FileNotFoundError:
+        return {"deploy_all_pipelines": False}
+
```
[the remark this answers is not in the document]
kenoir: Applied in part in b1afca0df8. A non-boolean value now exits with a clear error naming the file and the value, which also keeps this path consistent with discover.sh (a string "false" would previously have deployed everything here but only the latest pipeline via GitHub Actions). Malformed JSON is left to crash as before: this runs in CI and a loud parse failure pointing at the file is more useful than a silent fallback to the default.

.github/workflows/catalogue-graph-deploy.yml
StepanBrychta, line 1: Minor: I think we can consolidate the four retagging actions (`retag-unified-pipeline-task-image`, `retag-dated-unified-pipeline-task-image`, `retag-dated-unified-pipeline-lambda-image`, `retag-dated-unified-pipeline-lambda-image`) into two to reduce repeated code: 

```
  retag-prod-images:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        image_name: [ unified_pipeline_task, unified_pipeline_lambda ]
    steps:
      - uses: actions/checkout@v3
      - uses: ./.github/actions/retag-docker-image
        with:
          registry: 760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome
          image_name: ${{ matrix.image_name }}
          source_tag: ${{ inputs.deploy_tag }}
          target_tag: prod
          role_to_assume: ${{ secrets.CATALOGUE_GRAPH_CI_ROLE_ARN }}

  retag-dated-images:
    runs-on: ubuntu-latest
    needs: [ discover-pipeline-dates ]
    strategy:
      matrix:
        pipeline_date: ${{ fromJSON(needs.discover-pipeline-dates.outputs.pipeline_dates) }}
        image_name: [ unified_pipeline_task, unified_pipeline_lambda ]
    steps:
      - uses: actions/checkout@v3
      - uses: ./.github/actions/retag-docker-image
        with:
          registry: 760097843905.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome
          image_name: ${{ matrix.image_name }}
          source_tag: ${{ inputs.deploy_tag }}
          target_tag: env.${{ matrix.pipeline_date }}
          role_to_assume: ${{ secrets.CATALOGUE_GRAPH_CI_ROLE_ARN }}
```

A small downside is that the Lambda deploy jobs would have to wait until the ECS task images are tagged (instead of just the Lambda images), but this should have minimal impact in practice, since the retag operations run in parallel.

StepanBrychta (approved):

.github/workflows/catalogue-graph-deploy.yml
kenoir: Applied in e103236090. The four jobs are now retag-prod-images and retag-dated-images, each with an image_name matrix, and the two lambda deploy jobs depend on the consolidated jobs. This also drops the standalone configure-aws-credentials step from the prod jobs, since the retag-docker-image action configures its own credentials.

PR #3477, opened 2026-07-20
Change graph stack outputs and update downstream consumers

StepanBrychta:
## What does this change?

Change the outputs of the graph terraform stack to include outputs from all clusters rather than just the current production cluster. This change was originally included in the [new pipeline PR](https://github.com/wellcomecollection/catalogue-pipeline/pull/3465) and applied as part of the new terraform stack, but I moved it to a separate branch so that we can merge it independently and prevent terraform drift on the main branch. 

## How to test

* Run `terraform apply` on the graph stack and check outputs.
* Run `terraform apply` on the GitHub stack
* Verify that all consumers of the graph stack outputs access the new output names

* The GitHub action running integration tests should be able to run the tests against all clusters, not just the current production cluster
* The IAM permissions defined in each pipeline stack should be scoped to the right Neptune cluster, depending on the stack's `graph_date`

This change is low risk. Running `terraform plan` on the 2025-10-02 stack yields no changes.

### DIFF
--- a/builds/deploy_catalogue_pipeline.sh
+++ b/builds/deploy_catalogue_pipeline.sh
@@ -59,7 +59,8 @@ ENV_TAG="env.$PIPELINE_DATE" "$ROOT/builds/update_ecr_image_tag.sh" \
   uk.ac.wellcome/transformer_miro \
   uk.ac.wellcome/transformer_sierra \
   uk.ac.wellcome/transformer_tei \
-  uk.ac.wellcome/unified_pipeline_lambda
+  uk.ac.wellcome/unified_pipeline_lambda \
+  uk.ac.wellcome/unified_pipeline_task
   
 if [[ "$TASK" == "tag_images_and_deploy_services" ]]
 then
--- a/catalogue_graph/infra/graph/outputs.tf
+++ b/catalogue_graph/infra/graph/outputs.tf
@@ -1,15 +1,17 @@
-output "neptune_cluster_arn" {
-  value = module.catalogue_graph_neptune_cluster.neptune_cluster_arn
+# Output maps keyed by graph_date.
+
+output "neptune_cluster_arns" {
+  value = { for cluster in local.neptune_clusters : cluster.graph_date => cluster.neptune_cluster_arn }
 }
 
-output "neptune_cluster_resource_id" {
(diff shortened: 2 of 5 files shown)


builds/deploy_catalogue_pipeline.sh
```diff
@@ -59,7 +59,8 @@ ENV_TAG="env.$PIPELINE_DATE" "$ROOT/builds/update_ecr_image_tag.sh" \
   uk.ac.wellcome/transformer_miro \
   uk.ac.wellcome/transformer_sierra \
   uk.ac.wellcome/transformer_tei \
-  uk.ac.wellcome/unified_pipeline_lambda
+  uk.ac.wellcome/unified_pipeline_lambda \
+  uk.ac.wellcome/unified_pipeline_task
```
StepanBrychta, line 63: We forgot to include `unified_pipeline_task` here. Discovered while testing the new pipeline.

catalogue_graph/infra/graph/outputs.tf
```diff
@@ -1,15 +1,17 @@
-output "neptune_cluster_arn" {
-  value = module.catalogue_graph_neptune_cluster.neptune_cluster_arn
+# Output maps keyed by graph_date.
+# Note: the legacy production cluster uses an empty string ("") graph_date, so prod is keyed by "" in these maps.
+output "neptune_cluster_arns" {
```
kenoir, line 3: Since the old singular outputs are removed rather than kept alongside, the builds/terraform/github stack and the pipeline stacks won't plan cleanly until this stack has been applied with the new outputs. I assume the plan is: merge, apply the graph stack, then apply builds/terraform/github (which also needs an apply anyway for the CI role to pick up access to the dev and 2026-07-03 clusters). Worth adding that second apply to the "How to test" / rollout notes?

kenoir (approved):
Looks good. I checked that the map keys line up with what the pipeline stacks pass (the 2025-10-02 stack passes an empty graph_date, matching the prod key), that nothing outside the files updated here consumes the old singular outputs, and that the unified_pipeline_task ECR repo name matches ecr.tf. Indexing by var.graph_date is a nice improvement too, since a stack naming a nonexistent cluster now fails at plan time instead of silently getting prod access. One question inline about apply ordering for the downstream stacks.

PR #3478, opened 2026-07-21
Adapter-side reconciliation foundations and stateless deletion delivery (Phase A)

kenoir:
## What does this change?

Phase A of wellcomecollection/platform#6446: it moves Axiell reconciliation detection into the adapter and makes deletion delivery stateless. This is additive only. Nothing changes in prod behaviour yet.

The problem this sets up to fix: pipeline stacks share one reconciler store, so when a guid mapping changes, only the first pipeline to process the changeset emits the deletion. Committing the mapping destroys the diff for every pipeline that runs after it. Rather than give each pipeline its own store copy, detection moves into the adapter (which runs once) and delivery is driven from an append-only record instead of a live diff.

This PR adds the pieces, wired but not yet enabled in prod terraform:

- A new append-only deletion facts table and store (`DELETION_FACTS_ICEBERG_SCHEMA`, `DeletionFactsStore.append_facts`). Fact ids are deterministic (`"{record_id}/{changeset_id}"`) so a retry dedupes rather than double-writes.
- Shared guid computation in `adapters/steps/oai_pmh/guid_mapping.py`.
- The adapter-side detection step in `adapters/steps/oai_pmh/reconcile.py`: it reads changeset rows, diffs guid mappings against the reconciler store, writes facts first and then commits the mappings, and echoes `job_id`/`changeset_ids`/`covered_window_keys` for state machine pass-through. It includes a guid-handoff guard that suppresses a fact when the guid is claimed by a different record, excluding claimants that provably move off the guid in the same update.
- Delivery folded into the Axiell transformer via a new `AxiellStoreSource`. Fact rows stream after adapter rows on incremental runs only, become deleted works through the existing `ReconcilerWorkBuilder`, and are skipped with a warning if the guid is active again in the reconciler mappings (a liveness check so a stale fact cannot tombstone a live work). It tolerates a missing facts table, so it is a no-op until the table exists.
- `MarcXmlTransformer` gains a behaviour-preserving `_transform_row` hook so the transformer can dispatch to the store source.

The existing per-pipeline reconciler (`axiell_reconciler.py`, the `axiell_reconciler` transformer type, and its tests) is left untouched on purpose. It keeps running until Phase C so there is no gap at cutover. The apparent duplication between it and the new path is expected.

## How to test

- `cd catalogue_graph && uv run pytest tests/adapters/` (934 tests pass).
- `uv run ruff check` is clean.
- `uv run mypy .` is clean over 611 files.

New tests cover the added behaviour:
- `tests/adapters/utils/test_deletion_facts_store.py`
- `tests/adapters/extractors/oai_pmh/test_reconcile.py`
- `tests/adapters/transformers/axiell/test_axiell_facts_delivery.py`

Because none of this is enabled in prod terraform yet, there is no runtime behaviour to exercise in an environment. Phase B turns it on.

No prod-facing metric changes here. Success for this PR is that the new detection and delivery code lands with test coverage and no change to existing pipeline output. The end-to-end measure (every pipeline stack emitting the deletion for a guid change, not just the first) lands once Phase B enables the step.

- No behaviour change in prod: the reconcile step is not wired into the adapter state machine yet (Phase B, separate PR), and `AxiellStoreSource` is a no-op while the facts table does not exist.
- The old per-pipeline reconciler is intentionally kept so there is no reconciliation gap before Phase C removes it.
- Stale facts tombstoning live works is guarded by the liveness check against current reconciler mappings; guid handoffs between records are guarded by the handoff exclusion in the detection step.
- Retries are idempotent via deterministic fact ids and append dedup.

Rollout: Phase B enables the reconcile step in the adapter state machine terraform (separate PR). Phase C removes the old per-pipeline reconciler (separate PR).

Issue: wellcomecollection/platform#6446


catalogue_graph/src/adapters/steps/oai_pmh/reconcile.py
```diff
@@ -0,0 +1,294 @@
...
+    preserved, so the guid recomputes identically and produces no diff.
+    """
+    setup_logging(execution_context)
+
+    rows = runtime.adapter_store.get_records_by_changesets(
+        event.changeset_ids
+    ).to_pylist()
+    changeset_by_record_id = {row["id"]: row["changeset"] for row in rows}
+
+    mappings, skipped_ids = rows_to_mappings(rows, runtime.namespace)
```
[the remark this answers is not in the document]
kenoir: The adapter store upserts by id: an incremental update overwrites the existing row (overwrite filter on id) and retags it with the new changeset, so the table holds at most one row per record id. A record modified again in a later changeset belongs only to that newer changeset, so a multi-changeset read cannot return the same id twice.

catalogue_graph/src/adapters/steps/transformer.py
```diff
@@ -117,7 +119,38 @@ def build_transformer(
...
+                )
+                facts_store = DeletionFactsStore(facts_table, namespace="axiell")
+                reconciler_store = ReconcilerStore(reconciler_table, namespace="axiell")
+            except NoSuchTableError:
+                # Tolerated so the transform still runs before the adapter
+                # reconcile step has created the table (cutover, fresh dev
+                # environments).
+                logger.warning(
+                    "Deletion facts table not found; transforming without facts delivery"
+                )
```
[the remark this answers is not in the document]
kenoir: Reworded the message to cover both tables and included the exception detail in the log (6d81c0a55c). The tolerant fallback itself is intentional: it lets the transform run before the reconcile step has created the tables during cutover, and stays for fresh dev environments.

catalogue_graph/src/adapters/transformers/axiell_transformer.py
```diff
@@ -9,13 +19,44 @@ def __init__(
...
+        # Deletion facts (appended by AxiellStoreSource) carry a `guid` key;
+        # adapter rows never do. Each fact tombstones its superseded guid.
+        if "guid" in row:
+            builder = ReconcilerWorkBuilder(row["guid"], row["last_modified"])
+            yield (
+                row["id"],
+                builder.transform_deleted_work(deleted_reason=DeletedFromSource()),
+            )
+        else:
+            yield from super()._transform_row(row)
```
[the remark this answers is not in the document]
kenoir: Wrapped the fact branch in try/except recording via _add_error, matching the MARC row handling, with a test that a broken fact lands in the failure manifest while the run continues (6d81c0a55c). Facts are durable, so a recorded failure loses nothing.

catalogue_graph/src/adapters/transformers/axiell_store_source.py
```diff
@@ -0,0 +1,75 @@
...
+    Fact rows carry a `guid` key (the superseded guid), which adapter rows
+    never do; the transformer discriminates on that key. Facts are only read
+    for incremental runs: a full reindex writes into an empty index, so
+    historic deletions have nothing to overwrite.
+
+    Facts capture detection-time state but live forever, so each is
+    re-checked against the current reconciler mappings: a fact whose guid has
+    since been reclaimed (revert, handoff, or a redrive of an old changeset)
+    is skipped, as its tombstone would overwrite the live work now indexed
+    under that guid.
```
StepanBrychta, line 28: Are the 'revert' and 'handoff' scenarios possible? Assuming that a handoff involves reusing GUIDs between works (mapping some existing GUID to a new collect ID), it would break the GUID uniqueness constraint, which would have downstream consequences (e.g. the ID minter would assign the same ID to different works).

catalogue_graph/src/adapters/utils/facts_delivery.py
```diff
@@ -0,0 +1,70 @@
...
+            use_rest_api_table=use_rest_api_table,
+            create_if_not_exists=False,
+        )
+    except NoSuchTableError as e:
+        logger.warning(
+            "Deletion facts or reconciler table not found; "
+            "transforming without facts delivery",
+            error=str(e),
+        )
+        return None
```
StepanBrychta: Is there a good reason for soft failing when the tables don't exist? It risks hiding real issues: If the tables are missing in production, something is probably wrong, and it might make sense to stop the transformer.

catalogue_graph/src/adapters/steps/oai_pmh/reconcile.py
```diff
@@ -0,0 +1,288 @@
...
+    adapter_store: AdapterStore
+    reconciler_store: ReconcilerStore
+    facts_store: DeletionFactsStore
+    adapter_name: str
+    namespace: str
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+
+def handler(
```
StepanBrychta, line 66: This function spans 120 lines of code. Splitting it into smaller functions with descriptive names would make it easier to follow.

catalogue_graph/src/adapters/utils/facts_delivery.py
kenoir: Agreed, made it fail hard (e83d00dbcf). A missing table now raises NoSuchTableError, matching the FOLIO items store behaviour. The reconcile step creates both tables on its first run, so a fresh environment needs one reconcile pass before incremental transforms; the rollout ordering just needs the tables to exist before the new image serves incremental runs.

catalogue_graph/src/adapters/transformers/axiell_store_source.py
kenoir: Revert is real: the guid comes from record content, so it can be edited back. If g1 -> g2 and then g2 -> g1 land in one incremental run (or an old changeset is redriven), the first fact would tombstone the work just re-written under g1; this check skips it because g1 is an active mapping again. Handoff we have not observed, and agreed it would be a source data error with the minter consequences you describe. The reconcile step suppresses and logs handoff facts (guard_suppressed count), so it surfaces; this check just stops delivery deleting the live work under the guid's current owner.

catalogue_graph/src/adapters/steps/oai_pmh/reconcile.py
```diff
@@ -0,0 +1,288 @@
...
+            guard_suppressed += 1
+            logger.warning(
+                "Suppressed deletion fact for handed-off guid",
+                adapter=runtime.adapter_name,
+                job_id=event.job_id,
+                record_id=row["id"],
+                guid=row["guid"],
+                claimed_by=sorted(other_claimants),
+            )
+            continue
```
StepanBrychta: Following up on my other comment: I think this check is useful, but I also think the comment should be explicit about the fact that this can only happen as a result of a serious data quality issue in the source system. We might also want to stop the run so that we are alerted. GUID uniqueness is a core assumption and handoffs should never happen.

StepanBrychta (approved):

catalogue_graph/src/adapters/steps/oai_pmh/reconcile.py
kenoir: Split into _superseded_mappings, _guid_claimants, _reject_guid_handoffs and _build_deletion_facts, with the reasoning comments moved onto the helpers they explain (2d021cf887, reshaped further in 38c97b1aa2).

catalogue_graph/src/adapters/steps/oai_pmh/reconcile.py
kenoir: Agreed, made it fatal (38c97b1aa2). The check can only fire when two records have carried the same guid, so it now logs each affected record at error level and raises before any facts or mappings are written; the state machine run fails and alerts. The docstring spells out that this is a serious source data quality issue and why reverts never trip it (a record never claims its own old guid). guard_suppressed is gone from the response since a successful run can no longer suppress anything.

PR #3494, opened 2026-07-24
Accept same-version updates that re-link merge candidates in the matcher

kenoir:
## What does this change?

When a transformer change corrects a work's merge candidates, the work is re-processed at the **same version** (the source record itself hasn't changed). The matcher treated a same-version update with different merge candidates as a conflict and refused it: depending on the deployment the message was either dropped after an error log or retried until it reached the DLQ. Either way the corrected links never re-matched or re-merged until the source record next happened to be edited.

The matcher now accepts same-version updates. Last write wins, and components re-form or split through the normal update path. It still rejects strictly-older updates, and logs a warning whenever same-version content differs so the anomaly stays observable.

To make last-write-wins safe under concurrency, the matcher re-checks the stored work inside the lock and bails out if a fresher copy was matched while it waited, letting SQS retry with the current copy rather than reverting the corrected graph.

Behaviour applies to all pipelines on next deploy, and re-transform corrections become self-healing. One observability change to note: equal-version content divergence used to show up as matcher DLQ messages; it now shows up as a warning log ("resubmitted with different content"), so anything watching the matcher DLQ for this case should watch that log line instead.

Supports the full reindex work (wellcomecollection/platform#6445).

## How to test

The full matcher suite (docker-backed fixtures: localstack SQS, DynamoDB local, ES) passes locally, 71/71, including new tests for same-version re-link, split, idempotent redelivery, and the stale-copy bail-out. CI runs the same suite.

Re-transformed works whose merge candidates changed re-match and re-merge without a source-record edit, and no `different content` conflicts are dropped.

Last-write-wins could revert a concurrent update; the in-lock re-check guards against that by bailing out for SQS to retry. The warning log keeps genuinely-diverging same-version content visible.

### DIFF
--- a/pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
+++ b/pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
@@ -15,6 +15,7 @@ import weco.pipeline.matcher.models.{
 import weco.pipeline.matcher.services.LockingBuilder
 import weco.pipeline.matcher.storage.{WorkGraphStore, WorkNodeDao}
 import weco.pipeline.matcher.workgraph.WorkGraphUpdater
+import weco.pipeline_storage.Retriever
 import weco.storage.dynamo.DynamoConfig
 import weco.storage.locking.dynamo.DynamoLockDaoConfig
 import weco.storage.locking.{
@@ -34,7 +35,8 @@ trait WorkStubMatcher {
 
 class WorkMatcher(
   workGraphStore: WorkGraphStore,
-  lockingService: LockingService[MatcherResult, Future, LockDao[String, UUID]]
+  lockingService: LockingService[MatcherResult, Future, LockDao[String, UUID]],
+  retriever: Retriever[WorkStub]
 )(implicit ec: ExecutionContext)
     extends WorkStubMatcher
     with Logging {
@@ -46,6 +48,18 @@ class WorkMatcher(
 
     withLocks(work, initialLockIds) {
       for {
+        // The work we were given was fetched before we acquired any locks,
+        // so another process may have matched a newer copy of it in the
+        // meantime -- e.g. a re-transform that corrected its merge candidates
+        // at the same version.  If we carried on with a stale copy, we'd
+        // accept it as a last-write-wins update and quietly revert the
+        // corrected graph.
+        //
+        // Check the stored work is still the one we're processing -- if it's
+        // stale, we should bail out and wait for the SQS logic to retry us
+        // with a fresh copy rather than recover.
+        _ <- checkWorkIsCurrent(work)
+
         beforeNodes <- workGraphStore.findAffectedWorks(work.ids)
         afterNodes = WorkGraphUpdater.update(work, beforeNodes)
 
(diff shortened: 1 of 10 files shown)


pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
```diff
@@ -77,6 +91,24 @@ class WorkMatcher(
...
+              val t = new RuntimeException(
+                s"Error processing ${work.id}: stored work changed during matching"
+              )
+              Future.failed(t)
+
+            // If the work can't be retrieved (e.g. when the matcher is
+            // driven directly with a work that isn't in the store), carry
+            // on with the copy we were given.
+            case _ => Future.successful(())
+          }
```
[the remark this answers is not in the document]
kenoir: Both copies compared here come from the same retriever against the same index: the matcher is driven by work ids, and StoredWorksMatcher fetches the stub with the same ElasticWorkStubRetriever that checkWorkIsCurrent uses. They deserialise the same stored _source, so they can only differ if the document actually changed, which is exactly what this check is for. A mismatch is retryable and the retry processes the fresh copy, so a false positive costs one retry at worst. Comparing a subset of fields would risk carrying on with a genuinely stale copy, so full equality is deliberate.

pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/workgraph/WorkGraphUpdater.scala
```diff
@@ -58,10 +57,13 @@ object WorkGraphUpdater extends Logging {
...
-          s"update failed, work:${work.id} v${work.version} already exists with different content! update-ids:${work.mergeCandidateIds} != existing-ids:${existingMergeCandidateIds.toSet}"
-        debug(versionConflictMessage)
-        throw VersionUnexpectedConflictException(versionConflictMessage)
+        // This can happen when a transformer change corrects a work's merge
+        // candidates without the source record itself changing: the work is
+        // re-processed at the same version but with different candidates.
+        // We accept the update (last write wins) so the graph can re-form.
+        warn(
+          s"work:${work.id} v${work.version} resubmitted with different mergeCandidates (${existingMergeCandidateIds.toSet} -> ${work.mergeCandidateIds}); accepting update"
+        )
```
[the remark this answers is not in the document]
kenoir: Fair point: the PR description promises a warning whenever same-version content differs, and suppressed is the only other matcher-relevant field in SourceWorkData. The warning now also fires when suppression changes at the same version, with the old and new values in the message; acceptance behaviour is unchanged. Fixed in fa2c6bba14.

pipeline/matcher_merger/matcher/src/main/scala/weco/pipeline/matcher/matcher/WorkMatcher.scala
```diff
@@ -46,6 +48,18 @@ class WorkMatcher(
 
     withLocks(work, initialLockIds) {
       for {
+        // The work we were given was fetched before we acquired any locks,
```
agnesgaroux, line 51: What about moving the locking earlier in the flow, ie. moving lockingService from WorkMatcher to StoredWorksMatcher? That would remove the need for retriever as a WorkMatcher dependency + checkWorkIsCurrent check

kenoir: I had a go at this and I don't think the lock can move cleanly:

- The initial lock is over `work.ids`, which includes the merge candidate ids, and those come from the fetched content. So the full lock set can only be computed after the fetch, wherever the fetch lives, and the fetch-to-lock window stays open.
- We could lock on just the work id in StoredWorksMatcher and fetch under it, but WorkMatcher then can't re-acquire that id: `withLocks` creates a fresh lock context per call and the lock dao only re-grants a lock to the same context, so we'd have to split the lock set across the two classes.
- The locking service has to stay in WorkMatcher anyway for the second-phase subgraph locks in `writeUpdate`, which depend on the graph store results.

Given that, the in-lock re-check seemed like the smallest guard, and it also covers MatcherWorkerService, which drives `matchWork` directly with a pre-fetched stub. The retriever passed in is the same instance StoredWorksMatcher already uses. Can talk it through if you think the restructure is worth doing.

agnesgaroux: This works, and we can think about improving it when we pythonise the matcher-merger 😉

agnesgaroux (approved):

PR #3495, opened 2026-07-24
Ingest unmigrated calm records

### LINKED ISSUE #6448: Feed 3rd party unmigrated CALM records to new pipeline
There are some Calm records that CI do not want to migrate to Axiell, because they are other institutions' data, and the plan is to stop maintaining them and showing them on the website because we don't have the things (similar to Brought To Life). However, we want to give the owners plenty of notice before we stop showing it. More notice than we have time to give before Calm is switched off.

There is sufficient complexity in the records that it would be difficult to migrate them to Axiell.  Given that they are soon to be thrown away, a simpler solution is to leave them where they are.

There would therefore be a period when the records are not being managed in any CMS, but they should still be displayed on the website.

To enable this, the pipeline should continue to consume certain stale records sourced from CALM (list to be provided by VS).

A likely solution would be to use the CALM VHS to feed the existing CALM transformer.  However, other solutions may present themselves (in theory, data from anywhere before the matcher/merger would do, and if none of the records served are involved in merges with non-CALM data, then it might be possible to simply retain these records as they are in works-indexed)

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6448
Adds tooling to run `transformer_calm` locally for testing or batch-processing a subset of CALM records without using the deployed pipeline. 

This is useful for:
- Testing transformer changes against specific CALM records before deploying
- Ingesting a specific list of CALM IDs, like the ones in the linked issue
- Local development and debugging of transformer behaviour

### New files

- **`Dockerfile.local`** — Lightweight local dev image (JRE + Python + bash)
- **`local.docker-compose.yml`** — Docker Compose with:
  - `calm-transformer` — the Scala transformer
  - `elasticsearch` — local ES 7.17 for safe test indexing (default)
  - `localstack` — local SQS/SNS
- **`run_local/local_transformer.py`** — Python CLI with subcommands:
  - `start` — fetch ES config, build app, start services, stream logs
  - `enqueue` — read CALM IDs from a file, fetch metadata from DynamoDB, send to local SQS
  - `verify-completion` — wait for queue drain, check ES coverage, scan logs for errors
  - `fetch-es-env` — standalone ES config fetcher

### Workflow

```bash
# From pipeline/transformer/transformer_calm:
uv run --script run_local/local_transformer.py start 2026-07-03 --index-name works-source-local-test

# In another terminal:
uv run --script run_local/local_transformer.py enqueue /path/to/calm_ids.txt
uv run --script run_local/local_transformer.py verify-completion /path/to/calm_ids.txt
```

By default, writes go to local Docker Elasticsearch. Pass `--es-host public` to write to the deployed pipeline index instead.

### Checklist

- [x] Does this change the display model the catalogue API returns? **No** — this is dev tooling only, no changes to transformer logic or output model.

### Updates in latest commits

- Refines the local tooling after review feedback:
  - Uses `.env` wiring for local credentials/config in the container workflow
  - Improved start script: calm_transformer services depends on successful create index. Index is created with prod mappings. 
  - Handle service errors with retries and DLQ
  - batch messages 
  - Improved verify script: tighten log filtering and use --since, check DLQ
  - use ES 9.1.4

## How to test

1. Follow the README at `pipeline/transformer/transformer_calm/run_local/README.md`
2. Create a file with a few CALM IDs (one per line)
3. Run `start`, `enqueue`, `verify-completion`
4. Confirm documents appear in the local ES index
5. Confirm any missing IDs are written to `missing_ids.txt` as plain CALM IDs
6. Failure mode: hardcode a dummy S3 key: the message should fail twice, then go to the DLQ. The verify script should read the DLQ and write the failed id to the missing_ids report 

- Developers can run the local transformer workflow without errors
- The verify-completion step reports all enqueued IDs as indexed. Any ids in the input list that are not in the index are listed in `missing_ids.txt` at the end of the verify step 
- Fewer manual/debugging steps are needed to trial targeted CALM ingests before deployment

- **Accidental writes to production ES** — mitigated by defaulting to local ES (`--es-host local`). Public ES requires explicit `--es-host public` flag.
- **AWS credential exposure** — credentials are written to `.env` (gitignored) and injected into the container. Session tokens expire, limiting exposure window.


pipeline/transformer/transformer_calm/run_local/README.md
```diff
@@ -0,0 +1,130 @@
...
+  - For `--es-host local` (default), writes:
+    - `es_host=elasticsearch`, `es_port=9200`, `es_protocol=http`, empty `es_apikey`
+  - For `--es-host public`, reads:
+    - `elasticsearch/pipeline_storage_<pipeline_date>/public_host`
+    - `elasticsearch/pipeline_storage_<pipeline_date>/port`
+    - `elasticsearch/pipeline_storage_<pipeline_date>/protocol`
+    - `elasticsearch/pipeline_storage_<pipeline_date>/transformer/api_key`
+  - Writes `.env.local` containing:
+    - `es_host`, `es_port`, `es_protocol`, `es_apikey`, `es_index`
+    - resolved AWS session credentials for the container (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`)
```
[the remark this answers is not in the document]
agnesgaroux: Switched to .env file

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,552 @@
...
+        cwd=project_dir,
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    logs_text = logs_proc.stdout + logs_proc.stderr
+    error_pattern = re.compile(
+        r"DecodePayloadError|StoreReadError|TransformerError|ERROR|Exception"
+    )
+    error_lines = [
```
kenoir: `verify-completion` returns 1 on a clean run.

I ran the documented workflow against local ES with 50 CALM IDs pulled from the VHS. All 50 transformed and indexed, and it still failed:

```
Elasticsearch coverage: found=50 expected=50 missing=0
Potential transformer errors detected:
calm-transformer-1  | 13:17:26,957 |-INFO in ch.qos.logback.classic.model.processor.LoggerModelHandler - Setting level of logger [org.apache.http] to ERROR
calm-transformer-1  | 13:17:26,957 |-INFO in ch.qos.logback.classic.model.processor.LoggerModelHandler - Setting level of logger [io.netty] to ERROR
CLEAN_RUN_EXIT_CODE=1
```

Those two lines are logback's own startup output from `src/main/resources/logback.xml`, which sets `org.apache.http` and `io.netty` to level ERROR. The bare `ERROR` alternative on line 449 matches the level name inside that INFO message, so the check fires on every run before a single message has been processed.

Two changes would sort it:

- Drop the bare `ERROR` alternative, or match the level field rather than the word, so something like `\bERROR\b\s+w\.p\.transformer`. The `DecodePayloadError|StoreReadError|TransformerError` alternatives are specific enough to be useful on their own.
- Pass `--since` to `docker compose logs` on line 433. As written it reads the container's entire log, so errors from an earlier `enqueue` run count against the current verification.

Worth fixing because the exit code is the thing a batch run would key off, and right now it carries no signal.

pipeline/transformer/transformer_calm/local.docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+    image: "public.ecr.aws/localstack/localstack:3.5.0"
+    depends_on:
+      localstack:
+        condition: service_healthy
+    entrypoint: []
+    command:
+      - sh
+      - -c
+      - |
+        awslocal sqs create-queue --queue-name calm-transformer-queue --region eu-west-1
```
kenoir, line 71: The queue is created with no redrive policy, so a record that fails to transform never leaves the queue and `verify-completion` can only ever time out.

The queue as this line creates it, note there is no `RedrivePolicy` key:

```
$ aws --endpoint-url=http://localhost:4566 --region eu-west-1 sqs get-queue-attributes \
    --queue-url http://localhost:4566/000000000000/calm-transformer-queue --attribute-names All
{
    "Attributes": {
        "ApproximateNumberOfMessages": "0",
        "ApproximateNumberOfMessagesNotVisible": "0",
        "ApproximateNumberOfMessagesDelayed": "0",
        "CreatedTimestamp": "1784900654",
        "DelaySeconds": "0",
        "LastModifiedTimestamp": "1784900654",
        "MaximumMessageSize": "262144",
        "MessageRetentionPeriod": "345600",
        "QueueArn": "arn:aws:sqs:eu-west-1:000000000000:calm-transformer-queue",
        "ReceiveMessageWaitTimeSeconds": "0",
        "VisibilityTimeout": "30",
        "SqsManagedSseEnabled": "true"
    }
}
```

I tested the effect by putting a single message on the queue with an S3 key that does not exist, which is the `StoreReadError` path in `TransformerWorker`. The transformer throws, the message is never deleted, and it returns every 30 seconds indefinitely. Two of the four deliveries in the first 90 seconds, with the interleaved stack traces omitted:

```
13:20:29.590 [main-actor-system-pekko.actor.default-dispatcher-5] ERROR w.p.transformer.TransformerWorker - TransformerWorker: StoreReadError on CalmSourcePayload(poison-test-id,s3://wellcomecollection-vhs-calm-adapter/does-not-exist/1/nope.json,1,false)
13:21:00.222 [main-actor-system-pekko.actor.default-dispatcher-5] ERROR w.p.transformer.TransformerWorker - TransformerWorker: StoreReadError on CalmSourcePayload(poison-test-id,s3://wellcomecollection-vhs-calm-adapter/does-not-exist/1/nope.json,1,false)
```

The container stays healthy throughout, so nothing surfaces the stuck record. `verify-completion` run against that queue with a 30s timeout, showing the tail of an uncaught traceback:

```
Waiting for queue to drain...
Queue status: visible=0 in_flight=1
Queue status: visible=0 in_flight=1
Traceback (most recent call last):
    raise TimeoutError(
TimeoutError: Timed out waiting for queue to drain after 30s
```

So you wait out the full 600s default, get an uncaught traceback, and nothing tells you which record is stuck or why.

A DLQ with a `maxReceiveCount` would make failures terminal and visible:

```sh
awslocal sqs create-queue --queue-name calm-transformer-dlq --region eu-west-1
awslocal sqs create-queue --queue-name calm-transformer-queue --region eu-west-1 \
  --attributes '{"RedrivePolicy":"{\"deadLetterTargetArn\":\"arn:aws:sqs:eu-west-1:000000000000:calm-transformer-dlq\",\"maxReceiveCount\":\"3\"}"}'
```

`verify-completion` could then read the DLQ depth and fold those IDs into the report instead of raising, which would give you a real list of records that could not be processed.

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,552 @@
...
+            "localstack-setup",
+            "elasticsearch",
+            "calm-transformer",
+        ],
+        cwd=project_dir,
+        check=True,
+    )
+
+    if args.es_host == "local":
+        _ensure_local_es_index(env_file=env_file)
```
kenoir: The index is created after the transformer container has already started, and the transformer hard-fails if the index is missing at startup.

`ElasticIndexer.init()` runs when the stream starts, and throws if the index does not exist:

```scala
// common/pipeline_storage/src/main/scala/weco/pipeline_storage/elastic/ElasticIndexer.scala
  final def init(): Future[Unit] = {
    client.execute(indexExists(index.name)).map {
      response =>
        if (!response.result.isExists) {
          throw new RuntimeException(
            s"Indexer Initialisation error looking for index: ${index.name} resp: $response"
          )
        }
    }
  }
```

I confirmed the effect by deleting the index and restarting the container:

```
java.lang.RuntimeException: Indexer Initialisation error looking for index: test-calm-index resp: RequestSuccess(404,None,Map(X-elastic-product -> Elasticsearch, content-type -> application/json; charset=UTF-8, content-length -> 419),IndexExistsResponse(false))
```

```
$ docker inspect -f 'status={{.State.Status}} exit={{.State.ExitCode}}' transformer_calm-calm-transformer-1
status=exited exit=1
```

On my run the Python `PUT` got in before the JVM had finished starting, roughly two seconds, so it worked fine. The ordering is not guaranteed though, and there is no `restart:` policy on `calm-transformer`, so if the container gets there first it stays down, `enqueue` still succeeds against the queue, and `verify-completion` waits until it times out.

`elasticsearch` has no volume here and `start` always passes `--force-recreate`, so the index is gone at the beginning of every `start`, which means the race is run every time rather than only on first use.

The METS setup makes this a compose dependency instead, which removes the ordering question entirely:

```yaml
# pipeline/transformer/transformer_mets/local.docker-compose.yml
    depends_on:
      elasticsearch:
        condition: service_healthy
      localstack:
        condition: service_healthy
      es-setup:
        condition: service_completed_successfully
```

Moving index creation into an `es-setup` service that `calm-transformer` depends on would be the smaller change. Calling `_ensure_local_es_index` before `docker compose up` would also work, though it needs ES up first, so it would mean starting `elasticsearch` separately.

pipeline/transformer/transformer_calm/run_local/README.md
```diff
@@ -0,0 +1,130 @@
...
+- `local_transformer.py verify-completion`
+  - Waits until local input queue is drained (`visible=0` and `in_flight=0`)
+  - Checks transformer logs for errors
+  - Verifies every input ID has a corresponding ES document ID in the format:
+    - `Work[calm-record-id/<id>]`
+
+## Prerequisites
+
+- Docker + Docker Compose
+- Python 3
```
kenoir, line 55: `boto3` is missing from the prerequisites. It is the first thing that happens on a clean machine, with the checkout path shortened:

```
$ AWS_PROFILE=platform-developer ./run_local/local_transformer.py start 2026-07-03 --index-name test-calm-index --no-logs
Traceback (most recent call last):
  File ".../run_local/local_transformer.py", line 16, in <module>
    import boto3
ModuleNotFoundError: No module named 'boto3'
```

The `#!/usr/bin/env python3` shebang and the `run_local/local_transformer.py start` invocations in the README both suggest the script is directly runnable, so this is a slightly confusing first five minutes.

The repo standard elsewhere is uv, as in `catalogue_graph`. Inline script metadata would keep this a single self-contained file and make the documented invocation work as written:

```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.11"
# dependencies = ["boto3"]
# ///
```

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,552 @@
...
+        if err.code != 404:
+            raise
+
+    create_req = urllib.request.Request(
+        index_url,
+        data=b"{}",
+        headers=headers,
+        method="PUT",
+    )
+    with urllib.request.urlopen(create_req, context=context) as response:
```
kenoir: Creating the index with `{}` means the local run does not exercise the mapping the deployed index has, so it can pass locally and behave differently against the real thing.

The real mapping is `dynamic: false` with a `lowercase_normalizer` on the identifier fields, from `index_config/mappings.works_source.2026-03-25.json`:

```json
{
  "dynamic": "false",
  "properties": {
    "indexed_at": {
      "type": "date"
    },
    "data": {
      "properties": {
        "otherIdentifiers": {
          "properties": {
            "value": {
              "type": "keyword",
              "normalizer": "lowercase_normalizer"
            }
          }
        }
      }
    },
```

What I actually got locally after transforming 50 CALM records into an index created this way:

```
$ curl -s 'http://localhost:9200/test-calm-index/_mapping' | jq '[paths(type=="object" and has("type"))] | length'
139
```

```
$ curl -s 'http://localhost:9200/test-calm-index/_mapping' \
  | jq -r '[paths(objects) as $p | select(getpath($p).type=="date") | ($p|join("."))] | .[]' \
  | sed 's/.mappings.properties//;s/.properties//g'
test-calm-index.data.production.dates.range.from
test-calm-index.data.production.dates.range.to
test-calm-index.state.sourceModifiedTime
```

139 dynamically mapped fields where prod maps almost nothing, and date detection has typed fields that prod never types. A later record with a value that does not parse as a date in one of those fields would be rejected locally and accepted in prod.

METS creates its index with the real mapping for this reason, in the `es-setup` service in `pipeline/transformer/transformer_mets/local.docker-compose.yml`, including the `lowercase_normalizer` and `"dynamic": "false"`. Reading `index_config/mappings.works_source.<date>.json` and `analysis.works_source.<date>.json` and PUTting them would be closer still, and would keep working when the mapping is next revised.

pipeline/transformer/transformer_calm/local.docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+    depends_on:
+      localstack:
+        condition: service_healthy
+      localstack-setup:
+        condition: service_completed_successfully
+      elasticsearch:
+        condition: service_healthy
+
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.17.10
```
kenoir: Is 7.17.10 deliberate here? Everything else in the repo that stands up a local Elasticsearch is on 8.x, the METS transformer aside, which is where this file looks to have come from. `common/pipeline_storage/docker-compose.yml` is on 8.4.0, and that is the module that owns `ElasticIndexer` and does the writes here. Prod is on 9.1.4. It is not incompatible, I ran the full workflow against 7.17.10 and 50 records transformed and indexed fine, but should we try to match prod?

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,553 @@
...
+            "local.docker-compose.yml",
+            "up",
+            "-d",
+            "--build",
+            "--force-recreate",
+            "localstack",
+            "localstack-setup",
+            "elasticsearch",
+            "calm-transformer",
+        ],
```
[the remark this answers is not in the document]
agnesgaroux: I think that's fine

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,550 @@
...
+                f"es_apikey={es_apikey}",
+                f"es_index={index_name}",
+                f"AWS_ACCESS_KEY_ID={frozen_credentials.access_key}",
+                f"AWS_SECRET_ACCESS_KEY={frozen_credentials.secret_key}",
+                f"AWS_SESSION_TOKEN={frozen_credentials.token or ''}",
+                f"AWS_REGION={aws_region}",
+                f"AWS_DEFAULT_REGION={aws_region}",
+                "",
+            ]
+        ),
```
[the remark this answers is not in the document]
agnesgaroux: It does say `AWS_ACCESS_KEY_ID` but no plain test creds

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,552 @@
...
+        local_sqs.send_message(
+            QueueUrl=args.queue_url, MessageBody=json.dumps(wrapped_message)
+        )
+        return True, None
+
+    def process_batch(current_batch: list[str], current_batch_number: int) -> None:
+        nonlocal sent_count, missing_count
+        print(f"Enqueueing batch {current_batch_number} ({len(current_batch)} IDs)")
+        for calm_id in current_batch:
+            sent, warning = enqueue_id(calm_id)
```
kenoir: `batch_size` only changes log grouping. `process_batch` loops the batch calling `get_item` and `send_message` once per ID, so there is no `batch_get_item`, no `send_message_batch` and no pause between batches, and a batch of 250 does exactly the same number of round trips as a batch of 1. Since the README presents it as a tuning knob, it is worth either wiring it to `send_message_batch`, which takes 10 messages per call, or dropping the argument.

pipeline/transformer/transformer_calm/Dockerfile.local
```diff
@@ -0,0 +1,9 @@
+FROM eclipse-temurin:11-jre
+
+LABEL maintainer = "Wellcome Collection <digital@wellcomecollection.org>"
+
+RUN apt-get update && apt-get install -y bash python3 && rm -rf /var/lib/apt/lists/*
```
kenoir: `python3` is never used in this image. The entrypoint is the `transformer_calm` binary and the CLI runs on the host, so `bash` alone would do, which is what the equivalent METS `Dockerfile.local` installs.

pipeline/transformer/transformer_calm/local.docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+      - AWS_EC2_METADATA_DISABLED=true
+      - AWS_ENDPOINT_URL_SQS=http://localstack:4566
+      - AWS_ENDPOINT_URL_SNS=http://localstack:4566
+      - queue_url=http://localstack:4566/000000000000/calm-transformer-queue
+      - sns_topic_arn=arn:aws:sns:eu-west-1:000000000000:calm-transformer-topic
+      - metrics_namespace=calm-transformer
+      - batch_size=100
+      - flush_interval_seconds=30
+    env_file:
+      - .env.local
```
kenoir: The `.env` rename landed in `local_transformer.py` and the README but not in `env_file` here, so compose still looks for `.env.local` and `start` fails on a clean checkout at `4d4b560`:

```
$ docker compose -f local.docker-compose.yml config
env file /.../pipeline/transformer/transformer_calm/.env.local not found: stat /.../pipeline/transformer/transformer_calm/.env.local: no such file or directory
compose config exit=1
```

A leftover `.env.local` from an earlier run gets picked up silently with its old index name and credentials, so this may not show up locally. Changing this line to `- .env` should do it.

kenoir (commented):
This works. I ran the documented workflow end to end against local Elasticsearch with 50 real CALM IDs from the VHS, and all 50 transformed and indexed. The message shape matches what the transformer expects, and the DynamoDB read matches the live `vhs-calm-adapter` row shape including `isDeleted` being absent on normal rows.

Three things I think are worth picking up before merge: the `env_file` line still points at `.env.local` so `start` fails on a clean checkout, `verify-completion` exits 1 even when everything succeeded, and a record that fails to transform has nowhere to go so the queue never drains. The rest of what I have left on the diff is smaller.

I gathered the run evidence at d9eb0c4 and re-checked the findings against 4d4b560 after your two commits. They still hold.

kenoir:
Is this the whole story to address https://github.com/wellcomecollection/platform/issues/6448? I think it's a useful thing to have, but keeping this stuff might stick around for a bit and we may want a way to incorporate this into the reindex script and have the transformer run in AWS?

pipeline/transformer/transformer_calm/run_local/local_transformer.py
agnesgaroux: Removed bare  `ERROR` log matching
The enqueue script now write a start date which is used to filter container logs for the latest run

pipeline/transformer/transformer_calm/local.docker-compose.yml
agnesgaroux: Added redrive: 2 and DLQ 
The verify script checks the DLQ and adds any ids found to the missing_ids report
The DLQ is then purged to prevent false negatives on subsequent runs

pipeline/transformer/transformer_calm/run_local/local_transformer.py
agnesgaroux: Added es-setup service that creates the index, with mappings lifted from index_config

pipeline/transformer/transformer_calm/run_local/README.md
agnesgaroux: Moved to uv

pipeline/transformer/transformer_calm/run_local/local_transformer.py
agnesgaroux: See above, now creating index with prod mappings

pipeline/transformer/transformer_calm/local.docker-compose.yml
agnesgaroux: Align with prod `9.1.4`

pipeline/transformer/transformer_calm/run_local/local_transformer.py
agnesgaroux: Messages now sent in batches of 10

pipeline/transformer/transformer_calm/Dockerfile.local
agnesgaroux: Removed

pipeline/transformer/transformer_calm/local.docker-compose.yml
agnesgaroux: whoops, missed, fixed

agnesgaroux:
> Is this the whole story to address [wellcomecollection/platform#6448](https://github.com/wellcomecollection/platform/issues/6448)? I think it's a useful thing to have, but keeping this stuff might stick around for a bit and we may want a way to incorporate this into the reindex script and have the transformer run in AWS?

My understanding from Thursday's [Slack thread](https://wellcome.slack.com/archives/C02ANCYL90E/p1784821248917379) was that we were going the CLI-only route, to avoid keeping the service deployed/keep the new pipeline CALM-free from an infra perspective. Happy to integrate this into the start_reindex script if we prefer to keep the calm_transformer around. Will need to add it into the 2026-07-03 stack too
For the record I ran 1000 records locally and it took only a couple of minutes

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,624 @@
...
+    )
+    error_lines = [
+        line for line in logs_text.splitlines() if error_pattern.search(line) is not None
+    ]
+    if error_lines:
+        print("Potential transformer errors detected:")
+        for line in error_lines[:40]:
+            print(line)
+
+    if missing_ids or error_lines:
```
kenoir: The DLQ redrive policy and this log scan now encode different definitions of failure, and I think they can disagree.

`maxReceiveCount=2` means a first failure is explicitly not terminal: a message that hits a transient `StoreReadError` gets redelivered and can succeed on the second attempt. When that happens the record is indexed and nothing lands on the DLQ, so `missing_ids` is empty and the run genuinely succeeded. But the error line is still in the logs, so `error_lines` is non-empty and we return 1.

That's much better than where this started (every run exited 1), so not a blocker. But it's the same class of false alarm, and a verify step that cries wolf on a good run gets ignored fairly quickly.

Would gating the exit code on `missing_ids` alone work? The error lines are still useful, just as advisory output rather than a failure signal, since the DLQ and the index check between them already cover "did this actually work".

Separate but related: `Exception` is new in the pattern on line 512 and it's a bare substring match, which is what produced the original problem with `ERROR`. Have you run this against a full startup log? I'd expect JVM/Elastic4s startup chatter to be the risk.

pipeline/transformer/transformer_calm/run_local/local_transformer.py
```diff
@@ -0,0 +1,624 @@
...
+
+    project_dir = _project_dir()
+    enqueue_started_at = (
+        datetime.datetime.now(datetime.timezone.utc)
+        .replace(microsecond=0)
+        .isoformat()
+        .replace("+00:00", "Z")
+    )
+    enqueue_timestamp_file = _last_enqueue_timestamp_file(project_dir)
+    enqueue_timestamp_file.write_text(f"{enqueue_started_at}\n", encoding="utf-8")
```
kenoir, line 275: Nit: this lands at `pipeline/transformer/transformer_calm/.last_enqueue_timestamp` and isn't ignored, so it shows up as untracked after every `enqueue`.

`pipeline/transformer/.gitignore` is just `*.txt`, which covers `missing_ids.txt` but not this one:

```
$ git check-ignore -v pipeline/transformer/transformer_calm/missing_ids.txt
pipeline/transformer/.gitignore:1:*.txt
$ git check-ignore -v pipeline/transformer/transformer_calm/.last_enqueue_timestamp
(no match)
```

One line in that `.gitignore` sorts it.

kenoir (approved):
All nine of my earlier comments are addressed. I stood the compose stack up to check the ones that mattered: the works-source index now comes out with `dynamic: false`, the normalizer in place and 4 typed fields (was 139), the DLQ carries `maxReceiveCount=2` against `calm-transformer-dlq`, and `es-setup` is properly gated on `service_completed_successfully`.

Approving. The two comments I've just left are a question and a nit, neither blocking.

pipeline/transformer/transformer_calm/run_local/local_transformer.py
agnesgaroux: Removed EXCEPTION from the string match
Verify now passes on both queues empty and all ids indexed 
Error found are logged but do not fail the verify step

pipeline/transformer/transformer_calm/run_local/local_transformer.py
agnesgaroux: I'm writing it as .txt file so it's caught in the existing gitignire

PR #3501, opened 2026-07-27
Reinstate the matcher/merger ordering tests

### LINKED ISSUE #6078: Improve matcher/merger integration tests
Following https://github.com/wellcomecollection/catalogue-pipeline/pull/2926, and as a retro action following [20250702 retro](https://github.com/wellcomecollection/docs/blob/main/fortnotes/20250702.md).

We should improve our integration testing so that we can run tests that require multiple catalogue pipeline services to interact with one another, and external resources (e.g. elasticsearch, dynamodb).

**We now have some ignored tests that need to be reinstated**: https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/MergerIntegrationTest.scala#L821

We should look at reinstating these tests, and establish an improved pattern for integration testing requiring multiple services.

This is done when:

- [ ] The above tests are re-instated.

kenoir:
Closes https://github.com/wellcomecollection/platform/issues/6078

## What does this change?

Reinstates the three ignored METS/e-bib ordering scenarios in `MergerIntegrationTest`, by running the real matcher over a real DynamoDB graph table instead of a stub.

They were ignored when the matcher and merger moved to Lambda entrypoints. The tests were left driving `MatcherStub`, which returns pre-canned match groups and never writes the matcher graph, so no state carried from one work to the next. Those three scenarios are entirely about the order records arrive in, so with a stub matcher there was nothing left for them to test.

Order-dependent merge bugs are real in this area: the same file already carries a scenario for https://github.com/wellcomecollection/platform/issues/5375, where a record appeared twice, with a comment noting the bug only reproduces under certain orderings.

Both applications are now driven through their real Lambda entrypoints and share one work index, mirroring the way both read from the identified index in production. The transport between them stays faked: the matcher's output is handed straight to the merger rather than going through a queue. Works are processed one at a time, each taken all the way through both applications before the next starts, which is what makes the ordering scenarios meaningful.

One production change: `StoredWorksMatcher` takes a `Retriever[WorkStub]` rather than `ElasticWorkStubRetriever` specifically, so the tests can drive it from an in-memory index. The companion object still builds the Elasticsearch retriever for the deployed application.

The merger README now describes the pattern for the next pair of services that needs it.

### Checklist

- [x] Does this change the display model the catalogue API returns? No. This is matcher wiring and tests only, so the test documents under `catalogue_graph/document_generators/test_documents` are untouched.

## How to test

`./builds/run_sbt_tests.sh merger` starts the containers and runs the suite, as before: dynamodb-local was already declared in the merger's `docker-compose.yml`, so there is no new infrastructure. With the containers already up:

```
sbt "project merger" "testOnly weco.pipeline.merger.MergerIntegrationTest"
```

`MergerIntegrationTest` is 29 pass, 0 ignored (previously 26 pass, 3 ignored). Full matcher suite 72 pass, full merger suite 253 pass. Buildkite build #14994 passed on an earlier commit of this branch.

No ignored scenarios left in `MergerIntegrationTest`, and ordering bugs between METS works and Sierra e-bibs are caught before deployment rather than in production.

The production change is a widening of one constructor parameter's type, and the deployed application still passes the same Elasticsearch retriever.

The suite now depends on dynamodb-local, so it is slower and needs a container running, which the matcher's own tests already required.

<details><summary>How the new coverage was checked</summary>

- Ran the integration suite three times for stability, all green.
- Stopped the dynamodb-local container and re-ran: all 29 fail, so the graph store is genuinely exercised rather than incidentally satisfied.
- Suppressed the image notification in `MergerSQSLambda` and confirmed the new per-pass assertion fires. That file was restored and is not part of this diff.

</details>

### DIFF
--- a/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
+++ b/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
@@ -301,10 +301,19 @@ trait IntegrationTestHelpers
     notificationsBefore: Int
   )(implicit context: Context): Unit = {
     val newlyIndexed = indexedImageIds -- imagesBefore
-    val newlyNotified = context.imageDownstream.msgSender.messages
+    val newlyNotifiedBodies = context.imageDownstream.msgSender.messages
       .drop(notificationsBefore)
       .map(_.body)
-      .toSet
+    val newlyNotified = newlyNotifiedBodies.toSet
+
+    val duplicates = newlyNotifiedBodies
+      .groupBy(identity)
+      .collect { case (body, occurrences) if occurrences.size > 1 => body }
+
+    assert(
+      duplicates.isEmpty,
+      s"Images ${duplicates.mkString(", ")} were sent downstream more than once in a single pass"
+    )
 
     assert(
       newlyIndexed.subsetOf(newlyNotified),
--- a/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
+++ b/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
@@ -8,25 +8,49 @@ import weco.catalogue.internal_model.image.ImageData
 import weco.catalogue.internal_model.work.WorkState.{Identified, Merged}
 import weco.catalogue.internal_model.work.{Work, WorkState}
 import weco.catalogue.internal_model.work.generators.WorkGenerators
-import weco.lambda.Downstream
+import weco.lambda.{Downstream, SQSLambdaMessageFailure, SQSLambdaMessageResult}
 import weco.lambda.helpers.LambdaFixtures
 import weco.pipeline.matcher.MatcherSQSLambda
 import weco.pipeline.matcher.config.{MatcherConfig, MatcherConfigurable}
 import weco.pipeline.matcher.fixtures.MatcherFixtures
-import weco.pipeline.matcher.matcher.WorksMatcher
-import weco.pipeline.matcher.models.MatcherResult
+import weco.pipeline.matcher.matcher.StoredWorksMatcher
+import weco.pipeline.matcher.models.{MatcherResult, WorkStub}
 import weco.pipeline.merger.config.{MergerConfig, MergerConfigurable}
 import weco.pipeline.merger.{MergeProcessor, MergerSQSLambda}
+import weco.pipeline_storage.{Retriever, RetrieverMultiResult}
 import weco.pipeline_storage.memory.MemoryRetriever
 
 import java.time.Instant
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.collection.mutable
+import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
 import weco.fixtures.TestWith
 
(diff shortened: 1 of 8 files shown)


pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
```diff
@@ -186,52 +219,101 @@ trait IntegrationTestHelpers
...
+
+    assert(
+      newlyIndexed.subsetOf(newlyNotified),
+      s"Images ${(newlyIndexed -- newlyNotified).mkString(", ")} were saved but not sent downstream"
+    )
+
+    assert(
+      newlyNotified.subsetOf(indexedImageIds),
+      s"Images ${(newlyNotified -- indexedImageIds).mkString(", ")} were sent downstream but never saved"
+    )
```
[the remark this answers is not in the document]
kenoir: Good catch, applied in 71a765d403.

I checked first that duplicates cannot happen legitimately, so the assertion is not a flake risk: the matcher emits one MatcherResult per work stub and `processWork` sends a single work, so a pass carries one result, and the match groups inside it are disjoint sets of works. Verified by running the suite too. Of the 29 integration tests, 9 passes send an image notification and each sends exactly one.

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
```diff
@@ -186,52 +219,110 @@ trait IntegrationTestHelpers
...
+  private def indexedImageIds(implicit context: Context): Set[String] =
+    context.mergedIndex.values.collect { case Right(image) => image.id }.toSet
+
+  /** Only images result in a notification; works are picked up downstream by a
+    * window-based read of the merged index.
+    *
+    * Compares what this pass did rather than the whole scenario, so an image
+    * that was notified on an earlier pass can't cover for a later one.
+    */
+  private def assertImagesNotified(
```
agnesgaroux, line 306: assertImagesNotified guards inserts, not updates

newlyIndexed = indexedImageIds -- imagesBefore is keyed by image id, so a re-merged image (same id, new content) is excluded and never required to be re-notified. A regression that stopped notifying updates would still pass — and this is reachable today (Miro image → matching Sierra picture re-merges the component).

Not a prod bug: the merger already notifies on every merge success (successes.map(notifyDownstream)), updates included. Just a test gap.

Suggest: assert per-pass emitted images (snapshot merged-index values and require changed-or-new ⊆ notified, or surface MergerResponse.successes), and add a scenario that updates an image on a later pass.

agnesgaroux (approved):
Logic/tests look good but I haven't been able to run ./builds/run_sbt_tests.sh merger (Docker version 29.6.2) Might be incompatibility between my version of docker and sbt_wrapper, investigating

pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/fixtures/IntegrationTestHelpers.scala
kenoir: Good catch, and you're right it was inserts-only. The check now snapshots stored image values per pass and requires changed-or-new images to be notified, and there's a new scenario where a Miro image is re-merged by a later Sierra pass and must be notified both times. b38f8dc737

PR #3502, opened 2026-07-27
Refer to related concepts by an ID that has works

### LINKED ISSUE #6388: Related concepts link to concept IDs that 404
Related concepts on theme pages sometimes link to a concept ID that does not exist, so following the link returns a 404.

## Original report

With the `Show all fields on theme pages` toggle enabled, https://wellcomecollection.org/concepts/ynwzpctg contained a "Broader Than" link to [Campeche, Bay of (Mexico)](https://wellcomecollection.org/concepts/xhwbqawd), which returned a 404. Discovered while raising #6387.

That specific example no longer reproduces: `ynwzpctg` now returns empty `relatedConcepts`. `xhwbqawd` still 404s, and the underlying problem is still present across the index.

## Scale

Measured against the live catalogue API on 2026-07-27:

- 16 of 6,533 distinct related concept references resolve to a 404 (0.24%), sampled across 5,772 concepts
- a second sample of 3,218 references gave 5 broken (0.16%)
- affects `broaderThan`, `narrowerThan`, `relatedTo`, `people`, `frequentCollaborators` and `relatedTopics`, so it is not confined to the hierarchy fields
- separately, 13 of 1,172 sampled `sameAs` targets also 404

## Cause

This is not a filtering problem. The Neptune related concept queries already filter on `MATCH (related_concept)<-[:HAS_CONCEPT]-(work)`, so they only ever return concepts that have works.

The bad ID is introduced afterwards, in Python. `_get_related_concepts` merges synonymous results under a "primary" ID, and `_update_same_as_map` picks that primary as `sorted(same_as_ids)[0]`. Alphabetical order says nothing about whether a concept has works. `CatalogueConceptsGraphRemover` removes work-less concepts from the graph and `ingestor_deletions` then deletes them from the concepts index, so a reference to one resolves to a 404.

Worked example: works reference "Labor supply" as `wsg7zfsq` at the top level, and as `th3tx5an` only inside composite subjects such as "Labor supply - History - Sources". Nested component concepts have not been extracted since catalogue-pipeline commit `3a074a9c` (2025-07-10), so `th3tx5an` has no works. It still wins the alphabetical tie-break, so `wsg7zfsq` is the concept that gets indexed while related concepts point at `th3tx5an`.

## Fix

wellcomecollection/catalogue-pipeline#3502 emits an ID the graph actually returned when the primary has no works.

That stops new broken references being written. Documents already in the index keep their stale references until the concepts referring to them are next re-ingested, so clearing the existing ones needs a reindex, or incremental windows that happen to touch them.

kenoir:
## What does this change?

Closes wellcomecollection/platform#6388, which carries the problem statement, measurements and cause.

In short: the Neptune queries only return concepts that have works, but the Python merge step then reports the group under `sorted(same_as_ids)[0]`, and alphabetical order says nothing about whether a concept has works. Concepts with no works are removed from the graph and deleted from the concepts index, so the reference 404s.

The merge still happens under the primary. We now ask the graph which members of the `same as` group have works, using the `HAS_CONCEPT` join already in `CONCEPT_TYPE_QUERY`, and emit the primary whenever it has works itself, falling back to the alphabetically first member that does. No query text changes. Only groups whose primary has no works change, so index churn stays small and the emitted ID is stable across relationship types and across runs.

This stops new broken references being written. It does not repair the ones already in the index: those persist until the concepts referring to them are next re-ingested, which needs a reindex or incremental windows that touch them. Re-ingesting referring concepts when a node is removed is a separate change.

### Checklist

- [x] No change to the display model the catalogue API returns. Field shapes are untouched, only which concept ID a related concept points at, so the test documents under `catalogue_graph/document_generators/test_documents` do not need regenerating.

## How to test

`uv run --group dev pytest` from `catalogue_graph/` (1575 pass). New tests in `tests/ingestor/test_concepts_extractor.py` cover `_choose_target_id` and the merge path against canned Neptune responses. The regression test is non-vacuous: stubbing `_choose_target_id` back to the old behaviour reproduces the reported symptom, emitting `th3tx5an` instead of `wsg7zfsq`.

After deploy, open a theme page for a concept that has been re-ingested since and confirm its related concept links resolve.

Sample related concept references out of the concepts index and request each target from `/concepts/{id}`. The ~0.2% 404 rate should fall towards zero as concepts are re-ingested.

- Costs one extra `concept_type` query per relationship type that has related concepts, so up to 8 more per batch on top of the 40 that path already issues. Hoisting the lookup so all eight relationship types share one query would remove that, at the cost of restructuring how the queries are parallelised.
- Where the primary has no works, the related concept document now comes from a sibling, so its label and identifiers come from that sibling's source concepts. Synonymous, so labels should match, but wording can differ. If the sibling has no usable label at all, `_transform_related_concept` drops the related concept silently.
- 13 of 1,172 sampled `sameAs` targets also 404. Same family of cause, not addressed here.

### DIFF
--- a/catalogue_graph/src/ingestor/extractors/concepts/base_concepts_extractor.py
+++ b/catalogue_graph/src/ingestor/extractors/concepts/base_concepts_extractor.py
@@ -40,14 +40,7 @@ CONCEPTS_BATCH_SIZE = 40_000
 
 
 def _choose_target_id(primary_id: str, referenced_ids: set[str]) -> str:
-    """
-    Choose which ID to use when referring to a related concept.
-
-    `referenced_ids` are the IDs the graph returned for this related concept, all of which are connected to at least
-    one work (see `get_related_query`). `primary_id` is the primary of their 'same as' group, which may itself have
-    no works. Prefer the primary for consistency with the rest of the ingestor, but fall back to a referenced ID
-    when the primary is not among them.
-    """
+    """Refer to a related concept by an ID which has works. `referenced_ids` all do, the primary might not."""
     if primary_id in referenced_ids:
         return primary_id
 
@@ -167,9 +160,7 @@ class GraphBaseConceptsExtractor(GraphBaseExtractor, StreamingExtractor, ABC):
                     if related.get("relationship_type"):
                         entry["relationship_type"].add(related["relationship_type"])
 
-        # Results are merged under a primary ID, but the primary of a 'same as' group is chosen alphabetically and is
-        # not necessarily connected to any works. Concepts without works are removed from the graph (and therefore
-        # from the concepts index), so pointing at one produces a related concept which resolves to a 404.
+        # Concepts with no works are removed from the index, so referring to one would 404.
         for entries in merged_result.values():
             for primary_related_id, entry in entries.items():
                 entry["id"] = _choose_target_id(
--- a/catalogue_graph/src/ingestor/extractors/concepts/base_concepts_extractor.py
+++ b/catalogue_graph/src/ingestor/extractors/concepts/base_concepts_extractor.py
@@ -39,6 +39,21 @@ RelatedConcepts = dict[str, list[ExtractedRelatedConcept]]
 CONCEPTS_BATCH_SIZE = 40_000
 
 
+def _choose_target_id(primary_id: str, referenced_ids: set[str]) -> str:
+    """
+    Choose which ID to use when referring to a related concept.
+
+    `referenced_ids` are the IDs the graph returned for this related concept, all of which are connected to at least
+    one work (see `get_related_query`). `primary_id` is the primary of their 'same as' group, which may itself have
+    no works. Prefer the primary for consistency with the rest of the ingestor, but fall back to a referenced ID
+    when the primary is not among them.
+    """
+    if primary_id in referenced_ids:
+        return primary_id
+
+    return sorted(referenced_ids)[0]
+
+
 class GraphBaseConceptsExtractor(GraphBaseExtractor, StreamingExtractor, ABC):
     """Abstract base class for concept extraction from the catalogue graph.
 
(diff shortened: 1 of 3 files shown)


agnesgaroux (approved):

StepanBrychta:
Is it possible to write broken references by choosing the primary concept alphabetically? Concepts with no works are removed from the graph every 15 minutes, and this happens right after any deleted works (and corresponding HAS_CONCEPT edges) are removed. The ingestor runs right after the graph remover, so it should always see the graph in a 'clean' state (with no unconnected concepts). 

I think this does offer protection from writing broken references in cases where the ingestor runs at the same time as the bulk loader or the graph remover, in which case it might see the graph in an 'inconsistent' state, but this is rare.

I suspect the broken references we see today did not start out as broken references. They are likely the result of concepts being removed from the concepts index after the fact (after all works referencing them were removed).

kenoir:
You're right about the intent, and about the removal being global: `get_node_ids_to_remove` is an unfiltered `MATCH (n:Concept) WHERE NOT (n)-[:HAS_CONCEPT]-()`, so it isn't window-scoped, and `Graph removers` does run immediately before `Ingestors`.

The gap is that the removers aren't sequenced relative to each other. `catalogue_graph_removers_incremental_state_machine` is a single distributed `Map` with `MaxConcurrency = 10`, and "Catalogue Concept Nodes" and "Catalogue Work Nodes" are sibling items in it. When works are deleted, the concept remover has usually already run its scan, so the concepts those works orphaned survive that cycle's ingestor and are only removed in the next one.

That shows up in the 2025-10-02 remover logs:

- `2026-07-29T10:55:53Z`, a different remover in the same Map deleted 137 nodes
- `2026-07-29T11:10:44Z`, one cycle later, the concept remover deleted exactly 137 concepts

So the 10:55 ingestor ran against a graph holding 137 work-less `Concept` nodes. Across 288 concept-remover invocations in the 48 hours to 2026-07-30 09:00 BST, 5 were non-zero: 93, 1, 137, 5 and 3. It tracks work deletions rather than being confined to overlap with the bulk loader.

You're right that these don't start out as 404s. In that 10:55 run the doomed concepts still had index docs, because `get_consistent_batches` writes a doc for every member of the group. They only 404 in the next cycle, once the node goes and `ingestor_deletions` removes the doc. What this change stops is the ingestor choosing an ID that is already work-less at the point it writes it.

Your other case is real and this doesn't cover it: a concept that genuinely has works when it's referenced and loses them all later. Those references were valid when written, and only re-ingesting the referring concepts on removal will fix them. I'll raise that separately, since it's also the only thing that repairs the 404s already in the index.

Worth considering on its own: ordering the concept-node remover after the work removers rather than alongside them would close the window at source. I'll correct #6388, which currently reads as though the alphabetical tie-break is the whole story.

kenoir:
Follow-ups raised: wellcomecollection/platform#6466 for re-ingesting referring concepts on removal, and wellcomecollection/platform#6467 for ordering the concept node remover after the work removers.

PR #3503, opened 2026-07-28
Axiell to FOLIO sync: reconciliation deletes (soft-suppress + hard-delete)

anubhavijay:
## What does this change?

Implements **Pass 2 reconciliation deletes** for the Axiell to FOLIO sync step, completing the reconciliation pipeline specified in [RFC 090](https://github.com/wellcomecollection/docs/tree/main/rfcs/090-axiell-folio-sync).

The sync step now processes deletion facts from the Axiell adapter's reconcile phase in two independent modes (selected via `hard_delete` event field or `HARD_DELETE` env var):

1. **Soft-suppress (default, reversible)**
   - Sets `discoverySuppress=true` on Instance, Holdings, and Item
   - Sets `staffSuppress=true` on Instance only (holdings-storage and items don't support this field)
   - Idempotent: re-applying to already-suppressed records is a harmless PUT

2. **Hard-delete (opt-in, irreversible)**
   - Deletes entities child-first (Item → Holdings → Instance) to maintain referential integrity
   - Treats 404 responses as no-ops (safe under redelivery and races)
   - Aborts before parent if a child delete fails (preventing orphans)

### Changes include:

- **Event model**: added `hard_delete` field to `AxiellFolioSyncEvent` (defaults to `HARD_DELETE` env var, default `false`)
- **Response model**: added `total_deletions` to `AxiellFolioSyncResponse`
- **Report model**: added `SyncDeletionEntry` to track per-GUID actions (suppress or hard-delete) in S3 + CloudWatch reports
- **Deletion logic** (`upsert.py`): `suppress_by_guid()` and `delete_by_guid()` functions with entity-level result tracking; refactored for clarity and maintainability
- **Main loop** (`sync_to_folio.py`): two-pass architecture — Pass 1 upserts, Pass 2 processes deletion facts
- **Documentation** (`README.md`): extensive section on reconciliation semantics, idempotency, and the two delete modes
- **Tests** (`test_axiell_folio_sync.py`, `test_suppress.py`, `test_upsert.py`): comprehensive coverage for soft-suppress, hard-delete, and error scenarios

## How to test

1. **Unit tests**: `cd catalogue_graph && uv run pytest tests/adapters/steps/axiell_folio_sync/`
2. **Local dry run**: Test with a sample Axiell changeset (dry_run=true) to validate payload construction and plan actions without writing to FOLIO.
3. **Staging validation**: Deploy to staging Lambda/ECS and run against Staging FOLIO tenant to verify soft-suppress reversibility before enabling hard-delete.

- ✅ All unit tests pass (soft-suppress, hard-delete, upsert refactoring, and error cases)
- ✅ `PipelineReport` correctly records deletion facts (S3 JSON + CloudWatch metrics)
- ✅ Soft-suppress is idempotent: re-running deletion facts against already-suppressed records produces no changes
- ✅ Hard-delete child-first ordering is enforced (no orphaned holdings/items)
- ✅ Dry run produces correct plan without writing to FOLIO
- ✅ Redelivered deletion facts are safe (404s treated as no-ops, idempotent writes)
- ✅ Manual staging test: suppress a record, verify `discoverySuppress` and `staffSuppress` flags, then un-suppress via hard-delete (to prove reversibility pre-release)

- **Hard-delete is irreversible.** Must be opt-in via `hard_delete=true` event field or `HARD_DELETE=true` env var (defaults to false). Default behavior is soft-suppress (reversible).
- **Reconciliation logic correctness.** Deletion facts must accurately identify superseded GUIDs. Incorrect reconciliation in the Axiell adapter could suppress/delete the wrong record. Mitigated by staged deployment and manual staging validation.
- **FOLIO referential integrity.** Deletion must happen child-first (Item → Holdings → Instance) to avoid violating foreign keys. Cascade ordering is enforced in code; a child-delete failure aborts before the parent.
- **Lambda cold starts.** Both deletion modes read from FOLIO (for existing record lookups); ensures credentials are valid and FOLIO is reachable even in dry runs. No degradation vs. current behavior.
- **CloudWatch metrics namespace collision.** Metrics are already published to `catalogue_adapters`; deletion metrics follow the same pattern (no new namespace).

No breaking changes to the event schema (all new fields are optional with sensible defaults).

### DIFF
--- a/catalogue_graph/src/adapters/steps/axiell_folio_sync/README.md
+++ b/catalogue_graph/src/adapters/steps/axiell_folio_sync/README.md
@@ -125,7 +125,7 @@ AWS_PROFILE=platform-developer uv run python -m adapters.steps.axiell_folio_sync
 
 # Specific changesets; pass --live to disable dry run and write to FOLIO
 AWS_PROFILE=platform-developer uv run python -m adapters.steps.axiell_folio_sync.axiell_folio_sync \
-  --use-cli --job-id my-job-123 --changeset-ids 456429b2-6f0e-11f1-afea-525a8567ce81 --use-rest-api-table
+  --use-cli --job-id my-job-123 --changeset-ids <changeset-id> --use-rest-api-table
 ```
 
 Set the `OKAPI_*` environment variables first (see Configuration above).
--- a/catalogue_graph/src/adapters/steps/axiell_folio_sync/README.md
+++ b/catalogue_graph/src/adapters/steps/axiell_folio_sync/README.md
@@ -18,9 +18,8 @@ The design, record-selection rules and tombstone semantics are specified in
    (entrypoint `adapters.steps.axiell_folio_sync.axiell_folio_sync.lambda_handler`).
 3. The step reads the changed rows for the event's `changeset_ids` through `AxiellChangesetReader` (using the same
    `AXIELL_CONFIG` as the Axiell adapter, so S3 Tables in Lambda and a local sqlite catalog for local runs). The
-   reader also exposes the adapter's deletion facts via `iter_deletions()`, which is where authoritative FOLIO
-   deletes (platform#6440) will plug in.
-4. Each record is selected, mapped, and upserted:
+   reader also exposes the adapter's deletion facts via `iter_deletions()`, consumed by the reconciliation pass below.
+4. **Pass 1 — upsert.** Each record is selected, mapped, and upserted:
    * **Selection** (`is_selected_for_sync`): a record is synced only if it carries the harvest flag (MARC `980 $a`
      present) and is item-level (MARC `351 $c` == `ITEM`). Everything else is skipped.
    * **Mapping** (`mapping.py`): MARCXML → typed Pydantic payloads for Instance, Holdings and Item. FOLIO reference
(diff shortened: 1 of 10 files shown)


catalogue_graph/src/adapters/steps/axiell_folio_sync/README.md
```diff
@@ -105,7 +125,7 @@ AWS_PROFILE=platform-developer uv run python -m adapters.steps.axiell_folio_sync
 
 # Specific changesets; pass --live to disable dry run and write to FOLIO
 AWS_PROFILE=platform-developer uv run python -m adapters.steps.axiell_folio_sync.axiell_folio_sync \
-  --use-cli --job-id my-job-123 --changeset-ids <changeset-id> --use-rest-api-table
+  --use-cli --job-id my-job-123 --changeset-ids 456429b2-6f0e-11f1-afea-525a8567ce81 --use-rest-api-table
```
kenoir: Looks like a real changeset id got pasted over the `<changeset-id>` placeholder. The placeholder reads better as documentation, and a concrete id will tempt copy-paste runs against a changeset that no longer exists.

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping.py
```diff
@@ -166,12 +166,12 @@ def select_and_build(
...
     "audio-visual material - e-sound only": "sound recording",
     "audio-visual material - e-visual only": "video recording",
-    "published material": "Books",
+    "published material": "book",
     "archives": "unspecified",
 }
 
 # Fallbacks used when the MARC record carries no value for a resolved field.
-DEFAULT_MATERIAL_TYPE = "Books"
+DEFAULT_MATERIAL_TYPE = "book"
```
kenoir, line 174: This and the "published material" entry look like a genuine fix: FOLIO's shipped reference data names the type `book` (lowercase, singular), and the RefCache lookup lowercases but doesn't de-pluralise, so `"Books"` could never resolve and would raise `Unresolved material type` for any record taking the default. Was this hit while testing against the tenant? Since it's an upsert-path fix unrelated to the reconciliation deletes, a line in the PR description would help it not get lost, and it changes which records can sync at all, so worth calling out.

catalogue_graph/src/adapters/steps/axiell_folio_sync/sync_to_folio.py
```diff
@@ -102,12 +107,96 @@ def _build_error(
...
+# (not "suppressd"), so this cannot be derived by appending "d".
+_ACTION_COUNT_KEY = {
+    "create": "created",
+    "update": "updated",
+    "suppress": "suppressed",
+    "delete": "deleted",
+}
+
+
+def _tally_entity_actions(result: Any, counts: dict[str, int]) -> None:
```
kenoir: Could this be `result: UpsertResult | GuidCascadeResult` instead of `Any`? Both models have the same three `EntityResult` fields, so the union type-checks as-is, and it keeps mypy able to catch a wrong object being passed at the call sites.

catalogue_graph/src/adapters/steps/axiell_folio_sync/upsert.py
```diff
@@ -68,6 +86,71 @@ def _find_by_hrid(
...
+    rejects an unknown staffSuppress with a 422, and items silently drop it, so
+    sending it there is at best a no-op and at worst breaks the cascade.
+
+    Not-found → ``skip`` (the record is already gone). Idempotent: re-suppressing
+    a record whose flags are already true is a harmless PUT, so redelivered
+    deletion facts do not misbehave.
+    """
+    existing = _find_by_hrid(folio, search_path, hrid, list_key)
+    if not existing:
+        return EntityResult(action="skip")
```
kenoir, line 118: `_find_by_hrid` catches every exception and returns None, and in this path None means "the record is already gone". So a transient FOLIO failure during the lookup is reported as a successfully actioned deletion. I ran the cascade against a client whose GET always raises, simulating an outage:

```
cascade result:
  item: skip
  holdings: skip
  instance: skip
  errors: []

run_sync response:
  total_deletions: 1
  total_successful: 1
  total_errors: 0
```

The GUID lands in the manifest as skip/skip/skip with no error entry, so nothing prompts a rerun and the record stays live and discoverable. That feels riskier here than in the upsert path, where the next changeset touching the record gives it another chance; a deletion fact is only delivered with its changeset's run. In hard-delete mode there is a second effect: a swallowed item-lookup failure lets the cascade carry on to the holdings and instance deletes while the item may still exist.

Would it make sense for the cascade path to let lookup errors propagate, keeping the None return for a genuine empty result only? The try/except in `_cascade_by_guid` already aborts child-first and records an error entry, so a raising lookup would slot straight into the existing handling.

kenoir (commented):
This adds the reconciliation pass cleanly. The cascade design is the strong part: one child-first order table, a single try/except that both aborts before the parent and records the error, and careful accounting so an irreversibly deleted child doesn't vanish behind a GUID-level error. The tests match that care, down to asserting forbidden verbs on each path. I ran the suite locally (40 passed) and checked the FOLIO claims in the docstrings against the mod-inventory-storage schemas and the inventory client; they hold up.

I'm holding off approval until the `_find_by_hrid` exception swallowing is addressed: a swallowed lookup error reports a deletion fact as cleanly actioned with nothing prompting a rerun, and in hard-delete mode it can let the cascade carry on past a child that still exists, so deletes fire that should have been aborted (repro on the inline comment). The rest is a question about the bundled material-type fix and two small nits.

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping.py
anubhavijay: yes this has changed after the recent migration and therefore changed so that the item upsert does not fail with unresolved material type error

kenoir (approved):

PR #3508, opened 2026-07-29
Integrate unmigrated calm reindex

agnesgaroux:
## What does this change?

Following https://github.com/wellcomecollection/platform/issues/6448
We want the partial/list-based CALM reindex to be integrated in the reindex script
-> if doing a complete reindex on "all" sources or "calm", the script fails fast if the calm ids file is not present. 
-> the calm DDB table name has been removed to prevent it ever being scanned, ensuring we only use ids from the input file
This means deploying the calm_transformer infra to pipeline-2026-07-03 as well 
Includes ids list, as provided by VS

### Checklist

- [x] Does this change the display model the catalogue API returns? If so, regenerate the test documents under `catalogue_graph/document_generators/test_documents`. The `wellcomecollection/catalogue-api` repo copies them in with `copy_test_documents.py`, and its OpenAPI contract tests validate the published spec against them, so stale fixtures let the spec drift. Tooling only 

## How to test

`./run_terraform.sh plan` should give you
```
Plan: 26 to add, 12 to change, 0 to destroy.
```
26 changes all pertain to calm_transformer
12 changes are the usual transformer scaling_service churn. 

Unmigrated CALM archives can be reindexed along with other sources as part of the reindex script 

Calm ids input list is mandatory and the script fails fast if it's not present
The script cannot read from the CALM DDB for lack of a table name so will fail loudly if the logic is amended in a way that allows for a full calm reindex to be attempted


reindexer/scripts/start_reindex.py
```diff
@@ -239,26 +263,31 @@ def start_reindex(ctx, src, dst, mode, input_file):
     print(f"Starting a standard reindex {src} ~> {dst}")
 
     if mode == "complete":
-        total_segments = how_many_segments(table_name=SOURCES[src])
-        parameters = complete_reindex_parameters(total_segments)
+        if src == "calm":
+            print(f"Using specific IDs from {calm_input_file} for calm")
+            parameters = file_reader_reindex_parameters(calm_input_file)
+        else:
```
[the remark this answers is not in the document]
agnesgaroux: - file_reader_reindex_parameters now builds a list eagerly and filters blank lines with if line.rstrip() — a trailing newline or blank line in the file produces no empty ID.
- The calm branch collects the generator into a list and fails fast with a clear error if no valid IDs remain after filtering, rather than silently publishing nothing.

reindexer/scripts/start_reindex.py
```diff
@@ -239,26 +263,33 @@ def start_reindex(ctx, src, dst, mode, input_file):
...
+            specified_records_str = click.prompt(
+                "Which records do you want to reindex? (separate multiple IDs with spaces)",
+                type=str,
+            )
+            specified_records = specified_records_str.split()
+            if not specified_records:
+                return sys.exit("You need to specify at least 1 record ID")
+
+            verify_specific_ids(source=src, specific_ids=specified_records)
+            parameters = specific_reindex_parameters(specified_records)
```
kenoir, line 292: Worth noting this restructure fixes a latent bug rather than just moving code around. `--mode` is `required=True`, so it was always one of complete/partial/specific, which meant the old `elif input_file:` arm sat after three exhaustive branches and could never be reached. Passing `--input-file` before this PR silently ignored the file and prompted for IDs instead. Nesting it under `specific` makes the flag actually work.

reindexer/scripts/start_reindex.py
```diff
@@ -239,26 +263,33 @@ def start_reindex(ctx, src, dst, mode, input_file):
...
-        verify_specific_ids(source=src, specific_ids=specified_records)
-        parameters = specific_reindex_parameters(specified_records)
-    elif input_file:
-        parameters = file_reader_reindex_parameters(input_file)
-        if not parameters:
-            return sys.exit("Specified input file does not exist")
+        if input_file:
+            parameters = file_reader_reindex_parameters(input_file)
+            if not parameters:
+                return sys.exit("Specified input file does not exist")
```
kenoir: Now that this arm is reachable, the guard underneath it is not doing anything. `file_reader_reindex_parameters` is a generator function, so calling it returns a generator object, which is always truthy. The `return False` on `FileNotFoundError` does not run until the generator is first iterated, and even then it just ends iteration. A mistyped `--input-file` path gets past this check, publishes zero messages, and still scales the reindexer service up to 3 tasks.

The calm branch above sidesteps this by materialising with `list(...)`, so the same treatment would work here:

```py
parameters = list(file_reader_reindex_parameters(input_file))
if not parameters:
    return sys.exit(f"Input file {input_file!r} does not exist or contains no IDs")
```

Checked locally against a missing path with the same generator body:

```
truthy? True
not p -> False
list -> []
```

reindexer/scripts/start_reindex.py
```diff
@@ -207,8 +208,25 @@ def verify_specific_ids(*, source, specific_ids):
...
     help="A path to a file containing IDs to use (implies a specific-records reindex)",
 )
+@click.option(
+    "--calm-input-file",
+    help=(
+        "Path to a file of specific record IDs to reindex for calm. Required when "
+        "running a complete reindex that includes calm -- calm never runs a full "
+        "complete reindex."
+    ),
+)
```
kenoir, line 218: `reindexer/scripts/README.md` has not kept up. Its example is:

```
- Specific records from a file (one ID per line):
  - uv run start_reindex.py --src calm --dst catalogue --input-file ./ids.txt
```

which omits the now-required `--mode`, and does not mention `--calm-input-file` or the `third_party_archives.txt` list this PR adds. Could the README pick up the canonical invocation, something like `--src all --dst catalogue --mode complete --calm-input-file ./third_party_archives.txt`? A note on where `third_party_archives.txt` came from and when would help too; a bare list of 39,375 UUIDs is hard to date or re-derive six months from now.

reindexer/scripts/start_reindex.py
```diff
@@ -239,26 +263,33 @@ def start_reindex(ctx, src, dst, mode, input_file):
     print(f"Starting a standard reindex {src} ~> {dst}")
 
     if mode == "complete":
-        total_segments = how_many_segments(table_name=SOURCES[src])
-        parameters = complete_reindex_parameters(total_segments)
+        if src == "calm":
+            print(f"Using specific IDs from {calm_input_file} for calm")
+            parameters = list(file_reader_reindex_parameters(calm_input_file))
+            if not parameters:
+                sys.exit(f"--calm-input-file {calm_input_file!r} contains no valid IDs")
```
kenoir, line 270: On `reindexer/scripts/third_party_archives.txt` itself (commenting here because GitHub will not render a diff that large to anchor to):

Would S3 be a better home for it than the repo? It is a 1.4MB one-off data drop rather than code, git will carry it forever, and any correction to the list becomes a PR. The script could take an `s3://` URI as well as a local path, or the operator could pull the file down before running.

Not blocking, and I can see the argument for having the exact list under review alongside the change that uses it.

Contents check, for what it is worth: 39,375 lines, all unique, all 36-character UUIDs, no blank lines, no CRs. The missing trailing newline does not drop the last ID.

kenoir (approved):

reindexer/scripts/start_reindex.py
agnesgaroux: Yes, indeed

PR #3511, opened 2026-07-29
Use /authn/login-with-expiry OKAPI endpoint

anubhavijay:
### What does this change?

Switches the FOLIO OKAPI authentication from the deprecated `POST /authn/login` endpoint to `POST /authn/login-with-expiry`.

The old endpoint returned the token in an `x-okapi-token` response header. The new endpoint (used by Eureka/Keycloak in both production and the dev sandbox) delivers it exclusively via a `folioAccessToken` cookie. The token is still replayed as the `x-okapi-token` request header on every subsequent call, so downstream behaviour is unchanged.

Changes:
- `okapi_auth.py`: updated login URL and `_store_token` to read exclusively from the `folioAccessToken` cookie (header fallback removed after code review, as the new endpoint never returns a header token)
- `client.py`, `config.py`, `enrichment/runtime.py`: updated doc comments to reference the new endpoint
- `test_okapi_auth.py`, `test_client.py`: updated tests to exercise the cookie-based token flow
- `oai_pmh/folio/README.md`: updated documentation to reference the correct endpoint
- Added a `run_axiell_folio_sync_local.sh` helper script for local development runs

### How to test

Unit tests cover the updated token extraction logic:

```bash
cd catalogue_graph
uv run pytest tests/clients/folio_client/
```

End-to-end: run the Axiell→FOLIO sync or the FOLIO enrichment step against a FOLIO sandbox and confirm authentication succeeds without a 401 loop.

### How can we measure success?

No `OkapiLoginError` in CloudWatch logs after deployment; the Axiell→FOLIO sync and FOLIO enrichment Lambda runs complete without authentication failures.

### Have we considered potential risks?

- The `x-okapi-token` header fallback was deliberately removed after code review: `login-with-expiry` on both prod and sandbox always returns a cookie, never a header token, so the fallback was dead code that could mask misconfiguration.
- If the `folioAccessToken` cookie is absent, `OkapiLoginError` is raised with a clear message, so failures remain visible.

### DIFF
--- a/catalogue_graph/scripts/run_axiell_folio_sync_local.sh
+++ b/catalogue_graph/scripts/run_axiell_folio_sync_local.sh
@@ -15,7 +15,7 @@
 # CHANGESET_IDS wins if both are set. Any extra args are forwarded to the module:
 #   CHANGESET_IDS=<changeset_id> ./run_axiell_folio_sync_local.sh              # Dry run (default)
 #   CHANGESET_IDS=<changeset_id> ./run_axiell_folio_sync_local.sh --live       # Write to prod FOLIO
-#   CHANGESET_IDS="<changeset_id> def456" ./run_axiell_folio_sync_local.sh     # Multiple changesets
+#   CHANGESET_IDS="<changeset_id> <changeset_id2>" ./run_axiell_folio_sync_local.sh     # Multiple changesets
 #   SAMPLE_LIMIT=10 ./run_axiell_folio_sync_local.sh --live            # Sample 10 active records
 #   FOLIO_TARGET=dev CHANGESET_IDS=<changeset_id> ./run_axiell_folio_sync_local.sh  # Target dev FOLIO
 #   AWS_PROFILE=my-profile CHANGESET_IDS=<changeset_id> ./run_axiell_folio_sync_local.sh --live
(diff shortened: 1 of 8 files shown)


kenoir:
The description still describes a header fallback in `_store_token` ("with a fallback to the x-okapi-token header for compatibility") and calls that out as the source of backwards compatibility. That fallback was removed in the third commit, and `test_login_returning_x_okapi_token_header_only_raises` now checks that a header-only response raises `OkapiLoginError`. Worth updating the description so the risk section reflects what's actually shipping (no fallback, so any gateway that only returns the header would fail login outright).

kenoir:
`catalogue_graph/src/adapters/extractors/oai_pmh/folio/README.md:78` still says `POST /authn/login`. `enrichment/runtime.py:84`, a few lines above the code this doc describes, was updated in this PR to `/authn/login-with-expiry` — looks like this doc line got missed.

kenoir:
[nit] A few small things in this diff:
- `run_axiell_folio_sync_local.sh:31`: if the SSM secret is missing `url` (or `tenant`/`username`/`password`), `jq -r .url` prints the string `null` rather than empty, so `load_okapi_config`'s "missing field" check won't catch it — you'd get a confusing connection error against host `null`. Might be worth `jq -re '.url // empty'` or similar.
- `run_axiell_folio_sync_local.sh:18`: `CHANGESET_IDS="<changeset_id> def456"` mixes the placeholder with a stray literal value, unlike the other example lines.
- `test_okapi_auth.py:85`: the handler's non-login branch is unreachable — `_store_token` raises before the client sends a second request. `test_login_without_token_cookie_raises` right below shows the simpler pattern without it.

kenoir (approved):
Reviewed the auth change end to end: dropping the `x-okapi-token` header fallback in favour of the `folioAccessToken` cookie is correct per FOLIO's `login-with-expiry` docs and the RTR cookie pattern, and it's backed by a test that pins the new behaviour. Left a couple of comments: the PR description still describes the old fallback and should be updated, and there's a stale `/authn/login` reference in the FOLIO OAI-PMH README that this PR's docstring fix missed. Also a small nit batch on the new local sync script and a dead branch in one test. None of these block merge.

PR #3513, opened 2026-07-30
Quiesce the 2026-07-03 pipeline for the pre-reindex clear

kenoir:
## What does this change?

Phase 1 of wellcomecollection/platform#6461, quiescing the `2026-07-03` pipeline before its stateful parts are cleared for the full reindex.

The stack has been live-processing since 23 July. Its works-graph table holds 439,016 matcher entries and its indices hold 387,066 works from before the Axiell adapter store rebuild, including records the OAI feed no longer serves. All of that gets wiped, and nothing may be writing into it while that happens.

So the four schedule and trigger flags go off, along with `listen_to_reindexer`:

- `enable_adapter_transformer_trigger`, the EventBridge rule starting the transformer on `adapter.completed`
- `enable_id_minter_schedule`, the id-minter sweep and its id-pool top-up
- `enable_graph_pipeline_schedule`, the scheduled graph state machines
- `enable_image_inferrer_schedule`, the scheduled inferrer
- `reindexing_state.listen_to_reindexer`, which removes the reindex EventBridge rules and the Scala transformer queues' subscriptions to the reindex topics

`scale_up_tasks` and `scale_up_matcher_db` stay on. The reindex needs them, and the scale-down apply can only run once a day.

Missed harvest windows during the quiet period are superseded by the reindex.

This touches only the `2026-07-03` root, which has its own terraform state, so production cannot be affected by the apply.

### Checklist

- [x] Does this change the display model the catalogue API returns? No, it only toggles schedules and event rules.

## How to test

`terraform plan` in `pipeline/terraform/2026-07-03`. It should show EventBridge rules and schedules moving to disabled and the five reindex topic subscriptions being destroyed, with no resource replacements and nothing outside this pipeline date.

After the apply, no new executions should start on `transformer-2026-07-03`, `pipeline-2026-07-03_id_minter`, the `graph-*-2026-07-03` state machines, or the image inferrer.

The `2026-07-03` queues drain to zero and stay there, which is the precondition for the queue purge and the index wipe that follow.

The pipeline stops ingesting for the duration of the clear, and the staff preview (`?elasticCluster=axiell-collections-testing`) goes empty once phase 2 wipes the indices. It currently serves 88,592 works, so that is a visible change and stakeholders need telling.

This is reverted by the same one-line-per-flag change at the end of #6461, immediately before the reindex starts.

### DIFF
--- a/pipeline/terraform/2026-07-03/main.tf
+++ b/pipeline/terraform/2026-07-03/main.tf
@@ -1,8 +1,11 @@
 module "pipeline" {
   source = "../modules/pipeline_new"
 
+  # Quiesced while the pipeline's stateful parts are cleared, so nothing writes
+  # into the state being wiped. Back on for the reindex, along with the
+  # enable_* flags below.
   reindexing_state = {
-    listen_to_reindexer = true
+    listen_to_reindexer = false
     scale_up_tasks      = true
     scale_up_matcher_db = true
   }
@@ -21,10 +24,10 @@ module "pipeline" {
   # Base AMI for ECS instances
   ami_id = "resolve:ssm:arn:aws:ssm:eu-west-1:760097843905:parameter/imagebuilder/weco-al2023-ecs-optimised-x86_64/latest"
 
-  enable_adapter_transformer_trigger = true
-  enable_id_minter_schedule          = true
-  enable_graph_pipeline_schedule     = true
-  enable_image_inferrer_schedule     = true
+  enable_adapter_transformer_trigger = false
+  enable_id_minter_schedule          = false
+  enable_graph_pipeline_schedule     = false
+  enable_image_inferrer_schedule     = false
 
   pipeline_date = local.pipeline_date // namespaces services
   graph_date    = "2026-07-03"        // namespaces graph database


agnesgaroux (approved):

kenoir:
## Plan verified 2026-07-30

`0 to add, 19 to change, 5 to destroy`. No replacements, and every address is under `module.pipeline`, so nothing outside this pipeline date is touched. No other unapplied work is riding along.

**Nine schedules and rules flip `ENABLED` to `DISABLED`**, and nothing else about them changes:

| flag | resources disabled |
|---|---|
| `enable_id_minter_schedule` | `id_minter_schedule`, `minter_id_generator_schedule` |
| `enable_image_inferrer_schedule` | `image_inferrer_schedule` |
| `enable_graph_pipeline_schedule` | `catalogue_graph_pipeline_incremental`, `graph_pipeline_monthly` |
| `enable_adapter_transformer_trigger` | `adapter_transformer_trigger["axiell"]`, `["ebsco"]` |
| `listen_to_reindexer` | `reindex_transformer_trigger["axiell"]`, `["ebsco"]` |

**Five reindex subscriptions are destroyed**, all on `2026-07-03` queues: `calm_reindexer_topic`, `mets_reindexer_topic`, `catalogue_miro_reindex_topic`, `catalogue_sierra_reindex_topic`, `tei_reindexer_topic`. The matching five `aws_sqs_queue` policies drop the reindex topic from their `aws:SourceArn` condition, keeping the adapter output topic.

The five `aws_iam_role_policy.read_from_q` updates show an unknown `after` value, which is plan noise rather than a real change: `data.aws_iam_policy_document.read_from_queue` reads `aws_sqs_queue.q.arn`, and because the queue is being updated in place the data source defers. The queue keeps its name, so the ARN and the resulting policy are identical after the apply.

Addresses for phase 2, from `terraform state list`:

- `module.pipeline.module.matcher.aws_dynamodb_table.graph_table`
- `module.elastic.module.indices["<index>-2026-07-03"].elasticstack_elasticsearch_index.the_index` for each of the eight indices

The `settings_raw` drift that used to appear on every plan is gone, fixed by #3499.

PR #3521, opened 2026-07-30
Add new fields to work documents to support archive browsing

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6469

Add the following fields to final work documents to support archive browsing functionality in the frontend:
* `collection.isRoot` (display and filter sections): Stores `true` if a given work document is the root of a collection (defined by having at least one child work and no parent works). Required for the [Collections type listing page](https://www.figma.com/design/JUoyimGZGSELwBhMn2gQwj/Archive-browse?node-id=10877-7489&m=dev), since it shows a list of archive roots in a paginated view.
* `archive.category` (display and filter with a corresponding aggregation): If a work is part of an archive which has a recognised archive type (e.g. 'Personal Papers' for archives whose collection paths start with `PP/`), stores the ID/title of the archive type. Required for the [Collections type listing page](https://www.figma.com/design/JUoyimGZGSELwBhMn2gQwj/Archive-browse?node-id=10877-7489&m=dev), since it filters archive roots belonging to a specific archive type.
* `collection.root` (filter with a corresponding aggregation): Stores the ID/title of the collection root for works which belong to a collection. Unlike the `partOf` field, this field only stores the collection root (rather than all ancestors) and is also populated on the root document itself. Required to show full archive tree hierarchy on the new [Archive collections](https://www.figma.com/design/JUoyimGZGSELwBhMn2gQwj/Archive-browse?node-id=10877-8971&m=dev) page, which shows the archive tree in a fully open state (retrieved using pagination, with 50 works per page). 
* `collectionPath.sort` (query section only): Stores a zero-padded version of the collection path so that it can be sorted in a 'natural' way (e.g. `PPEBC/A/1/7` must come before `PPEBC/A/1/10`). Required to show fully open archive tree hierarchies on the new [Archive collections](https://www.figma.com/design/JUoyimGZGSELwBhMn2gQwj/Archive-browse?node-id=10877-8971&m=dev) page.

Two aspects of the designs are out of scope and will be implemented in separate PRs:
* https://github.com/wellcomecollection/platform/issues/6488
* https://github.com/wellcomecollection/platform/issues/6489

### Checklist

## How to test

I indexed the full dataset to a new `2026-07-30` works index in the production cluster. This index can be used locally with [this](https://github.com/wellcomecollection/catalogue-api/pull/942) catalogue API branch and [this](https://github.com/wellcomecollection/wellcomecollection.org/compare/sb/archive-filters-demo?expand=1) proof of concept frontend branch to test the change end to end.

The new fields should allow the works API to support the archive browse functionality (as specified in the Figma designs linked above) in a performant way.

Zero-padding numeric parts of collection paths to make them sortable in Elasticsearch is a valid approach, but there might be better approaches. Elasticsearch comes with a built-in ICU analysis plugin, which provides similar functionality via the `icu_collation_keyword` type. However, this plugin is disabled by default. To use it, we would have to ensure that the plugin is enabled in all future Elasticsearch clusters (both production and local). Whether this is preferable over zero-padding the collection path in the pipeline is an open question.

> [!WARNING]
> Do not merge until we are ready to switch to the new works index.


catalogue_graph/tests/models/pipeline/test_archive_type.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+
+def test_get_archive_type_strips_numeric_suffix_for_known_prefixes() -> None:
+    assert get_archive_type("OH1/B/3") == ArchiveType(id="OH", label="Oral History")
+
+
+def test_get_archive_type_keeps_numeric_suffix_for_other_prefixes() -> None:
+    # GC is not in _NUMERIC_SUFFIX_PREFIXES, so "GC176" is not stripped down to "GC"
+    assert get_archive_type("GC176/C/1") is None
```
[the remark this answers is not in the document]
StepanBrychta: The test works as intended. A work should only belong under the `GC` archive type when its collection path label begins with `GC/`, not `GC.../`. The test exists to make sure that the logic mapping collection path label prefixes to archive types is not too broad.

catalogue_graph/src/ingestor/transformers/work_query_transformer.py
```diff
@@ -124,13 +126,46 @@ def collection_path(self) -> str | None:
...
+        if self.hierarchy.is_collection_root:
+            return self.state.canonical_id
+        return None
+
+    @property
+    def collection_root_title(self) -> str | None:
+        if self.hierarchy.ancestors:
+            return self.hierarchy.ancestors[-1].work.properties.label
+        if self.hierarchy.is_collection_root:
+            return self.data.title
```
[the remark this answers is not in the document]
StepanBrychta: I don't see how this would make a difference. If `self.data.title` is `None`, the method will return `None` either way.

catalogue_graph/src/models/pipeline/archive_category.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
+
+import re
+
+from models.pipeline.collection_path import CollectionPath
+from models.pipeline.id_label import IdLabel
+
+# Maps collection path label prefixes (top-level reference number segments) to human-readable labels.
+COLLECTION_PATH_PREFIX_LABELS: dict[str, str] = {
```
kenoir, line 9: A fair amount of archive material comes out with no type. Counting against `works-indexed-2026-07-30`, 74,324 of the 269,281 works that have a collection path get no `archiveType`. Most of those are Sierra-style shelfmark hierarchies (`alpha_2003`, `b.20.n`) which I'd expect to have no archive type, but some look like archive families: RAMC (3,584 works), UGC (3,890), `MS.*` (13,814), and single-archive prefixes like HALDANE (6,548), JDW (6,521) and SB (5,213). Each has at least one work flagged `isCollectionRoot` (RAMC 1, UGC 3, `MS.*` 273), so they would appear on the collections listing page with no type to filter them by.

Separately, if UGC is one you'd want to add later it needs more than a table entry: its labels are `UGC 198/8/5/8`, so `label.split("/")[0]` returns `UGC 198`, and `([A-Z]+)(\d+)` doesn't match across the space.

catalogue_graph/src/ingestor/models/neptune/query_result.py
```diff
@@ -35,6 +35,15 @@ def ancestors_include_title(self, title: str) -> bool:
...
         return _standardise_work_title(title) in self._ancestor_titles
 
+    @property
+    def is_collection_root(self) -> bool:
+        """Returns true if the work is at the top of a collection hierarchy.
+
+        Collection hierarchies come from collection paths, so this includes hierarchies
+        which are not archives (such as those derived from Sierra 773/774 fields).
+        """
+        return len(self.ancestors) == 0 and len(self.children) > 0
```
kenoir: `is_collection_root` reads "no ancestors" as "top of a hierarchy", and ancestors come from the graph rather than from the path itself, so a work whose ancestor chain is broken counts as a root. 

In `works-indexed-2026-07-30` that is 17 of the 1,711 flagged roots. Fourteen are `PSY/xxx`, where there is no `PSY` record at all, so those are arguably the top record that exists. Two are under `WF/M/GB` (`WF/M/GB/09/11`, "Lancet Book 11"), where `WF/M/GB/09` has no record even though `WF/M/GB`, `WF/M` and `WF` all do, and one is `SA/FPA/A16/24/62-72`. All three I looked at have an empty `partOf` already, so this is the existing hierarchy gap showing through rather than something the PR introduces, and 17 works is not many. 

Is showing those on the collections listing page acceptable, or is it worth tightening the check?

kenoir (commented):
Two naming thoughts, both about how this reads from the API rather than about the implementation.

Can we avoid `ArchiveType`? Everything else in this area is "collection": `collectionPath`, `collectionRoot`, the `Collection`/`Section`/`Series` work types. `ArchiveType` would be the only member of its family, and I'm not sure the distinction earns a second vocabulary.

Relatedly, could `isCollectionRoot` stop being a root-level field and become part of a single `collection` block? Something like:

```json
// display
"collection": {
  "category": { "id": "PP", "label": "Personal papers", "type": "CollectionCategory" },
  "root": {
    "id": "uy0hyncn",
    "title": "Papers of Ernst Boris Chain",
    "referenceNumber": "PP/EBC",
    "totalParts": 2,
    "type": "Collection"
  },
  "isRoot": true
}

// filterableValues
"collection.category.id": "PP",
"collection.root.id": "uy0hyncn",
"collection.root.title": "Papers of Ernst Boris Chain",
"collection.isRoot": true

// aggregatableValues
"collection.category": { "id": "PP", "label": "Personal papers" },
"collection.root": { "id": "uy0hyncn", "label": "Papers of Ernst Boris Chain" }
```

`category` rather than `type` because display objects here use `type` as their discriminator, so a `collection.type` holding "PP / Personal papers" would sit next to the block's own `type`, and next to the work level `type: "Collection"` that archive works already carry. That's a suggestion rather than a strong preference, `family` or `grouping` would do the same job.

`collection.root` is the same shape as `DisplayRelation` (`ingestor/models/display/relation.py`), which is what `parts` and `partOf` already serialise to, so it could reuse that model and pick up `totalParts` for free.

Is there a case for the collection path living in the block as well? `display.referenceNumber` already carries `PP/EBC` and the query section has `collectionPath.path`, `.label` and `.sort`, so I'd only want this if `referenceNumber` moved into the block rather than a fourth copy of the same string appearing.

Last thought: `collection` looks like a candidate for the `include` param, in the same way `parts` and `partOf` already are. Filtering and aggregating read `filterableValues` and `aggregatableValues` rather than `display`, so gating the display block costs the browse page nothing beyond a query param.

catalogue-api#942 already has `archiveType`, `collectionRoot` and `isCollectionRoot` in `DocumentFilter`, `WorkAggregations`, `SortRequest`, `WorksIncludes`, the spec and the test fixtures, and `works-indexed-2026-07-30` holds 3,204,972 documents under the current names, so it would want re-ingesting.

catalogue_graph/src/models/pipeline/archive_category.py
StepanBrychta: > A fair amount of archive material comes out with no type.

That's on purpose. The designs only define 16 fixed archive types, which are "unlikely to change" (see [here](https://www.figma.com/design/JUoyimGZGSELwBhMn2gQwj/Archive-browse?node-id=10877-8961&m=dev)). Other archive families are left out on purpose. The `COLLECTION_PATH_PREFIX_LABELS` dictionary can always be expanded with more ID/label mappings if we want to add more recognised archive types in the future.

> Each has at least one work flagged isCollectionRoot (RAMC 1, UGC 3, MS.* 273), so they would appear on the collections listing page with no type to filter them by.

The would not appear on the collections type listing page precisely because they don't have an `archiveType`. The page only shows archive roots belonging to a specific archive type, so any query fetching the archive roots must include a filter like `archiveType=PP`. 

The `isCollectionRoot` field is purposefully more broad than the `archiveType` field. Not all hierarchical collections are archives, so it makes sense for some works to be collection roots but also have no archive type. 

> Separately, if UGC is one you'd want to add later it needs more than a table entry: its labels are UGC 198/8/5/8, so label.split("/")[0] returns UGC 198, and ([A-Z]+)(\d+) doesn't match across the space.

I'm aware that not all collection paths match the regex, but all collection paths which belong to one of the recognised archive types do. A strict regex protects against accidentally matching collection paths which shouldn't be matched.

catalogue_graph/src/ingestor/models/neptune/query_result.py
StepanBrychta: Well spotted, I'm not sure what the best approach is. Tightening the check here doesn't fully solve the underlying data issue, and the work pages would still look like collection roots.

On the fourteen `PSY/xxx` works: since we don't have a real `PSY` root, treating the fourteen works as collection roots might be the right thing to do, even if it means not respecting the 'root collection paths don't have a slash' convention.

StepanBrychta:
> Can we avoid ArchiveType? Everything else in this area is "collection": collectionPath, collectionRoot, the Collection/Section/Series work types. ArchiveType would be the only member of its family, and I'm not sure the distinction earns a second vocabulary.

The `archive` vs `collection` naming distinction is intentional. `collectionRoot`, `collectionPath`, and `isCollectionRoot` are all general fields applying to any hierarchical collection of works. `archiveType` applies more narrowly to archival collections. Naming it `collectionType` (or `collectionCategory`) would be less precise, and it would also make things more difficult for us in the future if we wanted to start categorising collections in some other way (since the `collectionType` field would already be taken).

> Relatedly, could isCollectionRoot stop being a root-level field and become part of a single collection block?

I think this is a good idea, nesting the fields into a shared block is cleaner than having all of them sit at the root of the display JSON.

> category rather than type because display objects here use type as their discriminator, so a collection.type holding "PP / Personal papers" would sit next to the block's own type, and next to the work level type: "Collection" that archive works already carry. That's a suggestion rather than a strong preference, family or grouping would do the same job.

Using `category` rather than `type` would mean adopting different terminology than the frontend. That's not a big issue, but I think it would be good to avoid it if we can, so that we don't add to the confusion of genres vs types/techniques and concept pages vs theme pages. (Of course, the frontend terminology can always change no matter what terminology we choose here, so maybe this is not a strong enough reason to keep using `type`.)

> Is there a case for the collection path living in the block as well? display.referenceNumber already carries PP/EBC and the query section has collectionPath.path, .label and .sort, so I'd only want this if referenceNumber moved into the block rather than a fourth copy of the same string appearing.

At the moment, this is a purely additive change which comes with forward compatibility for free. Moving the `referenceNumber` would mean keeping forward compatibility in mind when switching to the new index. (This is not a reason not to do it, just a consideration.)

> Last thought: collection looks like a candidate for the include param, in the same way parts and partOf already are. Filtering and aggregating read filterableValues and aggregatableValues rather than display, so gating the display block costs the browse page nothing beyond a query param.

I agree, the existing provisional display fields are already gated behind `include` on the [catalogue API branch](https://github.com/wellcomecollection/catalogue-api/pull/942).

kenoir:
> The archive vs collection naming distinction is intentional.

That makes sense, and perhaps also reframes the `type` discussion. Arguably the type of this collection is `archival`, and that makes me feel more strongly we should avoid `type` here even if it diverges from front-end terminology.

kenoir (approved):

PR #3535, opened 2026-08-04
Fix handling of date type q and partial years in EBSCO transformer

annak-k:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6482 explains the problem and how to verify it's fixed. 54 EBSCO records were failing the transformer deterministically and had no works in the pipeline, despite following valid (if messy) MARC date conventions.

The issue describes two failure classes, both fixed across this branch's commits:

1. **MARC 008 date type `q`** ("questionable date: between date 1 and date 2") was unhandled by `field008.py`, which raised `NotImplementedError: unexpected MARC008 date type value: q`. Fixed by treating it identically to the existing `d`/`m` range case, with a doctest example (`51cbbc512`).
2. **A messy/unparseable 260/264 date crashed `period.py`'s date-range builder** (`int()` on a non-numeric fragment), failing the whole record. Fixed at the call sites in `production.py`: `parse_period` failures are now caught and degrade to a label-only `Period` (`range=None`), and `extract_production`'s 008 fallback triggers whenever the 260/264 date has no usable range — not just when the dates list is empty — so a good 008 date is preferred over a label-only one. `axiell/production.py` already used this exact pattern, so this brings EBSCO in line with it.

Also fixed, found while testing the above against real records: a bracket landing between two digits of the same year (e.g. `"1[841-1849]"`) acted as a separator during digit extraction, breaking the year into a dropped single digit ("1") and a truncated one ("841") — silently mislabelling a record whose 008 field encoded the correct 1841 start date. `period.py` now discards bracket characters before extracting digits, rather than letting them interrupt a digit run.

Verified against the actual raw MARC values for real EBSCO records fetched from the live adapter store — `ebs144516e`, `ebs16268864e`, `ebs375800e`, `ebs28835937e`, and others — checking their 008/260/264 fields directly and confirming the transformer's output against them. Fields from these records were used as fixtures in the new tests covering this functionality, in `unit/test_production.py`.

### Checklist

- [ ] Does this change the display model the catalogue API returns? If so, regenerate the test documents under `catalogue_graph/document_generators/test_documents`. The `wellcomecollection/catalogue-api` repo copies them in with `copy_test_documents.py`, and its OpenAPI contract tests validate the published spec against them, so stale fixtures let the spec drift.
  - No schema change — `Period.range` was already optional. This only changes which values these 54 records now produce.

## How to test

- `uv run pytest --doctest-modules src/adapters/transformers/ebsco/parsers --no-cov` — parser doctests.
- `uv run pytest tests/adapters/transformers/ebsco --no-cov -q` — full EBSCO transformer suite, including regression tests in `unit/test_production.py` covering these records.
- To reproduce any of the previously-failing records directly against the adapter store, use `catalogue_graph/notebooks/adapter_source_tables.ipynb`.

The `ebsco-transformer-failures-2026-07-03` alarm should stop firing for these records. Re-running the changeset redrive (`c2dce996-824c-11f1-afb6-0a58a9feac02`) should show `failure_count` drop and `success_count` include the previously-missing 54 IDs.

- `parse_period` itself still raises `ValueError` on an unparseable date — this is now caught at each EBSCO call site (`production.py`), so the worst case for a genuinely unrecoverable date is a label-only `Period` (label preserved, no computed range), or a fallback to the record's 008 date if it has one, not a crash. Both outcomes are explicitly tested.
- This also fixes the equivalent dead code path in `axiell/production.py`'s `_parse_period_or_bare_label`, since `parse_period` is shared between the two transformers and previously never raised for it to catch.

### DIFF
--- a/catalogue_graph/src/adapters/transformers/ebsco/parsers/field008.py
+++ b/catalogue_graph/src/adapters/transformers/ebsco/parsers/field008.py
@@ -126,7 +126,7 @@ class Field008:
         >>> Field008("||||||m191u195u").maximal_date_range
         '1910-1959'
 
-        Date type q represents a questionable date between two years. 
+        Date type q represents a questionable date between two years.
         We treat it as a range like d and m
         >>> Field008("||||||q19251956").maximal_date_range
         '1925-1956'
@@ -163,7 +163,7 @@ class Field008:
             return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
 
         if date_type == "q":
-            # q means a questionable date between date 1 and date 2, 
+            # q means a questionable date between date 1 and date 2,
             # but we will treat it like d and m
             date_2 = self.raw_field.date_2
             return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
--- a/catalogue_graph/src/adapters/transformers/ebsco/parsers/field008.py
+++ b/catalogue_graph/src/adapters/transformers/ebsco/parsers/field008.py
@@ -125,6 +125,11 @@ class Field008:
         '1900-2009'
         >>> Field008("||||||m191u195u").maximal_date_range
         '1910-1959'
+
+        Date type q represents a questionable date between two years. 
+        We treat it as a range like d and m
+        >>> Field008("||||||q19251956").maximal_date_range
+        '1925-1956'
         """
         date_type = self.raw_field.date_type
         if date_type in "n|":
@@ -154,6 +159,12 @@ class Field008:
             # but for our purposes, they both represent a range.
             # d is an actual range of publication of a continuing resource
             # m is a pair of dates for a multipart item.
+            date_2 = self.raw_field.date_2
+            return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
+
+        if date_type == "q":
+            # q means a questionable date between date 1 and date 2, 
+            # but we will treat it like d and m
             date_2 = self.raw_field.date_2
             return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
 
--- a/catalogue_graph/src/adapters/transformers/ebsco/parsers/field008.py
+++ b/catalogue_graph/src/adapters/transformers/ebsco/parsers/field008.py
@@ -154,17 +154,12 @@ class Field008:
                 return f"{date_1.replace('u', '0')}-{date_1.replace('u', '9')}"
             return date_1
 
-        if date_type in "dm":
-            # these two have subtly different meanings in MARC,
-            # but for our purposes, they both represent a range.
+        if date_type in "dmq":
+            # d, m, and q all have subtly different meanings in MARC,
+            # for our purposes, we will treat all three the same, as a range of dates.
             # d is an actual range of publication of a continuing resource
-            # m is a pair of dates for a multipart item.
-            date_2 = self.raw_field.date_2
-            return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
-
-        if date_type == "q":
-            # q means a questionable date between date 1 and date 2,
-            # but we will treat it like d and m
+            # m is a pair of dates for a multipart item
+            # q is a questionable date between two years
             date_2 = self.raw_field.date_2
             return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
 
--- a/catalogue_graph/src/adapters/transformers/ebsco/parsers/period.py
+++ b/catalogue_graph/src/adapters/transformers/ebsco/parsers/period.py
@@ -9,7 +9,16 @@ from models.pipeline.identifier import Identifiable, Unidentifiable
 # or &amp;#41; (rparen) as the numbers 40 and 41 respectively.
 RE_DISCARD = re.compile(r"(&(amp;)?.+?;)")
 
-RE_KEEP = re.compile(r"\d{2,4}|-")
+# A digit run immediately followed by just enough dashes to pad it to a
+# 4-character year is a MARC convention for an unknown trailing digit, e.g.
+# "182-" (the 1820s), "19--" (the 1900s). The specific 3+1/2+2/1+3
+# alternatives must come before the general `\d{2,4}` fallback, otherwise a
+# *complete* 4-digit year followed by a real range-separator dash (e.g.
+# "1956-", meaning "1956 onwards") would be mistaken for a partial year.
+RE_KEEP = re.compile(r"\d{3}-(?!-)|\d{2}-{2}(?!-)|\d-{3}(?!-)|\d{2,4}|-")
+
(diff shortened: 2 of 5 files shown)


catalogue_graph/src/adapters/transformers/ebsco/parsers/period.py
```diff
@@ -9,7 +9,16 @@
...
 RE_DISCARD = re.compile(r"(&(amp;)?.+?;)")
 
-RE_KEEP = re.compile(r"\d{2,4}|-")
+# A digit run immediately followed by just enough dashes to pad it to a
+# 4-character year is a MARC convention for an unknown trailing digit, e.g.
+# "182-" (the 1820s), "19--" (the 1900s). The specific 3+1/2+2/1+3
+# alternatives must come before the general `\d{2,4}` fallback, otherwise a
+# *complete* 4-digit year followed by a real range-separator dash (e.g.
+# "1956-", meaning "1956 onwards") would be mistaken for a partial year.
+RE_KEEP = re.compile(r"\d{3}-(?!-)|\d{2}-{2}(?!-)|\d-{3}(?!-)|\d{2,4}|-")
```
kenoir: `\d{3}-(?!-)` can't distinguish a partial year from a complete 3-digit year followed by the real separator. The only difference is whether a digit follows the dash.

Parsing every 260/264 $c and $g string in the live EBSCO adapter store (188,981 records) under `origin/main` and under a908de4c, one record's range changes: `ebs16268864e`, `264$c = 1[841-1849]`, 008 `d18411849`.

```
main:        0841 to 1849
this branch: 1849 to 8419
008 says:    1841 to 1849
```

`1[` isn't a year token, so matching starts at `841-`, the new alternative claims it as a partial year, and `crack` falls to the no-separator branch where min/max string-compare `"841-"` against `"1849"`. The bounds swap, and `841-` resolves to 8419.

That reaches the index: `work_query_transformer.py:162` feeds `range.from_time` into `production.dates.range.from`, and `work_aggregate_transformer.py:108` buckets the from year for the facet. So this work's date facet moves from 841 to 1849 when the answer sitting in its 008 is 1841.

Adding `\d` to the three lookaheads, `(?![-\d])`, fixes the general case with 214 tests and 17 doctests still green, but for this record it only restores main's 0841. 

I think it comes out right only if `1[841-1849]` yields no range and your 008 fallback supplies it. The three records that newly get an 0001 or 9999 bound (`April 9--M.DCC.XCII. [1792]`, `--1794--`, `--1795--`) are the same shape: each has a single unambiguous 008 year the fallback would use. Those bounds aren't new, so this is about the fallback being a better answer than a guess rather than a regression.

catalogue_graph/src/adapters/transformers/ebsco/parsers/period.py
```diff
@@ -31,11 +40,23 @@ def parse_period(
...
+
+    Anything else degrades to a label-only Period rather than raising.
+    >>> parse_period("mid century")
+    Period(id=Unidentifiable(canonical_id=None, type='Unidentifiable'), label='mid century', type='Period', range=None)
     """
     range_fn = century_to_range if "century" in period else to_range
+    try:
+        date_range = range_fn(period)
+    except ValueError:
+        date_range = None
```
kenoir: Warning that this changes Axiell as well as EBSCO. `parse_period` is shared, and `axiell/production.py:54` catches the ValueError this now swallows, so that branch is unreachable and can go.

I measured the Axiell side against the live adapter store. Only 120 unique 264$c labels reach `parse_period` at all across 206,526 records, because a label with both 046 dates present takes the `_period_from_dates` path instead (42,677 unique labels do). Two of the 120 change, and both are the XSS/SQLi test records in the collect set, `collect:110206464` and `collect:110206466`, going from label-only to a nonsense range. No real Axiell record changes, so no objection from me.

**Worth a follow-up to move the parser out of the EBSCO namespace, given Axiell depends on it.** `adapters/transformers/marc/parsers/` already exists with `date_from_005.py` in it, and the marc tests for `other_identifiers`, `parents` and `subjects` note in their docstrings that those implementations are generic MARC sitting under ebsco, so `period.py` would be joining a migration that has already started. Happy to raise that issue so this PR stays focused.

catalogue_graph/src/adapters/transformers/ebsco/parsers/field008.py
```diff
@@ -157,5 +162,11 @@ def maximal_date_range(self) -> str | None:
             date_2 = self.raw_field.date_2
             return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
 
+        if date_type == "q":
+            # q means a questionable date between date 1 and date 2,
+            # but we will treat it like d and m
+            date_2 = self.raw_field.date_2
+            return f"{date_1.replace('u', '0')}-{date_2.replace('u', '9')}"
```
kenoir: Identical to the `dm` branch above. `if date_type in "dmq"`, with the extra sentence in that branch's comment, would do the same job.

catalogue_graph/src/adapters/transformers/ebsco/parsers/period.py
annak-k: My original idea was to add try/excepts to the parse_period calls in production.py so that any parse_period ValueError failure would result in falling back to 008. Is that the more appropriate strategy here? I wasn't sure how much we wanted to rely on 008.

kenoir: Falling back on ValueError wouldn't catch this one, because it doesn't raise. Running the branch's parser on that record:

```
'1[841-1849]'  ->  preprocess: '841- 1849'
                   range: 1849-01-01 -> 8419-12-31
                   parse_period(...).range is None?  False
```

The parse succeeds, it's just wrong.

The same goes for the check already on the branch: `if not any(date.range for date in productions260_264[0].dates)` doesn't fire either, because there *is* a range. 

I think it's ok to rely on 008 here. 008/07-14 is fixed-format and machine-readable and exists so software can filter and sort by date, whereas 260/264 $c is free text transcribed off the item. 

You might be able to use try/catch if you throw when you notice an implausible date range, then fall back to 008, but without consulting the dataset more thoroughly I don't know what would count as implausible.

annak-k: Sorry, I meant would it be most appropriate to revert all the changes in period.py, going back to the original `\d{2,4}|-`, then add a try/except for `ValueError` to the `parse_period` calls in production.py? Then we would fall back to 008 without having to figure out what date ranges are plausible and handle all of them. This would also make the axiell `except ValueError` branch live again.

kenoir: Yes, I think that makes. We need to make sure on the real data what the impact would be (i.e. are there records where the 008 fallback wouldn't work).

kenoir (approved):

PR #3539, opened 2026-08-04
Tei adapter deletion bug fix

agnesgaroux:
## What does this change?

See https://github.com/wellcomecollection/platform/issues/6487
Increase tei queues visibility timeout to prevent messages from being redriven before the delete_delay allows the service to process anything, and landing i nthe DLQ after a few tries
Also: align the DRS DB to the engine version already running in the deployed service

### Checklist

- [ ] Does this change the display model the catalogue API returns? If so, regenerate the test documents under `catalogue_graph/document_generators/test_documents`. The `wellcomecollection/catalogue-api` repo copies them in with `copy_test_documents.py`, and its OpenAPI contract tests validate the published spec against them, so stale fixtures let the spec drift.

## How to test

Run `terraform plan` in tei_adapter/terraform
```
Plan: 2 to add, 4 to change, 2 to destroy.
```
Add/destroy: the task defs whose env vars are changing 
module.tei_id_extractor_w.module.scaling_service.module.service.aws_ecs_service.service and module.tei_adapter_w.module.scaling_service.module.service.aws_ecs_service.service depends on task definitions 
service's read from queue policy, assuming linked to update task definition too 

https://wellcomecollection.org/works/bured6mf is not available in the public catalogue

- Slower recovery from a genuinely crashed consumer. If a task dies mid-processing for an unrelated reason (OOM, deploy, crash), the message now sits invisible for up to 12/40 minutes before anything can retry it — versus under a minute before.
- Slower DLQ escalation for truly broken messages. A message that will never succeed now takes far longer to exhaust maxReceiveCount and land in the DLQ (each failed attempt now "costs" up to 12/40 minutes instead of seconds), delaying any DLQ-based alerting.
- Reduced effective throughput under bursty load, since each in-flight message occupies a slot for much longer even when doing essentially no CPU work (just waiting out the delay) — a lower-probability concern given this is a low-volume queue, but worth naming.

~~The main issue that *needs* addressing is the scale down of the services based on number of visible messages: the "queue low → scale down" alarm has ample opportunity to fire and terminate the running task while the message is still mid-delay. Looking into this~~ The scale down adds up all SQS messages states so won't trigger until no message in flight

### DIFF
--- a/tei_adapter/terraform/rds_id_extractor.tf
+++ b/tei_adapter/terraform/rds_id_extractor.tf
@@ -56,7 +56,7 @@ module "tei_id_extractor_rds_serverless_cluster" {
   db_security_group_id     = aws_security_group.database_sg.id
   aws_db_subnet_group_name = aws_db_subnet_group.default.name
 
-  engine_version = "8.0.mysql_aurora.3.08.2"
+  engine_version = "8.0.mysql_aurora.3.10.3"
 
   snapshot_identifier = "aurora-mysql-v3-tei-pre-migration-24-15-08"
 }
--- a/tei_adapter/terraform/tei_id_extractor.tf
+++ b/tei_adapter/terraform/tei_id_extractor.tf
@@ -7,7 +7,7 @@ module "tei_id_extractor_w" {
   topic_arns = [module.tei_updater_lambda.topic_arn]
 
   queue_name                       = "tei-id-extractor"
-  queue_visibility_timeout_seconds = local.rds_lock_timeout_seconds + 30
+  queue_visibility_timeout_seconds = local.tei_id_extractor_delete_delay_seconds + local.visibility_timeout_buffer_seconds
 
   env_vars = {
     metrics_namespace = "${local.namespace}_tei_id_extractor"
@@ -15,7 +15,7 @@ module "tei_id_extractor_w" {
     bucket            = aws_s3_bucket.tei_adapter.id
     parallelism       = 10
     max_connections   = local.tei_id_extractor_max_connections
-    delete_delay      = "30 minutes"
+    delete_delay      = "${local.tei_id_extractor_delete_delay_seconds}s"
     database          = "pathid"
     table             = "pathid"
   }
--- a/tei_adapter/terraform/locals.tf
+++ b/tei_adapter/terraform/locals.tf
@@ -13,6 +13,17 @@ locals {
   tei_id_extractor_max_connections = 5
   rds_lock_timeout_seconds         = 10 * 60
 
+  # tei_id_extractor and tei_adapter both delay processing of *Deleted*
+  # messages, to avoid mistaking a file rename (delete-of-old + add-of-new)
+  # for a genuine deletion before the corresponding "changed" message has
+  # landed. Each queue's visibility timeout MUST exceed its own delete
+  # delay by a comfortable margin - otherwise SQS redelivers an in-flight
+  # delete message before the delay finishes, causing it to loop
+  # indefinitely and land in the DLQ without ever completing processing.
+  tei_adapter_delete_delay_seconds      = 2 * 60  # 2 minutes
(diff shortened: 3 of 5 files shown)


tei_adapter/terraform/tei_id_extractor.tf
```diff
@@ -7,15 +7,15 @@ module "tei_id_extractor_w" {
   topic_arns = [module.tei_updater_lambda.topic_arn]
 
   queue_name                       = "tei-id-extractor"
-  queue_visibility_timeout_seconds = local.rds_lock_timeout_seconds + 30
+  queue_visibility_timeout_seconds = local.tei_id_extractor_delete_delay_seconds + local.visibility_timeout_buffer_seconds
```
kenoir, line 10: The old value was `rds_lock_timeout_seconds + 30`, which I read as covering DB waits during processing. The new formula drops that term (the buffer only coincidentally equals 600s). Probably fine, and the old coupling looks shaky anyway (the `wait_timeout` parameter group doesn't seem attached to the Aurora 3 cluster), but was that deliberate? Keeping the old intent would be `delete_delay + rds_lock_timeout_seconds + buffer`.

tei_adapter/terraform/rds_id_extractor.tf
```diff
@@ -56,7 +56,7 @@ module "tei_id_extractor_rds_serverless_cluster" {
   db_security_group_id     = aws_security_group.database_sg.id
   aws_db_subnet_group_name = aws_db_subnet_group.default.name
 
-  engine_version = "8.0.mysql_aurora.3.08.2"
+  engine_version = "8.0.mysql_aurora.3.10.3"
```
kenoir, line 59: Nice to have, no action needed for this PR: this drift will recur next time AWS auto-upgrades the cluster. `ignore_changes = [engine_version]` (or pinning `auto_minor_version_upgrade`) would stop the churn.

kenoir (approved):
Nice fix, and the issue write-up is an excellent piece of diagnosis. I checked the Scala side: both services hold delete messages in an in-stream Pekko delay without acking, so raising visibility above the delay is the right fix, and "120s" parses the same as "2 minutes" via getDuration. Nothing blocking inline.

kenoir:
On the "queue low -> scale down" worry in the description: I think this is already covered. The `queue_low` alarm sums Visible + NotVisible + Deleted messages, and a message in its delete delay is NotVisible, so scale-down can't fire mid-delay. The remaining risk is a deploy or Fargate Spot interruption killing a task mid-delay, which with the new timeouts means slower redelivery, not the DLQ loop.

agnesgaroux:
> On the "queue low -> scale down" worry in the description: I think this is already covered. The `queue_low` alarm sums Visible + NotVisible + Deleted messages, and a message in its delete delay is NotVisible, so scale-down can't fire mid-delay. The remaining risk is a deploy or Fargate Spot interruption killing a task mid-delay, which with the new timeouts means slower redelivery, not the DLQ loop.

Yes, I also saw it in the task's events where it stays up for roughly the length of the delete_delay even when only one message was delivered

tei_adapter/terraform/tei_id_extractor.tf
agnesgaroux: The buffer acts as a catch_all "let the service process the message" time used in both adapter and id_extractor so the visibility is 
- wait for the delete_delay to expire
- then give it some time to do its thing (including waiting out lock timeout in the case of the extractor)

tei_adapter/terraform/rds_id_extractor.tf
agnesgaroux: > [auto_minor_version_upgrade](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#auto_minor_version_upgrade-1) - (Optional) Indicates that minor engine upgrades will be applied automatically to the DB instance during the maintenance window. **Defaults to true.** 🤨

PR #3546, opened 2026-08-06
Read every language on a TEI textLang instead of rejecting the record

### LINKED ISSUE #3545: TEI transformer rejects manuscripts whose textLang names both a main and other language
The TEI transformer rejects a whole manuscript when a single `textLang` element carries both `mainLang` and `otherLangs`. That is the standard TEI way to describe a manuscript written in more than one language, so any bilingual manuscript is unpublishable.

`TeiLanguages.parseLanguageId` accepts `mainLang` on its own, or `otherLangs` on its own, and returns a `Left` when both are present:

```scala
case _ =>
  Left(new RuntimeException(s"Multiple language IDs in $n"))
```

The `Left` propagates out of the fold in `parseLanguages`, out of `TeiNestedData`, and the transform fails, so the manuscript never reaches the catalogue.

## What this costs today

Three Arabic manuscripts are absent from the catalogue for this reason, confirmed from the transformer logs for the 2026-07-31 redrive:

| Manuscript | File | `textLang` | Label |
|---|---|---|---|
| `manuscript_16417` | `Arabic/MS_Arabic_773.xml` | `mainLang="ar" otherLangs="es"` | Arabic with interlinear Spanish |
| `manuscript_16464` | `Arabic/MS_Arabic_820.xml` | `mainLang="ar" otherLangs="ota"` | Arabic and Turkish |
| `manuscript_16530` | `Arabic/MS_Arabic_886.xml` | `mainLang="ar" otherLangs="fa"` | Arabic and Persian |

The markup in all three is correct, and the labels describe the manuscripts accurately.

## Proposed behaviour

Stop treating two attributes as an error. Read every id on the element, `mainLang` first and then each space-separated value in `otherLangs`, and offer each in turn to `TeiLanguageData` with the element's label:

- Every id that maps to a MARC language contributes a `Language`, in document order.
- If no id maps, the label becomes a language note, which is what already happens for a single id whose label cannot be matched.

For the three manuscripts above this produces a language note carrying the full label, because `TeiLanguageData` matches on the id and label together and no single label matches a combined description. That is the existing fallback rather than a new behaviour, and it publishes the work instead of discarding it. Where a combined element does carry a matchable label, for instance `mainLang="ar" otherLangs="ota"` labelled "Arabic", the Arabic language is kept.

Splitting the label into per-language parts is deliberately out of scope. The label is free text and the current design matches id and label together on purpose, to surface source data that needs fixing.

Split out from wellcomecollection/platform#6494, which covers the other seven manuscripts in the same dead-letter set, all of them missing `xml:id` on `msItem` elements.

kenoir:
## What does this change?

Closes #3545.

A `textLang` that carries both `mainLang` and `otherLangs` is the standard TEI way to describe a manuscript written in more than one language. `TeiLanguages.parseLanguageId` treated that combination as an error and returned a `Left`, which propagated out of the fold and failed the whole transform, so any bilingual manuscript was unpublishable. Three Arabic manuscripts are absent from the catalogue for this reason, found in the dead-letter set from the wellcomecollection/platform#6445 reindex.

Every id on the node is now read, `mainLang` first and then each space-separated value in `otherLangs`, and offered in turn to `TeiLanguageData` with the node's label. Ids that map contribute a `Language` in document order. If no id maps, the label becomes a language note, which is the existing fallback for a label that cannot be matched.

Splitting a combined label such as "Arabic and Persian" into per-language parts is deliberately not attempted. `TeiLanguageData` matches on id and label together on purpose, to surface source data that needs fixing, and guessing at free text would undermine that.

### Checklist

- [x] Does this change the display model the catalogue API returns? No. It changes which works are published and can add a language note, but the model itself is unchanged.

## How to test

`sbt "transformer_tei/test"`. `TeiLanguagesTest` gains four cases: both attributes with a matchable label keeps the language, a space-separated `otherLangs` still keeps the main language, both attributes with a combined label produce a language note, and a multi-language node no longer discards the languages found on other nodes in the same manuscript. The test that asserted the old error is gone, since that behaviour is what this removes.

Verified against the real files as well. Running the three dead-lettered manuscripts through `TeiXml(...).parse` on this branch returns `Right` for all three, where `main` fails them:

| Manuscript | File | `textLang` |
|---|---|---|
| `manuscript_16417` | `Arabic/MS_Arabic_773.xml` | `mainLang="ar" otherLangs="es"`, "Arabic with interlinear Spanish" |
| `manuscript_16464` | `Arabic/MS_Arabic_820.xml` | `mainLang="ar" otherLangs="ota"`, "Arabic and Turkish" |
| `manuscript_16530` | `Arabic/MS_Arabic_886.xml` | `mainLang="ar" otherLangs="fa"`, "Arabic and Persian" |

`manuscript_16417` comes out with its inner work carrying `Note(LanguageNote, "Arabic with interlinear Spanish")`, which is the intended outcome.

One local test fails on this branch and on `main` alike: `TeiTransformerEndToEndTest` needs localstack on port 4566.

The three manuscripts appear in the catalogue after the next TEI transform, and `Multiple language IDs` stops appearing in the TEI transformer logs.

The change is permissive, so a manuscript that used to fail now publishes. The risk is publishing a work with less language data than a cataloguer intended, since a combined label lands in a language note rather than as structured languages. That is strictly better than the current behaviour of publishing nothing at all, and the `Unable to map TEI language to catalogue language` warning still fires per unmapped id, so the source data is still visible as something to fix.

No alarm is needed. The failure mode this removes was silent by nature: it dead-lettered a record with nothing to tell the person who wrote it, which is being followed up separately.

### DIFF
--- a/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiLanguages.scala
+++ b/pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiLanguages.scala
@@ -30,55 +30,51 @@ object TeiLanguages extends Logging {
         case (n, Right((languageList, languageNoteList))) =>
           val label = n.text
 
-          val langId = parseLanguageId(n)
-
-          (langId, label) match {
-            case (Right(Some(id)), label) if label.trim.nonEmpty =>
-              appendLanguageOrNote(languageList, languageNoteList, id, label)
-            case (Right(None), label) if label.trim.nonEmpty =>
-              appendNote(languageList, languageNoteList, label)
-            case (Right(_), _) =>
+          (parseLanguageIds(n), label) match {
+            case (_, label) if label.trim.isEmpty =>
               warn(s"Missing label for language node $n")
               Right((languageList, languageNoteList))
-            case (Left(err), _) => Left(err)
+            case (Nil, label) =>
+              appendNote(languageList, languageNoteList, label)
+            case (ids, label) =>
+              appendLanguagesOrNote(languageList, languageNoteList, ids, label)
           }
 
         case (_, Left(err)) => Left(err)
       }
 
-  private def parseLanguageId(n: Node) = {
-    val mainLangId = (n \@ "mainLang").toLowerCase
-    val otherLangId = (n \@ "otherLangs").toLowerCase
-    (mainLangId, otherLangId) match {
-      case (id1, id2) if id2.isEmpty && id1.nonEmpty => Right(Some(id1))
-      case (id1, id2) if id1.isEmpty && id2.nonEmpty => Right(Some(id2))
-      case (id1, id2) if id2.isEmpty && id1.isEmpty =>
-        Right(None)
-      case _ =>
-        Left(new RuntimeException(s"Multiple language IDs in $n"))
-    }
-  }
+  /** A textLang may name a main language, other languages, or both. Both is how
+    * TEI describes a multi-language manuscript, so read every id rather than
+    * treating the combination as an error.
+    */
+  private def parseLanguageIds(n: Node): List[String] =
+    (Seq(n \@ "mainLang") ++ (n \@ "otherLangs").split("\\s+"))
+      .map(_.toLowerCase.trim)
+      .filter(_.nonEmpty)
+      .distinct
+      .toList
 
-  private def appendLanguageOrNote(
+  private def appendLanguagesOrNote(
     languageList: List[Language],
     languageNoteList: List[Note],
-    id: String,
+    ids: List[String],
     label: String
-  ) =
-    TeiLanguageData(id, label).fold(
-      err => {
-        warn("Could not parse language", err)
-        appendNote(languageList, languageNoteList, label)
-      },
-      appendLanguage(languageList, languageNoteList, _)
-    )
+  ) = {
+    val languages = ids.flatMap {
+      id =>
+        TeiLanguageData(id, label) match {
+          case Right(language) => Some(language)
+          case Left(err) =>
+            warn("Could not parse language", err)
+            None
+        }
+    }
 
-  private def appendLanguage(
-    languageList: List[Language],
-    languageNoteList: List[Note],
-    language: Language
-  ) =
-    Right((language +: languageList, languageNoteList))
+    // The label is shared by every id on the node, so it only becomes a note
+    // when it yielded no language at all.
+    if (languages.isEmpty) appendNote(languageList, languageNoteList, label)
+    else Right((languages ++ languageList, languageNoteList))
+  }
 
   private def appendNote(
     languageList: List[Language],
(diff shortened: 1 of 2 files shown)


pipeline/transformer/transformer_tei/src/test/scala/weco/pipeline/transformer/tei/transformers/TeiLanguagesTest.scala
```diff
@@ -77,18 +77,68 @@ class TeiLanguagesTest
     ))
   }
 
-  it("errors on languages that have more than one lang attribute") {
+  it("reads every language on a node that has both mainLang and otherLangs") {
     val xml =
       teiXml(
         languages = List(
-          <textLang mainLang="id1" otherLangs="id2">Sanskrit</textLang>
+          <textLang mainLang="ar" otherLangs="ota">Arabic</textLang>
```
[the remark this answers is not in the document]
kenoir: Good catch, you are right that this test only proved the mainLang mapping. I have added your case rather than swapping to it: the multi-language test now uses `mainLang="grc" otherLangs="el"` with the shared label "Greek", which maps both ids and asserts them in document order.

I kept the ar/ota example because it pins down a separate decision, renamed to "keeps the language of the id that maps when another id does not". That is the shape the three manuscripts this PR fixes actually have, and it proves the node contributes the mapped language and no duplicate language note. Fixed in 1ec0f96a4e.

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiLanguages.scala
```diff
@@ -30,55 +30,51 @@ object TeiLanguages extends Logging {
...
               warn(s"Missing label for language node $n")
               Right((languageList, languageNoteList))
-            case (Left(err), _) => Left(err)
+            case (Nil, label) =>
+              appendNote(languageList, languageNoteList, label)
+            case (ids, label) =>
+              appendLanguagesOrNote(languageList, languageNoteList, ids, label)
           }
 
         case (_, Left(err)) => Left(err)
```
agnesgaroux: Is this reachable?

agnesgaroux (approved):
Minor: the only Left in the fold is converted to warning/None so the final Left catch-all is unreachable

pipeline/transformer/transformer_tei/src/main/scala/weco/pipeline/transformer/tei/transformers/TeiLanguages.scala
kenoir: No, and well spotted. Nothing in the fold produces a `Left` any more, since every mapping failure becomes a warning plus a language note. It was only there because the accumulator was an `Either` and the match had to be total.

Removed in 52be2ca69b: the accumulator is a plain tuple now and `parseLanguages` wraps the result in a `Right` at the end, so the signature callers rely on is unchanged.

That commit also pins the id lowercasing to `Locale.ROOT`, which is the same locale-independence point Copilot raised on the METS access statuses in #3547.

PR #3548, opened 2026-08-06
Regenerate graph integration test fixtures

kenoir:
## What does this change?

Regenerates every fixture under `catalogue_graph/integration/graph/fixtures/` from the production graph using the repo's own `generate_fixtures.py`, and records the run in `REGENERATION_LOG.md`. Same treatment as #3417.

The main build for #3542 failed on `test_graph_query_matches_fixture[concept_people]`: three of the twenty related-concepts lists no longer match what the live graph returns, which puts the fixture under the harness's 0.9 match tolerance. 

### Checklist

- [ ] Does this change the display model the catalogue API returns? No. It only re-records integration test fixtures, so `document_generators/test_documents` is untouched.

## How to test

From `catalogue_graph/`, with graph credentials:

```
GRAPH_DATE=prod AWS_PROFILE=platform-developer uv run --group dev pytest -m integration integration/graph
```

All 27 tests pass on this branch. The same command on `main` reproduces the `concept_people` failure.

Main goes green and the deploy pipeline runs again.

The fixtures are a snapshot of a graph that keeps changing, so they will drift again and need the same regeneration. The 0.9 tolerance means a regeneration can also absorb a genuine query regression without anyone noticing, so it is worth a quick look at the diff to confirm the changes look like ordinary content drift rather than a query returning less than it used to. Nothing here ships to production.


kenoir:
Superseded by #3549, which regenerates the drifted fixtures with a targeted --fixture option rather than the full set.

PR #3549, opened 2026-08-06
--fixture option runs a subset of fixtures

agnesgaroux:
## What does this change?

- Add targeted --fixture option to the script, to target specific fixtures to regenerate. Avoids 22 file diff everytime the script is run
- Fix drift in fixtures v production graph. Safe, the result of this specific query is not exposed on the public pages

### Checklist

- [x] Does this change the display model the catalogue API returns? If so, regenerate the test documents under `catalogue_graph/document_generators/test_documents`. The `wellcomecollection/catalogue-api` repo copies them in with `copy_test_documents.py`, and its OpenAPI contract tests validate the published spec against them, so stale fixtures let the spec drift. Test fixtures only 

## How to test

From `catalogue_graph/`, with graph credentials:

```
GRAPH_DATE=prod AWS_PROFILE=platform-developer uv run --group dev pytest -m integration integration/graph
```

All 27 tests pass on this branch. The same command on `main` reproduces the `concept_people` failure.

Success integration tests


catalogue_graph/integration/graph/generate_fixtures.py
```diff
@@ -121,127 +127,170 @@ def row_to_ancestor_work_ids(item: dict[str, Any]) -> list[str]:
...
-    reason = confirm_regeneration()
-    append_regeneration_log(reason=reason)
-
-    graph_date = input("Enter the graph date (e.g. 2025-01-01): ").strip()
-    client = NeptuneClient(graph_date)
+@dataclass
+class FixtureSpec:
+    name: str
+    query: str
+    id_label: str  # "Concept" or "Work"
```
StepanBrychta: Minor: if `id_label` is always either `Concept` or `Work`, we can use a `Literal` type instead of a leaving a comment next to a broad string definition.

StepanBrychta (approved):

PR #3550, opened 2026-08-06
Restructure axiell_folio_sync: split mapping, upsert and folio into subpackages

anubhavijay:
Part of wellcomecollection/platform#6462.

## What does this change?

Restructures the `axiell_folio_sync` adapter step by splitting two large monolithic modules (`mapping.py` ~456 lines, `upsert.py` ~462 lines) into focused subpackages, and extracts the sync loop into its own module.

**New structure:**

- `mapping/` — MARC → FOLIO payload mapping:
  - `marc.py` — MARCXML extraction primitives + `CanonicalRecord` model (was `mapper.py`)
  - `config.py` — field mapping table (extracted from `mapping.py`)
  - `builders.py` — record-level payload builders
  - `payloads.py` — top-level payload assembly
- `folio/` — FOLIO client seam:
  - `okapi.py` — OKAPI credential resolution (extracted from `sync_to_folio.py`)
  - `callables.py` — `FolioInventoryOps` protocol (was `folio_callables.py`)
  - `ref_cache.py` — tenant reference-data UUID cache
- `upsert/` — FOLIO Inventory write layer:
  - `entities.py` — single-entity operations
  - `writer.py` — create/update logic
  - `reconcile.py` — suppress/delete cascade
- `run_axiell_folio_sync.py` — the select → map → upsert loop (extracted from `sync_to_folio.py`)
- `axiell_adapter_read.py` — adapter-store read (changed rows + superseded-GUID deletions)
- `results.py` — upsert outcome models (`UpsertResult`, `GuidCascadeResult`, …)

**Mapping corrections also included in this restructure:**

- **Material type** — source field switched from `MARC 949$c` (`material_type_code`) to `MARC 655$a` (`object_category`), with case-insensitive lookup; the `CanonicalRecord` field is renamed accordingly.
- **Permanent location** — resolution now goes through the shared `FieldMap`/`_resolve` pattern (`LOCATION_FIELD`) instead of an ad-hoc inline call, keeping location source and resolver in sync.
- **Local identifier** — `MARC 035$a` (`object_number`) is now extracted and wired into the instance `identifiers` payload as the AxC local identifier.

The README module table is updated to reflect the new layout.

## How to test

Run the existing tests:

```bash
cd catalogue_graph
uv run pytest tests/adapters/steps/axiell_folio_sync/ -v
```

Check for type errors:

```bash
uv run mypy src/adapters/steps/axiell_folio_sync/
```

- All existing tests pass.
- No new mypy errors in the restructured modules.
- Material type, permanent location, and local identifier fields are correctly populated in FOLIO after a live sync run.

- The mapping corrections (material type source field, local identifier) are behaviour changes — records already in FOLIO may receive updated values on the next sync.
- Import paths internal to the package have changed — any other module referencing the old flat paths (`mapping.py`, `upsert.py`, `mapper.py`, `folio_callables.py`, `sync_to_folio.py`) would break, but all such references are within this package and have been updated.


catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/config.py
```diff
@@ -0,0 +1,181 @@
...
+    label="identifier type",
+)
+
+# The full AxC-source table. Resolved rows are the named constants above; the
+# plain rows below are extracted straight onto the CanonicalRecord.
+FIELDS: tuple[FieldMap, ...] = (
+    FieldMap("source_id", marc="001"),  # Axiell GUID — identifies the record
+    FieldMap("title", marc="245$a"),  # → instance.title
+    FieldMap(
+        "object_number", marc="035$a"
```
kenoir: I don't think 035$a can be extracted with a plain first-value spec. The XSLT prefixes every 035 $a it emits: object_number as "(AltRefNo)value", related accession numbers as "(accession number)value", and alternative numbers as "(type)value", which includes Sierra bib numbers. extract() takes the first non-empty 035 in whatever order Adlib serialised the source elements, so a record with several 035s can get an accession or Sierra number as its "Local identifier", a record with no object_number at all gets one it shouldn't have, and even the right pick keeps the "(AltRefNo)" prefix. The test asserts a bare "AxC-12345" but builds CanonicalRecord directly, so extraction is never exercised. other_identifiers.py in the works transformer already parses these prefixes (partition on the closing paren, "AltRefNo" is in its table), so I think this should filter the 035s to the "(AltRefNo)" one and strip the prefix, the same way.

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/config.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+# ── normalization tables & defaults ─────────────────────────────────────────────
+
+# Axiell object_category (MARC 655$a) → FOLIO standard material-type name.
+# Keys carry the AxC object_category value in its real (mixed) case for readability;
+# matching is case-insensitive — MATERIAL_TYPE_FIELD folds these keys to lowercase
+# and `_resolve` lowercases the incoming value before lookup.
+# Values reflect the AxC object_category mappings agreed in the Folio Axiell Mapping.
+MATERIAL_TYPE: dict[str, str] = {
```
kenoir, line 36: Are "archive", "film", "audio format requestable" and "non-projected graphic" all defined as material types in our FOLIO tenant? They aren't FOLIO's shipped names, and wellcomecollection/platform#6462 proposed mapping to the standard set (sound recording, still image, unspecified, ...). Also, the table only covers the four Non Digital / Non-digital pairs: any other object_category (the issue lists digital and hybrid variants, 3-D Objects, Published Material, Sound only, ...) passes through raw to the tenant lookup and fails the record with MappingError, since the "book" default only applies when 655 $a is empty. Worth confirming against the current mapping spreadsheet and the tenant before this runs live.

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/config.py
```diff
@@ -0,0 +1,181 @@
...
+DEFAULT_LOAN_TYPE = "Can circulate"
+DEFAULT_LOCATION = "History of Medicine"
+DEFAULT_HOLDINGS_SOURCE = "MARC"
+AXIELL_LOCATION_NOTE_TYPE = "Axiell location"
+# FOLIO instance identifier type used for the AxC object_number (035$a).
+LOCAL_IDENTIFIER_TYPE = "Local identifier"
+
+# AxC current_location (MARC 852$b) values whose leading digits map to a fixed
+# FOLIO location, overriding the normal code/name lookup. First match wins.
+LOCATION_PREFIX_OVERRIDES: dict[str, tuple[str, ...]] = {
```
kenoir, line 59: wellcomecollection/platform#6462 has 852 $b values starting 215/183 going to a new Axiell location on the Euston Road campus, and "Deepstore" mapping to "hicon". This table sends 215/183 to "hicon" instead, and Deepstore or CLW values fall through to the normal tenant lookup, which raises MappingError unless the tenant knows them (the "History of Medicine" default only applies when 852 $b is empty). Has the spreadsheet moved on from the issue, or did the two rules get swapped here?

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/builders.py
```diff
@@ -0,0 +1,208 @@
...
+            instance_hrid=_instance_hrid(rec.source_id),
+            holdings_hrid=_holdings_hrid(rec.source_id),
+            item_hrid=_item_hrid(rec.source_id),
+            mapping_version=VERSION,
+            deleted=deleted,
+        ),
+    )
+
+
+def is_selected_for_sync(xml_content: str) -> bool:
```
kenoir, line 175: Removing the 980 $a harvest-flag gate changes record selection from the RFC 090 two-condition rule (harvest flag AND item-level) to item-level only, so every ITEM record in the collect set will sync to FOLIO on each adapter run, not just the ones CI opted in. Is this intentional?

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/marc.py
```diff
@@ -27,19 +27,12 @@ class CanonicalRecord:
...
-    call_number_prefix: str | None = None
-    shelving_order: str | None = None
-    barcode: str | None = None
-    material_type_code: str | None = None
-    loan_type_code: str | None = None
-    copy_number: str | None = None
-    volume: str | None = None
-    electronic_access_uri: str | None = None
+    title: str | None = None  # 245$a
+    object_number: str | None = None  # 035$a — AxC local identifier
```
kenoir: This drops call_number (852$h), call_number_prefix (852$c), shelving_order (852$j), copy_number (876$p), volume (876$t) and electronic_access_uri (856$u) from the mapping, but the PR description doesn't mention it, and wellcomecollection/platform#6462 lists call numbers among the currently mapped fields with the ask being to add fields, not remove them. Deliberate per the current spreadsheet? One thing to be aware of either way: updates merge the new payload over the existing record and the payloads exclude None values, so records already synced keep their old callNumber etc. rather than having them cleared, which leaves the two populations inconsistent. Also, the issue's table has access_status (949) as a CI-required field and I don't see it added here; is that deferred?

kenoir (commented):
The restructure itself looks good. I checked the moves against main: the old flat modules are properly deleted, mapper/ref_cache/folio_callables/sync_to_folio are clean renames, and the upsert split is faithful to the old code apart from log formatting. Tests, mypy and ruff are all green on the branch, and the Lambda entrypoint referenced from terraform is unaffected. Splitting the outcome models into results.py is a nice untangling of the old mapping/upsert dependency.

My concerns are all with the mapping changes that ride along with the restructure, which I've raised as inline comments: the removal of the 980 $a harvest-flag gate, the 035 $a local-identifier extraction, the material-type table, the location override, and the dropped holdings/item fields. Several of these diverge from wellcomecollection/platform#6462, and I can't tell from the PR which divergences are the spreadsheet moving on and which are accidental. I think the behaviour changes need to be confirmed against the current CI mapping spreadsheet (and the tenant reference data) before this merges, and the PR description updated to list them, since as it stands it reads mostly as a restructure.

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/config.py
anubhavijay: These mappings have been given but hasnt been added correctly on the Folio by the CI teams/migration, this is task for @tfolami in https://github.com/wellcomecollection/platform/issues/6251 to get the locations added so that the correct rules can be used here. This will have to be fixed in subsequent PRs

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/builders.py
anubhavijay: This is intentional for now but will be added back in subsequent PRs specially before we move from dry run to live. Have captured this in the docstring

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/marc.py
anubhavijay: call_numbers have been removed, I will update 6462 to match this and also add the spreadsheet there. Access Status again needs to be still figured out and this has been added to the list of things for @tfolami for https://github.com/wellcomecollection/platform/issues/6251

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/config.py
anubhavijay: These are the mappings given by CI, rest needs to be removed.
image

anubhavijay:
The mapping are still very much WIP so this PR has added  the mappings which we have so far, we would need to have subsequent PRs  as we get these and then defaults can be removed before we run these without dry run on dev/test instances. The current spreadsheet which is used with CI teams has been added to the ticket. I have also added a mapping doc in docs folder which can be kept upto date with whats in the code https://github.com/wellcomecollection/catalogue-pipeline/blob/av/restructure_mapping_axc_folio/catalogue_graph/docs/axiell-folio-field-mappings.md

catalogue_graph/src/adapters/steps/axiell_folio_sync/mapping/builders.py
kenoir: Ok, that makes sense. Have we got a timeline for the flag to appear in the migration? Are we expecting it this Friday?

kenoir (approved):

PR #3553, opened 2026-08-07
Allow the transformer to process records from a list of ids

### LINKED ISSUE #6498: Allow the transformer state machine to be triggered with record ids
The transformer state machine (`transformer-<pipeline_date>`) can only be told what to transform by changeset id. That works when we are replaying a known adapter run, but not when we have a list of records we want re-transformed: a set of works with a bad field, a handful of documents missing from the source index, or a spot fix after a transformer change. Today the only options are to find the changesets those records happened to arrive in, or to run a full reindex.

The OAI-PMH loader already has an id mode for the equivalent problem on the harvest side (`adapters.steps.oai_pmh.loader`, selected by an `ids` field on the event). This issue asks for the same affordance one stage later, so we don't need to refresh from source to redrive.

## Identifiers in the adapter

The adapter store record id: the OAI identifier for Axiell and FOLIO, the MARC 001 for EBSCO. This is the same id space the loader's id mode uses, so a recovery can hand ids straight from one step to the next.

## Scope

- `TransformerEvent` gains an `ids: list[str]` field. Mutually exclusive with `changeset_ids`: supplying both should be rejected rather than silently resolved. An explicitly supplied empty list should also be rejected, so an id query that came back empty fails instead of falling through to a full reindex.
- `AdapterStoreSource.stream_raw` gains a third branch that reads `get_namespace_records(In("id", ids), snapshot_id)`. Use the namespace read, not `get_active_namespace_records`: soft-deleted rows must be included so tombstones overwrite live documents, matching the changeset path.
- Thread `ids` through `build_transformer`, `SourceWorkTransformer`, and `AxiellChangesetReader` so all three adapters are covered. The FOLIO items join is already keyed by id per batch and should need no change.
- `TransformerReport` records the ids and stops labelling id runs as `reindex` in the S3 key. The loader's `idload-` job id prefix is a reasonable convention to follow.
- CLI: `--id` (repeatable) and `--ids-file`, matching the loader's `--ids` / `--ids-file`.

No Terraform change is expected. The state machine passes `$.detail` through to the Lambda unaltered, so an operator starts an id run the same way they start a changeset redrive:

```
aws stepfunctions start-execution \
  --state-machine-arn <transformer-YYYY-MM-DD> \
  --input '{"detail": {"transformer_type": "axiell", "job_id": "idload-...", "ids": ["..."]}}'
```

## Constraints worth knowing before starting

- Step Functions caps state payload at 256 KB, which is roughly 5,000 to 8,000 ids depending on id length. Validate against an explicit ceiling below that with a clear error, rather than letting the execution fail at the service limit. Do not reuse the loader's `MAX_IDS_PER_RUN = 50_000`; it is above what an execution input can carry. Reading ids from an S3 object would lift the ceiling and is a reasonable follow-up, but is not needed for the first cut.
- The adapter store is sorted on `id` precisely so id lookups prune Parquet row groups, so this read should be cheaper than the changeset read it sits alongside.
- The per-adapter `failure_count` alarm fires on transform failures regardless of how the run was started, so a hand-driven id run with bad records will page. Acceptable, but worth a note in the runbook.

## Done when

- An id run transforms exactly the named records into the pipeline's source works index, including soft-deleted rows as tombstones.
- Ids and changesets cannot be combined, and an empty id list is an error.
- Reports for id runs are distinguishable from reindex runs in S3.
- Tests cover the three-way source branch and the event validation.
- `catalogue_graph/src/adapters/transformers/README.md` documents the new event shape, the CLI flags, and the deletion-facts caveat.

annak-k:
## What does this change?

Closes wellcomecollection/platform#6498, which lays out the problem, the necessary changes, and the desired behaviour in detail.

In short: the transformer state machine could previously only be told what to transform by changeset id. That's a problem when a specific, known set of records needs re-processing: a set of works with a bad field, a handful of documents missing from the source index, or a spot fix after a transformer parsing change. This PR gives the transformer an `ids` field, mirroring the id mode the OAI-PMH loader already has one stage earlier (`adapters.steps.oai_pmh.loader`), so a known set of records can be redriven directly by adapter store record id — without needing to refresh from source or reindex everything.

**Identifiers**: `ids` means the adapter store record id. For Axiell and FOLIO, that's the OAI-PMH header identifier — the same value the OAI-PMH loader's own id mode (`adapters.steps.oai_pmh.loader`) already operates on, so a recovery there can hand ids straight from one step to the next without translation. For EBSCO, it's the record's MARC 001, falling back to a cleaned MARC 035 `$a` if 001 is blank.

**What changed, by area**:

- `TransformerEvent` (`adapters/steps/transformer.py`) gains `ids: list[str] | None`. Mutually exclusive with `changeset_ids` — supplying both raises `ValueError`. An explicitly empty list is also rejected, so an id query that came back empty fails loudly instead of silently falling through to a full reindex. New `MAX_IDS_PER_RUN` constant, derived from the Step Functions 256 KiB state payload limit — not the loader's separate, much larger constant of the same name, which would exceed what an execution input can carry.
- `AdapterStoreSource.stream_raw` (`adapters/utils/adapter_store_source.py`) gains a third branch: with `ids` set, it reads `get_namespace_records(In("id", ids), snapshot_id)` — the same namespace-scoped read the changeset path uses, so soft-deleted rows are included and can overwrite live documents downstream. The adapter store is sorted on `id`, so this prunes Parquet row groups rather than scanning the table.
- `build_transformer` threads `ids` through to all three concrete transformers. `EbscoTransformer` and `FolioTransformer` pass it straight into `AdapterStoreSource`/`FolioStoreSource`; FOLIO's item-enrichment join needed no changes, since it was already keyed by id per batch. `AxiellTransformer` goes through `AxiellChangesetReader` instead — both `build()` and `iter_records()` now accept and thread `ids`.
- `TransformerReport` records the `ids` list, and the S3 manifest key now uses an `idload__<job_id>` prefix for id runs — distinct from a changeset run's joined-changeset-id key and a full reindex's `reindex__<job_id>` key.
- `catalogue_graph/src/adapters/transformers/README.md` documents the new event shape, the CLI flags, and the deletion-facts caveat below.

**Deliberately out of scope, per the issue**: Axiell's deletion-fact delivery (superseded-GUID tombstones, via `AxiellChangesetReader.iter_deletions()`) does not extend to id-mode. Facts are recorded and read per-changeset; there's no equivalent per-id query, and id-mode's use cases (a bad field, a spot fix after a transformer change) don't need historical-deletion replay — those are already delivered by the changeset-driven run that originally detected them.

### Checklist

- [x] Does this change the display model the catalogue API returns? No.

## How to test

From `catalogue_graph/`: `uv run pytest tests/adapters` (1049 passing), plus `mypy .` and `ruff check`/`ruff format`, all clean.

Checked directly against the issue's "done when" list:

- **An id run transforms exactly the named records, including soft-deleted rows as tombstones** — covered end-to-end for all three transformer types (`test_transformer_id_run_transforms_only_named_ids`, `test_transformer_id_run_includes_deleted_rows_as_tombstones` in each of `tests/adapters/transformers/{axiell,ebsco,folio}/test_transformer.py`).
- **Ids and changesets cannot be combined, and an empty id list is an error** — `tests/adapters/transformers/test_transformer_event.py` (7 tests), plus `AdapterStoreSource`'s own independent check.
- **Reports for id runs are distinguishable from reindex runs in S3** — the `idload__<job_id>` manifest key, tested per adapter (`test_transformer_id_run_report_key_is_not_reindex`).
- **Tests cover the three-way source branch and the event validation** — `tests/adapters/utils/test_adapter_store_source.py` (11 tests) and `test_transformer_event.py`.
- **README documents the event shape, CLI flags, and the deletion-facts caveat** — done in `catalogue_graph/src/adapters/transformers/README.md`.

Operationally, an operator recovering a known set of bad/missing records can now invoke the transformer directly with `ids` instead of hunting down the changesets those records arrived in or running a full reindex:

- CLI: `uv run python -m adapters.steps.transformer --transformer-type <type> --id <id> [--id <id> ...]`, or `--ids-file <path>` for a file of ids, one per line. The two combine with each other if both are given, but both are mutually exclusive with `--changeset-id`.
- Lambda / state machine: pass `ids` instead of `changeset_ids` in the input. No Terraform change is needed — the state machine already passes `$.detail` through to the Lambda unaltered, so an id run is started the same way as a changeset redrive:

  ```bash
  aws stepfunctions start-execution \
    --state-machine-arn <transformer-YYYY-MM-DD> \
    --input '{"detail": {"transformer_type": "axiell", "job_id": "idload-...", "ids": ["..."]}}'
  ```

From the issue's constraints:

> The per-adapter `failure_count` alarm fires on transform failures regardless of how the run was started, so a hand-driven id run with bad records will page. Acceptable, but worth a note in the runbook.

The runbook update itself isn't part of this PR.

### DIFF
--- a/catalogue_graph/src/adapters/utils/adapter_store_source.py
+++ b/catalogue_graph/src/adapters/utils/adapter_store_source.py
@@ -2,6 +2,7 @@ from collections.abc import Generator
 from typing import Any
 
 import structlog
+from pyiceberg.expressions import In
 
 from adapters.utils.adapter_store import AdapterStore
 from core.source import BaseSource
@@ -21,10 +22,17 @@ class AdapterStoreSource(RecordSource):
         adapter_store: AdapterStore,
         changeset_ids: list[str],
         snapshot_id: int | None = None,
+        ids: list[str] | None = None,
     ):
+        if changeset_ids and ids:
+            raise ValueError(
+                "changeset_ids and ids are mutually exclusive; supply one or "
+                "the other, not both"
+            )
         self.adapter_store = adapter_store
         self.changeset_ids = changeset_ids
         self.snapshot_id = snapshot_id
+        self.ids = ids
 
     def stream_raw(self) -> Generator[dict[str, Any]]:
         if self.changeset_ids:
@@ -33,6 +41,16 @@ class AdapterStoreSource(RecordSource):
             # whole (possibly multi-changeset) table alongside the Arrow copy.
             table = self.adapter_store.get_records_by_changesets(
                 self.changeset_ids, self.snapshot_id
+            )
+            for batch in table.to_batches():
+                yield from self._process_rows(batch.to_pylist())
+        elif self.ids:
+            # Includes soft-deleted rows, for the same reason as the changeset
+            # path: tombstones must overwrite live documents downstream. The
+            # store is sorted on id, so this filter prunes row groups rather
+            # than scanning the whole namespace.
+            table = self.adapter_store.get_namespace_records(
+                In("id", self.ids), self.snapshot_id
             )
             for batch in table.to_batches():
                 yield from self._process_rows(batch.to_pylist())
--- a/catalogue_graph/src/adapters/steps/transformer.py
+++ b/catalogue_graph/src/adapters/steps/transformer.py
@@ -141,20 +141,13 @@ def build_transformer(
 
     snapshot_id = event.snapshot_id or adapter_store.current_snapshot_id()
 
-    if event.ids and event.transformer_type == "axiell":
-        # Not yet wired: AxiellChangesetReader doesn't thread ids through to
-        # its source, so falling through here would silently ignore ids and
-        # read every active record instead.
-        raise ValueError(
-            f"id-mode transforms are not yet supported for {event.transformer_type!r}"
-        )
-
     if event.transformer_type == "axiell":
         reader = AxiellChangesetReader.build(
             AXIELL_CONFIG,
             event.changeset_ids,
             use_rest_api_table=use_rest_api_table,
             snapshot_id=snapshot_id,
+            ids=event.ids,
             adapter_store=adapter_store,
         )
         return AxiellTransformer(reader)
--- a/catalogue_graph/src/adapters/steps/transformer.py
+++ b/catalogue_graph/src/adapters/steps/transformer.py
@@ -141,10 +141,10 @@ def build_transformer(
 
     snapshot_id = event.snapshot_id or adapter_store.current_snapshot_id()
 
-    if event.ids and event.transformer_type in ("axiell", "folio"):
-        # Not yet wired: AxiellChangesetReader and FolioTransformer don't
-        # thread ids through to their sources, so falling through here would
-        # silently ignore ids and read every active record instead.
+    if event.ids and event.transformer_type == "axiell":
+        # Not yet wired: AxiellChangesetReader doesn't thread ids through to
+        # its source, so falling through here would silently ignore ids and
+        # read every active record instead.
         raise ValueError(
             f"id-mode transforms are not yet supported for {event.transformer_type!r}"
         )
@@ -173,7 +173,11 @@ def build_transformer(
             use_rest_api_table=use_rest_api_table, create_if_not_exists=False
         )
         return FolioTransformer(
-            adapter_store, event.changeset_ids, snapshot_id, items_store=items_store
+            adapter_store,
+            event.changeset_ids,
+            snapshot_id,
+            ids=event.ids,
+            items_store=items_store,
         )
     raise ValueError(f"Unknown transformer type: {event.transformer_type}")
 
--- a/catalogue_graph/src/adapters/steps/transformer.py
+++ b/catalogue_graph/src/adapters/steps/transformer.py
@@ -11,7 +11,7 @@ from collections.abc import Callable
 from typing import Any, Literal, Protocol, cast
 
 import structlog
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 from adapters.extractors.ebsco import config as ebsco_config
 from adapters.extractors.ebsco import helpers as ebsco_helpers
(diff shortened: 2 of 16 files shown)


catalogue_graph/src/adapters/utils/adapter_store_source.py
```diff
@@ -36,6 +44,16 @@ def stream_raw(self) -> Generator[dict[str, Any]]:
...
             for batch in table.to_batches():
                 yield from self._process_rows(batch.to_pylist())
+        elif self.ids:
+            # Includes soft-deleted rows, for the same reason as the changeset
+            # path: tombstones must overwrite live documents downstream. The
+            # store is sorted on id, so this filter prunes row groups rather
+            # than scanning the whole namespace.
+            table = self.adapter_store.get_namespace_records(
+                In("id", self.ids), self.snapshot_id
+            )
```
kenoir, line 54: IDs that match no store row vanish silently here: the run succeeds, and the report can't tell you which ids were missed, because `successful_ids` holds ES document ids (`Work[axiell-guid/...]`), not adapter record ids. An operator recovering a known list (the main use case in the issue) has no way to spot a typo'd or genuinely-absent id short of diffing counts, and duplicate requested ids would confound even that, since they aren't deduped.

The loader's id mode handles both: `execute_id_loader` dedupes with `dict.fromkeys` and reports `requested`/`recovered`/`removed` per run. Would it be worth mirroring that here, even minimally: dedupe `ids` in the validator or `_read_ids`, track the distinct row ids actually yielded, and log a warning (or record the unmatched ids in the report) when some requested ids matched nothing? The empty-list rejection shows the design already cares about "the query came back empty" failing loudly; this is the same failure mode one notch down.

kenoir (approved):
This is a tidy implementation of wellcomecollection/platform#6498: the event validation is defence-in-depth (model validator plus an independent check in `AdapterStoreSource`), the deletion-facts caveat is both documented and locked in by the test that fails if the facts store is touched in id mode, and the sorted-on-id pruning claim checks out against the table configs. One small suggestion inline about making unmatched ids visible to the operator; nothing blocking.

PR #3557, opened 2026-08-10
Delete TEI inner works when their parent is deleted

kenoir:
## What does this change?

Fixes the parent-deletion half of wellcomecollection/platform#6471. Deliberately not a closing reference: the rename half needs a further change, described under Scope below, so the issue should stay open.

TEI manuscripts produce inner works from `msItem` and `msPart` elements. They are not source records: the merger synthesises them on every merge from the parent's `internalWorkStubs`. Nothing has ever deleted them, so deleting a parent left every child behind, still advertising a `partOf` pointing at a work the API now returns 410 for.

Confirmed live on 2026-08-10 while piloting the retraction in wellcomecollection/platform#6504: deleting `MS_197` took the parent `aknvf9qw` to 410 Gone and left `akpvxvj5` at 200. Retracting that folder would strand 462 such works, several of them phantom duplicates of the record they were meant to describe.

A deletion has no XML to parse, so the stubs are only knowable from the previously stored work. `TransformerWorker` already retrieves that work on every message, to decide whether re-sending would be a no-op, so the change is:

- `Transformer` gains a `reconcileWithStored` hook, concrete and defaulting to a no-op, so no other source is affected and the SAM conversion in the Sierra transformer's `Main` still compiles.
- `TeiTransformer` overrides it to carry the stored stubs onto a deleted work whose own stub list is empty.
- `Merger.merge` wraps the existing body and turns a deleted work's stubs into `Work.Deleted` children, deduplicated against what the merge already emitted.

The `isEmpty` guard on the override is what makes replays stable: once a reconciled delete is stored, reconciling again produces the same document, so `areEquivalent` holds and nothing resends in a loop.

Downstream needs no change. The id minter walks any node carrying a `sourceIdentifier` with no knowledge of `internalWorkStubs`, and minting is lookup-or-create, so deleted children keep the canonical ids they already have. Work nodes reach the graph from a source filtered on `{"match": {"type": "Visible"}}`, and `CatalogueWorkIdentifiersGraphRemover` drops path identifiers once disconnected, so the graph sheds the children on its own.

### Checklist

- [x] Does this change the display model the catalogue API returns? No. It changes which works are deleted.

## How to test

`sbt "transformer_tei/test" "transformer_common/test" "merger/test"`, with `docker compose up` in `pipeline/transformer/transformer_common` and `pipeline/matcher_merger/merger` first. All three suites pass: 165, 58 and 256 tests, the last including `MergerIntegrationTest`.

The new tests were checked against the mutations they exist to catch, rather than assumed to work:

- Changing `Right(Some((reconciledWork, key)))` back to `transformedWork` in `TransformerWorker` makes "indexes the work returned by reconcileWithStored, not the transformed one" fail. Without that test the substitution is silent: every TEI delete would ship with `internalWorkStubs: []` and every other test still passes.
- Removing the `isEmpty` guard from the override makes "keeps the stubs it already has rather than taking the stored ones" fail.

`PlatformMergerTest` keeps its existing "returns both Works unmodified if one of the Works is deleted", which is the guard that a stub-free deleted work still emits nothing extra.

## Scope

Only the parent-deletion case. The rename case in the same issue, where an `msItem` id changes and the old inner work is orphaned alongside the new one, needs somewhere to record the stubs that have gone, which means a new field on `WorkState.Source` and `WorkState.Identified`. That would put every stored `works-source` and `works-identified` document for all five Scala sources behind an assumption about circe `useDefaults` that I have not verified, so it belongs in its own change.

## Risks

`TransformerWorker` is shared by Sierra, Calm, METS, Miro, EBSCO and Axiell. The hook defaults to returning the new work unchanged, so those sources see no behaviour change, which the existing suites cover.

The one behaviour change beyond TEI is in the retriever's `recover` block. A genuine retrieval failure now logs at `warn` with the exception attached, and if the transformer says it needs the stored work, the message fails and dead-letters rather than being sent in a state we know to be wrong. Only TEI says that, and only for a deletion carrying no stubs of its own, so the other sources still send as before. Not-found is unchanged and still sends, since it fires on the first sighting of every record.

That guard exists because the failure was worse than a missed deletion. `batchIndexAndSendFlow` indexes before it sends, and `ElasticIndexer` writes with `ExternalGte` versioning, so a stub-free delete overwrites the stored work that held the stubs. A replay then reconciles against a delete with nothing in it and produces the same document, leaving the record exactly as the bug leaves it today. Dropping the deletion is permanent; dead-lettering the message is one redrive. Raised by @agnesgaroux in review.

### DIFF
--- a/pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
+++ b/pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
@@ -33,6 +33,8 @@ case class TransformerError[SourceData, Key](
   sourceData: SourceData,
   key: Key
 ) extends TransformerWorkerError(t.getMessage)
+case class StoredWorkRetrievalError[Key](t: Throwable, key: Key)
+    extends TransformerWorkerError(t.getMessage)
 
 trait SourceDataRetriever[Payload, SourceData] {
   def lookupSourceData(
@@ -104,10 +106,12 @@ trait TransformerEventProcessor[Payload <: SourcePayload, SourceData]
             case err: RetrieverNotFoundException =>
               debug(s"No stored work for $key: $err")
               Right(Some((transformedWork, key)))
-            // A work that needs the stored one is left to fail, so the message
-            // is retried rather than sent in a state we know to be wrong.
+            // A work that needs the stored one fails, so the message is
+            // retried rather than sent in a state we know to be wrong.
             case err: Throwable
-                if !transformer.requiresStoredWork(transformedWork) =>
+                if transformer.requiresStoredWork(transformedWork) =>
+              Left(StoredWorkRetrievalError(err, key))
+            case err: Throwable =>
               warn(
                 s"Unable to retrieve work $key, sending without reconciliation",
                 err
@@ -255,6 +259,11 @@ final class TransformerWorker[Payload <: SourcePayload, SourceData, SenderDest](
               case TransformerError(t, sourceData, key) =>
                 error(
                   s"$transformerName: TransformerError on $sourceData with $key ($t)"
+                )
+              case StoredWorkRetrievalError(t, key) =>
+                error(
+                  s"$transformerName: StoredWorkRetrievalError on $key; this work cannot be sent without the stored work to reconcile against",
+                  t
                 )
             }
 
--- a/pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
+++ b/pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
@@ -11,7 +11,11 @@ import weco.catalogue.internal_model.Implicits._
 import weco.json.JsonUtil.fromJson
 import weco.messaging.sns.NotificationMessage
 import weco.pipeline_storage.Indexable._
-import weco.pipeline_storage.{PipelineStorageStream, Retriever}
+import weco.pipeline_storage.{
+  PipelineStorageStream,
+  Retriever,
+  RetrieverNotFoundException
+}
 import weco.storage.{Identified, ReadError, Version}
 import io.circe.optics.JsonPath._
 import weco.typesafe.Runnable
@@ -84,8 +88,10 @@ trait TransformerEventProcessor[Payload <: SourcePayload, SourceData]
           .apply(workIndexable.id(transformedWork))
           .map {
             storedWork =>
-              if (shouldSend(transformedWork, storedWork)) {
-                Right(Some((transformedWork, key)))
+              val reconciledWork =
+                transformer.reconcileWithStored(transformedWork, storedWork)
+              if (shouldSend(reconciledWork, storedWork)) {
+                Right(Some((reconciledWork, key)))
               } else {
                 info(
                   s"$transformerName: from $key transformed work with id ${transformedWork.id}; already in pipeline so not re-sending"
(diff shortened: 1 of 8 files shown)


pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
```diff
@@ -94,8 +100,14 @@ trait TransformerEventProcessor[Payload <: SourcePayload, SourceData]
               }
           }
           .recover {
+            // Not-found is the normal first sighting of a record, so it stays quiet.
+            case err: RetrieverNotFoundException =>
+              debug(s"No stored work for $key: $err")
+              Right(Some((transformedWork, key)))
```
agnesgaroux, line 108: Are we sending the non-reconciled work downstream if the storedWork is not found? Would that silently drop the workStub deletion?

agnesgaroux: mmh that would be blocking in a reindex scenario :/

kenoir: Good question, and the two branches turn out to differ, which is also the answer to the reindex worry.

Not-found is safe, and you are right that failing it would block a reindex. The retriever and the indexer are built over the same works-source index, so not-found means this pipeline has never stored the parent, and children only exist because a stored parent carrying stubs went through the merger. There is nothing to delete. A reindex into a fresh pipeline recreates neither the parent nor the children, since a currently-deleted TEI record produces a delete with no visible parent. So that path keeps sending, unchanged.

The second branch, a genuine retrieval error, is a real hole and it does drop the deletion. Worse, it does not self-heal: `batchIndexAndSendFlow` indexes before it sends and `ElasticIndexer` writes with `ExternalGte`, so the stub-free delete overwrites the stored work that held the stubs, and a replay then reconciles against a delete with nothing in it and produces the same document. The record ends up exactly as the bug leaves it today, orphaned children pointing at a 410 parent, until a reindex or a manual cleanup. A whole-cluster outage is not the case to worry about, since the indexer writes to the same cluster and a failed write leaves the message unacked to be retried. What is left is a single failed mget, a timeout, or a stored document that will not decode.

So I have made that path fail rather than log, in da3ecaf. The guard is a default-false method on `Transformer` that only TEI overrides, and only for a deletion with no stubs of its own, so not-found still sends, reindexes are unaffected, and the other six sources are untouched. A dropped deletion is permanent, whereas a dead-lettered message is one redrive.

Thank you for pushing on this.

pipeline/transformer/transformer_common/src/main/scala/weco/pipeline/transformer/TransformerWorker.scala
```diff
@@ -94,8 +100,14 @@ trait TransformerEventProcessor[Payload <: SourcePayload, SourceData]
...
             case err: Throwable =>
-              debug(s"Unable to retrieve work $key: $err")
+              warn(
+                s"Unable to retrieve work $key, sending without reconciliation: $err"
+              )
```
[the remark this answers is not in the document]
kenoir: Applied in 3a6e66b. The two-argument form is what the rest of the codebase uses, for example `TeiLanguages.scala:73`.

agnesgaroux (approved):
Works because it bubbles up to the outer Left but would be better with its own error type, eg. 
```
Left(StoreRetrievalError(err, key)) 
``` 
For the case where transformer.requiresStoredWork(transformedWork) is true

kenoir:
Applied in 24cc7a9, but named `StoredWorkRetrievalError`, as `StoreRetrievalError` reads too close to the existing `StoreReadError` (which is about the source-data store, not the stored work).

One correction on the mechanism: it does not reach the outer `Left` today. The `recover` block has no case for this combination, so the partial function is undefined, the Future stays failed with the raw exception, and `run()`'s per-error logging never runs. I checked by asserting on `processEvent` before the change: it failed with `java.lang.RuntimeException: BOOM!` rather than returning a `Left`. So the cost is the log, which comes out as "Unrecognised failure while: ..." with no key. The message dead-letters either way, so behaviour is unchanged.

PR #3562, opened 2026-08-11
Retire works-indexed-2026-03-03 from the 2025-10-02 stack

kenoir:
## What does this change?

Removes the works entry from the `2026-03-03` index config on the 2025-10-02 stack, which destroys `works-indexed-2026-03-03`. Its concepts index stays, because that is still the production concepts index. The `2026-07-30` comment is updated to describe current state rather than a future prod index. Part of wellcomecollection/platform#6511, following the switch in wellcomecollection/platform#6509.

Merge only after the new index has soaked. Target is 2026-08-18, one week after the switch. All five of these should hold first:

- [x] Seven days of green incremental windows on the 2025-10-02 stack, with failures unrelated to the index not counting against it -> one failure on August 15, internal server error at the graph load stage. Unrelated
- [x] No API error-rate increase or user-reported regressions attributable to works queries since the 2026-08-11 flip
- [x] Rank tests passing on every catalogue-api deploy since the flip
- [x] No open question that would need `works-indexed-2026-03-03` for rollback
- [x] Fresh spot check that `works-indexed-2026-07-30` still tracks `works-denormalised` counts

## How to test

Applying takes two passes, because deletion protection has to come off before the destroy can run:

1. On current main, before merging, set `allow_delete_indices = true` locally in `pipeline/terraform/2025-10-02/main.tf` and `terraform apply`, so the index resources drop `deletion_protection` in place.
2. Merge this PR and apply again. The destroy of `works-indexed-2026-03-03` now proceeds.
3. Revert `allow_delete_indices` to `false` locally and apply once more, or fold that into a follow-up commit.

The alternative is deleting the index with `cleanup_unused_api_indexes.py` in catalogue-api and reconciling terraform with `terraform state rm`. The choreography above keeps everything terraform-native.

Deletion is unrecoverable, so rollback past that point means a fresh full re-ingest into a recreated index. The API stopped reading the index on 2026-08-11 and nothing has written to it since the repoint on 2026-08-10.


agnesgaroux:
NOTE: applying this will also destroy "works-source-2026-01-12" which should have gone with https://github.com/wellcomecollection/catalogue-pipeline/pull/3552 but I forgot to disable index deletion protection first

agnesgaroux:
get works-indexed-2026-07-30/_count
// 3205869
get works-denormalised-2025-10-02/_count
// 3205875
Close enough

agnesgaroux (approved):

PR #3565, opened 2026-08-12
Add workflow syncing test documents to catalogue-api

kenoir:
## What does this change?

Adds `.github/workflows/sync-test-documents.yml`. When a push to `main` changes `catalogue_graph/document_generators/test_documents/`, it opens a pull request against `wellcomecollection/catalogue-api` updating its copy of the fixtures under `common/search/src/test/resources/test_documents/`, replacing manual runs of that repo's `copy_test_documents.py`.

Because CI here already regenerates the fixtures and fails on an uncommitted diff, every display model change produces a fixture diff on `main`, so this workflow turns every display model change into a catalogue-api PR whose OpenAPI contract tests stay red until `reference/catalogue.yaml` documents the new shape. That closes the manual gap described in wellcomecollection/platform#6514.

Two details differ from the `sync-openapi-spec.yml` / `sync-prismic-types.yml` precedents:

- The copy is `rsync --archive --delete`, so fixtures removed here are also removed downstream. The target directory contains nothing but copies of these files (the file sets are currently identical), so deletion is safe.
- The change check uses `git status --porcelain` rather than `git diff --quiet`, because new fixture files are untracked in the downstream checkout and invisible to `git diff`.

Also updates the PR template checklist and the `document_generators` README to describe the automated flow.

Before the workflow can run it needs setup that isn't in this PR:

- the sync GitHub App (the one used by catalogue-api's `sync-openapi-spec.yml`) needs `catalogue-api` added to its installation's repository access
- this repo needs `SYNC_APP_CLIENT_ID` and `SYNC_APP_PRIVATE_KEY` secrets, matching the values held by catalogue-api

## How to test

Once the secrets are in place, trigger the workflow manually (`workflow_dispatch`) from the Actions tab. With fixtures already in sync it should complete without opening a PR. Then merge any change that regenerates a fixture and check that an `auto-sync/test-documents-*` PR appears on catalogue-api with review requested from digital-platform.

Display model changes surface as catalogue-api PRs within minutes of merging here, instead of waiting for someone to run `copy_test_documents.py`. The spec-drift window described in platform#6514 (for example the `Genre` concept gap) should stop recurring for works and images.

- A red auto-PR on catalogue-api is the designed behaviour when the display model changes shape, not a failure; the PR body says so and points at the spec.
- If two fixture-changing merges land close together the second PR will contain both diffs; the first can be closed.
- The app token is scoped to `catalogue-api` only and the workflow has read-only repo permissions otherwise.


.github/workflows/sync-test-documents.yml
```diff
@@ -0,0 +1,96 @@
...
+              "base": "main",
+              "body": "This PR copies `catalogue_graph/document_generators/test_documents/` from `wellcomecollection/catalogue-pipeline` into `common/search/src/test/resources/test_documents/`, replacing a manual run of `copy_test_documents.py`.\n\nIf `OpenApiSpecResponseTest` fails on this PR, the pipeline display model has changed shape and `reference/catalogue.yaml` needs updating on this branch before merge.\n\n**Triggered by commit:** wellcomecollection/catalogue-pipeline@${{ github.sha }}\n\nSee wellcomecollection/platform#6514 for context."
+            }' | jq -r '.number')
+
+          # Request review from digital-platform team
+          curl -s -X POST \
+            -H "Authorization: Bearer ${{ steps.app-token.outputs.token }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/wellcomecollection/catalogue-api/pulls/$PR_NUMBER/requested_reviewers \
+            -d '{"team_reviewers": ["digital-platform"]}'
```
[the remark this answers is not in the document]
kenoir: Good catch, this was copied from sync-openapi-spec.yml and inherited its failure mode. Replaced both curl calls with a single gh pr create in 13f721ef9b, which fails the step on API errors and requests the team review in the same call.

.github/workflows/sync-test-documents.yml
```diff
@@ -0,0 +1,98 @@
...
+          cd catalogue-api
+
+          # Configure git
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+          # Create branch
+          TIMESTAMP=$(date +%Y%m%d-%H%M%S)
+          BRANCH_NAME="auto-sync/test-documents-$TIMESTAMP"
+          git checkout -b $BRANCH_NAME
```
annak-k: Claude suggests a fixed branch name instead of the timestamp, so the auto-PRs don't stack while one sits open waiting for a `catalogue.yaml` update. `gh pr create` errors when a PR for that head is already open, so the create step would need a `gh pr list --head ... --state open` guard, and committing on top of the branch rather than force-pushing would preserve any spec work already pushed there. The downside is that "Triggered by commit" would then only name the first commit, which a `gh pr comment` on later runs could cover.

kenoir: Implemented in b8aec9af04. The workflow now fetches auto-sync/test-documents and commits on top when it already exists, checking it out before the rsync so change detection is relative to what the open PR already has; gh pr create only runs when gh pr list shows no open PR for that head, and otherwise the run comments the new trigger commit on the open PR.

agnesgaroux (approved):

PR #3566, opened 2026-08-12
Add a fixture covering rarely-populated display fields

kenoir:
## What does this change?

catalogue-api is adding a test that asserts every field documented in its OpenAPI spec appears in at least one of the pipeline-generated test documents (part of wellcomecollection/platform#6514). Fourteen documented field paths are never populated by the deterministic generators, so they sit on a named allowlist there. This PR teaches the generators to populate ten of them, so the allowlist can shrink:

- `Work.physicalDescription`, `Work.referenceNumber`, `Work.createdDate`, `Work.currentFrequency`
- `Item.title`, `Item.note`
- `AccessCondition.terms`, `AccessCondition.note`
- `ProductionEvent.function`
- `Agent.identifiers` (via an identified contributor agent)

Rather than sprinkling fields across many fixtures, it adds one deliberately maximal work, `work.visible.rare-fields.json`, generated by a new `create_work_with_rarely_populated_fields` step. The generators gain pass-through parameters (`physical_description`, `reference_number`, `created_date`, `current_frequency` on works; `title` and `note` on items) with defaults of `None`, and the new step runs last, so the seeded RNG sequence is untouched and no existing fixture changes.

This changes fixture content only, not the display model itself.

The remaining four allowlisted paths, `Period.id`, `Period.identifiers`, `Place.id` and `Place.identifiers`, cannot be exercised from the generators today. The spec only documents Period and Place as components of production events and `createdDate`, and `DisplayWorkTransformer` builds label-only concepts in both places, dropping ids and identifiers even for identified concepts. Surfacing them would be an ingestor behaviour change, out of scope here; the new fixture already carries identified production places, agents and dates, so if the transformer ever passes ids through, the fields will start appearing without further generator work. Once the sync workflow (#3565) is live, this merge will propagate the new fixture to catalogue-api automatically.

### Checklist

No display model change, but the fixture set gains one document, regenerated with both `create_test_work_documents` and `create_test_image_documents`.

## How to test

From `catalogue_graph/`, run `uv run python -m document_generators.create_test_work_documents` and `uv run python -m document_generators.create_test_image_documents` and check `git status` is clean, which is what CI does. Then inspect `document.display` in `test_documents/work.visible.rare-fields.json` and confirm each of the ten fields above is present (display keys are camelCase). `uv run pytest tests/ingestor`, ruff and mypy all pass.

catalogue-api's fixture coverage test can drop the ten fields from its `allowedUnexercised` list after the next fixture copy, leaving only the four transformer-limited Period and Place paths (plus the fields the API splices in or no longer emits).

Low risk: no pipeline runtime code changes, only test document generators and one new fixture. Existing fixtures are byte-for-byte unchanged, so no downstream test in either repo should move.

### DIFF
--- a/catalogue_graph/document_generators/create_test_work_documents.py
+++ b/catalogue_graph/document_generators/create_test_work_documents.py
@@ -33,7 +33,7 @@ from models.pipeline.access_condition import (
 from models.pipeline.access_method import AccessMethod
 from models.pipeline.access_status import AccessStatus, AccessStatusRelationship
 from models.pipeline.collection_path import CollectionPath
-from models.pipeline.concept import Concept, Genre, Period, Subject
+from models.pipeline.concept import Concept, Contributor, Genre, Period, Subject
 from models.pipeline.format import (
     FORMAT_LABEL_MAPPING,
     Audio,
@@ -42,7 +42,7 @@ from models.pipeline.format import (
     Journals,
     Pictures,
 )
-from models.pipeline.id_label import Id, Language
+from models.pipeline.id_label import Id, Label, Language
 from models.pipeline.identifier import Identified, SourceIdentifier
 from models.pipeline.production import ProductionEvent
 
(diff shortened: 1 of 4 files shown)


annak-k (approved):
Looks good to me and all the suggested tests pass.

PR #3568, opened 2026-08-12
Generate deterministic concept test documents for catalogue API tests

kenoir:
## What does this change?

Step 3 of wellcomecollection/platform#6514: the catalogue-api concepts service is gaining response-schema validation against the OpenAPI spec, using pipeline-generated fixtures instead of its current hand-built circular ones. The pipeline already generates deterministic test documents for works and images but had nothing for concepts.

This adds `catalogue_graph/document_generators/create_test_concept_documents.py`, which builds six `concepts.*.json` fixtures through the real `IndexableConcept` Pydantic models, so the fixtures are exactly what the concepts ingestor indexes. The indexer serializes with `exclude_none=True` (`generate_operations` in `src/utils/elasticsearch.py`), the same as works and images, and the fixture writer matches that.

The fixtures cover the fields the new catalogue-api coverage tests assert on:

- `concepts.person`: every field populated, including a description, three identifiers with full identifierType objects, alternative labels, `sameAs`, two `displayImages` IIIF locations (one bare, one carrying a full License with id/label/url, credit, linkText, and an access condition with method, status, terms, and note), and related concepts with `relationshipType` both set (`has_sibling`) and omitted.
- `concepts.organisation`: `foundedBy` (with `relationshipType: has_founder`) and `people`.
- `concepts.genre`: `narrowerThan`, `broaderThan`, and `relatedTopics`, with no description.
- `concepts.place`: identifiers from multiple sources and non-empty `relatedTo`.
- `concepts.period`: a label-derived identifier.
- `concepts.minimal`: empty lists, no identifiers, no description.

Between them, all six concept types (Person, Organisation, Genre, Place, Period, Concept) and all eight `relatedConcepts` lists appear non-empty at least once.

The generator is wired into the `check-test-documents` CI job, which reruns all generators and fails on any diff. Reruns are deterministic via the shared seeded RNG and `freeze_time`; running the works and images generators after this change produces zero diff.

Once catalogue-pipeline#3565 (the test-documents sync workflow) is live, these fixtures propagate to catalogue-api automatically, since the sync copies the whole `test_documents` directory.

### Checklist

## How to test

From `catalogue_graph/`, run `uv run python -m document_generators.create_test_concept_documents` twice and confirm the fixtures do not change. `uv run pytest tests/ingestor`, ruff, and mypy all pass.

The catalogue-api concepts contract tests can validate the published OpenAPI spec against realistic pipeline-generated documents, and fixture drift is caught by the regeneration check in CI.

No production code paths change; this only adds test fixtures, a generator script, and a CI check line. The main risk is fixtures drifting from real indexed documents, which is mitigated by building them through the same Pydantic models and serialization settings the ingestor uses.


catalogue_graph/document_generators/create_test_concept_documents.py
```diff
@@ -0,0 +1,303 @@
...
+        description="an organisation concept with founders and people",
+        doc_id="concepts.organisation",
+    )
+
+
+def create_genre_concept_document() -> None:
+    concept = create_indexable_concept(
+        label="Emu engravings",
+        concept_type="Genre",
+        identifier_sources=["lc-gmgpc"],
```
annak-k: Is `lc-gmgpc` a possible source for a concept? It's not in `ConceptSource`, so maybe `lc-subjects` or `label-derived` here instead.

annak-k (commented):
Fixtures regenerate clean for me: running all three generators leaves `git status` empty on `test_documents/`. Nothing below blocks, two questions and a nitpick.

**Nitpick on the description:** it says the fixtures cover "all six concept types", but `ConceptType` has nine values, with Agent, Meeting and Subject unrepresented. Six looks right, since all nine serialise to an identical shape and three more fixtures would add no field coverage, so maybe make that explicit in the PR description for future readers.

catalogue_graph/document_generators/create_test_concept_documents.py
```diff
@@ -0,0 +1,303 @@
...
+    """Mirrors the concepts transformer, which builds iiif-image locations from weco-authority image URLs."""
+    image_ref = f"V{random_alphanumeric(7).lower()}"
+    return DisplayDigitalLocation(
+        url=f"https://iiif.wellcomecollection.org/image/{image_ref}/full/full/0/default.jpg",
+        locationType=DisplayLocationType.from_id("iiif-image"),
+        accessConditions=[],
+    )
+
+
+def create_licensed_display_image() -> DisplayDigitalLocation:
```
annak-k: Concept images only ever come out with a url, `iiif-image` and empty access conditions, so the license, credit and linkText here can't appear in a real document. The spec points concept `displayImages` at the shared `DigitalLocation` schema, which was written for works, and I realise that leaves the coverage test wanting an example from somewhere. Would a narrower schema for concept images be better, so we're not documenting a license that never appears?

catalogue_graph/document_generators/create_test_concept_documents.py
```diff
@@ -0,0 +1,303 @@
...
+) -> ConceptRelatedTo:
+    return ConceptRelatedTo(
+        label=label,
+        id=random_canonical_id(),
+        relationshipType=relationship_type,
+        conceptType=concept_type,
+    )
+
+
+def create_related_concepts(**overrides: list[ConceptRelatedTo]) -> RelatedConcepts:
```
annak-k, line 66: Nitpick: a mistyped key is dropped silently (`narrowerthan=[...]` gives `narrowerThan: []`), and it'd surface as a spec failure in catalogue-api rather than an error here. Maybe assert the keys are a subset of `RelatedConcepts.model_fields`.

catalogue_graph/document_generators/create_test_concept_documents.py
kenoir: Good catch, it is not. Switched the genre fixture to label-derived in 540b45dd71; I think that fits a genre better than lc-subjects, and lc-subjects is still exercised by the place fixture.

catalogue_graph/document_generators/create_test_concept_documents.py
kenoir: Agreed, documenting a license that can never appear defeats the point of validating the spec against real documents. I have reverted the enriched image in 540b45dd71, so concepts.person carries the bare url, iiif-image locationType and empty accessConditions shape the transformer actually emits.

On the catalogue-api branch I have added a narrow ConceptImage schema (wellcomecollection/catalogue-api@2116e53a) documenting just those fields, with accessConditions declared as an always-empty array, and pointed Concept.displayImages at it. The works-side DigitalLocation is unchanged and still exercised by the works fixtures, and the coverage test passes with the allowlist still empty.

catalogue_graph/document_generators/create_test_concept_documents.py
kenoir: Done in 540b45dd71: create_related_concepts now asserts its keys are a subset of RelatedConcepts.model_fields, so a mistyped key fails here rather than as a spec failure downstream.

agnesgaroux (approved):

PR #3570, opened 2026-08-13
Align mappings.works_indexed.2026-07-30 with deployed index

agnesgaroux:
## What does this change?

Adds the following fields to `index_config/mappings.works_indexed.2026-07-30.json` to match the mappings that were already manually applied to the live index:

**`aggregatableValues`**
- `archiveRoot` (nested)
- `archiveType` (nested)
- `collectionRoot` (nested)

**`filterableValues`**
- `archiveRoot`
- `archiveType`
- `collectionRoot`
- `isArchiveRoot` (boolean)
- `isCollectionRoot` (boolean)

These fields were manually applied to the `works-indexed-2026-07-30` index in the 2025-10-02 pipeline. The `index_config` file was out of sync, causing Terraform plan to show a diff. Since the mappings are strict and the apply would be subtractive, we need to include the defunct properties in the index config. This brings the file into alignment with the deployed state so that `terraform apply` on the 2025-10-02 pipeline is a no-op for this mapping.

~~We'll also want to apply these mappings in the 2026-07-03 pipeline to align it with the 2025-10-02 pipeline.~~
These added mappings properties were relevant at an intermediate stage of development but not part of the final mappings. The 2026-07-03 pipeline and `works-indexed-2026-07-03` index use a separate index config, which reflect the desired state, ie without the properties that were eventually decided against.

## How to test

Run `terraform plan` against the 2025-10-02 pipeline — the `works_indexed.2026-07-30` mapping should show no diff, apart from the expected "false" -> false on the dynamic properties 

Verify the added fields match what `GET works-indexed-2026-07-30/_mapping` returns for the live index.

Run `terraform plan` against the 2026-07-03 pipeline — the `works_indexed.2026-07-03` mapping should show no diff: the correct/final version mappings have already been applied 

`terraform plan` on the 2025-10-02 pipeline produces no mapping diff for `works_indexed.2026-07-30`, except for the "false" -> false (9 indices)
`terraform plan` on the 2026-07-03 pipeline produces no mapping diff for `works_indexed.2026-07-03`.

Low risk — this is a purely additive change to the mapping config, aligning it with what is already deployed. No index reindexing required.

### DIFF
--- a/index_config/analysis.works_indexed.2026-07-03.json
+++ b/index_config/analysis.works_indexed.2026-07-03.json
@@ -1,0 +1,385 @@
+{
+  "filter": {
+    "pattern_replace_g_j": {
+      "pattern": "g",
+      "type": "pattern_replace",
+      "replacement": "j"
+    },
+    "pattern_replace_j_i": {
+      "pattern": "j",
+      "type": "pattern_replace",
+      "replacement": "i"
+    },
+    "french_elision": {
+      "type": "elision",
+      "articles": [
+        "l",
+        "m",
+        "t",
+        "qu",
+        "n",
+        "s",
+        "j",
+        "d",
+        "c",
+        "jusqu",
+        "quoiqu",
+        "lorsqu",
+        "puisqu"
+      ],
+      "articles_case": "true"
+    },
+    "pattern_replace_uu_w": {
+      "pattern": "uu",
+      "type": "pattern_replace",
+      "replacement": "w"
+    },
+    "pattern_replace_vv_w": {
+      "pattern": "vv",
+      "type": "pattern_replace",
+      "replacement": "w"
+    },
+    "hindi_stemmer": {
+      "type": "stemmer",
+      "language": "hindi"
+    },
+    "pattern_replace_v_u": {
+      "pattern": "v",
+      "type": "pattern_replace",
+      "replacement": "u"
+    },
+    "german_stemmer": {
+      "type": "stemmer",
+      "language": "light_german"
+    },
+    "english_stemmer": {
+      "type": "stemmer",
+      "language": "english"
+    },
+    "italian_elision": {
+      "type": "elision",
+      "articles": [
+        "c",
+        "l",
+        "all",
+        "dall",
+        "dell",
+        "nell",
+        "sull",
+        "coll",
+        "pell",
+        "gl",
+        "agl",
+        "dagl",
+        "degl",
+        "negl",
+        "sugl",
+        "un",
+        "m",
+        "t",
+        "s",
+        "v",
+        "d"
+      ],
+      "articles_case": "true"
+    },
+    "asciifolding": {
+      "type": "asciifolding"
+    },
+    "possessive_english": {
+      "type": "stemmer",
+      "language": "possessive_english"
+    },
+    "spanish_stemmer": {
+      "type": "stemmer",
+      "language": "light_spanish"
+    },
+    "arabic_stemmer": {
+      "type": "stemmer",
+      "language": "arabic"
+    },
+    "french_stemmer": {
+      "type": "stemmer",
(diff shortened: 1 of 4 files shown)


kenoir (changes requested):
Thanks for chasing this down. The file change itself looks right to me: the live index can't have those fields removed without a reindex, so this is the only way to get a clean plan on 2025-10-02.

I'd push back on this bit though:

> We'll also want to apply these mappings in the 2026-07-03 pipeline to align it with the 2025-10-02 pipeline.

I don't think we want that. The flat fields (`archiveRoot`, `archiveType`, `collectionRoot`, `isArchiveRoot`, `isCollectionRoot`) only ever existed in intermediate commits of #3521 (5d490b3 through dde5045); before it merged, the "Reorganise new collection and archive fields" refactor replaced them with the nested shape (`archive.category`, `collection.root`, `collection.isRoot`), which is what the ingestor serialises today (`catalogue_graph/src/ingestor/models/filter/work.py`, `models/aggregate/work.py`). The same rename happened in catalogue-api before its merge (c794e387), so nothing deployed queries the flat paths either; the only consumer was pre-merge branch code via the archive-browse-testing preview entry, removed in wellcomecollection/catalogue-api#950. I also checked the frozen index itself: an `exists` query on each of the five flat fields returns zero documents, so they're unpopulated mapping leftovers from a manual `PUT _mapping` during development.

Since `pipeline/terraform/2026-07-03/elastic.tf` points at this same file, I think we should decouple rather than align: keep this edited file as the accommodation for the hand-patched frozen index, add a `mappings.works_indexed.2026-07-03.json` with the pre-this-PR content (the clean nested-only state from #3521), and point the 2026-07-03 pipeline at that. Happy to open that as a follow-up if you'd rather land this as-is.

agnesgaroux:
> Thanks for chasing this down. The file change itself looks right to me: the live index can't have those fields removed without a reindex, so this is the only way to get a clean plan on 2025-10-02.
> 
> I'd push back on this bit though:
> 
> > We'll also want to apply these mappings in the 2026-07-03 pipeline to align it with the 2025-10-02 pipeline.
> 
> I don't think we want that. The flat fields (`archiveRoot`, `archiveType`, `collectionRoot`, `isArchiveRoot`, `isCollectionRoot`) only ever existed in intermediate commits of #3521 ([5d490b3](https://github.com/wellcomecollection/catalogue-pipeline/commit/5d490b3efb51a6962fcb19569d1f8f447659a65d) through [dde5045](https://github.com/wellcomecollection/catalogue-pipeline/commit/dde5045858c0488b4ef0917508936ae044415e09)); before it merged, the "Reorganise new collection and archive fields" refactor replaced them with the nested shape (`archive.category`, `collection.root`, `collection.isRoot`), which is what the ingestor serialises today (`catalogue_graph/src/ingestor/models/filter/work.py`, `models/aggregate/work.py`). The same rename happened in catalogue-api before its merge (c794e387), so nothing deployed queries the flat paths either; the only consumer was pre-merge branch code via the archive-browse-testing preview entry, removed in [wellcomecollection/catalogue-api#950](https://github.com/wellcomecollection/catalogue-api/pull/950). I also checked the frozen index itself: an `exists` query on each of the five flat fields returns zero documents, so they're unpopulated mapping leftovers from a manual `PUT _mapping` during development.
> 
> Since `pipeline/terraform/2026-07-03/elastic.tf` points at this same file, I think we should decouple rather than align: keep this edited file as the accommodation for the hand-patched frozen index, add a `mappings.works_indexed.2026-07-03.json` with the pre-this-PR content (the clean nested-only state from #3521), and point the 2026-07-03 pipeline at that. Happy to open that as a follow-up if you'd rather land this as-is.

Agreed, we should not carry the defunct properties forward 👍

kenoir (approved):

PR #3583, opened 2026-08-19
Create replacement ES API keys before destroying the old ones

kenoir:
## What does this change?

Replacing a pipeline Elasticsearch API key currently gives the catalogue API a burst of 5xx (wellcomecollection/platform#6528). Terraform destroys before it creates, so Elasticsearch invalidates the old key before the new one exists or has reached Secrets Manager, and a consumer that re-fetches the secret on a 401 has nothing valid to fetch. With `create_before_destroy` the new key is created and stored first, so re-reading the secret is enough to recover.

Elasticsearch allows duplicate API key names, and the elasticstack provider's own rotation example uses this lifecycle. Since provider 0.16.1 a role-descriptor change updates the key in place rather than replacing it (confirmed by a plan on the 2026-07-03 stack on 2026-08-19), so this lifecycle only matters for the replacements that remain, such as a change to `name` or `expiration`.

### Checklist

- [ ] Does this change the display model the catalogue API returns? No. It changes pipeline terraform only.

## How to test

Run `terraform plan` for a pipeline stack and confirm the API key resources report no changes, so nothing is replaced by merging this on its own. To exercise the new ordering, force a replacement in a non-prod stack (for example by changing `expiration`) and check the plan shows the key created before the old one is destroyed, then confirm the value in Secrets Manager authenticates against the cluster after the apply.

The next key replacement lands without a 5xx spike on the catalogue API, and consumers recover by re-reading the secret rather than needing a redeploy.

Both keys are briefly valid during a replacement, which is the point but does mean the old key is usable for a short window after the new one is stored. There is no automated terraform apply here, so this is inert until someone applies it for each pipeline stack.

### DIFF
--- a/pipeline/terraform/modules/pipeline_es_api_key/main.tf
+++ b/pipeline/terraform/modules/pipeline_es_api_key/main.tf
@@ -2,20 +2,7 @@ locals {
   read_descriptor = {
     indices = [
       {
-        # allow_restricted_indices is false by default, but if you leave it out,
-        # terraform will recreate a new API key every time, because it compares
-        # the value reported by ElasticStack (which contains this default value)
-        # with the one you are trying to insert (which, in that case, would not).
-        # This then breaks all of your running pipeline applications because they
-        # are still trying to authenticate with the old key.
-        #
-        # This could (and possibly also should) be avoided by the careful application
-        # of replace_triggered_by in the corresponding modules, but it is unlikely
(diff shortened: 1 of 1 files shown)


agnesgaroux (approved):
Change is sensible but testing steps not run because of an ongoing reindex in the non-prod pipeline, preventing clean plan/apply

PR #3584, opened 2026-08-19
Archive browsing access info

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6489

Add new fields to the work model to support access-related archive browsing functionality. 

### Design context
The [archive browsing designs](https://www.figma.com/design/JUoyimGZGSELwBhMn2gQwj/Archive-browse) specify the following access-related requirements:
1. Each archive root page should include an 'Access' section, displaying one of the following:
    * `X items available online`: If at least one work belonging to the archive renders the 'View' CTA on its work page. The CTA is rendered when the work has a digital item with a non-closed, non-restricted IIIF presentation (different rules apply for some users who are members of staff, but `X` should reflect how many works are available online to regular users).
    * `Y items available by request`: If the above is false and at least one work belonging to the archive has a physical item showing an access label of 'Online request' or 'Manual request'. These labels map to the access method values `online-request` and `manual-request`. ('Manual request' items are very rare, there are only 3 across all archive works.)
    * `Items cannot be requested`: If the above is false.
2. Archive tree works should display an `Online` badge if their work pages render the 'View' CTA. 
3. The new collections content table should include an 'Access' column, displaying one of three labels:
    * `Online`: If the work page renders the 'View' CTA (same as above)
    * `In-person`: If the work page is available by an online/manual request (same as above)
    * `Can't be requested`: If the work page shows an access label of 'Not requestable' (mapping to the `not-requestable` access method).

### Implementation

**1. Archive root page 'Access' section**

The catalogue API needs to provide value `X` in `X items available online` and value `Y` in `Y items available by request`. These values come from two different aggregatable fields and can be retrieved via a search request:
* Value `X` can be retrieved from the `online` bucket of the `availabilities` aggregation. This aggregation is already available and no changes are needed in this PR. 
* Value `Y` can be obtained by summing the values from the `online-request` and the `manual-request` aggregation buckets of the `items.locations.accessConditions.method` aggregation. This is a new aggregation added in this PR (along with a paired filter). 

The full search request would look like this:
```
https://api.wellcomecollection.org/catalogue/v2/works?aggregations=availabilities%2Citems.locations.accessConditions.method&include=partOf,items&pageSize=50&sortOrder=asc&sort=collectionPath&collection.root=xqc9qs4x
```

The `collection.root` filter exists to exclude works which do not belong to the archive, and `partOf`/`items` fields are included so that only one request is needed to both retrieve the aggregations and to get the first 50 works to show in the collection contents table (see below). 

The `items.locations.accessConditions.method` aggregation also includes a `view-online` value, but this cannot reliably be used as `X`, since it includes closed and restricted digital items. The `online` bucket of the `availabilities` aggregation is technically broader than the design requirements, including non-IIIF presentation digital items accessible via different CTAs (e.g. `View Z images` for IIIF images). In practice, all archival works with `availabilities=online` include a IIIF presentation, so the requirements are met, and deriving online availability from the  `availabilities` field is consistent with the rest of the website.

Note that retrieving `X` and `Y` from search request aggregations does not necessarily imply an extra request from the frontend. The frontend already needs to make a search request to populate the [collection contents](https://www.figma.com/design/JUoyimGZGSELwBhMn2gQwj/Archive-browse?node-id=11068-111772&m=dev) table, and the aggregations can be retrieved via this request.

**2. `Online` badge in archive trees**

To support this functionality, this PR adds a new `availabilities` field to `Work` nodes in the catalogue graph and uses that field to populate a new `isAvailableOnline` field on each non-series `DisplayRelation` (which represents archive tree items in the `parts` and `partOf` fields).

Open question: Instead of deriving the `isAvailableOnline` field, each `DisplayRelation` could carry the full `availabilities: list[DisplayIdLabel]` field. This would make API responses larger (some works have 100+ parts), but it would give us more flexibility in the future.

**3. 'Access' column in the collection contents table**

This column can already be derived from existing display model data (`availabilities` and `items` fields). No changes needed here.

### Checklist

- [x] Does this change the display model the catalogue API returns? If so, regenerate the test documents under `catalogue_graph/document_generators/test_documents`. After merge, `sync-test-documents.yml` copies them into `wellcomecollection/catalogue-api` as an auto-PR, where OpenAPI contract tests validate the published spec against them; expect that PR to fail CI until the spec documents the new shape.

## How to test

The graph change can be tested against the 'dev' cluster. I already used the bulk loader to populate the `availabilities` field on all `Work` nodes. Ingestor can be tested by running it locally against a local Elasticsearch index.

All access-related archive browsing functionality should be supported, pending a downstream catalogue API PR. 

Summing the `online-request` and `manual-request` access method aggregation buckets to get `Y` can theoretically overcount in cases where a work has two physical items, one accessible via an online request, and one via a manual request. However, `manual-request` items are very rare: only 3 archival works have them, and never in combination with an `online-request` item, so in practice this limitation affects 0 works.

## Alternative approaches & future work

The approach implemented in this PR relies on values retrieved from aggregations, which can add query time overhead. This overhead is minimal at the moment due to the dual purpose of the search query (explained above).

However, to avoid using aggregations, it is possible to store values `X` and `Y` as denormalised counts on the archive root. The best way of achieving this would likely involve adding `isAvailableOnline` and `isRequestable` booleans to all `Work` nodes in the graph and then running Neptune queries in the ingestor loader computing the final counts to store on the root document. This approach is more involved than the aggregation approach and would likely impact ingestor loader performance (archive roots would have to be reprocessed whenever any of their descendants change, and the largest archives have 10,000+ works), but we can explore it in the future, possibly as part of a broader ticket treating collections as distinct from works (see Slack conversation [here](https://wellcome.slack.com/archives/C02ANCYL90E/p1787134174642889?thread_ts=1787125777.120019&cid=C02ANCYL90E)).

## Release plan

1. Note current time `T`. 
2. Bulk load all `Work` nodes into the production graph by running the graph extractor locally on this branch, followed by the bulk loader. Once this is done, `Work` nodes should have an `availabilities` field. (Optionally repeat this step for the non-production `2026-07-03` graph to prevent stale data in the new pipeline.)
3. Create a new works index with the new mapping (`2026-08-20`) via a targeted terraform apply. Run the works ingestor in `full` mode locally against the index. (Also apply the new mappings to the non-production `2026-07-03` index.)
4. Pause the `2025-10-02` incremental pipeline. 
5. Merge this PR and wait for all services to deploy.
6. Resume the incremental pipeline.
7. Backfill missing pipeline windows between `T` and now.
8. Switch the catalogue API to the new index.
9. Merge a downstream catalogue API PR adding support for the new filter/aggregation.


annak-k (approved):
Looks good to me. I ran the ingestor against the dev graph and a local index built from the new mapping, and both isAvailableOnline and the new access method aggregation are correct.

PR #3601, opened 2026-08-25
Add short_description field for archive collection roots

agnesgaroux:
## What does this change?

Adds a `short_description` field to the `DisplayWork` model, populated for archive collection roots only. The value is always a substring of the existing `description` -- the first sentence that describes the collection rather than its cataloguing -- so no text is generated.

The extraction logic lives in a new `short_description.py` transformer:

- HTML is parsed into block-level segments, so a paragraph with no terminal punctuation ends at the paragraph boundary rather than running into the next one.
- `pysbd` handles sentence boundary detection, with an abbreviation-merging pass on top (`c.`, `vol.`, `pp.`, etc.) to repair splits the segmenter gets wrong.
- Boilerplate sentences are filtered out: interim/temporary/summary description notices, custody and access notes, and sentences about the state of cataloguing rather than the collection.
- A `_join_stubs` pass reattaches colon-terminated lead-ins to the sentence they introduce.
- The value is never truncated here; display truncation is a frontend decision.

The second commit adds comments showing the (currently commented-out) guard for TEI manuscripts, which will be enabled once we know whether those browse cards also need this field.

## How to test

Unit tests are in `catalogue_graph/tests/ingestor/test_short_description.py`, covering the sentence splitter, HTML block parser, abbreviation merging, boilerplate filter, and notice-splitting logic. Run from `catalogue_graph/`:

```
uv run pytest tests/ingestor/test_short_description.py
```

To check the field appears on a real record, query the display index for an archive collection root and confirm `short_description` is populated.

Browse cards for archive collections show a short, human-readable description rather than a blank. The field is absent (null) for non-archive works, so there is no regression risk for other record types.

- The value is always a verbatim substring of the existing catalogued description, so there is no risk of generating incorrect text.
- The `pysbd` dependency is pinned in `uv.lock`. It is a pure-Python library with no network calls.
- Filtering is conservative: when all sentences are boilerplate, the field is null rather than surfacing a notice as a description.


catalogue_graph/src/ingestor/transformers/short_description.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def _merge_abbreviations(sentences: list[str]) -> list[str]:
+    """Rejoin sentences the splitter broke after an abbreviation such as "c." or "Vol."."""
+    merged: list[str] = []
+    for sentence in sentences:
+        if merged and _ends_with_abbreviation(merged[-1]):
+            merged[-1] = f"{merged[-1].rstrip()} {sentence.lstrip()}"
+        else:
+            merged.append(sentence)
```
kenoir, line 160: The notice split only looks at `sentences[0]`, so a colon-separated notice that follows another note loses the whole description. I ran `derive_short_description("This collection is uncatalogued. Interim description pending cataloguing: Papers of the Society.")` and got `None`: the first sentence is filtered as boilerplate, the second still contains "interim description" and is filtered whole, and nothing is left. Same for `<p>Uncatalogued</p><p>Interim description pending cataloguing: Papers of the Society.</p>`. Applying the split to every sentence rather than just the first gives `"Papers of the Society."` for both. I think that is the cheaper fix, since the notice regex is already anchored at the start of a sentence.

catalogue_graph/src/ingestor/transformers/short_description.py
```diff
@@ -0,0 +1,250 @@
+"""Derive a short description from the description a cataloguer already wrote.
+
+The value is always a substring of the existing description, never generated text. Scoped to
+archive collection roots, which is the population the archive browse cards need.
+
+The difficulty is not finding the sentence boundary. Two independent splitters agree on 97% of
+collection roots. It is that a correctly extracted first sentence is often a note about the state
+of the cataloguing rather than a description of the collection, which is what the boilerplate
+filter below is for. See catalogue_graph/notebooks/short_description_extraction.ipynb for the
```
kenoir: The docstring points at `catalogue_graph/notebooks/short_description_extraction.ipynb`, but that file is not in this PR or on `main`. Could it be added, or the reference switched to the experiment repo (`wellcomecollection/catalogue-short-description-experiment`) where the measurements live?

catalogue_graph/src/ingestor/transformers/work_display_transformer.py
```diff
@@ -36,6 +37,16 @@
 
 
 class DisplayWorkTransformer(WorkBaseTransformer):
+    @property
+    def short_description(self) -> str | None:
+        """Collection roots, excluding TEI manuscripts."""
+        TEI_IDENTIFIER_TYPE = "tei-manuscript-id"
+        if not self.is_collection_root:
+            return None
+        if self.state.source_identifier.identifier_type.id == TEI_IDENTIFIER_TYPE:
```
kenoir, line 46: The commented-out TEI guard reads as dead code above the docstring. Since `archive` is derived from `COLLECTION_PATH_PREFIX_LABELS`, which has no TEI prefixes, the `self.archive is None` check already excludes TEI manuscripts; the guard only becomes necessary if the scope widens beyond archives. I think a one-line TODO naming that condition would say the same thing with less to maintain.

agnesgaroux: My understanding is that there are roots others than those listed in `COLLECTION_PATH_PREFIX_LABELS` that we want short descriptions for.
Maybe @StepanBrychta can give us his 2 cents after talking with the XP team

catalogue_graph/document_generators/test_documents/works.archive.PPEBC.root.json
```diff
@@ -52,6 +52,7 @@
       "title": "Papers of Ernst Boris Chain",
       "alternativeTitles": [],
       "description": "<p>Papers of the biochemist Sir Ernst Boris Chain (1906-1979). Includes correspondence, laboratory notebooks and photographs.</p>",
+      "shortDescription": "Papers of the biochemist Sir Ernst Boris Chain (1906-1979).",
```
kenoir, line 55: The `sync-test-documents` workflow will open a PR on catalogue-api copying this fixture, and `OpenApiSpecResponseTest` ("cover every field the pipeline's display documents contain") will fail there because `reference/catalogue.yaml` has no `shortDescription` on `Work`. The API itself passes `display` through as untyped JSON, so the field will reach consumers with no Scala change; the spec is the only thing that needs updating. I think the minimal fix is adding `shortDescription` under `Work.properties` next to `description` on the `auto-sync/test-documents` branch before that PR is merged. The regenerated `catalogue-api.d.ts` on wellcomecollection.org follows from the spec automatically, but the hand-written `Work` and `WorkBasic` types there will need the field added by hand before the browse cards can use it. Worth a line in the PR body so it isn't a surprise.

catalogue_graph/pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "cryptography>=48.0.1",
     "yoyo-migrations>=8.2.0",
     "nh3>=0.3",
+    "pysbd>=0.3.4",
```
kenoir: Could this be `pysbd==0.3.4` rather than `>=0.3.4`? The package is pure Python with no dependencies and a single maintainer, and the only release is from 2021-02-11, so any new version appearing on PyPI would be a surprise worth reviewing rather than picking up automatically. The exact pin matters more than usual here because `lambda.Dockerfile` and `task.Dockerfile` install with `uv pip install --system .`, which resolves from `pyproject.toml` and does not read `uv.lock` (the comment above that line says otherwise). That applies to every dependency, not just this one, so I have raised it separately as wellcomecollection/platform#6661.

catalogue_graph/src/ingestor/transformers/short_description.py
agnesgaroux: Addressed in [7cf828eec1661a75c3428dd653702dd3db17181f](https://github.com/wellcomecollection/catalogue-pipeline/pull/3601/changes/7cf828eec1661a75c3428dd653702dd3db17181f) and covered by test

catalogue_graph/pyproject.toml
agnesgaroux: 📌

catalogue_graph/document_generators/test_documents/works.archive.PPEBC.root.json
agnesgaroux: It's nice that we get to exercise this new automation. 
The steps should be then:
- make some change that affects the catalogue api response (could be in cat pipeline or api repo)
- once merged, the sync test document PR is opened automatically
- before merging the sync test document PR, add whatever has changed to reference/catalogue.yaml

agnesgaroux: Opened https://github.com/wellcomecollection/catalogue-api/pull/967
and added a reference to the api spec in the PR template here
Checklist now reads
`Does this change the display model the catalogue API returns? If so, regenerate the test documents under `catalogue_graph/document_generators/test_documents`. After merge, `sync-test-documents.yml` copies them into `wellcomecollection/catalogue-api` as an auto-PR, where OpenAPI contract tests validate the published spec against them; expect that PR to fail CI until the spec >>>>> (`catalogue-api/reference/catalogue.yaml`) <<<<< documents the new shape.`

kenoir (approved):

kenoir (approved):

PR #3604, opened 2026-08-25
Add a runbook for clearing catalogue entities from a Neptune graph

kenoir:
## What does this change?

Adds a runbook for clearing catalogue entities from a dated Neptune graph while keeping the ontology, distilled from the round 1 and round 2 clears (wellcomecollection/platform#6461, wellcomecollection/platform#6503). It records what deletes and what stays, the access path, the incident-edge pre-mapping that explains the HAS_PARENT drop, the hard rules (never `_reset_database`, assert the endpoint carries the graph date), label-scoped verification with reference counts, and a duration estimate from measured throughput against the cluster's real population.

Written so the round 3 clear (wellcomecollection/platform#6623 phase 4) can be run from the repository alone, without relying on issue-thread archaeology.

## How to test

Documentation only; nothing executes. The counts and rates quoted are from the 2026-07-30 clear and a label-scoped census of `catalogue-graph-2026-07-03` on 2026-08-25.

### DIFF
--- a/catalogue_graph/docs/graph-clear-runbook.md
+++ b/catalogue_graph/docs/graph-clear-runbook.md
@@ -1,0 +1,56 @@
+# Clearing catalogue entities from a Neptune graph
+
+How to clear the catalogue-derived population from a dated Neptune cluster while keeping the ontology, so a pipeline can be rebuilt without re-running the monthly authority load. Run for real on `catalogue-graph-2026-07-03` during the round 1 and round 2 migration-testing clears (wellcomecollection/platform#6461 phase 4, wellcomecollection/platform#6503 phase 4).
+
+How long it takes: the 2026-07-30 clear removed ~183k catalogue nodes in about 4 minutes, roughly 45k nodes a minute. Scale that by the current population, which you should count first anyway (see the verification section): on 2026-08-25 the cluster held 2,169,565 catalogue nodes (Work 1,171,074, Concept 617,395, Image 123,457, PathIdentifier 257,639), which at the measured rate is about 45 to 50 minutes. Edge-heavy nodes delete slower, so treat it as a floor.
+
+## What gets deleted and what stays
+
+Each dated cluster holds two populations:
+
+- Catalogue-derived nodes, written by the graph pipeline from works: `Work`, `Concept`, `Image`, `PathIdentifier`. These are what the clear removes.
+- The ontology, loaded by the monthly authority job: `SourceConcept`, `SourceLocation`, `SourceName` and the authority edges (`SAME_AS`, `NARROWER_THAN`, `RELATED_TO`, `HAS_FIELD_OF_WORK`, `HAS_FOUNDER`, and the ontology share of `HAS_PARENT`). These stay.
+
+The split works because every `HAS_SOURCE_CONCEPT` edge originates on the `Concept` side, so `DETACH DELETE` on the catalogue labels severs the boundary while leaving every ontology node standing.
+
+## Access
+
+Runs locally, no tunnel needed: the clusters sit in public subnets with port 8182 open and IAM database auth. `platform-developer` carries `neptune-db:DeleteDataViaQuery` on the dated clusters. If in doubt, verify with `iam simulate-principal-policy` against the cluster's `neptune-db` resource id (not the RDS-style ARN).
+
+```python
+# AWS_PROFILE=platform-developer, from catalogue_graph/ with PYTHONPATH=src
+from clients.neptune_client import NeptuneClient
+
+client = NeptuneClient("2026-07-03")  # the graph date, never empty
+assert "2026-07-03" in client.neptune_endpoint
+```
+
+The endpoint assertion matters: client-setup patterns copied from notebooks default to the production graph when the date is empty.
+
+## Before deleting: map the incident edges
+
+Count the edges each doomed label touches, so the post-clear check is an exact prediction instead of an impression:
+
+```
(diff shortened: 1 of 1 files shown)


catalogue_graph/docs/graph-clear-runbook.md
```diff
@@ -0,0 +1,56 @@
...
+Each dated cluster holds two populations:
+
+- Catalogue-derived nodes, written by the graph pipeline from works: `Work`, `Concept`, `Image`, `PathIdentifier`. These are what the clear removes.
+- The ontology, loaded by the monthly authority job: `SourceConcept`, `SourceLocation`, `SourceName` and the authority edges (`SAME_AS`, `NARROWER_THAN`, `RELATED_TO`, `HAS_FIELD_OF_WORK`, `HAS_FOUNDER`, and the ontology share of `HAS_PARENT`). These stay.
+
+The split works because every `HAS_SOURCE_CONCEPT` edge originates on the `Concept` side, so `DETACH DELETE` on the catalogue labels severs the boundary while leaving every ontology node standing.
+
+## Access
+
+Runs locally, no tunnel needed: the clusters sit in public subnets with port 8182 open and IAM database auth. `platform-developer` carries `neptune-db:DeleteDataViaQuery` on the dated clusters. If in doubt, verify with `iam simulate-principal-policy` against the cluster's `neptune-db` resource id (not the RDS-style ARN).
```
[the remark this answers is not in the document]
kenoir: This is ok, it's inferable elsewhere, but we can rephrase.

catalogue_graph/docs/graph-clear-runbook.md
```diff
@@ -0,0 +1,56 @@
...
+
+```python
+# AWS_PROFILE=platform-developer, from catalogue_graph/ with PYTHONPATH=src
+from clients.neptune_client import NeptuneClient
+
+client = NeptuneClient("2026-07-03")  # the graph date, never empty
+assert "2026-07-03" in client.neptune_endpoint
+```
```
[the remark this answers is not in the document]
kenoir: Agreed, the endpoint hostname is not guaranteed to carry the date. The snippet now guards on the graph date value itself (empty or prod selects the legacy production cluster via NeptuneClient.namespace), which is the actual trap. Fixed in the latest commit.

catalogue_graph/docs/graph-clear-runbook.md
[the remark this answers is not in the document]
kenoir: Fair point for a public repo. The network specifics are gone; the doc now just says direct connections with IAM database auth work. Fixed in the latest commit.

catalogue_graph/docs/graph-clear-runbook.md
```diff
@@ -0,0 +1,57 @@
+# Clearing catalogue entities from a Neptune graph
+
+How to clear the catalogue-derived population from a dated Neptune cluster while keeping the ontology, so a pipeline can be rebuilt without re-running the monthly authority load. Run for real on `catalogue-graph-2026-07-03` during the round 1 and round 2 migration-testing clears (wellcomecollection/platform#6461 phase 4, wellcomecollection/platform#6503 phase 4).
```
StepanBrychta: Presumably this file will outlive the migration, so it might be good to remove references to specific tickets and graph dates to avoid the documentation becoming stale.

catalogue_graph/docs/graph-clear-runbook.md
```diff
@@ -0,0 +1,57 @@
...
+## The clear
+
+Use `delete_all_nodes_with_label` (batched `DETACH DELETE`, 10k per batch):
+
+```python
+for label in ["Work", "Concept", "Image", "PathIdentifier"]:
+    client.delete_all_nodes_with_label(label)
+```
+
+Never use `_reset_database`. It is hard-disabled for a reason: it would destroy the ontology and force a monthly-load re-run.
```
StepanBrychta: I haven't checked, but using `_reset_database` might actually be faster. A full monthly pipeline run is very time consuming, but it wouldn't be necessary. The monthly bulk load files already exist in S3, and so we would only have to re-run the bulk loaders (via the `graph-bulk-loaders-monthly-2026-07-03` state machine). Based on previous run durations, this would probably take under 30 minutes.

The approach documented in this file is valid, and it probably doesn't matter if it's a few minutes slower than some alternatives. But I think that "never use `_reset_database`" is too strong, and it might even be useful to document that approach as an alternative (i.e. run `_reset_database` followed by monthly bulk loaders).

catalogue_graph/docs/graph-clear-runbook.md
kenoir: Agreed, I'll genericise the graph dates to a placeholder. I'd keep the measured timings and counts as evidence the procedure has actually run, but I'll move them into a dated "measured runs" section at the end so the body stays evergreen.

catalogue_graph/docs/graph-clear-runbook.md
kenoir: Fair, and at the current population (~2.2M catalogue nodes, 45+ min of label deletes) your route may be faster. I'll soften "never" to what the code enforces (`ALLOW_DATABASE_RESET = False`, so it's a deliberate code change) and document reset plus bulk loaders as an alternative. On the incrementally-loaded authority edges: no gap. They are Concept to SourceConcept edges, re-emitted by the incremental pipeline during the post-clear reindex from the persisted monthly outputs in S3, and authority-to-authority edges reload from those same files, so the alternative needs no extra authority-load step.

StepanBrychta (approved):

PR #3605, opened 2026-08-25
Add the interesting-works CSV generator for migration user testing

kenoir:
## What does this change?

Adds `scripts/user_testing/generate_interesting_works.py`, which builds the interesting-works CSV that collections staff use when testing a migration round through the toggle preview (wellcomecollection/platform#6631).

The CSV is a per-round artefact rather than a committed file: canonical ids for Axiell-only works re-mint on every id-minter respin, so the script selects records by stable source keys (GUID, RefNo, search phrase) and resolves canonical ids and site links at generation time, after the round's reindex. Classes cover the migration's interesting behaviours: subjects that carried vocabulary markers, structured dates, deep archive hierarchies, Sierra and Axiell merges, digitised works with images (which doubles as b-number load evidence for wellcomecollection/platform#6542), and an absence check for the resource guides.

## How to test

From `scripts/user_testing/`: `AWS_PROFILE=platform-read_only uv run generate_interesting_works.py --pipeline-date 2026-07-03 --out interesting-works.csv`. Against the current round 2 index this produces 18 rows across 6 classes; the sample is seeded and candidate order is sorted, so re-runs against an unchanged index agree.


scripts/user_testing/generate_interesting_works.py
```diff
@@ -0,0 +1,183 @@
...
+        request_timeout=60,
+    )
+
+
+def rows_for(es, index, entry, cls, sample=False):
+    r = es.search(
+        index=index,
+        size=50 if sample else 5,
+        query=entry["query"],
+        source=["display.title", "query.identifiers.value"],
```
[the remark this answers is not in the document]
kenoir: In elasticsearch-py 8.x the typed search API takes source= for source filtering and serialises it to _source in the request body, so this is the documented form rather than the query-string parameter. The runs in the PR description used this code and returned only the requested fields, which confirms the filtering is applied. Leaving as is.

scripts/user_testing/generate_interesting_works.py
```diff
@@ -0,0 +1,216 @@
...
+# ///
+"""Generate the interesting-works CSV for migration user testing.
+
+Selects works by stable source keys (RefNo, priref, GUID, class queries) and
+resolves canonical ids and links at generation time, because Axiell-only
+canonical ids re-mint on every id-minter respin. Run it AFTER the round's
+reindex, against that round's pipeline date:
+
+    AWS_PROFILE=platform-read_only uv run generate_interesting_works.py \
+        --pipeline-date 2026-07-03 --out interesting-works.csv
```
agnesgaroux, line 13: Does this work for you? I'm having to instantiate a boto session with the `platform-developer` profile in the script or I can't retrieve the secret

kenoir: No, it does need developer. Fixed the docstring and README to say platform-developer, and added a --profile flag so the script takes the profile directly. Verified against the live index after the change.

agnesgaroux (approved):

PR #3607, opened 2026-08-26
Merge audiovisual clusters one e-bib at a time

### LINKED ISSUE #6643: Merger attaches every METS manifest to one e-bib when an AV recording has several digitised bibs
## Summary

When an audiovisual recording is catalogued as one physical bib plus two or more digitised e-bibs, the merger attaches every METS manifest to a single e-bib and leaves the others with no manifest. Which e-bib wins is arbitrary, so the same cluster produced opposite results in the 2025-10-02 and 2026-07-03 pipelines.

Reported by Collections Information for "Recent advances in musico-rheumotherapeutics and Hors d'oevres":

- Side A, b2472497x, https://wellcomecollection.org/works/eb5qwcxy
- Side B, b24724981, https://wellcomecollection.org/works/a6g3k8eq
- Physical bib b24724956, https://wellcomecollection.org/works/zbrsq46n

In production Side B carries both IIIF manifests (b24724981 and b2472497x) and Side A has only its archive.org 856 link.

## Cause

Both e-bibs have `776$w (UkLW)b24724956`, so the matcher puts all three Sierra bibs and both METS works into one cluster.

For that cluster:

1. `TargetPrecedence` ranks `sierraDigitisedAv` (Sierra, audiovisual format, no identified items) above `physicalSierra`. Both e-bibs satisfy it, so `works.find` returns whichever arrived first.
2. `ItemsRule.mergeMetsIntoSierraTarget` appends every METS source item to a zero-item Sierra target and redirects both METS works into it.
3. `Sources.findFirstLinkedDigitisedSierraWorkFor` returns `None` for audiovisual targets (wellcomecollection/platform#4876), so the other e-bib and the physical bib are not redirected and survive as separate works, one of them with no manifest.

The METS files and DDS manifests are correct. The b2472497x manifest has the Side A title and `homepage` eb5qwcxy.

Evidence, both merged 2026-08-25:

- 2025-10-02 pipeline (serving the API): a6g3k8eq has `redirectSources = [mets b24724981, mets b2472497x]`; eb5qwcxy has no manifest.
- 2026-07-03 pipeline: eb5qwcxy has both manifests; a6g3k8eq has none.
- Matcher graph node for eb5qwcxy lists the five-work component: a6g3k8eq, bdxvvz9v, br2ux6n5, eb5qwcxy, zbrsq46n.

## Fix

AV recordings are deliberately kept as separate catalogue records: the physical bib on its own and each e-bib on its own. That is the policy already encoded in `Sources.findFirstLinkedDigitisedSierraWorkFor` (wellcomecollection/platform#4876), and the 776 links stay as they are.

The change is in `Merger.merge`. When the cluster target would be a Sierra work, each AV e-bib is carved out into its own group with the non-Sierra works linked directly to it by merge candidate, and each group is merged on its own. The rest of the cluster (physical bib, unlinked METS, non-AV pairs) merges as before, so clusters with a Calm, Axiell or TEI target and METS records that name the physical bib keep their current outcome. A work linked to several AV e-bibs is left in the main group with a warning.

## Affected records

Any multi-part AV recording catalogued as physical plus several e-bibs with 776 links. Worth a count against the works-identified index once the fix is deployed.

kenoir:
Closes wellcomecollection/platform#6643.

## What does this change?

A physical AV bib and its e-bibs land in one matcher cluster because of their 776 links, but AV Sierra bibs are never merged with each other (`Sources.findFirstLinkedDigitisedSierraWorkFor`, wellcomecollection/platform#4876). Merging the cluster as a whole elected one arbitrary e-bib as the target and `ItemsRule.mergeMetsIntoSierraTarget` handed that one work every METS manifest, leaving the other e-bibs with none. The example in the issue is a two-sided cassette, b2472497x and b24724981, where the 2025-10-02 and 2026-07-03 pipelines picked opposite e-bibs for the same cluster.

`Merger.merge` now runs a `partitionAudiovisual` step first. When the cluster target would be a Sierra work, each AV e-bib is carved out along with the non-Sierra works linked directly to it by merge candidate in either direction, and each group is merged on its own. Whatever is left over merges as before. A work linked to more than one AV e-bib has no single owner, so it stays in the main group and logs a warning.

The partition is narrow on purpose. Clusters with a Calm, Axiell or TEI target never reach it, and non-AV physical and e-bib pairs pulled into an AV cluster stay in the remainder group, so both keep their current outcome. A METS record that names the physical bib also stays in the remainder; when nothing is carved out that merges as before, and when the e-bibs are carved out the remainder elects the physical bib, so the manifest lands on the bib the METS record names.

### Checklist

- [ ] Does this change the display model the catalogue API returns? No. It changes which works merge together, not the shape of the documents we produce.

## How to test

`sbt merger/test` with a local DynamoDB running. The full merger suite passes locally, 263 tests. New `PlatformMergerTest` cases cover: two e-bibs with a METS work each, run table-driven over two input orders so an order-dependent target fails the test; a METS work naming the physical bib with a single bare e-bib, which still merges into the e-bib; the same with the e-bibs carved out, where it merges into the physical bib; and a METS work linked to two e-bibs, which does not follow the carved-out one.

Once this is deployed and the affected records have moved through, eb5qwcxy and a6g3k8eq should each carry their own IIIF manifest instead of one carrying both.

The issue asks for a count of affected multi-part AV recordings against `works-identified` after deployment. There is no ongoing metric beyond that.

The partition changes cluster membership, so a predicate that is scoped too widely would re-merge unrelated works without raising an error. `sierraDigitisedAv` requires a Sierra work with an audiovisual format and no identified items, and the carve-out only pulls in non-Sierra works that hold a direct merge-candidate link, which keeps the blast radius to AV clusters. A mistake here surfaces as works gaining or losing items rather than as a failure, so it is worth looking at AV records again after the next reindex.

### DIFF
--- a/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
+++ b/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
@@ -940,7 +940,7 @@ class PlatformMergerTest
   }
 
   it(
-    "keeps a METS work linked to several audiovisual e-bibs in the main group"
+    "redirects a METS work linked to several audiovisual e-bibs to the e-bib that was not carved out"
   ) {
     val physical = sierraIdentifiedWork()
       .format(Format.Audio)
@@ -971,9 +971,47 @@ class PlatformMergerTest
       .map(w => w.id -> w.redirectTarget.canonicalId)
       .toMap
     redirects(metsForFirstEbib.id) shouldBe ebibs(0).state.canonicalId
-    redirects.get(ambiguousMets.id) should not contain ebibs(
-      0
-    ).state.canonicalId
+    redirects(ambiguousMets.id) shouldBe ebibs(1).state.canonicalId
+  }
+
+  it(
+    "redirects a METS work linked to several audiovisual e-bibs to the physical bib once every e-bib is carved out"
+  ) {
+    val physical = sierraIdentifiedWork()
+      .format(Format.Audio)
+      .items(List(createIdentifiedPhysicalItem))
+    val ebibs = (1 to 2).map {
+      _ =>
+        sierraIdentifiedWork()
+          .format(Format.Audio)
+          .mergeCandidates(List(createSierraPairMergeCandidateFor(physical)))
+    }.toList
+    val metsForEbibs = ebibs.map {
+      ebib =>
+        identifiedWork(sourceIdentifier = createMetsSourceIdentifier)
+          .mergeCandidates(List(createMetsMergeCandidateFor(ebib)))
+          .items(List(createDigitalItem))
+          .invisible()
+    }
+    val ambiguousMets =
+      identifiedWork(sourceIdentifier = createMetsSourceIdentifier)
+        .mergeCandidates(ebibs.map(createMetsMergeCandidateFor))
+        .items(List(createDigitalItem))
+        .invisible()
+
+    val result = merger
+      .merge(physical :: ebibs ++ metsForEbibs :+ ambiguousMets)
+      .mergedWorksWithTime(now)
+
+    val redirects = result
+      .collect { case w: Work.Redirected[Merged] => w }
+      .map(w => w.id -> w.redirectTarget.canonicalId)
+      .toMap
+    redirects shouldBe Map(
+      metsForEbibs(0).id -> ebibs(0).state.canonicalId,
+      metsForEbibs(1).id -> ebibs(1).state.canonicalId,
+      ambiguousMets.id -> physical.state.canonicalId
+    )
   }
 
   it("gives each audiovisual e-bib its own METS work") {
--- a/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
+++ b/pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
@@ -891,6 +891,91 @@ class PlatformMergerTest
     redirected.head.redirectTarget.canonicalId shouldBe workForEbib.state.canonicalId
   }
 
+  it(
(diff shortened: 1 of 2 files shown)


pipeline/matcher_merger/merger/src/test/scala/weco/pipeline/merger/services/PlatformMergerTest.scala
```diff
@@ -863,6 +863,212 @@ class PlatformMergerTest
...
+      .head
+      .data
+      .items
+      .flatMap(_.locations) should contain theSameElementsAs
+      physical.data.items.flatMap(_.locations) ++ metsForPhysical.data.items
+        .flatMap(_.locations)
+  }
+
+  it(
+    "keeps a METS work linked to several audiovisual e-bibs in the main group"
```
agnesgaroux: Could this be "redirects a METS work linked to several audiovisual e-bibs to the physical bib"? and this
```
redirects.get(ambiguousMets.id) should not contain ebibs(
      0
    ).state.canonicalId
```
be 
```
redirects.get(ambiguousMets.id) should be the physical bib
```

agnesgaroux: As in, we need to assert that it still being attached to the physical bib, if ambiguous

kenoir: Good call on the negative assertion, I have made it positive. In this fixture the physical bib is not the target though: ebibs(1) has nothing singly attached to it so it stays in the main group, and TargetPrecedence ranks digitised AV above physical Sierra, so the ambiguous METS redirects to ebibs(1). The test now asserts that and is renamed to match. If both e-bibs had been carved out the remainder would be the physical bib plus the METS, and it would land there instead, which is now covered by a second test.

agnesgaroux (approved):

PR #3611, opened 2026-08-27
Scale up tasks and the ES cluster for the round 3 reindex

kenoir:
Pre-staged for wellcomecollection/platform#6624, mirroring round 2's #3576. Merge and apply after the phase 7 re-enable, before the reindex starts: `infrastructure/critical` via `./run_terraform.sh` first (the ES resize), then `pipeline/terraform/2026-07-03`.

`scale_up_tasks` on, ES to 30g across 2 zones. The matcher DynamoDB scale-up is a separate PR so it goes on last and comes off first. Reverting after the comparison: the ES scale-down apply can only run once a day.

Will need a rebase once the quiesce and re-enable PRs have merged, since they touch the same block.

### DIFF
--- a/infrastructure/critical/es_cluster.tf
+++ b/infrastructure/critical/es_cluster.tf
@@ -7,8 +7,11 @@ module "es_cluster_2026_07_03" {
 
   cluster_date = "2026-07-03"
 
-  memory     = "4g"
-  node_count = 3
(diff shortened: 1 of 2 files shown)


agnesgaroux:
Superseeded by https://github.com/wellcomecollection/catalogue-pipeline/pull/3626

PR #3612, opened 2026-08-27
Scale up the matcher DynamoDB for the round 3 reindex

kenoir:
Pre-staged for wellcomecollection/platform#6624, mirroring round 2's #3577. Apply last, immediately before the reindex starts, and revert first after the comparison signs off: the provisioned capacity is a flat ~$135/day. Note the flag scales the whole matcher stage (lambda memory, concurrency and batch size), not only the table.


agnesgaroux:
Superseeded by https://github.com/wellcomecollection/catalogue-pipeline/pull/3626

PR #3624, opened 2026-09-03
Add a role the Identifiers API can assume to read the id-minter registry

### LINKED ISSUE #6680: Create a platform-account role the Identifiers API can assume for Data API reads
## Summary

An IAM role in the platform account (760097843905) that the Identifiers API Lambda can assume in order to read the id-minter registry over the RDS Data API.

#6531 puts the Lambda in the catalogue account, and both registry clusters are in the platform account. The Data API rejects a cluster owned by another account ("The AWS account ID 756629837203 doesn't own DB cluster identifiers-v2-serverless-2026-07-03"), and the cluster's master secret is encrypted with an AWS managed key so it cannot be shared across accounts either. The Lambda therefore has to make the call as a platform-account identity. See #6531 for the surrounding design.

## Key decisions

- The role lives with the cluster, in `infrastructure/critical/modules/id-minter-rds`, so each cluster grants separately and #6658 can later select a role per database.
- The trust names the consumer's account with an `ArnEquals` condition on `aws:PrincipalArn` rather than naming the role as a principal. A role ARN written as a principal is resolved when the policy is stored, which would mean the Lambda's execution role had to exist first, and would break silently if that role were ever recreated.

## Scope of work

- [x] A role in a new `iam.tf` in that module, with `rds-data:ExecuteStatement` on the cluster and `secretsmanager:GetSecretValue` on its secret
- [x] A module variable taking the consuming role ARNs, granted `lambda-role-identifiers-api-stage` on the 2026-07-03 cluster only, leaving production until #6543
- [x] The role ARN added to the `id_minter_rds` output map, which catalogue-api already reads through remote state

## Watch for

`rds-data:ExecuteStatement` covers writes as well as reads, and IAM has no read-only form of it, so this role's name describes intent rather than something the policy enforces. Until #6533 creates a SELECT-only database user, the guarantee rests on the application's own guard, and the secret in the second statement is the cluster master credential. Point it at #6533's secret when that lands, and do not add further consumers before then.

The policy names the cluster's current master secret, and each testing round's rebuild creates a new one, so `infrastructure/critical` needs re-applying after every respin or the role points at a secret that no longer exists.

## Done when

The role is applied, and the Lambda's execution role can assume it and read the 2026-07-03 cluster, which is the invoke check in #6531.

annak-k:
## What does this change?

Adds a role in the platform account that the Identifiers API's Lambda can assume in order to read the id-minter registry over the RDS Data API, and grants it on the 2026-07-03 cluster.

wellcomecollection/platform#6680. The Identifiers API runs in the catalogue account, and the Data API rejects a cluster owned by another account, so the Lambda has to make the call as a platform-account identity.

- `modules/id-minter-rds/iam.tf` creates the role, with `rds-data:ExecuteStatement` on the cluster and `secretsmanager:GetSecretValue` on its secret. Nothing is created when no consumer is named, so the production cluster is unaffected.
- The trust names the consumer's account with an `ArnEquals` condition on `aws:PrincipalArn`, so it does not depend on the Lambda's role existing yet.
- `rds_id_minter.tf` grants the stage Lambda on the 2026-07-03 cluster only. Production comes with wellcomecollection/platform#6543.
- The role ARN joins the `id_minter_rds` output map, which catalogue-api already reads.

## Applying it

Can be applied at any time, in either order relative to the catalogue-api side. The whole path is verified by the lookup check in wellcomecollection/platform#6531, which needs the Lambda deployed too, so this does not close wellcomecollection/platform#6680.

## Notes

- `rds-data:ExecuteStatement` covers writes as well as reads, and IAM has no read-only form of it. The read-only guarantee comes from the guard in the application, and then the SELECT-only user from wellcomecollection/platform#6533, which should also replace the master secret here.
- The production cluster's entry in the output map is null until it is granted.
- The policy names the cluster's current master secret, and a respin creates a new one, so this stack needs re-applying after each round's rebuild.

### DIFF
--- a/infrastructure/critical/modules/id-minter-rds/iam.tf
+++ b/infrastructure/critical/modules/id-minter-rds/iam.tf
@@ -1,0 +1,46 @@
+locals {
+  identifiers_api_read_count = length(var.data_api_consumer_role_arns) > 0 ? 1 : 0
+}
+
+resource "aws_iam_role" "identifiers_api_read" {
+  count = local.identifiers_api_read_count
+
+  name               = "identifiers-api-registry-read${local.hyphen_suffix}"
+  assume_role_policy = data.aws_iam_policy_document.identifiers_api_assume[0].json
+}
+
+data "aws_iam_policy_document" "identifiers_api_assume" {
+  count = local.identifiers_api_read_count
+
+  statement {
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "AWS"
+      identifiers = var.data_api_consumer_role_arns
+    }
+  }
+}
+
+resource "aws_iam_role_policy" "identifiers_api_read" {
+  count = local.identifiers_api_read_count
+
+  role   = aws_iam_role.identifiers_api_read[0].name
+  policy = data.aws_iam_policy_document.identifiers_api_read[0].json
+}
+
+data "aws_iam_policy_document" "identifiers_api_read" {
+  count = local.identifiers_api_read_count
+
+  statement {
+    actions   = ["rds-data:ExecuteStatement"]
+    resources = [module.identifiers_v2_serverless_rds_cluster.rds_cluster_arn]
+  }
+
+  # The Data API reads the credential on the caller's behalf, so the caller needs
+  # access to the secret as well as to the cluster.
+  statement {
+    actions   = ["secretsmanager:GetSecretValue"]
+    resources = [module.identifiers_v2_serverless_rds_cluster.master_user_secret_arn]
+  }
+}
--- a/infrastructure/critical/modules/id-minter-rds/iam.tf
+++ b/infrastructure/critical/modules/id-minter-rds/iam.tf
@@ -1,5 +1,9 @@
 locals {
   identifiers_api_read_count = length(var.data_api_consumer_role_arns) > 0 ? 1 : 0
+
+  data_api_consumer_account_roots = distinct([
+    for arn in var.data_api_consumer_role_arns : "arn:aws:iam::${split(":", arn)[4]}:root"
+  ])
 }
 
 resource "aws_iam_role" "identifiers_api_read" {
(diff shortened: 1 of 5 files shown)


infrastructure/critical/modules/id-minter-rds/iam.tf
```diff
@@ -0,0 +1,46 @@
...
+data "aws_iam_policy_document" "identifiers_api_read" {
+  count = local.identifiers_api_read_count
+
+  statement {
+    actions   = ["rds-data:ExecuteStatement"]
+    resources = [module.identifiers_v2_serverless_rds_cluster.rds_cluster_arn]
+  }
+
+  # The Data API reads the credential on the caller's behalf, so the caller needs
+  # access to the secret as well as to the cluster.
```
agnesgaroux, line 51: Can we add a comment here such as
```
  # This is the cluster master secret, so the grant is admin-level on the registry
  # until https://github.com/wellcomecollection/platform/issues/6533 creates a
  # SELECT-only user with its own secret. Point this at that secret when it lands,
  # and do not add further consumers to data_api_consumer_role_arns before then.
```

agnesgaroux:
I'd have the`"aws_iam_policy_document"` like:
```
data "aws_iam_policy_document" "identifiers_api_assume" {
  count = local.identifiers_api_read_count

  statement {
    actions = ["sts:AssumeRole"]

    principals {
      type        = "AWS"
      identifiers = local.data_api_consumer_account_roots
    }

    condition {
      test     = "ArnEquals"
      variable = "aws:PrincipalArn"
      values   = var.data_api_consumer_role_arns
    }
  }
}
```
and in locals
```
data_api_consumer_account_roots = toset([
    for arn in var.data_api_consumer_role_arns : "arn:aws:iam::${split(":", arn)[4]}:root"
  ])
```

The Identifiers API Lambda role doesn't exist yet and can't, because #6531 needs this role's ARN before it can configure the Lambda, while a direct-principal trust policy needs the Lambda's role before it can be written. Matching by condition breaks that cycle so critical applies first and the API side second. Effective grant is unchanged: that one role ARN and nothing else.

agnesgaroux (approved):
Granted 😄

PR #3627, opened 2026-09-07
Folio transformer: add physical description and edition fields

StepanBrychta:
## What does this change?

Addresses https://github.com/wellcomecollection/platform/issues/6544

Implements FOLIO→Work mapping for the physical description field and the edition field, mirroring Sierra transformer logic. Translates existing Scala unit tests into Python BDD tests to verify that Python behaviour matches Sierra behaviour. Cases where Python implementation deliberately diverges from Scala are codified at the end of each feature file.

### Checklist

## How to test

I transformed a snapshot of all Folio works and compared the extracted fields against those stored on Sierra works in the source index:
* Physical description: 761,576 matches and 48 differences, plus 88 differences due to a trailing whitespace in the Sierra field (a known divergence). All 48 differences are upstream, not attributable to transformer logic differences.
* Edition:  761,613 matches and 2 differences, both upstream of the transformer. Another 9 differences were due to Sierra not using unicode NFC consistently (see [here](https://github.com/wellcomecollection/catalogue-pipeline/pull/3628) for more info).

The physical description and edition fields on Folio works should match their counterparts on Sierra works.


agnesgaroux (approved):
👍

PR #3628, opened 2026-09-07
Folio transformer: add title BDD tests

StepanBrychta:
## What does this change?

Addresses https://github.com/wellcomecollection/platform/issues/6545

The ticket asks to determine FOLIO→Work mapping for the title field. This mapping already exists: the Folio transformer extracts titles from MARC 245 (via the shared `MarcXmlWorkBuilder` class). This PR translates existing Scala unit tests into Python BDD tests to verify that the extracted Python titles match those extracted by the Sierra transformer. Two tests at the end of the feature file codify cases where the Python behaviour deliberately diverges from the Scala behaviour.

### Checklist

## How to test

I transformed a snapshot of all Folio works and compared the extracted titles against those stored on Sierra works in the source index. This produced 761,585 matches and 39 differences (nearly a 100% match), with one caveat (see below). All 39 differences are upstream, not attributable to transformer logic differences.

**NFC normalisation**
All Sierra titles were normalised using `unicodedata.normalize("NFC", title)` before they were compared to Folio titles. Sierra uses mixed unicode normalisation, where some diacritics are stored as a decomposed combining sequence (base character followed by separate combining accent or modifier), and some diacritics are precomposed. 

Folio titles use pure NFC, which is an improvement and fixes an active bug which makes some Sierra works difficult to find in the frontend. For example, users currently cannot find [this work](https://wellcomecollection.org/works/bdmkg6dc) by typing `muṟaikaḷ`, since most standard keyboards produce NFC, and the work is only returned when the query uses the combining-mark form.

The title field on Folio works should match its counterpart on Sierra works.

### DIFF
--- a/catalogue_graph/tests/adapters/transformers/folio/BDD/features/title.feature
+++ b/catalogue_graph/tests/adapters/transformers/folio/BDD/features/title.feature
@@ -1,0 +1,79 @@
+Feature: title (MARC 245)
+  The title is built from MARC 245 subfields ǂa, ǂb, ǂc, ǂh, ǂn and ǂp,
+  joined with a space in the order they appear in the field. Anything in
+  square brackets is stripped from ǂh, and a ǂh with nothing after it to
+  join to is dropped. Title is mandatory: a record without a usable 245
+  fails to transform.
+
+  These scenarios mirror the unit tests of the Scala transformer this
+  replaces (MarcTitleTest.scala), including its test data, so the two can
+  be compared directly.
+
+  http://www.loc.gov/marc/bibliographic/bd245.html
+
+  Background:
+    Given a MARC record with field 001 "abc123"
+    And the MARC record has a 999 field with indicators "f" "f" with subfield "i" value "10000000-0000-0000-0000-000000000001"
+
+  # Ported from MarcTitleTest.scala.
+
+  Scenario: The title is built from the 245 field
+    Given the MARC record has a 245 field with subfield "a" value "hello"
+    When I transform the MARC record
+    Then the work's title is "hello"
+
+  Scenario: The first 245 is used if a record erroneously has more than one
+    Given the MARC record has a 245 field with subfield "a" value "hello world"
+    And the MARC record has another 245 field with subfield "a" value "hello dolly"
+    When I transform the MARC record
+    Then the work's title is "hello world"
+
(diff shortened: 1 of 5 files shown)


catalogue_graph/tests/gherkin_steps/marc.py
```diff
@@ -51,6 +51,17 @@ def add_field(
...
 
+empty_field_step_regex = parsers.re(
+    r"the MARC record has (?:a|another) (?P<tag>\d{3}) field with no subfields"
+)
+
+
+@given(empty_field_step_regex)
+def add_empty_field(marc_record: Record, tag: str) -> None:
+    """A malformed data field: present in the record, but wholly empty."""
+    marc_record.add_field(Field(tag=tag))
```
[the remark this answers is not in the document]
StepanBrychta: It does not leave `subfields` as `None`. `Field(tag="245").subfields` produces `[]`.

catalogue_graph/tests/adapters/transformers/folio/BDD/features/title.feature
```diff
@@ -0,0 +1,83 @@
...
+  Scenario: The first 245 is used if a record erroneously has more than one
+    Given the MARC record has a 245 field with subfield "a" value "hello world"
+    And the MARC record has another 245 field with subfield "a" value "hello dolly"
+    When I transform the MARC record
+    Then the work's title is "hello world"
+
+  Scenario: The title is built from subfields ǂa, ǂb, ǂc, ǂh, ǂn and ǂp
+    Given the MARC record has a 245 field with subfield "a" value "cyntaf" and subfield "b" value "ail" and subfield "c" value "trydydd" and subfield "h" value "pedwerydd" and subfield "n" value "pumed" and subfield "p" value "chweched" and subfield "k" value "saithfed"
+    When I transform the MARC record
+    Then the work's title is "cyntaf ail trydydd pedwerydd pumed chweched"
```
agnesgaroux, line 34: Why Welsh though? 😂

agnesgaroux (approved):

catalogue_graph/tests/adapters/transformers/folio/BDD/features/title.feature
agnesgaroux: Paul, that's why

PR #3631, opened 2026-09-08
Folio transformer: add former frequency field

StepanBrychta:
## What does this change?

Addresses https://github.com/wellcomecollection/platform/issues/6551

Implements FOLIO→Work mapping for the former frequency field, mirroring Sierra transformer logic. Translates existing Scala unit tests into Python BDD tests to verify that Python behaviour matches Sierra behaviour.

### Checklist

## How to test

I transformed a snapshot of all Folio works and compared the extracted current frequencies against those stored on Sierra works in the source index. This produced 761,624 matches and 0 differences.

The former frequency field on Folio works should match its counterpart on Sierra works.


kenoir (approved):
Checked the subfield ordering against scala-libs: `subfieldsWithTags` is a filter on tag membership so it keeps document order, and pymarc `get_subfields` does the same, so the join is order-identical to the Scala. The stripping and empty-dropping divergences are the same ones `edition.feature` already documents, and I think stating them in the feature file the way you have is the right way to carry that decision forward. Folio suite, ruff and mypy all pass locally.

PR #3632, opened 2026-09-08
Folio transformer: add current frequency field

StepanBrychta:
## What does this change?

Addresses https://github.com/wellcomecollection/platform/issues/6550

Implements FOLIO→Work mapping for the current frequency field, mirroring Sierra transformer logic. Translates existing Scala unit tests into Python BDD tests to verify that Python behaviour matches Sierra behaviour. Cases where Python implementation deliberately diverges from Scala are codified at the end of the feature file.

### Checklist

## How to test

I transformed a snapshot of all Folio works and compared the extracted current frequencies against those stored on Sierra works in the source index. This produced 761,622 matches and 2 differences. Both differences were due to the Sierra transformer sometimes emitting a double space when joining the contents of multiple subfields (e.g. `6 no. a year,  <v. 26, issue 6->` instead of `6 no. a year, <v. 26, issue 6->`). The Python transformer deliberately diverges.

The current frequency field on Folio works should match its counterpart on Sierra works.


kenoir (approved):
Looks good. I checked the four whitespace scenarios in the feature file by running the Scala `MarcCurrentFrequency` body against the same inputs, and the Python output matches on every parity case and diverges only where the feature file says it does. Moving the extractor under `marc/` also tidies up EBSCO slightly: a blank ǂa or ǂb used to produce a double space there and now doesn't, which only touches records with empty 310 subfields.

PR #3635, opened 2026-09-08
Folio transformer: add designation field

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6552

Implements FOLIO→Work mapping for the designation field, mirroring Sierra transformer logic. Translates existing Scala unit tests into Python BDD tests to verify that Python behaviour matches Sierra behaviour. Cases where Python implementation deliberately diverges from Scala are codified at the end of the feature file.

Designation field extractor is reused from EBSCO, with the addition of logging if multiple instances of non-repeatable subfield $a are found. This is a deliberate divergence from Scala, which makes the work invisible if multiple $a instances are found in a single field. At the moment, there are no works affected by this divergence.

### Checklist

## How to test

I transformed a snapshot of all Folio works and compared the extracted designations against those stored on Sierra works in the source index. This produced 761,624 matches and 0 differences.

The current designation field on Folio works should match its counterpart on Sierra works.


catalogue_graph/src/adapters/transformers/marc/designation.py
```diff
@@ -0,0 +1,30 @@
...
+
+def _designation(field: Field) -> str:
+    """362 ǂa is a non-repeatable field. If it does repeat, log an error and take the first value."""
+    values = field.get_subfields("a")
+    if len(values) > 1:
+        logger.error(
+            "Repeated non-repeating subfield $a",
+            tag=field.tag,
+            count=len(values),
+        )
```
kenoir, line 30: This log is now the only signal that a value got dropped, where the Scala failed the record loudly. As it stands I don't think you could act on it: `tag` is always "362" because `_designation` is only called on 362 fields, so `count` is the only variable, and there's no record id to chase. `ExecutionContext` binds `trace_id`, `pipeline_step` and `started_at` only (`src/utils/logger.py:103-115`), and the transformer passes `row_id=` by hand where it wants it (`src/adapters/transformers/marcxml_transformer.py:43,64`), so nothing is bound per record here.

`marc/other_identifiers.py:58` handles the same problem by logging `field=str(field)`, which at least gives you the values to search for. Would that work here, or is the record id worth threading through? Not something this PR created, `marc/title.py:53-58` has the same gap, so happy for it to be a follow-up.

kenoir: Separate from the payload, there's the level. `marc/title.py:53-58` handles the same situation, a non-repeatable MARC element that repeats where we take the first and carry on, and logs it at `warning`. Either is defensible, error matches the fact that the Scala rejected the record outright, I'd just rather the two matched. No strong view on which way.

catalogue_graph/src/adapters/transformers/marc/designation.py
```diff
@@ -0,0 +1,30 @@
...
+
+logger = structlog.get_logger(__name__)
+
+
+def extract_designation(record: Record) -> list[str]:
+    return non_empty(_designation(field) for field in record.get_fields("362"))
+
+
+def _designation(field: Field) -> str:
+    """362 ǂa is a non-repeatable field. If it does repeat, log an error and take the first value."""
```
kenoir: Small wording thing: ǂa is a non-repeatable subfield, and 362 itself is repeatable, which is what the rest of the function relies on. The feature file gets this right ("362 is repeatable... ǂa should not repeat within a field"). Maybe "ǂa does not repeat within a 362"?

catalogue_graph/src/adapters/transformers/marc/designation.py
StepanBrychta: I added `field=str(field)` to the log so that the value is searchable in the adapter store. I also changed the `warning` in `marc/title.py` to `error`. I think that's the right level, since a repeating non-repeatable subfield is not valid.

catalogue_graph/src/adapters/transformers/marc/designation.py
StepanBrychta: Yes, changed to "362 ǂa is a non-repeatable **sub**field".

kenoir (approved):
Two non-blocking things left inline, on the log payload and the log level. Neither needs to hold this up, and the log payload one is arguably a follow-up across `marc/title.py` too.

PR #3637, opened 2026-09-09
Give the Identifiers API its own read-only registry credential

annak-k:
## What does this change?

Gives the Identifiers API its own read-only credential on the id-minter registry, so it no longer needs the cluster master secret. Three of the six items on wellcomecollection/platform#6533.

The role added in #3624 grants `GetSecretValue` on the cluster's master secret, which is admin-level on the registry. This points that grant at a secret holding `identifiers_api_read`, a MySQL user with `SELECT` on `identifiers.identifiers` and nothing else.

`modules/id-minter-rds/identifiers_api_credential.tf` generates the password and stores the credential in a secret. `iam.tf` points the grant at that secret, and its ARN is exported through the `id_minter_rds` map so #6531's Terraform can read it through remote state.

Creating a MySQL user is SQL, so `create_identifiers_api_user.sh` does that part: it reads the secret, creates the user if it is missing, sets its password to match, revokes everything and grants `SELECT`. It runs as the master user over the Data API, so it needs no VPC access, and re-running it is safe.

The password does not rotate. To change it, replace `random_password.identifiers_api_read`, apply, and run the script again. #6533's `with rotation` is therefore unmet. Terraform generating the password also means it is held in Terraform state.

### Checklist

- [x] Does this change the display model the catalogue API returns? No, nothing here touches the pipeline or the transformers.

## How to test

Plan first:

```bash
cd infrastructure/critical && ./run_terraform.sh plan
```

Expect three additions, a `random_password`, a secret and its version, plus one in-place update to the inline IAM policy. Nothing should be replaced or destroyed.

After applying, set the database up to match:

```bash
./create_identifiers_api_user.sh \
  identifiers-v2-serverless-2026-07-03 \
  rds/identifiers-v2-serverless-2026-07-03/identifiers_api_read
```

The script checks its own work: it confirms the new user can run one of the API's two lookups, and that a `DELETE` is refused by the database.

A lookup succeeds with the new credential, and a write is refused at the database rather than only by the `_assert_read_only` guard in the application. Both are checked by the script.

The rest of #6533 needs #6531's Lambda, which is what makes the deployed API read through this credential, so #6533 stays open after this merges.

`infrastructure/critical` is shared with the production id-minter and carries deletion protection. Every new resource shares the `count` driven by `data_api_consumer_role_arns`, which is set only on the 2026-07-03 cluster, so the production cluster gets nothing. The only change to an existing resource is one statement in an inline IAM policy.

Narrowing that grant is safe now because no Lambda holds the role yet, so no running service loses access to a secret it is still configured to read.

Changing the password later is a two-step operation with an outage between the steps. Applying a replaced `random_password` leaves the secret holding a password the database does not accept, and lookups fail until the script runs. The order is forced, because the script reads the password out of the secret. That is the operational cost of not rotating, and it does not apply yet because the Lambda from #6531 does not exist.

### DIFF
--- a/infrastructure/critical/README.md
+++ b/infrastructure/critical/README.md
@@ -23,3 +23,17 @@ aws stepfunctions start-execution \
 ```
 
 The `id` field must contain only letters, digits, and hyphens, and the resulting `ExportTaskIdentifier` (`id-exp-{id}`) must be at most 60 characters. When triggered by EventBridge, the event ID (a UUID) is used automatically.
+
+## Identifiers API read-only credential (id-minter)
+
+The Identifiers API reads the registry as `identifiers_api_read`, a user with `SELECT` on one table, rather than as the master user. Terraform creates the secret and configures its rotation; the database user and the secret's first value come from `create_identifiers_api_user.sh`, so that no password passes through terraform state.
+
+After applying to a cluster with `data_api_consumer_role_arns` set, run the script with platform account credentials, passing the cluster and the secret terraform created:
+
+```bash
+./create_identifiers_api_user.sh \
+  identifiers-v2-serverless-2026-07-03 \
+  rds/identifiers-v2-serverless-2026-07-03/identifiers_api_read
+```
+
+Each run resets the password and rewrites the secret, so it is safe to repeat, and it needs repeating for any other cluster the API reads. Rotation does not fire on creation because the secret is empty until the script has run, so trigger the first one with `aws secretsmanager rotate-secret`.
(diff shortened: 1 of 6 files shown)


kenoir:
I think the infra complexity here to enable secret rotation doesn't stand up to the real risk of compromise. The credentials are read-only, the database is not available in a public subnet, and there is no private or secret data in the database so there is no data exfiltration risk.

I would remove the rotation and simplify things a bit.

kenoir (approved):

PR #3640, opened 2026-09-10
Folio transformer: add description field

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6546
https://github.com/wellcomecollection/platform/issues/6568

Makes a few tweaks to the Python MARC `description` extractor. Translates existing Scala unit tests into Python BDD tests to verify that Python behaviour matches Scala behaviour. Cases where Python implementation deliberately diverges from Scala are codified at the end of the feature file.

The following description extractor tweaks are implemented:
* Improved blank field handling: do not emit empty paragraphs (`<p></p>`) or stray blank separators
* Non-repeatable subfield handling: when encountering repeated instances of a non-repeatable subfield, take the first one and log an error. This is stricter than current Python logic (which concatenates values from all repeated subfields) but it's less strict than Scala (which refuses to transform the whole work), and it matches what the Python transformer does in other places. There are no live Axiell or EBSCO records with a repeating non-repeatable field in MARC 520, so this change does not make a difference with current production data.

## How to test

I transformed a snapshot of all Folio works and compared the extracted descriptions against those stored on Sierra works in the source index. This produced 761,604 matches and 20 differences, after normalising Sierra data to NFC (see below). All 20 differences were attributable to raw data, not the transformer.

**NFC normalisation**
All Sierra descriptions were normalised using `unicodedata.normalize("NFC", description)` before they were compared to Folio descriptions. Sierra uses mixed unicode normalisation, where some diacritics are stored as a decomposed combining sequence (base character followed by separate combining accent or modifier), and some diacritics are precomposed. Folio descriptions use pure NFC.

The description field on Folio works should match its counterpart on Sierra works.


catalogue_graph/src/adapters/transformers/utils/html.py
```diff
@@ -91,10 +91,10 @@ def is_url(maybe_url: str) -> bool:
...
-    """Format $u subfield contents as an HTML link if valid URL, otherwise return contents unchanged."""
+    """Format $u subfield contents as an HTML link if valid URL, otherwise return the trimmed contents."""
     trimmed = maybe_url.strip()
     if is_url(trimmed):
         return f'<a href="{trimmed}">{trimmed}</a>'
 
-    logger.warning("$u subfield doesn't look like a URL", value=maybe_url)
-    return maybe_url
+    logger.warning("$u subfield doesn't look like a URL", value=trimmed)
+    return trimmed
```
kenoir, line 100: This helper is also what `marc/notes.py` uses for ǂu, so the trim lands in notes too. That is a small step away from the Scala `MarcNotes`, which leaves a non-URL ǂu untrimmed inside the note and only trims the note as a whole, whereas `MarcDescription` trims it the way you have here. I think the trimmed version is the better output in both places, so no change requested.

catalogue_graph/src/adapters/transformers/marc/description.py
```diff
@@ -10,32 +10,51 @@
...
+        if code not in ("a", "b", "c"):
+            continue
+        if code in seen:
+            logger.error(
+                "Repeated non-repeating subfield in field 520",
+                subfield=code,
+                field=str(field),
+            )
+            continue
+        seen.add(code)
```
kenoir, line 56: One edge I noticed while poking at this: if the first ǂa is blank and a later one is populated, the blank one wins and is then filtered out, so the populated value is lost (`$a="   " $b="x" $a="real"` gives `<p>x</p>`). Still far kinder than the Scala, which drops the whole work. Could you let me know whether you'd rather take the first non-blank occurrence, or why the current behaviour is the right call? I'm fine with either, I'd just like it to be a deliberate choice.

kenoir (approved):
Looks good. The port is faithful where it claims to be and the divergences from the Scala are documented and accurate. One question inline on the repeated-subfield handling that I'd like a response to before merging, either way is fine.

catalogue_graph/src/adapters/transformers/marc/description.py
StepanBrychta: I think that taking the first non-blank value is preferable. I added two new reusable helpers (`non_repeatable_subfield` and `non_repeatable_subfields`), allowing us to handle repeated instances of non-repeatable fields consistently across all transformers (take the first non-blank value and log an error). I think it's unlikely this combination of edge cases exists in the real data, but handling it can't hurt.

PR #3643, opened 2026-09-10
Connect axiell folio sync to folio dev sandbox

anubhavijay:
### What does this change?

The `axiell_folio_sync` Lambda could only ever write to the production FOLIO tenant. This lets a
run target the FOLIO dev sandbox instead, so mapping and upsert behaviour can be exercised against
a throwaway tenant using the deployed function, its IAM role and the real Step Functions path.

**Choosing a target.** A run picks its FOLIO instance from `folio_target` on the event (`"prod"` or
`"dev"`), falling back to the `FOLIO_TARGET` environment variable, then to `"prod"`. Each target has
its own SSM SecureString, and there is no fallback between them: a `dev` run whose parameter is
missing fails rather than quietly writing to production. The resolved target is logged, and returned
on the response and in the run manifest, so every run records which instance it wrote to.

**Reaching the sandbox.** The sandbox has no public endpoint, so the Lambda needs to be attached to
the VPC. That is gated behind the `folio_dev_target_enabled` input, which defaults to false, and
places the ENIs in the catalogue VPC private subnets. With it off, `vpc_config` is `null` and
nothing about the production path changes.

**Scheduled runs are insulated.** The state machine maps `folio_target` and `hard_delete` through to
the Lambda, and resolves `folio_target` from `folio_default_target` at apply time rather than
reading the Lambda's environment. Scheduled runs therefore stay on production while someone is
testing against the sandbox.

**Two fixes that affect production as well as the sandbox:**

- `lambda_memory_mb` goes from 512 to 2048. Loading the reference-data cache alone peaks at around
  870 MB against the prod tenant, so 512 MB was OOM-killing runs before any record was processed.
- The `cloudwatch:PutMetricData` grant was scoped to a namespace (`AxiellFolioSync`) that the report
  does not use; it publishes to `catalogue_adapters`. Dry runs skip metric publishing, so this was
  only reachable on a live run, where it failed *after* the FOLIO writes had landed. The first real
  production write would have hit it.

**Security group ownership.** The `:8000` ingress rule on `folio-sandbox-sg` lives in
[`folio-dev-server`](https://github.com/wellcomecollection/folio-dev-server) rather than here,
behind a `sync_lambda_security_group_id` variable, because that repo declares the group's ingress
with inline blocks and so owns it exclusively. A rule created from this repo would be deleted by its
next apply. **That change needs to merge alongside this one**, or the dev path has no route to Kong.

Also included: a `--folio-target` flag on the local CLI, tests for target resolution and the handler
wiring, and `docs/axiell-folio-sync-lambda-dev-instance.md` covering the setup, the network path and
the failure modes seen while building it.

## How to test

1. `uv run pytest tests/adapters/steps/axiell_folio_sync/` — 71 tests.
2. `terraform plan` in `catalogue_graph/infra/adapters`. With `folio_dev_target_enabled` unset, the
   sandbox resources are not created and `vpc_config` is null.
3. Against production, read-only:
   ```bash
   aws lambda invoke --region eu-west-1 \
     --function-name axiell-folio-sync-adapter-lambda \
     --cli-binary-format raw-in-base64-out \
     --payload '{"job_id":"smoke-prod","folio_target":"prod","dry_run":true,"sample_limit":1}' \
     /dev/stdout
   ```
4. Against the sandbox, following the enabling steps in the doc, with `"folio_target":"dev"`.

Verified against the live account on 2026-09-14:

| Run | Result |
| --- | --- |
| prod, `dry_run: true` | `total_errors: 0`, only `api-wellcome.folio.ebsco.com` contacted, 723 MB |
| dev, `dry_run: true` | `total_errors: 0`, only `172.31.134.237:8000` contacted, 867 MB |
| dev, `dry_run: false` | `updated: 3`, `total_errors: 0`, manifest written, 870 MB |

Ongoing, `folio_target` on every response and manifest makes the instance a run wrote to visible
without reading logs, and its absence identifies a deployment too old to honour the field.

- **Production is unchanged when the flag is off.** `vpc_config` is null, no dev parameter is
  created, and `FOLIO_TARGET` is `prod`. The exceptions are the two fixes above, both of which
  correct failures that already affected production runs.
- **Attaching to a VPC removes the Lambda's default internet egress.** The catalogue private subnets
  provide NAT for SSM, KMS, S3 Tables and CloudWatch, and an S3 gateway endpoint so Iceberg reads
  bypass NAT. Both were confirmed against the account.
- **Writing to the wrong tenant.** Targets never fall back to one another, the SecureStrings are
  separate, the scheduled path resolves its target at apply time, and `folio_default_target` is
  clamped to `prod` unless `folio_dev_target_enabled` is also set.
- **The dev password never enters Terraform.** The dev parameter is seeded with placeholders and
  filled in out of band, so no credential is read into a plan or into `adapters.tfstate`.
- **The ENI security group is not destroyed when the flag is turned off.** Lambda deletes its ENIs
  asynchronously, so destroying the group during detachment fails with `DependencyViolation`. It is
  left in place, empty and free, which also keeps its id stable for `folio-dev-server`.
- **The sandbox is shut down outside working hours** by its own scheduler, so a dev-targeted run
  against a stopped instance fails with a connection timeout. This only affects runs that opt in.

### DIFF
--- a/catalogue_graph/docs/axiell-folio-sync-lambda-dev-instance.md
+++ b/catalogue_graph/docs/axiell-folio-sync-lambda-dev-instance.md
@@ -1,0 +1,110 @@
+# Connecting the sync Lambda to the FOLIO dev server
+
+The `axiell_folio_sync` Lambda can write to either FOLIO instance:
+
+| Target | Instance | Reached over |
+| --- | --- | --- |
+| `prod` (default) | EBSCO SaaS, `api-wellcome.folio.ebsco.com` | public internet |
+| `dev` | the `folio-sandbox` EC2 instance, Kong on `:8000` | the catalogue VPC |
+
+The sandbox has no public IP and no inbound ports, so the Lambda has to be attached to the VPC to
+reach it. That is off by default and gated behind `folio_dev_target_enabled`.
+
+## How a run picks its target
+
+Highest precedence first:
+
+1. `folio_target` on the event, either `"prod"` or `"dev"`
+2. the `FOLIO_TARGET` environment variable on the Lambda
+3. `"prod"`
+
+The two entry points take their default from different places:
+
+- **Scheduled runs** (Axiell adapter, then `axiell.adapter.completed`, then EventBridge, then Step
+  Functions) resolve `folio_target` inside the state machine definition, which is baked at apply
+  time from `folio_default_target`. These runs never read the Lambda's environment.
+- **Direct `lambda invoke`** falls through to `FOLIO_TARGET`, which
+  `scripts/folio_dev_session.sh` sets.
+
+Each target has its own SSM SecureString, and there is no fallback between them. A `dev` run whose
+parameter is missing or incomplete fails instead of writing to production.
+
+## What is deployed
+
+| Where | What |
+| --- | --- |
+| `infra/adapters/folio_dev_sandbox.tf` | `folio_dev_target_enabled`, the `folio-sandbox-sg` lookup, the ENI security group, its egress rule, and the `:8000` ingress rule |
+| `infra/adapters/main.tf` | `folio_default_target`, and the subnets and security groups passed to the module |
+| `modules/axiell_folio_sync/lambda.tf` | `vpc_config` when the flag is set and `null` otherwise, plus `FOLIO_TARGET` and `OKAPI_DEV_SECRET_PARAM` in the environment |
+| `modules/axiell_folio_sync/ssm.tf` | `…/okapi_credentials_dev`, seeded with placeholders |
+| `modules/axiell_folio_sync/iam.tf` | `ssm:GetParameter` extended to the dev parameter |
+| `modules/axiell_folio_sync/state_machine.tf` | `folio_target` and `hard_delete` mapped into the Lambda payload |
+| `folio/okapi.py` | `resolve_folio_target`, and `load_okapi_config(target)` choosing the parameter |
+| `models.py` | `FolioTarget`, and `folio_target` on the event |
+| `scripts/folio_dev_session.sh` | starts the sandbox, populates the dev SecureString, switches `FOLIO_TARGET` |
+
+The Lambda runs at 2048 MB. Loading the reference-data cache on its own peaks at roughly 870 MB
+against the prod tenant and 720 MB against `diku`, so 512 MB is not enough to finish a run.
+
+## Network path
+
+With `folio_dev_target_enabled` set, the Lambda's ENIs sit in
+`catalogue_vpc_delta_private_subnets`. Those are the same subnets as the adapter ECS tasks, and
+the same VPC as the sandbox, so traffic to it stays local. Attaching to a VPC removes the Lambda's
+default internet egress, and these subnets replace it:
+
+| Hop | Provided by |
+| --- | --- |
+| Lambda ENI to the sandbox on `:8000` | `172.31.0.0/16 → local`, plus the ingress rule below |
+| SSM, KMS, S3 Tables, CloudWatch metrics, FOLIO SaaS | NAT gateway on the subnets' route table |
+| S3 (manifests, Iceberg parquet) | S3 gateway endpoint on the same route table, bypassing NAT |
+| CloudWatch Logs | written by the Lambda service, not through the ENI |
+
+`folio-sandbox-sg` carries exactly one ingress rule, created here: tcp `:8000` from the sync
+Lambda's security group. FOLIO's gateway listens on `0.0.0.0:8000` over plain HTTP, so the OKAPI
+url is `http://<private-ip>:8000`.
+
+## Enabling it
+
+1. Set `folio_dev_target_enabled = true` in `folio_dev_sandbox.tf` and apply.
+2. Run `scripts/folio_dev_session.sh on`. That starts the sandbox if it is stopped, writes the dev
+   SecureString using the instance's live private IP and the
+   `folio-sandbox/diku-admin-password` Secrets Manager entry, and sets `FOLIO_TARGET=dev`.
+3. Invoke the Lambda directly:
+
+   ```bash
+   aws lambda invoke --region eu-west-1 \
+     --function-name axiell-folio-sync-adapter-lambda \
+     --cli-binary-format raw-in-base64-out \
+     --payload '{"job_id":"folio-dev-smoke","dry_run":true,"sample_limit":1}' \
+     /dev/stdout
+   ```
+
+`scripts/folio_dev_session.sh off` reverses steps 2 and 3. `status` reports the sandbox state and
+the current target.
+
+A successful run logs `folio target resolved` along with the url it chose. That is the quickest
+way to confirm which instance was written to.
+
+## Credentials
+
+Terraform creates `…/okapi_credentials_dev` with placeholder values and ignores later changes to
+it. The real value is written outside Terraform by `folio_dev_session.sh`, so the sandbox password
+is never read into a plan or into `adapters.tfstate`. The script rewrites the parameter on every
+`on`, because the url follows the instance's private IP and that changes whenever the sandbox is
+rebuilt.
+
+## Known failure modes
+
+| Symptom | Cause |
+| --- | --- |
+| `ConnectTimeout` on the first FOLIO call | the sandbox is stopped, or Kong has not finished starting. Kong lags the instance reaching `running` by several minutes |
+| `ConnectTimeout` while the sandbox is up | the dev SecureString holds a stale url after a rebuild. Rerun `folio_dev_session.sh on` |
+| A run reaches production despite `folio_target: "dev"` | the deployed image predates the field, and Pydantic drops unknown keys silently. Check whether the `folio target resolved` line appears in the logs |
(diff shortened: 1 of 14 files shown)


catalogue_graph/infra/adapters/folio_dev_sandbox.tf
```diff
@@ -0,0 +1,59 @@
...
+# modules/axiell_folio_sync/ssm.tf, and populated by
+# scripts/folio_dev_session.sh. A run reaches the sandbox by passing
+# {"folio_target": "dev"} on the event. Runs without it go to prod.
+#
+# See docs/axiell-folio-sync-lambda-dev-instance.md.
+locals {
+  folio_dev_target_enabled = true
+
+  folio_dev_api_port = 8000 # Kong/Eureka gateway, plain HTTP
+}
```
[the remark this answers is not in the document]
kenoir: [concern] Agreeing with this one. `folio_dev_target_enabled = true` at line 16 means the next apply of the adapters root attaches the production sync Lambda to the VPC, creates the dev SecureString, and adds the ingress rule, whether or not anyone is testing.

The PR description and the doc both read as though it merges as `false`: the test steps say "with `folio_dev_target_enabled = false` the only change should be the Lambda's `memory_size` and `FOLIO_TARGET=prod`", and "Enabling it" step 1 in the doc is "set `folio_dev_target_enabled = true` ... and apply". I think this is just a local test value that got committed, but it is worth flipping before merge since the whole "production is unchanged when the flag is off" argument depends on it.

catalogue_graph/infra/adapters/folio_dev_sandbox.tf
```diff
@@ -0,0 +1,59 @@
...
+
+# The sandbox's security group, owned by
+# wellcomecollection/aws-account-infrastructure and matched by name.
+data "aws_security_group" "folio_dev" {
+  count = local.folio_dev_target_enabled ? 1 : 0
+
+  name   = "folio-sandbox-sg"
+  vpc_id = local.vpc_id
+}
```
kenoir, line 32: [suggestion] Small correction: `folio-sandbox-sg` is created by `wellcomecollection/folio-dev-server` (`terraform/main.tf`, `resource "aws_security_group" "folio"`, named `"${var.name}-sg"`), not by `aws-account-infrastructure`. Same attribution in the PR description and in the doc, worth updating all three so the next person looks in the right place.

Related, while the flag is on this data source makes the adapters root depend on the sandbox existing. `aws_security_group` is a singular data source, so no match is a hard error ("no matching EC2 Security Group found", terraform-provider-aws `internal/tfresource/not_found_error.go:100`), which would block a plan of the whole root including unrelated adapter changes. Not an issue once the flag defaults to `false`, but it is an argument for keeping dev sessions short.

catalogue_graph/infra/adapters/folio_dev_sandbox.tf
```diff
@@ -0,0 +1,59 @@
...
+# reach Kong on :8000.
+resource "aws_vpc_security_group_ingress_rule" "folio_dev_from_sync" {
+  count = local.folio_dev_target_enabled ? 1 : 0
+
+  security_group_id            = data.aws_security_group.folio_dev[0].id
+  referenced_security_group_id = aws_security_group.folio_sync_dev[0].id
+  from_port                    = local.folio_dev_api_port
+  to_port                      = local.folio_dev_api_port
+  ip_protocol                  = "tcp"
+  description                  = "Kong/Eureka API from the Axiell to FOLIO sync Lambda"
```
kenoir, line 63: [concern] The risk you flagged in the description ("if that repo manages ingress inline, its next apply could remove the rule") is not hypothetical, it is what will happen. `wellcomecollection/folio-dev-server` `terraform/main.tf` declares the SG with inline rules:

```hcl
resource "aws_security_group" "folio" {
  name = "${var.name}-sg"
  egress { ... }
  dynamic "ingress" {
    for_each = var.allowed_ssh_cidr != "" ? [1] : []
    ...
  }
}
```

`allowed_ssh_cidr` defaults to `""`, so as far as that config is concerned the SG has zero ingress rules. Inline rules are authoritative, and the provider docs are explicit: "You should not use the `aws_security_group` resource with in-line rules ... in conjunction with the `aws_vpc_security_group_egress_rule` and `aws_vpc_security_group_ingress_rule` resources ... Doing so may cause rule conflicts, perpetual differences, and result in rules being overwritten" (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group).

I think the fix is to move the rule into folio-dev-server behind its own variable, so one repo owns that SG's ingress and there is no conflict to reason about. Documenting it here as a known failure mode would work too, but it leaves a rule that silently disappears whenever someone applies the other repo, which is the kind of thing that costs an afternoon the third time it happens.

catalogue_graph/docs/axiell-folio-sync-lambda-dev-instance.md
```diff
@@ -0,0 +1,110 @@
...
+`on`, because the url follows the instance's private IP and that changes whenever the sandbox is
+rebuilt.
+
+## Known failure modes
+
+| Symptom | Cause |
+| --- | --- |
+| `ConnectTimeout` on the first FOLIO call | the sandbox is stopped, or Kong has not finished starting. Kong lags the instance reaching `running` by several minutes |
+| `ConnectTimeout` while the sandbox is up | the dev SecureString holds a stale url after a rebuild. Rerun `folio_dev_session.sh on` |
+| A run reaches production despite `folio_target: "dev"` | the deployed image predates the field, and Pydantic drops unknown keys silently. Check whether the `folio target resolved` line appears in the logs |
```
kenoir, line 103: [suggestion] This row documents the one path where a dev-intended run can write to the production tenant, and the check for it is "go and read the logs afterwards". Since `AxiellFolioSyncResponse` is what `lambda invoke` prints, would it be worth adding `folio_target` to the response (and to the manifest) as well as logging it? On a stale image the field is simply absent from the response, so the caller sees the problem in the output they are already looking at rather than having to remember to check CloudWatch. Dry run is the default so the blast radius is small, but `dry_run: false` plus a stale image is the case this would catch.

catalogue_graph/scripts/folio_dev_session.sh
```diff
@@ -0,0 +1,132 @@
...
+    --environment "{\"Variables\":$env_json}" >/dev/null
+  aws_ lambda wait function-updated --function-name "$FUNCTION_NAME"
+  echo "FOLIO_TARGET is now '$target'"
+}
+
+# Populate the dev OKAPI SecureString, which Terraform only seeds with
+# placeholders. Rewritten on every
+# `on`, because a rebuilt sandbox changes the url and a stale one fails as a
+# connection timeout.
+refresh_dev_credentials() {
```
kenoir, line 73: [suggestion] Stepping back from the individual lines: I think most of this script could be designed away, and it might be worth doing before it becomes the documented workflow.

The credential refresh exists because the sandbox's private IP is treated as volatile, but it mostly is not. EC2 keeps an instance's primary private IPv4 for the life of the instance, including stop and start, so the daily `on`/`off` cycle never moves it. Your comment at line 31 has the real trigger right, it is a rebuild. And folio-dev-server already goes out of its way to avoid rebuilds, with `ignore_changes = [ami, user_data]` on `aws_instance.folio` so the weekly AMI bake does not replace a working box.

If you pin the address so it survives a rebuild too, by setting `private_ip` on that instance, or by moving it onto a standalone `aws_network_interface` that outlives the instance, then the dev SecureString stops being a moving target. Terraform can seed it once from a `aws_secretsmanager_secret_version` data source on `folio-sandbox/diku-admin-password`, which also removes the `"placeholder"` values and the `jq`-password handling here.

That leaves two jobs. Starting and stopping the instance is already covered by `scripts/up.sh` and `down.sh` in folio-dev-server, alongside the 18:00 auto-stop schedule. And on the `FOLIO_TARGET` flip, is it needed at all? `resolve_folio_target` puts the event field ahead of the env var, so a manual invoke can pass `{"folio_target": "dev"}` and reach the sandbox without mutating the deployed function. That mutation is the one thing here that leaves the Lambda out of step with Terraform, which your header comment already flags as a hazard, so dropping it would remove the hazard rather than document it.

kenoir (changes requested):
Read through this one carefully. The target-resolution design is the strong part: separate SSM parameters per target with no fallback between them, and tests that assert the prod parameter is never even fetched on a dev run, is a good way to make "wrote to the wrong tenant" hard to do by accident. `terraform validate` and `fmt` are clean and all 59 tests pass locally.

Requesting changes on one thing only. `folio_dev_target_enabled` is hard-coded to `true`, so the next apply of the adapters root attaches the production sync Lambda to the VPC and creates the security groups and the ingress rule, with nobody testing anything. Both the PR description and the doc describe the merge state as `false`, so I read this as a local test value that got committed. Flipping it clears the request.

Two smaller points in the threads. `folio-sandbox-sg` is owned by folio-dev-server rather than aws-account-infrastructure, and that repo declares the group with inline rules, which means its next apply removes the ingress rule added here. Separately, I think the session script could mostly be designed away by pinning the sandbox's private IP, though that is a suggestion rather than a blocker.

kenoir (approved):

PR #3644, opened 2026-09-10
Take the terms-of-use access word from the status, not the subfield the date came from

agnesgaroux:
Axiell holds the restricted-until and closed-until date in the same `506 $g` subfield, confirmed with collections staff. The transformer was reading `506 $g` as a closed-until date and `540 $g` as a restricted-until date, so a Restricted record whose date sat in `506 $g` matched neither the Closed branch (its status is not Closed) nor the Restricted branch (no `540 $g` date). It fell through to the catch-all, which appended a closed-until sentence to text saying the item is restricted:

> This item is restricted. You will need to complete a Restricted Access form agreeing to anonymise personal data before you will be allowed to view this item. [...] **Closed until 1 January 2054.**

That affected 5,564 records. With one date field and the status deciding which word introduces it, the catch-all now fires on 15.

## The status can no longer be derived from the date

`extract_access_status` had a fallback: a record with no usable status but a future closed-until date was treated as Closed. That worked when the date could only be a closed-until date. Now that `506 $g` carries either kind, a date on its own no longer tells us whether the item is closed or restricted, so the fallback is gone.

The consequence is that a date needs a status to label it. Where the status says neither closed nor restricted, the date is dropped and the conditions text is published unchanged, rather than asserting an access status the record does not give.

## The 15 records that still reach the catch-all

| `506 $f` | records | what the conditions say |
|---|---:|---|
| `OPEN` | 7 | "This item was previously restricted until 01/01/2026. It was reassessed..." |
| `MISSING` | 7 | Winnicott Trust permission, restricted-access forms, one "This item is closed and cannot be accessed" |
| `PERMISSIONREQUIRED` | 1 | Winnicott Trust permission, but the conditions do not mention restrictions, so the guard does not admit it |

Dropping the date is right for the seven `OPEN` records: the item is open now and the conditions already explain that it used to be restricted until that date, so appending the sentence would contradict the status.

The seven `MISSING` ones are worth a look by collections rather than in code. `MISSING` maps to `Unavailable`, yet their conditions describe live access arrangements, including one asserting the item is closed. Ids are in the issue.

## Why removing the fallback is safe on the current data

863 records have no access status the transformer can resolve: 25 carrying the unmapped value `DATAISSUES`, and 838 with no `506 $f` at all. **None of them has a `506 $g` date**, so no record in the corpus was relying on the fallback to acquire a status. All 8,843 Closed items in the round 3 index get their status from a genuine `CLOSED` value in `$f`, against production's 8,828, so #6476's parity does not depend on it either.

The fallback's own comment described it as a bridge held "until the next data load restores the full status vocabulary". On this data that vocabulary is present.

If a status-less record with a date does appear in a future load, the catch-all warning is what surfaces it. That is why the warning keeps its existing wording rather than being renamed to describe one specific cause.

## Also in this branch

`IGNORED_PREFIXES` in `marc/other_identifiers.py` contained `"AV  barcode"` with two spaces, and all 412 AV barcode identifiers in the corpus (on 311 records) use one, so none matched and each logged an "Unknown identifier prefix" warning. Those identifiers are dropped either way, since the prefix has no type mapping; the entry only controls whether the drop is logged. The one-character fix removes 2,040 warning lines, 27% of the transformer's warnings over a four-day window. Discussed with the team and small enough to carry here.

## Tests

`extract_terms_of_use` collapses from four branches to one, since the only thing that varied between Closed, Restricted and PermissionRequired was the word introducing the date. Verified byte-identical against the pre-refactor code across all 209,608 records in the Axiell adapter store.

Four tests were passing without exercising the behaviour they described, and mutation testing was what surfaced them:

- two wrote their date to `540 $g` using a `dd/mm/yyyy` value the ISO parser rejects, so the record had no date at all and took the "no dates" branch
- two supplied no `506 $f`, so they were relying on the removed status fallback rather than testing de-duplication

The fixture now takes a single `until` parameter writing `506 $g`, matching the code. The tests for the removed fallback are gone, in `test_access_status.py` and in `items.feature`, along with the feature header sentence describing it. Two scenario names claimed a precedence between `$f` and `$g` that no longer exists and now say what they check.

### DIFF
--- a/catalogue_graph/src/adapters/transformers/axiell/access_status.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/access_status.py
@@ -5,12 +5,9 @@ Extract access status from field
 https://www.loc.gov/marc/bibliographic/bd506.html
 """
 
-from datetime import date
-
 import structlog
 from pymarc.record import Record
 
-from adapters.transformers.axiell.dates import extract_closed_until_date
 from adapters.transformers.marc.common import non_empty_subfields
 from adapters.transformers.marc.identifier import extract_id
 from models.pipeline.access_status import (
@@ -64,14 +61,6 @@ def extract_access_status(record: Record) -> AccessStatus | None:
 
     if status in ACCESS_STATUS_MAPPING:
         return ACCESS_STATUS_MAPPING[status]
-
-    # Most closed material does not (yet) carry a CLOSED status value: the Calm-to-Axiell
-    # migration delivered Calm's 'Closed' as 'Certain restrictions apply' until the next
-    # data load restores the full status vocabulary. For records without a usable status,
-    # a 'closed until' date in the future marks the item as closed.
-    closed_until = extract_closed_until_date(record)
-    if closed_until and closed_until >= date.today():
-        return Closed
 
     if status is not None:
         logger.warning(
--- a/catalogue_graph/src/adapters/transformers/axiell/dates.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/dates.py
@@ -28,12 +28,8 @@ def _date_from(record: Record, field: str, subfield: str) -> date | None:
     return _try_parse_date(value) if value else None
 
 
-def extract_closed_until_date(record: Record) -> date | None:
+def extract_restricted_or_closed_until_date(record: Record) -> date | None:
     return _date_from(record, "506", "g")
-
-
-def extract_restricted_until_date(record: Record) -> date | None:
-    return _date_from(record, "540", "g")
 
 
 def _parse_production_date(value: str, month: int, day: int) -> date | None:
--- a/catalogue_graph/src/adapters/transformers/axiell/terms_of_use.py
+++ b/catalogue_graph/src/adapters/transformers/axiell/terms_of_use.py
@@ -5,13 +5,15 @@ import structlog
 from pymarc.record import Record
 
 from adapters.transformers.axiell.access_status import extract_access_status
-from adapters.transformers.axiell.dates import (
-    extract_closed_until_date,
-    extract_restricted_until_date,
-)
+from adapters.transformers.axiell.dates import extract_restricted_or_closed_until_date
 from adapters.transformers.marc.common import first_non_empty_subfield
 from adapters.transformers.marc.identifier import extract_id
-from models.pipeline.access_status import Closed, PermissionRequired, Restricted
+from models.pipeline.access_status import (
+    AccessStatus,
+    Closed,
+    PermissionRequired,
+    Restricted,
+)
 
 logger = structlog.get_logger(__name__)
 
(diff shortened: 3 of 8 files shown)


catalogue_graph/src/adapters/transformers/axiell/access_status.py
```diff
@@ -65,14 +62,6 @@ def extract_access_status(record: Record) -> AccessStatus | None:
     if status in ACCESS_STATUS_MAPPING:
         return ACCESS_STATUS_MAPPING[status]
 
-    # Most closed material does not (yet) carry a CLOSED status value: the Calm-to-Axiell
-    # migration delivered Calm's 'Closed' as 'Certain restrictions apply' until the next
-    # data load restores the full status vocabulary. For records without a usable status,
-    # a 'closed until' date in the future marks the item as closed.
```
agnesgaroux, line 71: "CLOSED status" fix landed in round 3

catalogue_graph/src/adapters/transformers/marc/other_identifiers.py
```diff
@@ -48,7 +48,7 @@ def extract_other_identifiers(record: Record) -> list[SourceIdentifier]:
     "Previous number",
     "Previouse number",
     "Archivematica UUID",
-    "AV  barcode",
+    "AV barcode",
```
agnesgaroux, line 51: Unrelated piggy backing, as discussed
This was firing warnings in the transformer for no good reason

catalogue_graph/src/adapters/transformers/axiell/dates.py
```diff
@@ -28,14 +28,10 @@ def _date_from(record: Record, field: str, subfield: str) -> date | None:
     return _try_parse_date(value) if value else None
 
 
-def extract_closed_until_date(record: Record) -> date | None:
+def extract_restricted_or_closed_until_date(record: Record) -> date | None:
     return _date_from(record, "506", "g")
```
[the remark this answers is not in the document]
agnesgaroux: Done in https://github.com/wellcomecollection/catalogue-pipeline/pull/3644/changes/8f19364ca8a7bf28ad81a700899bd8c0dcf26101

catalogue_graph/tests/adapters/transformers/axiell/test_access_status.py
```diff
@@ -31,20 +33,8 @@ def test_restrictionsapply_maps_to_restricted() -> None:
...
+    # whatever else the 506 field carries.
     record = make_axiell_record()
-    add_506(record, "g", "2999-01-01")
-    assert extract_access_status(record) == Closed
-
-
-def test_no_status_with_past_closed_until_is_none() -> None:
-    record = make_axiell_record()
-    add_506(record, "g", "2001-01-01")
     assert extract_access_status(record) is None
```
kenoir, line 40: I think this test would still pass if the date fallback came back: with no `506 $g` on the record, the old code returned `None` here too. I tried re-adding the fallback locally and the whole Axiell suite stayed green (213 passed), so nothing currently pins the "a date on its own no longer gives a status" behaviour the PR description leans on. Would it be worth giving this record a future `$g` (and maybe keeping one case with an unrecognised `$f` plus a future `$g`, which the deleted `test_unrecognised_status_with_future_closed_until_is_closed` covered from the other side)?

catalogue_graph/src/adapters/transformers/axiell/terms_of_use.py
```diff
@@ -51,75 +53,63 @@ def _has_restrictions(text: str) -> bool:
...
+        return "Restricted"
+    return None
+
+
+def _already_stated(conditions: str, label: str, until: date) -> bool:
+    """Whether the conditions text already carries both the word and the date."""
+    stated = (
+        _has_restrictions(conditions)
+        if label == "Restricted"
+        else label.lower() in conditions.lower()
```
kenoir, line 85: Was widening the Restricted de-duplication to `_has_restrictions` deliberate? The old Restricted branch and the Scala `CalmTermsOfUse` both matched only the word "restricted", with "restrictions" reserved for the PermissionRequired guard. The only shape it changes is conditions that say "restrictions ... until <date>" without "restricted", where the sentence is now suppressed rather than appended. Seems like a reasonable improvement to me, and if the 209,608-record parity check covered it then there is nothing to do; I just could not tell from the diff whether it was intended.

kenoir (approved):
Read through the logic change and the tests, and ran the Axiell and MARC transformer suites plus ruff and mypy on the branch: 287 passed, all clean. I also compared old and new `extract_terms_of_use` on a set of synthetic 506 records and every difference is the intended one, so the collapse to `_until_label` plus `_already_stated` is faithful to the four branches it replaces. Nice to see the mutation testing surface the four tests that were passing without exercising anything.

Two comments inline. The one I think is worth acting on before merge is the `test_no_access_status_field_maps_to_none` case: as written it does not pin the removal of the date fallback, and re-adding the fallback locally left the whole suite green. The other is a question about the Restricted de-duplication now matching "restrictions" as well as "restricted", which reads like an improvement but was not called out in the description.

Approving on the basis that the first comment gets a response before this merges.

catalogue_graph/src/adapters/transformers/axiell/terms_of_use.py
agnesgaroux: Deliberate, but only as a side effect of keeping PermissionRequired identical, and you're right that the diff doesn't show that.

The two branches de-duplicated differently. Restricted required the word: `"restricted" in conditions.lower() and _contains_date(...)`. PermissionRequired checked only the date, because its branch guard had already required `_has_restrictions`. Collapsing them into one `_already_stated` meant picking one test, and the exact word would have changed PermissionRequired for conditions saying "restrictions" but not "restricted", which its guard admits.

So the widening lands only where you say: status Restricted, an until date, and conditions carrying the date but the word "restrictions" rather than "restricted". Before, the sentence was appended to text that already had the date. Now it isn't.

The 209,608-record parity check does cover it, in the sense that it found zero differences, but that is weaker than it sounds: I have since counted the records that actually reach this shape and there are **0**. So the check passing tells us the shape is absent from the corpus, not that the behaviour is unchanged. Worth stating plainly rather than leaning on the parity number.

Happy to revert it to the exact word and give PermissionRequired its own date-only check instead, if you would rather the refactor changed nothing at all. My preference is to keep it, since the appended duplicate is the worse output of the two, but it is a behaviour change and it should have been in the description.

catalogue_graph/tests/adapters/transformers/axiell/test_access_status.py
agnesgaroux: You're right, and I reproduced it: re-adding the fallback left all 213 green. The test used a record with no `506 $g`, so the old code never reached the inference and the assertion held either way.

Fixed in 21dcd9a and fde5c29, both sides as you suggested:

- `test_future_until_date_without_a_status_is_not_closed`: no `$f`, future `$g`
- `test_future_until_date_with_an_unrecognised_status_is_not_closed`: `$f=PRIVATE`, future `$g`, which is what the deleted `test_unrecognised_status_with_future_closed_until_is_closed` covered

With the fallback restored locally both now fail; without it, 215 pass.

Worth noting the same gap existed in my own mutation run, which is how it got through. I mutated `extract_access_status_value` and `_contains_date` but never tried re-adding the deleted branch, so the mutation set could not reach the thing the test was supposed to pin.

catalogue_graph/src/adapters/transformers/axiell/terms_of_use.py
kenoir: Keep it 👍

PR #3650, opened 2026-09-11
Trial: build an app image on GitHub Actions

kenoir:
Experiment, not for merging. Does an app image build on a hosted runner?

Builds sierra_merger and stops there: no role available to Actions can push these ECR repositories.


kenoir:
Findings captured; superseded by the matcher workflow.

PR #3651, opened 2026-09-11
Trial: run sbt on the runner instead of the wrapper container

kenoir:
Experiment, not for merging. Runs transformer_marc_xml's tests with sbt on the runner instead of the wrapper container, so no Docker, ECR or AWS.

Compare its duration with wellcomecollection/catalogue-pipeline#3648.


kenoir:
Findings captured; superseded by the matcher workflow.

PR #3656, opened 2026-09-11
Protect the CALM adapter store against deletion and add a recovery window

kenoir:
## What does this change?

Turns on DynamoDB deletion protection and point-in-time recovery for `vhs-calm-adapter`, using variables added in v4.3.0 of the VHS module (wellcomecollection/terraform-aws-vhs#22). Nothing else changed in the module between v4.0.5 and v4.3.0.

With the adapter stopped, nothing can rebuild this store, and production reads CALM works straight out of it. Deletion protection has to live in terraform, because it isn't computed and any apply that doesn't declare it turns it off.

## How to test

Applied from this branch on 2026-09-14, and deletion protection and PITR are both confirmed on. The plan also re-read three IAM policies and added default tags to three scalable targets. Two took the tags. The `calm_adapter` target had no ARN, so it couldn't hold them. It gained one when #3659 changed its capacity, so its tag diff should clear on the next apply. Permissions and capacities were confirmed unchanged.

Merge this soon. Until it lands, an apply of this root from main turns deletion protection back off.

Decommissioning CALM will need protection set to `false` and applied before the table can be destroyed.


annak-k (approved):
Looks good. Note that the plan does say there will be one change in place, adding `default_tags` to the `aws_appautoscaling_target`, but the tags cannot be written to that object.

PR #3657, opened 2026-09-11
Add a tool to snapshot a VHS to parquet

kenoir:
## What does this change?

Adds `catalogue_graph/scripts/snapshot_vhs.py`, which writes the latest version of every record in a VHS to parquet by scanning the DynamoDB index and reading the S3 object each row points at. It covers all three VHS stores.

Details worth reviewing:

- The S3 key is read from the row, never rebuilt from the version. The deletion checker bumps the version without writing a new object, which left 29,762 CALM records with a version one ahead of their key.
- The whole DynamoDB row is kept in `index_row`, because Miro holds hand-curated suppression state on the row and nowhere else.
- Sierra's `-deleted` companion table is scanned too, or 557,348 records go missing.
- Unreadable bodies are collected rather than raised on, and fail the run at the end unless `--allow-unreadable` covers them.

## How to test

`uv run pytest tests/test_snapshot_vhs.py` (28 tests). The CALM snapshot was taken from this branch on 2026-09-14 (408,710 rows, 371MB, 13 minutes), checked independently of the tool, and uploaded to `s3://wellcomecollection-platform-infra/vhs_snapshots/vhs-calm-adapter/`.

It only reads. `--limit` writes a file covering part of a store, so it warns, rejects values below 1 and refuses `--upload-to`.

`last_modified` is the S3 object's timestamp, so for a deleted record it is when the last live body was written, not when the record was deleted.

### DIFF
--- a/catalogue_graph/scripts/snapshot_vhs.py
+++ b/catalogue_graph/scripts/snapshot_vhs.py
@@ -1,0 +1,394 @@
+"""Write a parquet snapshot of the latest version of every record in a VHS.
+
+A VHS (versioned hybrid store) keeps an index in DynamoDB and the record
+bodies in S3. The DynamoDB row is the only thing that says which S3 object is
+current: the bucket holds every version a record has ever had, so the snapshot
+scans the table, reads the object each row points at, and writes one parquet
+row per record.
+
+This exists because a VHS becomes the only copy of its source dataset once its
+adapter stops harvesting, which is happening to the CALM store under
+wellcomecollection/platform#6689. The other two stores have the same shape and
+will reach the same point as their sources are retired.
+
+The snapshot is written to a `.partial` file and moved into place only on a
+complete, successful run, so a file at the output path is always whole.
+
+Usage:
+    uv run python scripts/snapshot_vhs.py --store calm --output-path /tmp/calm.parquet
+    uv run python scripts/snapshot_vhs.py --store calm --output-path /tmp/calm.parquet --upload-to s3://bucket/prefix/
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import time
+from collections.abc import Iterator
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+from typing import Any
+
+import boto3
+import pyarrow as pa
+import pyarrow.parquet as pq
+import structlog
+from botocore.config import Config as BotocoreConfig
+from pyiceberg.io.pyarrow import schema_to_pyarrow
+from pyiceberg.schema import Schema
+from pyiceberg.types import (
+    BooleanType,
+    IntegerType,
+    NestedField,
+    StringType,
+    TimestamptzType,
+)
+
+from utils.logger import ExecutionContext, get_trace_id, setup_logging
+
+logger = structlog.get_logger(__name__)
+
+BATCH_SIZE = 10_000
+"""Records per parquet row group. Each batch's bodies are held in memory while
+it is assembled, so this bounds peak usage rather than the record count."""
+
+SCAN_SEGMENTS = 16
+"""Parallel DynamoDB scan segments."""
+
+FETCH_WORKERS = 32
+"""Concurrent S3 GETs. The bodies are small and the run is one-off, so this is
+set for a sensible wall-clock rather than to saturate anything."""
+
+PROGRESS_LOG_EVERY = 25_000
+
+
+# The first six fields match ADAPTER_STORE_ICEBERG_SCHEMA in
+# adapters/utils/schemata.py, so selecting them gives a table that loads
+# through the adapter store's own snapshot path. `version` and `s3_key` are
+# VHS-specific provenance on top.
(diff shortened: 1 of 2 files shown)


catalogue_graph/scripts/snapshot_vhs.py
```diff
@@ -0,0 +1,394 @@
...
+                    logged_at = written
+    except BaseException:
+        # A partial file left next to the output is the one thing that could
+        # be mistaken for a snapshot, so never leave one behind.
+        if os.path.exists(partial_path):
+            os.remove(partial_path)
+        raise
+
+    os.replace(partial_path, output_path)
+    return written
```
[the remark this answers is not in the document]
kenoir: Fixed in 4173ef70c5. `os.replace` now sits inside the cleanup scope, so a failing move removes the partial rather than stranding it. Added a regression test that puts a directory at the output path to make the move fail; it fails without the fix, leaving `calm.parquet.partial` behind exactly as you describe.

catalogue_graph/scripts/snapshot_vhs.py
```diff
@@ -0,0 +1,394 @@
...
+    rows = scan_index(session.resource("dynamodb"), config.table_name)
+
+    if limit is not None:
+        logger.warning(
+            "Smoke test only. The file this writes covers part of the store and "
+            "is not a snapshot of it.",
+            limit=limit,
+            index_rows=len(rows),
+        )
+        rows = rows[:limit]
```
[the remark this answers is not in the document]
kenoir: I think this one is already covered. `snapshot_vhs` rejects the value before it reaches the slice: `if limit < 1: raise ValueError(f"--limit must be at least 1, got {limit}")` runs about ten lines above `rows = rows[:limit]`, and nothing else slices on it. Checked against the running tool rather than by reading: `--limit -1` exits with `ValueError: --limit must be at least 1, got -1` and writes no file. `test_snapshot_vhs_rejects_a_limit_below_one` covers both 0 and -1. Happy to be shown a path that gets past it.

catalogue_graph/scripts/snapshot_vhs.py
```diff
@@ -0,0 +1,496 @@
...
+    """
+    location = item.get("payload") or item.get("location")
+    if not location:
+        raise SnapshotError(f"Row {item.get('id')} has no payload or location")
+
+    return IndexRow(
+        id=item["id"],
+        version=int(item["version"]),
+        bucket=location["bucket"],
+        key=location["key"],
```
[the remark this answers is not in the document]
kenoir: A full scan of both Sierra VHS tables found all 2,377,143 rows use `payload.bucket`/`payload.key`, none use `namespace`/`path`. That fallback handles a shape that no longer exists, so I have left the parser as is.

catalogue_graph/scripts/snapshot_vhs.py
```diff
@@ -0,0 +1,502 @@
...
+    return written
+
+
+def verify_snapshot(output_path: str, rows: list[IndexRow]) -> None:
+    """Re-read the written file and check it against the index it came from.
+
+    This reads the file back from disk rather than trusting the writer, so it
+    catches a truncated or unreadable parquet as much as a miscount.
+    """
+    table = pq.read_table(output_path, columns=["id", "deleted", "content"])
```
agnesgaroux, line 359: Do we need to read "content" here if we're not using it in the verification?

catalogue_graph/scripts/snapshot_vhs.py
```diff
@@ -0,0 +1,502 @@
...
+PROGRESS_LOG_EVERY = 25_000
+
+EPOCH = datetime(1970, 1, 1, tzinfo=UTC)
+"""Stand-in `last_modified` for a row whose object could not be read. The
+column is non-optional, and a row like that is reported separately anyway."""
+
+
+# The first six fields match ADAPTER_STORE_ICEBERG_SCHEMA in
+# adapters/utils/schemata.py, so selecting them gives a table that loads
+# through the adapter store's own snapshot path. A test in
```
agnesgaroux: This makes it sound like we can use the here snapshots to load an iceberg table. However I don't think we have this set up? If we did, would it be an issue that the Sierra snapshot contains duplicated rows from the main table + deleted table

catalogue_graph/scripts/snapshot_vhs.py
kenoir: Reading it proves the column decodes, so a corrupt content column fails verify rather than surfacing later. I have added a comment saying so.

agnesgaroux (approved):

catalogue_graph/scripts/snapshot_vhs.py
kenoir: Good catches. Nothing loads these into Iceberg, so I reworded that comment in 4bf32f9eab. And yes, 95,749 Sierra ids are in both tables, so the combined file would have duplicated them. #3663 snapshots each table to its own file and makes verify fail on duplicate ids.

PR #3658, opened 2026-09-11
Report CALM deletions without marking them

kenoir:
## What does this change?

Adds `catalogue_graph/scripts/report_calm_deletions.py`, which reports the CALM records our adapter store still holds that CALM no longer has. It writes nothing to DynamoDB, publishes nothing, and leaves a CSV to decide from.

The deletion check was switched off along with the adapter, so anything deleted in CALM between the last weekly run and the read-only cutover was never detected and sits in our store as a live work. Closing that gap means running the check, and the check as it stands is not safe to point at a CALM that is being wound down.

`ApiDeletionChecker.test` searches for a batch of record ids and counts the hits, and `defectiveRecords` treats every id the search does not return as deleted. If CALM answers but finds nothing, because the database has been detached or our credentials lose their scope, `d == items.size` and the entire batch is marked deleted and published to `calm_deletion_checker_output`, which the production pipeline's CALM transformer subscribes to. Nothing in that path has a threshold. A silent CALM failure would take the archive out of the live catalogue.

Three guards, in the order they run:

- A control probe searches for record ids drawn from our own store and requires CALM to return all of them. This is the one that catches a CALM answering but finding nothing.
- Any batch that comes back wholly missing aborts the run. At 512 records that is a far better match for an API fault than for reality.
- Records already marked deleted are skipped, matching the service, which diverts them before testing. That is 150,075 of the 408,710, so the sweep covers 258,635.

`find_missing` is a port of the Wang et al group-testing algorithm (arxiv.org/abs/1407.2283) the service uses, so a sweep here costs the same API calls the real run would and measures what that would cost. A batch with no deletions is a single call.

Applying whatever this finds is a separate decision. `DeletionCheckInitiator.specific_records` already publishes `SpecificReindexParameters`, so the result can be applied to named ids rather than by sweeping the store again.

## How to test

`uv run pytest tests/test_report_calm_deletions.py` (85 tests). The algorithm port is checked against ground truth over 60 random shapes rather than against itself, and separately that it costs fewer calls than testing each record individually. The guards each have a test for the failure they exist to catch.

Verified against production on the read-only profile as far as its permissions reach: the store scan reports 258,635 live and 150,075 already deleted, matching the table. `platform-read_only` cannot read `calm_adapter/calm_api/*`, so the CALM half has not been exercised against the live API yet and wants a first run with `--limit`.

It only reads, and the CALM API calls are searches. The cost is API load on a system being wound down, which the group-testing algorithm keeps to roughly one call per 512 records where nothing has been deleted.

The report is a point-in-time answer. CALM is read-only, so it should not drift, but that assumption is worth re-testing if the file is acted on much later.

### DIFF
--- a/catalogue_graph/scripts/report_calm_deletions.py
+++ b/catalogue_graph/scripts/report_calm_deletions.py
@@ -1,0 +1,410 @@
+"""Report which CALM records our adapter store holds that CALM no longer has.
+
+This reads. It writes nothing to DynamoDB, publishes nothing, and produces a
+CSV for a human to act on.
+
+The real `calm_deletion_checker` service does the same search and then marks
+what it finds, which is why this exists separately. Its `test` is a search for
+a batch of record ids and a count of hits, and it treats every id the search
+does not return as deleted. If CALM answers but returns nothing, because the
+database is detached or the credentials get scoped down during the
+decommission, the service marks the whole batch deleted and publishes those
+deletions to the production catalogue. Nothing in it guards against that, and
+CALM has just been frozen ahead of being switched off, which is when it is
+least predictable. So: look first, decide, then mark.
+
+Safety, in the order it runs:
+
+ 1. A control probe searches for record ids we already hold and requires CALM
+    to return all of them. If it cannot find records that certainly exist, the
+    run stops and nothing is reported as deleted.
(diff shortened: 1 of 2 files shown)


kenoir:
Closing as not required.

PR #3659, opened 2026-09-14
Scale the CALM adapter to zero while harvesting is off

kenoir:
## What does this change?

Scales the CALM adapter to zero while `harvesting_enabled` is false, as the deletion checker already does. #3655 left it at max 2 so a re-harvest would still work. With CALM read-only and the store snapshotted to `vhs_snapshots/vhs-calm-adapter/`, a re-harvest could only pick up edits the switch-off exists to exclude, and the adapter was the last thing able to write to the store.

## How to test

A plan from main should show one change: the `calm_adapter` scalable target, max capacity 2 to 0.

The adapter has no running tasks and can't start any.

Apply from main only, after #3656 merges. This branch still pins VHS module v4.0.5, so applying it would turn deletion protection back off.


annak-k (approved):
Terraform currently outputs `Plan: 0 to add, 5 to change, 0 to destroy.` but this should become the expected `Plan: 0 to add, 1 to change, 0 to destroy.` with only `max_capacity = 2 -> 0` if https://github.com/wellcomecollection/catalogue-pipeline/pull/3656 is merged first

PR #3660, opened 2026-09-15
Import only Item-level records, and decode XML in MARC references

agnesgaroux:
Two fixes to the Axiell b number import builder, both found in the failures from the 2026-09-10 import run for https://github.com/wellcomecollection/platform/issues/6542.

## Item level only

The run failed on 516 rows, every one of them above Item level. Axiell refuses to save those with "A location may not be set for this type of record", even though the import sets no location.

The build now takes the level from MARC 351 `$c` and withholds anything that is not Item to `wrong_level.csv`, with a per-level count in `report.md` so the shape of what is held back is visible. `--all-levels` imports them anyway, which is only useful for inspecting the full set.

The level is read the way `adapters/transformers/axiell/organisation_and_arrangement.py` reads it: the first non-empty `$c` across all 351 fields, compared case-insensitively. 351 is repeatable and its casing varies across the migrated data, so taking only the first field recorded an empty level for a record whose first 351 carries no `$c`, withholding a valid Item row.

## XML in subfield values

MARC subfield values were being used raw. The 53 records whose AltRefNo contains an ampersand went into the CSV as `MSS.1055-1061 &amp; 7126`, and Axiell reported them as having no match. Subfield text (RefNo, AltRefNo, b numbers, title) is now unescaped before use.

The store holds `etree.tostring` output from the OAI-PMH harvest (`record_writer.py`), so it is XML in which only the five predefined entities can appear, and the decoder is `xml.sax.saxutils.unescape`. `html.unescape` gives identical results on this input, but it is the HTML5 decoder and the wrong spec to point at XML.

## Item part is withheld, and that part is untested

Level distribution across the 208,776 store records carrying a 907 RecordID, scanned on 2026-09-15:

| count | level |
|---|---|
| 187,070 | Item |
| 8,745 | Series |
| 5,918 | Sub-series |
| 2,844 | Section |
| 2,359 | Sub-section |
| 1,162 | Collection |
| 562 | Item part |
| 116 | no level at all |

The match on Item is exact, so the 562 Item part records (CALM's old "piece" level) are withheld along with everything above Item. Nothing here shows Axiell rejects them: 2026-09-10 attempted every level and its 516 failures are all above Item, so either that load carried no Item part rows or Axiell saved them fine. 562 is the ceiling on what excluding them costs, and fewer in practice since only records paired with a Sierra b number reach the import. Worth settling with a handful of test rows before the switchover load, not in this PR.

Two incidental findings from the same scan. The 516 failures are far smaller than the 21,028 above-Item records in the store, which fits pairs coming from Sierra bibs and skewing item-level. And every 351 `$c` value in the store is title case today, so the case-insensitive match guards the `work_type` BDD requirement rather than anything in the current data.

## Testing

Checked the 351 and 035 patterns over MARC samples with and without the `marc:` namespace prefix: a first 351 carrying only `$b`, an empty first `$c`, a whitespace-only `$c`, no 351 at all, lowercase `item`, and an 852 `$c` shelfmark that must not be mistaken for a level. Confirmed against lxml's own output that the XML decoder round-trips an escaped ampersand, a literal `&amp;` and a semicolon-less `&not` back to the original text. `ruff check` and `ruff format` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kenoir (approved):

agnesgaroux:
NOTE: VW confirmed "all digitised records should sync to Item level"

PR #3662, opened 2026-09-15
Folio transformer: add duration field

StepanBrychta:
## What does this change?

Addresses https://github.com/wellcomecollection/platform/issues/6549

Implements FOLIO→Work mapping for the duration field, mirroring Sierra transformer logic. Translates existing Scala unit tests into Python BDD tests to verify that Python behaviour matches Sierra behaviour. Cases where Python implementation deliberately diverges from Scala are codified at the end of the feature file.

The extractor reads the first 306 $a as `hhmmss` and returns the playing time in seconds. It deliberately diverges from Scala by requiring exactly six digits, as the field definition specifies. Scala accepts invalid five-digit values and misinterprets them (determined by cross-checking parsed values against duration text stored in 300 $a). There are 23 such five-digit values across the Sierra dataset. The Python extractor ignores them instead of trying to interpret them.

## How to test

Testing is deferred. Only two Folio records currently have a 306 field, because none of the Sierra material types that carry a duration have been migrated to Folio yet. Both records match their Sierra counterpart (2 matches, 0 differences). A full validation run against Folio data will only be possible after these material types are loaded.

In the meantime, I ran the extractor over all 2,926 visible Sierra works with a 306 field and compared the results against the Scala transformer's output in the source index. This produced 2,903 matches and 23 differences (all due to the five-digit handling difference described above).

The duration field on Folio works should match its counterpart on Sierra works.

There will be additional Folio data loads which might change what the raw MARC data looks like. The material types which carry a duration are not yet in Folio, so the extractor has not been exercised against real Folio data at scale. The Folio pipeline is non-production, and we will spot data changes in upcoming validation runs before releasing.


catalogue_graph/src/adapters/transformers/marc/duration.py
```diff
@@ -0,0 +1,32 @@
...
+from pymarc.record import Record
+
+# Six numeric characters in the pattern hhmmss, as required by the field definition.
+HHMMSS = re.compile(r"[0-9]{6}")
+
+
+def extract_duration(record: Record) -> int | None:
+    """Total playing time in seconds, from the first 306 ǂa read as hhmmss"""
+    values = [
+        value
```
annak-k: Should this strip whitespace and drop empty values, like `designation.py` and `former_frequency.py` do? A 306 ǂa of
"011012 " will give no duration, and a blank first ǂa masks a valid later one. I see that this would be a change from the Scala.

catalogue_graph/src/adapters/transformers/marc/duration.py
```diff
@@ -0,0 +1,32 @@
...
+    return _parse_hhmmss(values[0])
+
+
+def _parse_hhmmss(value: str) -> int | None:
+    # Ignore all values which are not exactly six digits.
+    if not HHMMSS.fullmatch(value):
+        return None
+
+    hours, minutes, seconds = (int(value[i : i + 2]) for i in range(0, 6, 2))
+    return hours * 3600 + minutes * 60 + seconds
```
annak-k, line 30: Do we need to worry about minutes and seconds > 60?

annak-k (approved):
Looks good! I just had two non-blocking questions.

catalogue_graph/src/adapters/transformers/marc/duration.py
StepanBrychta: Yes, good idea. It also makes the implementation simpler, since we can use the existing `first_non_empty_subfield` helper.

catalogue_graph/src/adapters/transformers/marc/duration.py
StepanBrychta: We could refuse to parse them, and it would technically be more correct. The Scala transformer doesn't, and there are a few Sierra works with duration values that are technically invalid (minutes > 60) but still give the duration the cataloguer intended when parsed. For example, `b28511414` has an invalid duration of `008000` (giving 80 minutes when parsed), and a separate field (`1 videocassette (80 min.)`) shows that the parsed value is correct.

I don't know how many of these invalid values will exist in Folio (most works which have durations have not been migrated yet). Maybe we can keep the permissive logic for now and revisit this after the next Folio migration. I added a TODO comment.

PR #3663, opened 2026-09-15
Snapshot the Sierra deleted table to its own file

kenoir:
## What does this change?

Following #3657, `snapshot_vhs.py` now snapshots Sierra as two files, one per table. `sierra` reads only the live table and the new `sierra-deleted` store reads only the deleted table, marking every row deleted. `verify_snapshot` now also fails on duplicate ids.

A review comment on #3657 asked whether the combined Sierra file would hold duplicate rows, and it would have. RFC 043 moved bibs deleted before 2018 into the deleted table, but a 2023 reharvest recreated 95,749 of them in the live table. Verify missed this because its count and id-set checks see duplicates the same way on both sides. A full comparison of all 95,749 pairs on 2026-09-15 found every pair equivalent, so anyone combining the files later can dedupe on id. In the live file `deleted` stays false even for deleted bibs because Sierra rows carry no isDeleted flag, which a comment in the code notes.

### Checklist

- [ ] Does this change the display model the catalogue API returns? No. It only changes an operator-run snapshot script.

## How to test

`uv run --group dev pytest tests/test_snapshot_vhs.py` from `catalogue_graph/` passes 29 tests, and the new duplicate-id test fails without the guard. With `--limit` on the read-only profile against the live tables, `sierra-deleted` scanned 557,348 rows and marked every sampled row deleted, and `sierra` scanned 1,819,867 rows and marked none.

Nothing is snapshotted by this PR. The Sierra adapter is still running, so this makes the tool correct ahead of that run.


catalogue_graph/scripts/snapshot_vhs.py
```diff
@@ -364,6 +370,12 @@ def verify_snapshot(output_path: str, rows: list[IndexRow]) -> None:
         )
 
     snapshot_ids = set(table.column("id").to_pylist())
+    if len(snapshot_ids) != table.num_rows:
+        raise SnapshotError(
+            f"Snapshot has {table.num_rows - len(snapshot_ids)} duplicate id(s). "
+            "A single table cannot hold these, so the rows came from more than "
+            "one place."
+        )
```
agnesgaroux, line 378: Nice! 🙌

agnesgaroux (approved):
So fresh and so clean 👌

PR #3666, opened 2026-09-16
Version indexed works on both source and merge time

### LINKED ISSUE #6686: Ingestor re-types Redirected works to Invisible despite a resolvable target
Found by the round 3 comparison, #6642. Part of the round 3 remediation, #6632.

## What happens

484 METS works are typed `Redirected` at `works-denormalised` with a resolvable `redirectTarget`, and the ingestor writes them to `works-indexed` as `Invisible` with the target dropped. The redirect survives on the target's side, so nothing downstream looks obviously broken.

Traced end to end, `n9btd4g3` into `enarb2as`:

| stage | source work | target work |
|---|---|---|
| `works-denormalised` | Redirected, `redirectTarget` set to `enarb2as` | Visible, lists `n9btd4g3` in `redirectSources` |
| `works-indexed` | **Invisible, `redirectTarget` gone** | Visible, still lists `n9btd4g3` |

All 484 targets are present and Visible at both stages, 460 Sierra and 24 committed CALM, none absent. So this is not a failure to resolve the target.

Production does not do this at all: it holds 4,280 Invisible METS works at `works-denormalised` and the same 4,280 at `works-indexed`.

## History

Round 1 (#6464) reported roughly 2,800 works re-typed between the two stages and left it unexplained, noting a possible interaction with the SAME_AS traversal bound from wellcomecollection/catalogue-pipeline#3529. Round 3 is the same defect at 484.

## Why it matters

These 484 are a quarter of the 999-work Invisible excess against production. #6542 currently attributes that excess to digitised METS works awaiting the b-number import, but an ingestor re-typing cannot be fixed by an import, and 460 of the 484 targets are Sierra works that Axiell has no bearing on. Clearing this is a prerequisite for #6542's re-measure meaning anything.

## Reproduce

Scan `works-indexed-2026-07-03` for `type: Invisible`, keep the ones whose `debug.source.identifier.identifierType.id` is `mets`, then compare `type` for the same canonical ids in `works-denormalised-2026-07-03`.

## Done when

Works that are Redirected at `works-denormalised` with a present target are indexed as Redirected, and the Invisible count at `works-indexed` matches `works-denormalised` the way production's does.

agnesgaroux:
## What does this change?

#3649 moved the works version from `sourceModifiedTime` to `mergedTime`. The reason was sound: a re-merge leaves the source record untouched, so two merges of the same work carried the same version and `external_gte` let a concurrent run overwrite a newer merge with an older one.

That fixed one case and opened another. A stale in-flight message carries old source data with a fresh merge time, so under merge-time ordering the stale write now wins. Ordering by either timestamp alone gets one of the two cases wrong, and we need both.

The version is now lexicographic on `(sourceModifiedTime, mergedTime)`, both to whole seconds, packed into one long: source seconds in the high digits, merge seconds since 2020-01-01 in the low nine.

```
sourceModifiedTime 2025-06-30T16:20:14Z
mergedTime         2025-08-19T19:23:53.532022Z
_version           1751300414177794633   ->  1751300414 | 177794633
```

A later source modification always wins. The merge time only breaks ties between merges of the same source content, which is exactly the re-merge case #3649 was after.

`get_modified_time` becomes `get_version` on `IndexableRecord`, so each record type owns its own ordering rather than `generate_operations` assuming epoch millis for everything. Images, concepts and augmented images keep epoch millis, and the `max(100, ...)` floor moves into `version_from_modified_time` where it now applies only to the record types it was ever for.

### Why the encoding looks like this

An external version is a single non-negative int64, and that drives every choice in `version.py`:

- Two bare second epochs are 10 digits each, which is 20 and does not fit in the 19 of `Long.MAX`. The merge field therefore goes in as an offset from a fixed base, which keeps it to 9 digits. **That base cannot move later**: every version would drop and every document would refuse writes until its index was rebuilt.
- Second precision on the source field is what makes room for the merge field at all. 19 digits will not hold source epoch millis (13) plus anything useful, since the remaining 5 digits span one day. Some sources do emit microsecond `sourceModifiedTime`, so this is a real if narrow loss: two source modifications inside one wall-clock second tie, and the merge time decides between them.
- Records with a `sourceModifiedTime` at the Unix epoch need the source field floored, which is what `MIN_SOURCE_TIME_SEC` is for. Packed unfloored they land around `2.1e8`, below the epoch-millis version those same documents already hold, and Elasticsearch would refuse every later write to them. Nothing about the record has to change for that to bite; the document simply becomes unwritable.
- The merge field is clamped at both ends. Past the base window it saturates, which loses the tie-break but keeps the source ordering intact. Unclamped it would carry into the source field and invert the order outright.

### Checklist

- [x] Does this change the display model the catalogue API returns? No. `_version` is index metadata and no document field changes.

## How to test

`catalogue_graph/tests/ingestor/test_indexable_version.py` states the ordering as invariants rather than examples: lexicographic ordering across a table of states, int64 bounds, the equal-version re-ingest case, and the migration property that a packed version always beats the epoch-millis version it replaces, parametrised over source times including epoch 0. That last one is the epoch-0 floor expressed as a property, so it fails if anyone later drops the floor.

The encoding was also exercised against a local Elasticsearch 9.1.4 while working this out, covering nine ordering scenarios and all 24 arrival orders of four `(source, merge)` states. Arrival order never decided the winner under the new encoding, while the current merge-millis encoding got 12 of those 24 wrong and the pre-#3649 source-millis encoding got the other 12. That harness is not in this PR. Everything it asserted is either a property of Elasticsearch itself, which cannot regress when `version.py` changes, or a property the unit tests already cover, and it ran against a hand-pinned cluster behind a deselected marker, so it would not have caught an Elasticsearch change either. If a version stops being accepted, the ingestor raises on the next run, since `ingestor_indexer.py` treats any bulk error that is not a version conflict as fatal.

Two findings from it are worth recording, since neither is obvious from the code:

- A 19-digit version survives `elasticsearch.helpers.bulk` and comes back exactly, which is worth knowing because it is past the 2^53 a double holds.
- The bulk API takes the integer part of a decimal version and succeeds, with no error and no conflict. Only the index API rejects one, where the version travels as a URL query parameter. A decimal of source seconds and merge millis, which was one reading of the original suggestion, would therefore have been accepted on the path the ingestor uses and quietly degraded to ordering by source seconds alone. Nothing in the shipped code can emit one, since the version is integer arithmetic and mypy enforces the `int`.

`elasticsearch.docker-compose.yml` moves from 8.12.0 to 9.1.4, matching the deployed `pipeline_storage` clusters.

Worth confirming independently before merge: my reading is that no production write path reads `_version` back. The Python ingestor is the only writer of works-indexed, and the Scala `ElasticIndexer` versions the earlier stages separately through its own `Indexable.version`.

`IndexerReport.version_conflict_count` is the signal. Expect it to rise, and that is the change working: stale writes that used to overwrite a newer merge now bounce as version conflicts, which the indexer already treats as benign and counts. A count that stays flat at zero through an overlapping reindex would suggest the new versions are not doing anything.

Fixes wellcomecollection/platform#6686

**The switch is one way.** Packed versions beat the epoch-millis versions already stored, so the live index takes the new scheme in place and no reindex is needed. Going back does mean rebuilding the index, because every version would drop and nothing could be written. Worth being deliberate about the deploy for that reason, which is why this is a draft.

**Epoch-0 source records.** The floor covers them, and `test_a_packed_version_beats_the_epoch_millis_version_it_replaces` is parametrised over source times including epoch 0, so it fails if anyone later drops the floor. The blast radius is also smaller than it first looks, since only documents the ingestor has rewritten since #3649 landed on 2026-09-11 are exposed. One still holding its pre-#3649 version of `100` takes any new value fine. I have not counted how many documents that is, and it is awkward to count: `sourceModifiedTime` is not indexed in any pipeline index, since `works_indexed` maps `debug` as `dynamic: false` and `works_denormalised` indexes only `state.mergedTime`, so it needs a `_source` scroll rather than a query.

**Two source modifications inside one second.** These tie on the source field and fall through to the merge time, so an older source can still win inside that window. I think the trade is right, since the alternative is no merge-time tie-break at all, but it is a real narrowing rather than a full fix.

**Field headroom.** The merge field runs to 2051-09 and the source field to 2262. Both are past the life of any single index, and the failure mode at the merge boundary is a lost tie-break rather than a wrong order.

`debug.source.version`, the adapter's own counter, was considered as the ordering key. It orders source states exactly and would leave room for a millisecond merge tie-break. I dropped it because second precision is acceptable here, so the extra precision buys nothing, and an adapter store rebuild is always followed by a full reindex into a fresh index, which is the only case where its counters restarting would have mattered.


agnesgaroux:
The harness the description mentioned is now in the PR, under `catalogue_graph/integration/es_versioning/`, with a README covering what each part establishes and the known limits. 23 tests, 19 seconds against a local cluster.

`uv run pytest -m elasticsearch`

Three things worth a reviewer's attention:

**They live under `integration/` rather than `tests/`.** `tests/conftest.py` has an autouse fixture that replaces `elasticsearch.Elasticsearch` and `helpers.bulk` with mocks for every test, so a live-cluster harness there would quietly test the mock. `integration/` has no such conftest and is already a deselected-by-default testpath.

**They are deselected by default.** New `elasticsearch` marker, added to the existing `-m 'not integration'` in `addopts`. They need Docker and pull a large image, so running them on every PR did not seem worth the minutes. Easy to change if you would rather CI covered them.

**`elasticsearch.docker-compose.yml` moves from 8.12.0 to 9.1.4**, the version the deployed clusters run, so local behaviour matches production. Kibana follows because it has to match the Elasticsearch major. 

The shipped encoding is imported from `ingestor.models.indexable.version` rather than restated, so the tests cannot drift from what the ingestor writes. `version_encodings.py` holds the rest for comparison: the two encodings we have shipped, both readings of the original suggestion, and the bit-packed alternative that was considered and not shipped.

agnesgaroux:
Correcting something in the description.

**A decimal version is not rejected on the path the ingestor uses.** The description says an external version is an integer and implies a decimal bounces. That holds for the index API, where the version is a URL query parameter and refuses to parse. The bulk API is different: the version is a value in the action metadata, and that parser takes the integer part and discards the rest.

```
_version 1619481599.1761916  -> accepted, stored _version = 1619481599
_version 100.9               -> accepted, stored _version = 100
```

No error, no conflict, nothing in the indexer report. So the reading of the original suggestion that looked merely invalid would in fact have been accepted by the ingestor and quietly degraded to ordering by source seconds alone, which is the behaviour #3649 set out to fix. That is a worse outcome than a rejection, and it is now tested in its own right rather than inferred from the query-parameter result. 26 tests.

It does not change the encoding this PR ships, which was already an integer. It does change why: the integer is not a formality that Elasticsearch enforces for you, it is the thing standing between us and a silent truncation.

kenoir:
Under #3649 the merge time orders your examples and the Redirected version wins, and this PR gives the same answer because the source times are equal.

Where this PR would differ is a later merge carrying older source data, and I think the place that can start is the id-minter. The Python minter writes `works-identified` with plain bulk index actions and no version guard, in both pipelines, and minter runs do overlap. The Scala minter versioned that index with `external_gte` on `sourceModifiedTime`, and I think the guard was lost in the port rather than dropped on purpose: #3266 doesn't mention it, and #3347 restored it for the ingestor only.

So I'd rather fix it there. Restoring `external_gte` on source millis to the minter's bulk actions stops stale data before `works-identified`, so identified, denormalised and indexed all stay consistent, and #3649 already orders the ingestor by merge time. Guarding only `works-indexed` would leave the two upstream indexes holding the stale copy and reject every later merge of that work, including a METS manifest arriving, until the record is next edited. The minter change is tracked in wellcomecollection/platform#6720, and I don't think we should take this one.

kenoir:
Closed, superseded by https://github.com/wellcomecollection/catalogue-pipeline/pull/3693/changes

PR #3668, opened 2026-09-18
Retry stranded OAI-PMH windows and quieten the gap alert

### LINKED ISSUE #6713: OAI-PMH windows that fail behind the cursor are never retried
## What happened

On 2026-09-17 the Axiell OAI endpoint returned empty bodies from about 13:45Z to 16:29Z. The 13:30 to 14:00Z windows failed seconds before it recovered, and the later windows succeeded in the same run. The trigger resumes from the last published window, so those two were never retried. The gap alert fired on every 15 minute run until the gap was reloaded by hand on 2026-09-18. 38 records were missing from the adapter store in the meantime.

## Done

The gap is closed and the 38 records have been through reconcile and the 2026-07-03 transformer. The window table shows no gaps or failures.

## Fix

wellcomecollection/catalogue-pipeline#3668 changes both OAI-PMH adapters.

- The trigger retries gaps behind the cursor, within the look-back and the lag tolerance.
- A gap being retried is reported after three failed runs, then once a day.
- An operator can backfill a range by starting the state machine with a `window` input, which runs reconcile, enrichment and publish. The local reloader does none of those.

## Done when

The PR is merged and deployed. The new behaviour has not run against live infrastructure, so the first check is the next window that fails behind the cursor. The following run should start at that window and the alert should stay quiet.

kenoir:
## What does this change?

On 2026-09-17 the Axiell OAI endpoint returned empty bodies from about 13:45Z to 16:29Z. Two windows failed seconds before it recovered, so they sat behind the trigger's cursor and were never retried. The gap alert fired on every 15 minute run until the gap was reloaded by hand on 2026-09-18, and 38 records were missing from the adapter store in the meantime.

- The trigger starts its range at the oldest gap behind the cursor, within the look-back and the lag tolerance, so a stranded window is retried instead of waiting for an operator. This wires up `AUTO_RETRY_FAILED_WINDOWS`, which both adapter configs declared but nothing read.
- The harvester no longer reuses success windows that are already published, so the wider range does not re-publish their changesets.
- The gap alert is throttled without storing state. A gap being retried is reported once three scheduled runs have failed to close it, and after that once a day.
- The trigger accepts an operator `window` in its event, so a backfill runs through reconcile, enrichment and publish. The alert and the README point at that rather than the local reloader, which publishes nothing.
- I left the empty-body retry budget alone, because the client already retries three times and a larger budget would be paid by every backlog window during an outage.

Closes wellcomecollection/platform#6713

### Checklist

- [ ] Does this change the display model the catalogue API returns? No, this only touches adapter harvesting and alerting.

## How to test

`uv run pytest tests/adapters/extractors/oai_pmh tests/adapters/utils` in `catalogue_graph` passes 465 tests, with ruff and mypy clean, and new tests cover the retry limits, the off switch, the alert timing, the operator window and the harvester skipping published windows. Nothing has been run against live infrastructure. It is Python only with no terraform and applies to both the Axiell and FOLIO adapters, so after deploy the next window that fails behind the cursor should be picked up by the following run, with no alert unless three runs fail to close it.

A stranded window should close itself within a run or two, and a gap that does need an operator should reach Slack once rather than every 15 minutes.

- Throttling can delay an operator seeing a gap that the retry cannot close, by up to three runs.
- The operator `window` path skips the lag check, so a badly chosen range is harvested as asked.


catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
```diff
@@ -283,11 +344,19 @@ def lambda_handler(
         trace_id=get_trace_id(context),
         pipeline_step=f"{config.config.pipeline_step_prefix}_trigger",
     )
+    # A "window" key marks an operator backfill of that range.
+    window = (
+        IncrementalWindow.model_validate(event["window"])
+        if event.get("window")
+        else None
+    )
```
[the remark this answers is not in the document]
kenoir: Fixed in aa162ab50e. Any non-null `window` is now validated, so `{"window": {}}` fails with the missing `end_time` error, and there is a test for it.

catalogue_graph/src/adapters/utils/window_notifier.py
```diff
@@ -88,6 +103,41 @@ def notify_if_gaps(
...
+    ) -> list[CoverageGap]:
+        """Return the gaps to report on this run, or none if no report is due."""
+        if trigger_time is None:
+            return gaps
+
+        interval = timedelta(minutes=self.window_minutes)
+        is_digest_run = (
+            trigger_time.hour == self.DIGEST_HOUR_UTC
+            and trigger_time.minute < self.window_minutes
+        )
```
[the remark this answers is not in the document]
kenoir: Fixed in aa162ab50e. The digest slot is compared after converting the trigger time to UTC, with a test using a +02:00 timestamp.

catalogue_graph/src/adapters/utils/window_reporter.py
```diff
@@ -352,7 +355,13 @@ def _calculate_coverage_gaps(
             gap_start = merged_intervals[i][1]
             gap_end = merged_intervals[i + 1][0]
             if gap_start < gap_end:
-                coverage_gaps.append(CoverageGap(start=gap_start, end=gap_end))
+                coverage_gaps.append(
+                    CoverageGap(
+                        start=gap_start,
+                        end=gap_end,
+                        stranded_at=self._stranded_at(successful_rows, gap_end),
```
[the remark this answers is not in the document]
kenoir: Fixed in aa162ab50e. The gap before the first successful window now gets `stranded_at` from that window, and a reporter test covers both gap positions.

agnesgaroux:
On `_determine_retry_start` (`trigger.py:89`): `now` is the only datetime in this path that isn't normalised on the way in, and this change is the first thing to compare it directly against a gap boundary.

`horizon = now - timedelta(...)` is compared to `gap.start`, which the reporter has already made tz-aware. If `now` is naive that raises `TypeError: can't compare offset-naive and offset-aware datetimes`. It can be naive from the CLI, where `datetime.fromisoformat(args.at.replace("Z", "+00:00"))` only produces an aware value if you typed the `Z`: the `--at` help text shows one in the example but nothing enforces it. Same for a hand-written state machine input, since `ScheduledEvent.time` and `OAIPMHTriggerEvent.now` are both bare `datetime` fields. EventBridge always sends `Z`, so scheduled runs are unaffected.

It didn't surface before because the other consumers launder it: `_determine_start(...).astimezone(UTC)` and `end_time = now.astimezone(UTC)` treat a naive value as local time rather than raising, so a naive `--at` silently shifted the window by the local offset instead of failing.

Every other datetime crossing this boundary goes through `ensure_datetime_utc`: `IncrementalWindow.start_time`/`end_time`, the reporter's row timestamps, `WindowSummary` parsing. A single `now = ensure_datetime_utc(now)` at the top of `build_window_request` would match that and fix the drift as well as the crash. `window_notifier.py:133` has the same shape, where `age = trigger_time - gap.stranded_at` uses the raw `trigger_time` although `trigger_utc` is computed two lines above.

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
```diff
@@ -97,20 +128,34 @@ def build_window_request(
...
         job_id: Optional job identifier (generated if not provided).
+        window: Operator-supplied range, used as given when present.
 
     Returns:
         OAIPMHLoaderEvent to pass to the loader step.
 
     Raises:
         RuntimeError: If adapter is too far behind (lag check) or no windows ready.
     """
+    if window is not None:
```
agnesgaroux, line 166: The operator window is passed through with only `IncrementalWindow`'s shape checks: `end_time` present, `start_time` defaulted, `start_time <= end_time`. Nothing requires either boundary to sit on a `window_minutes` boundary, and `generate_windows` sets its cursor to `start_time` verbatim before snapping to the epoch grid. A hand-typed `13:37 -> 15:00` therefore produces a first sub-window of `13:37-13:45`.

That key matches no stored row. The harvester looks rows up by exact `to_iso_string()`, and the row covering that period is filed under `13:30-13:45`. So the sub-window is classed as pending, harvested, and written as a second overlapping status row, and its records go out in a fresh changeset although the enclosing window is already published. That is the case the new `published_at` skip just below exists to prevent.

Coverage reporting survives it, since `_merge_intervals` merges the overlap before summing, so the cost is duplicated harvest work and duplicated changesets rather than a wrong gap report. The notifier's suggested execution input is grid-aligned, so this only bites a hand-typed window, which is what the new README section invites.

Fix: reject a `start_time` that isn't aligned to `window_minutes` here. Snapping it back to the enclosing boundary would also work, but I think silently widening the range an operator asked for is its own surprise.

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
```diff
@@ -74,6 +75,36 @@ def _determine_start(
...
+    now: datetime,
+    window_lookback_days: int,
+    max_lag_minutes: int,
+) -> datetime | None:
+    """Return the start of the oldest gap behind the cursor worth retrying.
+
+    Gaps older than the lookback are left for an operator. So is a backlog
+    larger than the lag tolerance, which would delay new windows for too long.
+    """
+    horizon = now - timedelta(days=window_lookback_days)
```
agnesgaroux, line 91: now is the only datetime in this path that isn't normalised on the way in, and this change is the first thing to compare it directly against a gap boundary.

horizon = now - timedelta(...) is compared to gap.start, which the reporter has already made tz-aware. If now is naive that raises TypeError: can't compare offset-naive and offset-aware datetimes. It can be naive from the CLI, where datetime.fromisoformat(args.at.replace("Z", "+00:00")) only produces an aware value if you typed the Z: the --at help text shows one in the example but nothing enforces it. Same for a hand-written state machine input, since ScheduledEvent.time and OAIPMHTriggerEvent.now are both bare datetime fields. EventBridge always sends Z, so scheduled runs are unaffected.

It didn't surface before because the other consumers launder it: _determine_start(...).astimezone(UTC) and end_time = now.astimezone(UTC) treat a naive value as local time rather than raising, so a naive --at silently shifted the window by the local offset instead of failing.

Every other datetime crossing this boundary goes through ensure_datetime_utc: IncrementalWindow.start_time/end_time, the reporter's row timestamps, WindowSummary parsing. A single now = ensure_datetime_utc(now) at the top of build_window_request would match that and fix the drift as well as the crash. window_notifier.py:133 has the same shape, where age = trigger_time - gap.stranded_at uses the raw trigger_time although trigger_utc is computed two lines above.

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
kenoir: Agreed, and rejecting is the right call over snapping. I reproduced it: `13:37 -> 15:00` generates a first sub-window of `13:37-13:45`, which misses the status lookup and gets harvested again. The trigger now raises if an operator `start_time` is off the `window_minutes` grid, and the error names the enclosing boundary. I left `end_time` alone because the generator already floors it to the grid, so it can't produce an off-grid key. The README backfill section says the start must sit on a boundary. Fixed in d90ef183c2.

catalogue_graph/src/adapters/steps/oai_pmh/trigger.py
kenoir: Agreed. `build_window_request` now runs `now` through `ensure_datetime_utc` first, so a naive `--at` or execution input is read as UTC like every other datetime here, and the notifier uses the normalised value for the gap age as well as the digest check. `generate_job_id` had the same problem: it used `astimezone(UTC)`, which reads a naive value as local time, so a naive `--at` from a laptop in BST got a job id an hour off its window. That goes through `ensure_datetime_utc` too. Tests cover a naive `now` in the trigger, a naive `trigger_time` in the notifier, and the job id. Fixed in 54462135ff.

agnesgaroux (approved):

PR #3671, opened 2026-09-18
Move Axiell Folio sync to Folio Dev Server

anubhavijay:
## What does this change?

- Configure the Axiell Folio sync Lambda to run in the catalogue VPC.
- Add the supporting Lambda, networking, trigger, EventBridge, and state-machine Terraform wiring required for the Folio Dev sync flow.
- Update the completion rule and related variables for the new deployment path.

## How to test

- Run Terraform formatting and validation in `catalogue_graph/infra/adapters`.
- Review `terraform plan` for the intended environment, paying particular attention to Lambda VPC settings, IAM, EventBridge, and Step Functions changes.
- Deploy to Folio Dev and run an end-to-end Axiell Folio sync.
- Confirm completion events, retries, failures, and Lambda CloudWatch logs behave as expected.

- The sync Lambda runs in the catalogue VPC and reaches all required dependencies.
- Axiell Folio syncs complete successfully in Folio Dev.
- Existing completion events and state-machine behaviour remain functional, with no regression in sync success rate, latency, retries, or error visibility.

- Incorrect subnet, security-group, route, or endpoint configuration could prevent the Lambda from reaching Folio or AWS services.
- EventBridge trigger changes could cause missed or duplicate invocations.
- Terraform changes could alter existing state-machine or Lambda behaviour.
- Mitigate these risks with `terraform plan`, staged deployment, an end-to-end sync test, and CloudWatch monitoring.


catalogue_graph/infra/adapters/modules/axiell_folio_sync/event_trigger.tf
```diff
@@ -42,6 +42,14 @@ resource "aws_cloudwatch_event_rule" "axiell_adapter_completed" {
...
     }
   })
+
+  # Trigger-window schedules own this after apply; ignore state changes so an
+  # out-of-hours apply cannot re-enable the trigger against a stopped sandbox.
+  state = local.trigger_window_enabled ? "DISABLED" : "ENABLED"
+
+  lifecycle {
+    ignore_changes = [state]
+  }
```
kenoir, line 52: I think the rollback path can leave this rule disabled with nothing to turn it back on. `ignore_changes = [state]` applies in both modes, so if the window is removed while the rule is off, the two schedules are destroyed, the rule stays `DISABLED`, and the plan shows no diff because Terraform only uses ignored arguments on create (https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle). That is any apply without `-var 'folio_dev_target_enabled=true'` after 17:30 or at a weekend, which covers the Rollback step in `docs/axiell-folio-sync-lambda-dev-instance.md` and someone applying this root for an unrelated adapter change.

catalogue_graph/infra/adapters/modules/axiell_folio_sync/trigger_window.tf
```diff
@@ -0,0 +1,99 @@
...
+# enableRule and disableRule take the rule name and bus, not an ARN.
+locals {
+  trigger_rule_input = local.trigger_window_enabled ? jsonencode({
+    Name         = aws_cloudwatch_event_rule.axiell_adapter_completed.name
+    EventBusName = var.event_bus_name
+  }) : null
+}
+
+resource "aws_scheduler_schedule" "trigger_on" {
+  count = local.trigger_window_enabled ? 1 : 0
```
kenoir, line 64: Is it intended that changes harvested while the rule is off never reach the sandbox? The adapter only publishes `axiell.adapter.completed` when it has changesets, the sync reads only the changesets named on that event, and EventBridge does not keep events for a disabled rule (https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_DisableRule.html). So anything harvested between 17:30 and 09:00 or over a weekend is skipped for good, and an edit saved at about 17:20 can miss the window if its harvest finishes after 17:30. I think that is fine for exercising the write path, but it means the sandbox will drift from Axiell a little every day, so any check that compares the two needs to allow for it. Could you add a line to the comment here or the doc saying this is accepted?

catalogue_graph/infra/adapters/modules/axiell_folio_sync/state_machine.tf
```diff
@@ -16,7 +16,7 @@ locals {
             job_id           = "{% $states.input.detail.job_id %}"
             transformer_type = "{% $exists($states.input.detail.transformer_type) ? $states.input.detail.transformer_type : null %}"
             sample_limit     = "{% $exists($states.input.detail.sample_limit) ? $states.input.detail.sample_limit : null %}"
-            dry_run          = "{% $exists($states.input.detail.dry_run) ? $states.input.detail.dry_run : ${var.dry_run_default} %}"
+            dry_run          = "{% $exists($states.input.detail.dry_run) ? $states.input.detail.dry_run : ${local.dry_run_default} %}"
```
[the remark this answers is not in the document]
kenoir: I think Copilot is right about the mechanics but this is how main already behaves: an event that names `dry_run: false` and `folio_target: "prod"` has always won over the defaults. What changes in this PR is the comment in `main.tf`, which now says the module "can only ever write while the target is dev". Could that be reworded to say the clamp covers scheduled runs and the Lambda's env default, and that a hand-built event can still override both? If we do want a hard guard against prod writes during validation, I think it belongs in the handler (refuse `dry_run=false` unless the resolved target is `dev`) and could be a follow-up.

kenoir (approved):

PR #3681, opened 2026-09-21
Append new window status rows instead of overwriting them

kenoir:
## What does this change?

Long OAI-PMH catch-ups slow down with every window they harvest. On 2026-09-21 the folio adapter had a five day gap to cover (483 windows). Empty windows cost 2 seconds each at the start of the run and 9 seconds by window 200.

Each window writes one row to the window status table with `overwrite` and a `window_key` filter. pyiceberg plans the delete half of an overwrite by reading every manifest in the table, even when no row matches, and every commit adds a manifest. Commit n therefore reads about n manifests from S3. Scheduled runs never notice, because they write one window and S3 Tables compacts the table between runs.

The harvester already loads the status rows for the range before it starts. For a window with no row in that map, the first write is now a plain append, which reads no manifests. A `flush_every` batch appends when every window in it is new. Later writes for the same window, retries of failed windows, reprocess mode and `mark_published` still overwrite. An append cannot duplicate a row another writer added in the meantime: the commit asserts the branch head it started from and fails with `CommitFailedException`, as an overwrite would. A test pins that behaviour, since the approach depends on it.

Windows that contain records still pay for one overwrite when their final row replaces the progress row. I think merging manifests on commit (`commit.manifest-merge.enabled`) is worth a separate look, because it would bound the cost of every remaining overwrite, but it needs a property change on the live tables.

## How to test

`uv run pytest tests/adapters/utils/test_window_store.py tests/adapters/utils/test_window_harvester.py`. The new tests count manifest reads rather than checking snapshot operations, because an overwrite that matches nothing is also recorded as an append.

Commit spacing in the window status table's snapshot log should stay flat across the next long catch-up.

A duplicate row would need an append for a key that already exists. The status map covers every candidate window's start time, a key is forgotten before its first write so a write of unknown outcome is never retried as an append, and a concurrent writer fails the commit.


kenoir:
Close as won't do.

